PR #34, opened 2022-11-10
fix(thfe): update public key parameters

jborfila:


IceTDrinker (approved):
Glad we caught that, thanks for the fix

PR #52, opened 2022-11-30
fix(doc): update pk encryption example for shortint

jborfila:


IceTDrinker (approved):
thanks for the cherry-pick !

PR #55, opened 2022-11-30
chore(ci): do not parse report dir when walking subdirectories

soonum:


ci/benchmark_parser.py
```diff
@@ -52,6 +52,8 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix=""):
             continue
         for subdir in dire.iterdir():
             if walk_subdirs:
+                if subdir.name == "report":
+                    continue
                 subdir = subdir.joinpath("new")
             elif subdir.name != "new":
```
IceTDrinker, line 58: what about checking if subdir exists ? and if it does not continue ? Handles all the potential dirs in there

soonum: I don't quite understand your point.

IceTDrinker: her our issue is that report/new did not exist so my proposal is to do

```python
subdir = subdir.joinpath("new")
if not subdir.exists():
    continue
```

this way we avoid all dirs that may not be benchmark dirs

what do you think ?

soonum: Oh you're right.

IceTDrinker (approved):
thanks !

PR #64, opened 2022-12-16
fix(tools): fix arch detection script for aarch64

dartdart26:
On Linux with Apple M1, the output of `uname -a` is:

```
Linux ... aarch64 aarch64 aarch64 GNU/Linux
```

Therefore, recognize that output as aarch64.

### DIFF
--- a/scripts/get_arch_feature.sh
+++ b/scripts/get_arch_feature.sh
@@ -4,7 +4,7 @@ set -e
 
 ARCH_FEATURE=x86_64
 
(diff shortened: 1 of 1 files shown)


IceTDrinker:
please use the conventional commit format https://www.conventionalcommits.org/en/v1.0.0/

Here the commit message can be

```
fix(tools): fix arch detection script for aarch64
```

IceTDrinker (approved):

PR #67, opened 2023-01-03
Compute key sizes in benchmarks

soonum:
I didn't found another way than making the struct public to expose the internals of the boolean server key.

Moreover, I didn't find a way to avoid having two for loops (one for boolean keys and one for shortint). I've tried to implement a function using a generic input argument to refactor the common code but without success.

For now, the measurement is not implemented in the workflow. I should probably create an independent workflow just for this task since it's measuring both boolean and shortint keys in one go, what do you think?

### DIFF
--- a/.github/workflows/boolean_benchmark.yml
+++ b/.github/workflows/boolean_benchmark.yml
@@ -64,7 +64,7 @@ jobs:
 
       - name: Run benchmarks
         run: |
-          RUSTFLAGS="-C target-cpu=native" cargo +nightly bench --bench boolean-bench --features=x86_64-unix,boolean,internal-keycache -p tfhe
+          make bench_boolean
 
       - name: Parse results
         run: |
@@ -84,7 +84,7 @@ jobs:
 
       - name: Run benchmarks with AVX512
         run: |
-          RUSTFLAGS="-C target-cpu=native" cargo +nightly bench --bench boolean-bench --features=x86_64-unix,boolean,internal-keycache,nightly-avx512 -p tfhe
+          make AVX512_SUPPORT=ON bench_boolean
 
       - name: Parse AVX512 results
         run: |
@@ -92,6 +92,16 @@ jobs:
           --name-suffix avx512 \
           --append-results
 
+      - name: Measure key sizes
+        run: |
+          make measure_boolean_key_sizes
+
+      - name: Parse key sizes results
+        run: |
+          python3 ./ci/benchmark_parser.py tfhe/boolean_key_sizes.csv ${{ env.RESULTS_FILENAME }} \
+          --key-sizes \
+          --append-results
+  
       - name: Upload parsed results artifact
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/shortint_benchmark.yml
+++ b/.github/workflows/shortint_benchmark.yml
@@ -64,7 +64,7 @@ jobs:
 
       - name: Run benchmarks
         run: |
-          RUSTFLAGS="-C target-cpu=native" cargo +nightly bench --bench shortint-bench --features=x86_64-unix,shortint,internal-keycache -p tfhe
+          make bench_shortint
 
       - name: Parse results
         run: |
@@ -85,13 +85,23 @@ jobs:
 
       - name: Run benchmarks with AVX512
         run: |
-          RUSTFLAGS="-C target-cpu=native" cargo +nightly bench --bench shortint-bench --features=x86_64-unix,shortint,internal-keycache,nightly-avx512 -p tfhe
+          make AVX512_SUPPORT=ON bench_shortint
 
       - name: Parse AVX512 results
         run: |
           python3 ./ci/benchmark_parser.py target/criterion ${{ env.RESULTS_FILENAME }} \
           --walk-subdirs \
           --name-suffix avx512 \
+          --append-results
+
+      - name: Measure key sizes
+        run: |
+          make measure_shortint_key_sizes
+
+      - name: Parse key sizes results
+        run: |
+          python3 ./ci/benchmark_parser.py tfhe/shortint_key_sizes.csv ${{ env.RESULTS_FILENAME }} \
+          --key-sizes \
           --append-results
 
       - name: Upload parsed results artifact
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,17 @@ RS_BUILD_TOOLCHAIN:=$(shell \
 	( (echo $(TARGET_ARCH_FEATURE) | grep -q x86) && echo stable) || echo $(RS_CHECK_TOOLCHAIN))
 CARGO_RS_BUILD_TOOLCHAIN:=+$(RS_BUILD_TOOLCHAIN)
 MIN_RUST_VERSION:=1.65
+AVX512_SUPPORT?=OFF
 WASM_RUSTFLAGS:=
 # This is done to avoid forgetting it, we still precise the RUSTFLAGS in the commands to be able to
 # copy paste the command in the terminal and change them if required without forgetting the flags
 export RUSTFLAGS:=-C target-cpu=native
+
+ifeq ($(AVX512_SUPPORT),ON)
+		AVX512_FEATURE=nightly-avx512
+else
+		AVX512_FEATURE=
+endif
 
 .PHONY: rs_check_toolchain # Echo the rust toolchain used for checks
 rs_check_toolchain:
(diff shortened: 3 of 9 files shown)


tfhe/src/boolean/engine/bootstrapping.rs
```diff
@@ -80,8 +80,8 @@ impl Memory {
 /// * `key_switching_key` - a public key, used to perform the key-switching operation.
 #[derive(Clone)]
 pub struct ServerKey {
-    pub(crate) bootstrapping_key: FourierLweBootstrapKeyOwned,
-    pub(crate) key_switching_key: LweKeyswitchKeyOwned<u32>,
+    pub bootstrapping_key: FourierLweBootstrapKeyOwned,
+    pub key_switching_key: LweKeyswitchKeyOwned<u32>,
 }
```
IceTDrinker, line 85: we can add a function in impl ServerKey that would be `pub fn bootstrapping_key_size_elements(&self) -> usize` or even `pub fn bootstrapping_key_size_bytes(&self) -> usize` that basically does

```rust
        self.bootstrapping_key.as_ref().len() * std::mem::size_of::<concrete_fft::c64>()
```

what do you think ?

IceTDrinker (commented):
For the question about having a separate worklow, the key_sizes can be split into boolean_key_size.rs and shortint_key_size.rs, this way you call the one you want in the benchmarking workflow, what do you think ?

tfhe/src/boolean/engine/bootstrapping.rs
IceTDrinker: and then an equivalent function for the ksk

soonum: Yes, I like the idea. But in this case, we should do it for Shortint keys too right ?

IceTDrinker: yes that's fine by me !

.github/workflows/boolean_benchmark.yml
```diff
@@ -92,6 +92,16 @@ jobs:
           --name-suffix avx512 \
           --append-results
 
+      - name: Measure key sizes
+        run: |
+          RUSTFLAGS="-C target-cpu=native" cargo +nightly --example boolean_key_sizes --features="x86_64-unix boolean internal-keycache"
```
IceTDrinker, line 97: can you create a target in the makefile perhaps ?

.github/workflows/shortint_benchmark.yml
```diff
@@ -94,6 +94,16 @@ jobs:
           --name-suffix avx512 \
           --append-results
 
+      - name: Measure key sizes
+        run: |
+          RUSTFLAGS="-C target-cpu=native" cargo +nightly --example shortint_key_sizes --features="x86_64-unix shortint internal-keycache"
```
IceTDrinker, line 99: same thing, maybe a make target, what do you think ?

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -534,6 +534,22 @@ impl ServerKey {
             engine.create_trivial_assign(self, ct, value).unwrap()
         })
     }
+
+    pub fn  bootstrapping_key_size_elements(&self) -> usize {
+        self.bootstrapping_key.as_view().data().as_ref().len()
+    }
+
+    pub fn bootstrapping_key_size_bytes(&self) -> usize {
+        self.bootstrapping_key.as_view().data().as_ref().len() * std::mem::size_of::<concrete_fft::c64>()
```
IceTDrinker, line 543: let's call bootstrapping_key_size_elements here to avoid having to update potential formulas twice :)

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -534,6 +534,22 @@ impl ServerKey {
...
+    }
+
+    pub fn  key_switching_key_size_elements(&self) -> usize {
+        self.key_switching_key.as_ref().len()
+    }
+
+    pub fn key_switching_key_size_bytes(&self) -> usize {
+        self.key_switching_key.as_ref().len() * std::mem::size_of::<u64>()
```
IceTDrinker, line 551: same thing here

tfhe/src/boolean/engine/bootstrapping.rs
```diff
@@ -80,8 +80,26 @@ impl Memory {
...
+    pub (crate) key_switching_key: LweKeyswitchKeyOwned<u32>,
+}
+
+impl ServerKey {
+    pub fn  bootstrapping_key_size_elements(&self) -> usize {
+        self.bootstrapping_key.as_view().data().as_ref().len()
+    }
+
+    pub fn bootstrapping_key_size_bytes(&self) -> usize {
+        self.bootstrapping_key.as_view().data().as_ref().len() * std::mem::size_of::<concrete_fft::c64>()
```
IceTDrinker, line 93: let's call bootstrapping_key_size_elements directly

tfhe/src/boolean/engine/bootstrapping.rs
```diff
@@ -80,8 +80,26 @@ impl Memory {
...
+    }
+
+    pub fn  key_switching_key_size_elements(&self) -> usize {
+        self.key_switching_key.as_ref().len()
+    }
+
+    pub fn key_switching_key_size_bytes(&self) -> usize {
+        self.key_switching_key.as_ref().len() * std::mem::size_of::<u64>()
```
IceTDrinker, line 101: same thing

IceTDrinker (commented):
A few questions, to make sure your PR complies with the CI you can:

```
make fmt
make pcc
# Fix lints until this goes green
```

Make sure the commit message follows the format we use, github actions `Check commit and PR compliance` from workflow check_commit.yml

tfhe/src/boolean/engine/bootstrapping.rs
IceTDrinker: calling the keyswitch function of course

.github/workflows/boolean_benchmark.yml
soonum: We could also create target for benchmarks no ?

IceTDrinker: ah yes ! don't know why I did not think of this !

soonum: By the way, do we need the nightly Rust toolchain to be used for the tests?

soonum: I believe benchmarks with AVX support needs nightly. And it does seems that the Makefile support installation of a nightly tool chain. Am I wrong ?

IceTDrinker: avx-512 support needs nighlty yes

IceTDrinker: the check toolchain is always a nighlty one so you can use that one as toolchain, if you want we can have a quick call

Makefile
```diff
@@ -205,6 +212,30 @@ test_nodejs_wasm_api_in_docker: build_nodejs_test_docker
 test_nodejs_wasm_api: build_node_js_api
 	cd tfhe && node --test js_on_wasm_tests
 
+.PHONY: bench_shortint # run benchmarks for shortint
+bench_shortint: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) bench \
```
IceTDrinker, line 217: this needs to be the check toolchain not the build toolchain

Makefile
```diff
@@ -205,6 +212,30 @@ test_nodejs_wasm_api_in_docker: build_nodejs_test_docker
 test_nodejs_wasm_api: build_node_js_api
 	cd tfhe && node --test js_on_wasm_tests
 
+.PHONY: bench_shortint # run benchmarks for shortint
```
IceTDrinker, line 215: run -> Run (nitpick but the toolchain selection needs a fix, so I write it here)

Makefile
```diff
@@ -205,6 +212,30 @@ test_nodejs_wasm_api_in_docker: build_nodejs_test_docker
...
+bench_shortint: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) bench \
+	--bench shortint-bench \
+	--features=$(TARGET_ARCH_FEATURE),shortint,internal-keycache,$(AVX512_FEATURE) -p tfhe
+
+.PHONY: bench_boolean # Run benchmarks for boolean
+bench_boolean: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) bench \
```
IceTDrinker, line 223: Wrong toolchain

IceTDrinker (commented):
A mistake for the toolchain in benches to use avx512, I think pcc won't be green (I could be wrong)

IceTDrinker:
apparently needs a rebase as I merged an unrelated change in the Makefile

IceTDrinker:
Still some PCC errors https://github.com/zama-ai/tfhe-rs/actions/runs/3831202720/jobs/6520033070#step:4:243

I'll add notifications for those in Slack soon

soonum:
Ah this is related to your recent commit (after rebasing)

IceTDrinker:
the toolchain was updated in main a while ago but yeah the last rebase must have made them pop, sorry about that

soonum:
Should be good now.

IceTDrinker (approved):
Looks good thanks !

PR #88, opened 2023-01-23
fix(shortint): add degree management in KS-PBS

jborfila:


tfhe/src/shortint/engine/mod.rs
```diff
@@ -163,12 +165,18 @@ impl ShortintEngine {
...
+
         // This accumulator extracts the carry bits
         for i in 0..modulus_sup {
             let index = i * box_size;
             accumulator_u64[index..index + box_size]
                 .iter_mut()
                 .for_each(|a| *a = f(i as u64) * delta);
+            if f(i as u64) > max_value {
+                max_value = f(i as u64);
+            }
```
IceTDrinker, line 179: this should work ok:

```rust
max_value = max_value.max(f(i as u64))
```

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -445,6 +448,7 @@ impl ShortintEngine {
         let acc = self.generate_accumulator(server_key, |x| x % modulus)?;
 
         self.keyswitch_programmable_bootstrap_assign(server_key, ct, &acc)?;
+        println!("Message Extract Done");
```
IceTDrinker, line 451: leftover debug to remove

tfhe/src/shortint/engine/mod.rs
jborfila: I read that the max for Iterator was slower due to the Option it returns. Do you have an opinion on that?

IceTDrinker: here what I propose is not on an iterator, max_value is e.g. a u64, so max_value.max(other) is just comparing two values and updating max value, basically your piece of code but f is potentially only evaluated once instead of twice (though normally the compiler can see there is no side effect and should call f only once)

PR #89, opened 2023-01-24
chore(doc): update README twitter badge

aquint-zama:
twitter API closed to 3rd party
see https://github.com/badges/shields/issues/8837


IceTDrinker (approved):
thanks

PR #95, opened 2023-01-30
refactor(all): derive serde::{Serialize, Deserialize}

tmontaigu:
This replaces our manual implementations of serde's Serialize and Deserialize trait with 'derives'.

The manual implementetions were needed when using concrete-core but as tfhe-rs does not use concrete-core's engines we can simply derive the implementations.

Closes zama-ai/tfhe-rs-internal#16


IceTDrinker:
perhaps this https://github.com/marketplace/actions/s3-cache-for-github-actions ?

IceTDrinker:
could you add a "closes: #link_to_issue_number" in the description of your PR so that it gets closed automatically ?

We may be missing a PR template at the moment for this

IceTDrinker:
maybe there is a way to make bincode optional and only enable it for internal-keycache, c-api and wasm-api

tmontaigu:
> maybe there is a way to make bincode optional and only enable it for internal-keycache, c-api and wasm-api

Adding bincode as an optional dep and making the relevant c-api / wasm-ap feature enable bincode should work

IceTDrinker:
for C and wasm we have the __c_api and __wasm_api features

IceTDrinker:
we'll probably encounter the keycache issue again, I'll purge the cache on the S3 bucket and let this PR rebuild it

IceTDrinker (approved):
Thanks a lot ! This helps declutter our code a bit, though the discussion about magic numbers to avoid issues during deserialization may force us to re-write/prepare a macro to manage that

PR #99, opened 2023-02-03
refactor(polynomials): try plugging karatsuba

IceTDrinker:


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
Plugged the karatsuba polynomial multiplication algorithm with a small fix on the stop condition. Removed key cache as it's faster to generate than to download and check (and removes the burden of managing the key cache format + cache startegies between branches)

Reworked the shortint test script to better parallellize the tests (as we were taking a worst case scenario that did not apply to all tests)

M1 times down from ~3h to 1h20 x86 times down from ~1h to about 40 minutes

tmontaigu (approved):

PR #102, opened 2023-02-09
chore(tfhe): bump criterion version to remove outdated dep from dep tree

IceTDrinker:
closes https://github.com/zama-ai/tfhe-rs/issues/3


IceTDrinker:
https://github.com/zama-ai/tfhe-rs/actions/runs/4141977033
https://github.com/zama-ai/tfhe-rs/actions/runs/4141976137

Benches look good and grafana has gotten the data it seems so upgrading to criterion looks safe

PR #111, opened 2023-02-22
chore(ci): benchmark pbs with cost per dollar spent

soonum:
Here we benchmark a fixed number of PBS with boolean and shortint flavors on AWS EC2 instance. Once measurements are done, we compute the number of operations per dollar we can perform for a given set of cryptographic parameters and EC2 instance type. Data are then set to Slab that in turn send them to a database to be plotted in Grafana.


IceTDrinker:
failure is a known flaky test (and is in our backlog)

tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -0,0 +1,184 @@
...
+use tfhe::boolean::parameters::{BooleanParameters, DEFAULT_PARAMETERS, TFHE_LIB_PARAMETERS};
+use tfhe::core_crypto::prelude::*;
+use tfhe::shortint::parameters::*;
+use tfhe::shortint::Parameters;
+
+macro_rules! named_param {
+    ($param:ident) => {
+        (stringify!($param), $param)
+    };
+}
```
IceTDrinker, line 11: Try using the already implemented NamedParam trait from

tfhe::shortint::keycache

then you should be able to call e.g. PARAM_MESSAGE_2_CARRY_2.name()

ci/benchmark_parser.py
```diff
@@ -37,16 +39,19 @@
                     help='Check for results in subdirectories')
 parser.add_argument('--key-sizes', dest='key_sizes', action='store_true',
                     help='Parse only the results regarding keys size measurments')
+parser.add_argument('--ops-per-dollar', dest='ops_per_dollar', action='store_true',
+                    help='Compute and append number of operations per dollar')
```
IceTDrinker, line 43: I don't see the argument being used when launching the benches ?

ci/benchmark_parser.py
```diff
@@ -68,6 +73,12 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix=""):
                 test_name_parts = list(filter(None, [test_name, stat_name, name_suffix]))
                 result_values.append({"value": value, "test": "_".join(test_name_parts)})
 
+                if stat_name == "mean" and hardware_cost:
+                    test_name_parts.append("ops-per-dollar")
+                    result_values.append({"value": compute_ops_per_dollar(value, hardware_cost),
+                                          "test": "_".join(test_name_parts)})
+                    print("_".join(test_name_parts), ":", compute_ops_per_dollar(value, hardware_cost))  # DEBUG
```
IceTDrinker, line 80: I see a `# DEBUG` comment ?

ci/benchmark_parser.py
```diff
@@ -115,6 +126,18 @@ def parse_key_sizes(result_file):
...
+def compute_ops_per_dollar(data_point, product_hourly_cost):
+    """
+    Compute numbers of operations per dollar for a given ``data_point``.
+
+    :param data_point: timing value measured during benchmark
+    :param product_hourly_cost: cost in dollar per hour of hardware used
+
+    :return: number of operations per dollar
+    """
+    return ONE_HOUR_IN_NANOSECONDS / (product_hourly_cost * data_point)
```
IceTDrinker, line 138: this formula looks wrong

Let's say I have a cost = 10 $/hour

let's say data_point is the mean time to run a PBS in nanoseconds so let's say we run 1 PBS per nanosecond we have data_point = 1 PBS/ns

cost_per_nanosecond = cost / ONE_HOUR_IN_NANOSECONDS it's expressed in $/ns

and now we can compute our PBS per dollar :

data_point / cost_per_nanosecond = (PBS / ns) / ($ / ns) = (PBS * ns) / ($ * ns) = PBS/$ 

so I think your full formula is currently computing the cost of a PBS instead of the number of PBS per dollar

ci/benchmark_parser.py
```diff
@@ -177,7 +200,12 @@ def check_mandatory_args(input_args):
     raw_results = pathlib.Path(args.results)
     if not args.key_sizes:
         print("Parsing benchmark results... ")
-        results = recursive_parse(raw_results, args.walk_subdirs, args.name_suffix)
+        hardware_cost = None
+        if args.ops_per_dollar:
+            ec2_costs = json.loads(pathlib.Path("ci/ec2_products_cost.json").read_text())
```
IceTDrinker, line 205: no API call to get this information somewhere ? (storing it in a file is fine to avoid API failures, but we'll need to keep it up to date)

ci/benchmark_parser.py
```diff
@@ -68,6 +73,12 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix=""):
                 test_name_parts = list(filter(None, [test_name, stat_name, name_suffix]))
                 result_values.append({"value": value, "test": "_".join(test_name_parts)})
 
+                if stat_name == "mean" and hardware_cost:
```
IceTDrinker, line 76: `and hardware_cost is not None`

let's be explicit in python

IceTDrinker (commented):
I think there is a mistake in the formulat to compute PBS per dollar and it does not seem to be called yet.

A few other comments, other than that thanks for the PR, this will prove useful !

ci/benchmark_parser.py
soonum: I think I got it right, here's the rationale:

- criterion measure as time per benched function (thus `nanoseconds/PBS`)
- I have hourly cost for AWS (thus  `$/hour`)

Since criterion results are in nanoseconds and we want `PBS/$`, I do:

`one_hour_in_ns (ns/h) / hourly_cost ($/h) = duration_per_dollar (ns/$)`
then
`duration_per_dollar (ns/$) / data_point (ns/PBS) = pbs_per_dollar (PBS/$)`

It seems that my dimensional analysis is right. At least it yields coherent values during parsing.
From what I see, your first hypothesis is wrong a data_point is not PBS per nanosecond but rather nanoseconds per PBS.

ci/benchmark_parser.py
soonum: Indeed I could do an API call for that but as you said It would require handling of network issues. Moreover, it introduce the need to use a third-party library like boto3 to perform such call, which is not installed on the machine.
IMHO, we should stay with a hardcoded list as long it doesn't grow (too much), prices won't change drastically anyway.

ci/benchmark_parser.py
soonum: My bad.

ci/benchmark_parser.py
soonum: It is used only for PBS benchmark for now. Is it meaningful to perform such computation on boolean and shortint benchmarks also ? What do you think @jborfila ?

IceTDrinker: ah no you are right, I missed it

ci/benchmark_parser.py
IceTDrinker: you are right 😵 I completely mixed the meaning of data point

tfhe/benches/core_crypto/pbs_bench.rs
soonum: Sadly no since parameters with `.*_CARRY_0` have no parameter name and thus panic when we call `name()` from `NamedParam` trait.

IceTDrinker: we can add the implementation for the parameters missing the impl

soonum: Resolved by adding the parameters used in the corresponding macro.

ci/benchmark_parser.py
```diff
@@ -68,6 +76,18 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix=""):
                 test_name_parts = list(filter(None, [test_name, stat_name, name_suffix]))
                 result_values.append({"value": value, "test": "_".join(test_name_parts)})
 
+                if stat_name == "mean" and throughput is True:
```
IceTDrinker, line 79: just `throughput` it's a boolean :) or to be a bit more explicit rename it to "compute_throughput"

ci/benchmark_parser.py
```diff
@@ -177,7 +220,14 @@ def check_mandatory_args(input_args):
     raw_results = pathlib.Path(args.results)
     if not args.key_sizes:
         print("Parsing benchmark results... ")
-        results = recursive_parse(raw_results, args.walk_subdirs, args.name_suffix)
+        HARDWARE_COST = None
```
IceTDrinker, line 223: why use HARDWARE_COST in caps ? it has kind of a special meaning in python

IceTDrinker (commented):
Two small things but I think we should be good

ci/benchmark_parser.py
soonum: Apparently, the way I use it seems to looks like a Constant for pylint. Should I ignore the warning sent by the linter ?
I've to admit, this warning makes no sense to me though

IceTDrinker: avoid using all caps if it's not supposed to be a runtime constant

ci/benchmark_parser.py
```diff
@@ -68,6 +77,18 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix=""):
                 test_name_parts = list(filter(None, [test_name, stat_name, name_suffix]))
                 result_values.append({"value": value, "test": "_".join(test_name_parts)})
 
+                if stat_name == "mean" and compute_throughput is True:
```
IceTDrinker, line 80: no need to explicitely check "is True" for a boolean

IceTDrinker (approved):
Looks good thanks a lot !

IceTDrinker:
still that flaky crt lut generation test 🤔

PR #112, opened 2023-02-24
Compute throughput on boolean and shortint benchmarks

soonum:
This commit series makes `--throughput` CLI argument for benchmark result parser optional.
In addition an AWS Slab profile for PBS benchmarks had been added. It couldn't run without it... :sweat_smile:


IceTDrinker (approved):
Approved !

PR #121, opened 2023-02-28
chore(thfe): add integer workflow to make it availble for slab-ci

IceTDrinker:


soonum (approved):
You're good to go ! Thanks.

PR #124, opened 2023-03-02
fix(shortint): use correct value to shift msg in bivariate bps

tmontaigu:
fixes zama-ai/tfhe-rs-internal#32


IceTDrinker:
should we go ahead with this review ?

tmontaigu:
Hm, I would rather propose as a PR the smaller change, and later do a refactor of the function_bivariate functions

IceTDrinker:
ok !

tmontaigu:
closed as #141 is a better fix

Issue #115, opened 2023-02-27
Unexpected result from keyswitch_programmable_bootstrap_bivariate


okuyiga:
**To Reproduce**
https://gist.github.com/okuyiga/a8528d96a500ef079f98245cfad7876e

**Expected behaviour**
`is_functional_bivariate_pbs_possible` should provide reliable data on when `keyswitch_programmable_bootstrap_bivariate` can be safely used

**Evidence**
See snippet for reproduction

IceTDrinker:
Thanks for the report, we'll be checking this on our end, but it seems that there is something a bit odd going on indeed

IceTDrinker:
this should have been closed by https://github.com/zama-ai/tfhe-rs/pull/141 if you are still experiencing issues don't hesitate to open a new issue 🙂

PR #131, opened 2023-03-10
fix(shortint): fix comparisons on trivial ciphertexts

tmontaigu:


tfhe/src/shortint/engine/server_side/comp_op.rs
```diff
@@ -311,9 +306,8 @@ impl ShortintEngine {
     ) -> EngineResult<()> {
         let modulus = (ct_right.degree.0 + 1) as u64;
         let modulus_msg = ct_left.message_modulus.0 as u64;
```
IceTDrinker, line 308: as discussed here I think it makes sense to rename modulus -> modulus_right and modulus_msg -> modulus_left

IceTDrinker: same for other occurrences

IceTDrinker (approved):
Thanks ! Looks good can be worth checking if #115 is seeing improvements with this fix

tmontaigu:
> Thanks ! Looks good can be worth checking if #115 is seeing improvements with this fix

I don't think it will fix it, the problem in #115, is I think what I described elsewhere: we have incoherences in how we "shift" the lhs in the case of bivariate pbs, some places do a shift by message_modulus, others by degree + 1 and if we are in a case where `degree + 1 != message_modulus` things will go bad.

tmontaigu:
Yes the fix to this is larger than what is in this PR

tmontaigu:
```Rust
    fn generate_accumulator_bivariate_with_engine<F>(
        server_key: &ServerKey,
        f: F,
    ) -> EngineResult<Accumulator>
    where
        F: Fn(u64, u64) -> u64,
    {
        let modulus = server_key.message_modulus.0 as u64;
            let lhs = (input / modulus) % modulus;
            let rhs = input % modulus;

            f(lhs, rhs)
        };
        ShortintEngine::generate_accumulator_with_engine(server_key, wrapped_f)
    }
```

```Rust
    /// Verify if a bivariate functional pbs can be applied on ct_left and ct_right.
    pub fn is_functional_bivariate_pbs_possible(&self, ct1: &Ciphertext, ct2: &Ciphertext) -> bool {
        //product of the degree
        let final_degree = ct1.degree.0 * (ct2.degree.0 + 1) + ct2.degree.0;
        final_degree < ct1.carry_modulus.0 * ct1.message_modulus.0
    }
```

```Rust

    pub(crate) fn keyswitch_programmable_bootstrap_bivariate_assign(
        &mut self,
        server_key: &ServerKey,
        ct_left: &mut Ciphertext,
        ct_right: &Ciphertext,
        acc: &Accumulator,
    ) -> EngineResult<()> {
        let modulus = (ct_right.degree.0 + 1) as u64;

        // Message 1 is shifted to the carry bits
        self.unchecked_scalar_mul_assign(ct_left, modulus as u8)?;

        // Message 2 is placed in the message bits
        self.unchecked_add_assign(ct_left, ct_right)?;

        // Compute the PBS
        self.keyswitch_programmable_bootstrap_assign(server_key, ct_left, acc)?;

        Ok(())
    }

```

The 3 function above have problem,  `generate_accumulator_bivariate_with_engine` shift by message_modulus,
whereas `keyswitch_programmable_bootstrap_bivariate_assign` shift by degree + 1 and `is_functional_bivariate_pbs_possible` checks that by shifting by degree + 1, we don't lose information.

`generate_accumulator_bivariate_with_engine` would need an input degree to correctly shift, but that also means that 
the generated accumulator only works for this particular degree thus, is very limited in terms of usage

tmontaigu:
changing `generate_accumulator_bivariate_with_engine`  to
(adding as input degree of ct_rights)

```rust

    fn generate_accumulator_bivariate_with_engine<F>(
        server_key: &ServerKey,
        f: F,
        ct_right_degree: Degree,
    ) -> EngineResult<Accumulator>
    where
        F: Fn(u64, u64) -> u64,
    {
        let ct_right_modulus = ct_right_degree.0 as u64 + 1;
        let message_modulus = server_key.message_modulus.0 as u64;
            let lhs = (input / ct_right_modulus) % message_modulus;
            let rhs = input % ct_right_modulus;

            f(lhs, rhs)
        };
        ShortintEngine::generate_accumulator_with_engine(server_key, wrapped_f)
    }
```

Fixes #115 

But that means that this generated accumulator is only valid on invocations of `server_key.keyswitch_programmable_bootstrap_bivariate_assign(&mut res, ct_right, &acc);` where `ct_right.degree == generate_accumulator_bivariate_with_engine::ct_right_degree` .

Which makes me worry that it becomes very easy to misuse, and we may also be losing the 'advantage' of precaculating the accumulator, since the accumulator is more specific.

And so, it would be better to use `unchecked_functional_bivariate_pbs` however, as this 'bivariate' function takes a univariate closure, it is the user who has the responsibility to use the correct shift, however, I don't think  the user can know the correct value unless he looks at the internal code

IceTDrinker:
good to merge ?

tmontaigu:
Yes

PR #133, opened 2023-03-10
refactor(wopbs): manage LUTs for wopbs to avoid copies

IceTDrinker:


tfhe/src/integer/wopbs/mod.rs
```diff
@@ -22,6 +24,53 @@ pub struct WopbsKey {
...
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl std::ops::DerefMut for IntegerWopbsLUT {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
```
tmontaigu, line 72: Its probably fine for this but this kind of use of Deref is generally seen as an anti-pattern https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

IceTDrinker: ok got it, I prefer to avoid known anti patterns, which shape should it take, via a trait providing access to the underlying struct ?

tmontaigu (approved):

PR #139, opened 2023-03-13
chore(makefile): add --all-targets switch to build command

tmontaigu:
This adds the --all-targets to the cargo build commands invoked by the makefile so that when running
`make build_boolean`, lib, tests, benches, examples are built.

See the `cargo help build`

```
--all-targets
    Build all targets. This is equivalent to specifying --lib --bins --tests --benches --examples.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,27 +117,27 @@ gen_key_cache: install_rs_build_toolchain
 .PHONY: build_boolean # Build with boolean enabled
 build_boolean: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) build --release \
-		--features=$(TARGET_ARCH_FEATURE),boolean -p tfhe
+		--features=$(TARGET_ARCH_FEATURE),boolean -p tfhe --all-targets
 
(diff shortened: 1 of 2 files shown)


IceTDrinker (approved):
thanks !

IceTDrinker:
we have a build all targets thing I think already

https://github.com/zama-ai/tfhe-rs/blob/385c90780707f4ea37c1abc9adacb161e728c9b9/Makefile#L102 isn't it enough ?

tmontaigu:
Well, since the clippy_all_targets builds with all features enabled you can't catch if you set / forget a `#[cfg]` on some code to make it compile when the corresponding feature is enabled

IceTDrinker:
Understood, thanks !

aquint-zama:
@cla-bot check

IceTDrinker:
ready for review ?

tmontaigu:
yes i put it bact to a WIP PR to assert if it was really useful, and came across a case where I modified a test in typed_api/integers an ran `make build_integer` but no error was reported until a ran make clippy_all_targets, so it has some usefulness but not a lot

IceTDrinker:
I would say good to merge ?

PR #141, opened 2023-03-14
fix issue #115

tmontaigu:
A bivariate PBS is a univariate PBS where we encode
the lhs, and rhs values into a singular value:
`univariate_value = (lhs * shift) + rhs`

Some places shifted the lhs by the parameter's message modulus
while others shifted by rhs.degree + 1, this could leed to incoherences
and wrong result in some cases.

The commits adds a `BivariateAccumulator` that stores the shift
value that was used to create the LUT, to avoid said incoherences.

Also, bivariate function family that expected
a univariate closure `Fn(u64) -> u64` will now expect a
bivariate closure `Fn(u64, u64) -> u64` so that they are less
error prone as the user does not need to figure out the
shift to be used.

### DIFF
--- a/tfhe/src/shortint/engine/mod.rs
+++ b/tfhe/src/shortint/engine/mod.rs
@@ -17,6 +17,9 @@ use crate::shortint::ServerKey;
 use std::cell::RefCell;
 use std::fmt::Debug;
 
+use super::parameters::MessageModulus;
+use super::server_key::BivariateAccumulator;
+
 mod client_side;
 mod public_side;
 mod server_side;
@@ -236,21 +239,32 @@ impl ShortintEngine {
         })
     }
 
+    /// Generates a bivariate accumulator
     fn generate_accumulator_bivariate_with_engine<F>(
         server_key: &ServerKey,
         f: F,
-    ) -> EngineResult<Accumulator>
+        left_message_scaling: MessageModulus,
+    ) -> EngineResult<BivariateAccumulator>
     where
         F: Fn(u64, u64) -> u64,
     {
-        let modulus = server_key.message_modulus.0 as u64;
+        // Depending on the factor used, rhs and / or lhs may have carries
+        // (degree >= message_modulus) which is why we need to apply the message_modulus
+        // to clear them
+        let factor_u64 = left_message_scaling.0 as u64;
+        let message_modulus = server_key.message_modulus.0 as u64;
         let wrapped_f = |input: u64| -> u64 {
-            let lhs = (input / modulus) % modulus;
-            let rhs = input % modulus;
+            let lhs = (input / factor_u64) % message_modulus;
+            let rhs = (input % factor_u64) % message_modulus;
 
             f(lhs, rhs)
         };
-        ShortintEngine::generate_accumulator_with_engine(server_key, wrapped_f)
+        let accumulator = ShortintEngine::generate_accumulator_with_engine(server_key, wrapped_f)?;
+
+        Ok(BivariateAccumulator {
+            acc: accumulator,
+            ct_right_modulus: left_message_scaling,
+        })
     }
 
     /// Return the [`BuffersRef`] and [`ComputationBuffers`] for the given `ServerKey`
--- a/tfhe/src/shortint/server_key/mod.rs
+++ b/tfhe/src/shortint/server_key/mod.rs
@@ -68,9 +68,36 @@ pub struct ServerKey {
     pub max_degree: MaxDegree,
 }
 
+/// Returns whether it is possible to pack lhs and rhs into a unique
+/// ciphertext without exceeding the max storable value using the formula:
+/// `unique_ciphertext = (lhs * factor) + rhs`
+fn ciphertexts_can_be_packed_without_exceeding_space(
+    lhs: &Ciphertext,
+    rhs: &Ciphertext,
+    factor: usize,
+) -> bool {
+    let final_degree = (lhs.degree.0 * factor) + rhs.degree.0;
+    final_degree < lhs.carry_modulus.0 * lhs.message_modulus.0
+}
+
 pub struct Accumulator {
     pub acc: GlweCiphertextOwned<u64>,
     pub degree: Degree,
+}
+
+pub struct BivariateAccumulator {
+    // A bivariate accumulator is an univariate accumulator
+    // where the message space is shared to encode
+    // 2 values
+    pub acc: Accumulator,
+    // By how much we shift the lhs in the LUT
+    pub ct_right_modulus: MessageModulus,
+}
+
+impl BivariateAccumulator {
+    pub fn is_bivariate_pbs_possible(&self, lhs: &Ciphertext, rhs: &Ciphertext) -> bool {
+        ciphertexts_can_be_packed_without_exceeding_space(lhs, rhs, self.ct_right_modulus.0)
+    }
 }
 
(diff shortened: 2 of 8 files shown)


tfhe/src/shortint/server_key/mod.rs
```diff
@@ -73,6 +73,15 @@ pub struct Accumulator {
     pub degree: Degree,
 }
 
+pub struct BivariateAccumulator {
+    // A bivaraite accumulator is a univariate accumulator
```
tmontaigu, line 77: typo `bivariate`

tfhe/src/shortint/engine/mod.rs
```diff
@@ -239,18 +242,25 @@ impl ShortintEngine {
     fn generate_accumulator_bivariate_with_engine<F>(
         server_key: &ServerKey,
         f: F,
-    ) -> EngineResult<Accumulator>
+        factor: MessageModulus,
```
IceTDrinker, line 245: is there a way to have a better name for this ? I'm guessing it's linked to the various usage in the code base

tfhe/src/shortint/engine/mod.rs
```diff
@@ -239,18 +242,25 @@ impl ShortintEngine {
...
         F: Fn(u64, u64) -> u64,
     {
-        let modulus = server_key.message_modulus.0 as u64;
+        let factor_u64 = factor.0 as u64;
+        let message_modulus = server_key.message_modulus.0 as u64;
         let wrapped_f = |input: u64| -> u64 {
-            let lhs = (input / modulus) % modulus;
-            let rhs = input % modulus;
+            let lhs = (input / factor_u64) % message_modulus;
+            let rhs = (input % factor_u64) % message_modulus;
```
IceTDrinker, line 254: I feel like this may not work if factor_u64 > message modulus ? As the modulo message_modulus is going to remove some data from the input % factor_u64 or like in the original code maybe it's not needed ?

Does it mean we need some asserts ?

IceTDrinker (commented):
Looks good, thanks for the change, maybe some documentation could be welcome on the generation function and how to use the "factor" term.

I mentionned an assert somewhere, looks like you put it in a function a bit higher

soonum:
@slab-ci cpu_test

tfhe/src/shortint/engine/mod.rs
tmontaigu: No it is needed:

Here's me trying to hopefully explain it without being wrong 🤔 

# base

For bivariate pbs, we pack the messages into one with the formula

`v = (lhs * f) + rhs` 
with:
  * ` v < (Carry Modulus * Message Modulus),`
 *  ` rhs < f`


To get back our lhs and rhs from v we do:

lhs:
`v / f = ((lhs * f) / f) + (rhs / f)`
since we have `rhs < f` and we are on unsigned integers: `rhs /f = 0`
`v / f = (lhs * f / f) + (0)`
`v / f = lhs`


rhs:

`v % f = ((lhs * f) + rhs) % f`
`v % f = rhs`

---

Lets say we are doing a call to functionnal bivariate bps,
so we give bivariate function, and in 'one go' it will internally 
generate a LUT and apply it via PBS.

And we are using some param_message_x_carry_y with y >= x.


In that scenario, the factor f that is going to be used is `rhs.degree + 1`

The precondition (is_functional_bivariate_pbs_possible) is:
`(lhs.degree * (rhs.degree + 1)) + rhs.degree < 2^(x+y)`
i.e: can we do `v = (lhs * (rhs.degree + 1)) * rhs` without overflowing and losing information

## Case: the precondition is true:

Then here we don't know if `rhs.degree + 1` is inferior or superior to 2^x the message modulus
ie are their non zero carry bits in either lhs or rhs (or both potentially) ? 

If rhs has some carry bits our factor `f` (`rhs.degree + 1`) we have `f > 2^x` the message modulus

meaning that when doing `v % f = rhs` we have `0 <= rhs < f`
but since `f > message_modulus` when applying the users function we need to apply the message modulus
(ie clean carries) otherwise this will lead to incorrect result:

## Example:

Lets consider some clean and non clean ciphertexts as input for a bivariate pbs:

Pamameters: param_message_1_carry_3,
lhs = 1 % 2^1 = 1 => 0b000|1
rhs = 2 % 2^1 = 0 => 0b001|0

rhs.degree = 3 => f = rhs.degree + 1 = 4

v = (lhs * f) + rhs = (1 * 4) + 2 = 6
v <= 2^4
lhs = v / f = 6 / 4 = 1
rhs = v % f = 6 % 4 = 2

now say the bivariate function is `<` if we dont apply the message_mod
lhs < rhs ?
-> 1 < 2
-> true

if we apply modulo:
(lhs & message_mod) < (rhs % message_mod) ? 
-> (1 * 2) < (2 % 2)
-> 1 < 0
-> false

And since param_message_1_carry_3 means 1 bit of message,
semantically, the message stays on one bit so we have to apply the modulus to
stay correct.

(same idea if lhs has some carry bits)

## Case: the precondition is false

If the precondition is false, we do a clean of the carries (message_extract)
on both of the ciphertext, which will make it so that our 2 ciphertexts will
have a degree of message_mod - 1, thus we can safely do the computations described above
is just that in that case `f = rhs.degree + 1 = message_mod -1 + 1` => `f = message_mod`.


And in fact the whole idea of using `rhs.degree + 1` as the factor `f` instead
of `message_mod` as I think we had in earlier version of concrete-shortint, is that we can avoid
doing some pbses that are just meant to clean some carries if we can afford to clean them durring
the bivariate pbs.

IceTDrinker: Ok, I think I was missing the part where the degree > message_modulus in my reasoning, thanks for the detailed answer

tmontaigu:
I do feel like this PR is a bit incomplete, i think it will still potentialle be easy for users to make mistake when generating an accumulator and reusing it.

I think ideally we should have a `smart_functionnal_bivariate_pbs` function, the `BivariateAccumulator` type should have a
`is_bivariate_bps_possible` method as well

tfhe/src/shortint/engine/mod.rs
tmontaigu: Yeah i'm not a huge fan of this name but I don't have another proposition for now

IceTDrinker:
> I do feel like this PR is a bit incomplete, i think it will still potentially be easy for users to make mistake when generating an accumulator and reusing it with a ciphertext for which the degree does not fit will with the accumulator
> 
> I think ideally we should have a `smart_functionnal_bivariate_pbs` function and it should be the preffered advertised way of doing a bivarPBS, the `BivariateAccumulator` type should have a `is_bivariate_bps_possible` method as well

I was facing something a bit similar for the wopbs lut management : https://github.com/zama-ai/tfhe-rs/pull/133 if you want some potential inspiration

IceTDrinker:
Notably how the ShortintWopbsLut is a wrapper for the base but populating it to match the invariants of shortint functions, not sure it's a good idea to have several lut types though for the bivariate PBS in that case

aquint-zama:
@cla-bot check

tfhe/src/shortint/engine/mod.rs
IceTDrinker: left_message_scaling ? as it's effectively by how much we scale the left message to put it in the MSB/carry part of the ciphertext ?

PR #150, opened 2023-03-17
fix(integer): stop decomposing before overflow

soonum:
This only happens on binary scalar operations over 64bits of precision.

### DIFF
--- a/tfhe/src/integer/server_key/radix/tests.rs
+++ b/tfhe/src/integer/server_key/radix/tests.rs
@@ -45,6 +45,8 @@ create_parametrized_test!(integer_smart_scalar_add);
 create_parametrized_test!(integer_unchecked_scalar_sub);
 create_parametrized_test!(integer_unchecked_scalar_add);
 
+create_parametrized_test!(integer_unchecked_scalar_decomposition_overflow);
+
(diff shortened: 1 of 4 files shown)


IceTDrinker (commented):
The change don't make much sense to me at the moment, can we add tests that failed before and now pass please ?

soonum:
> The change don't make much sense to me at the moment (as I don't know the algorithms well), can we add tests that failed before and now pass please ?

I'll try to generate such tests.

tmontaigu (approved):

tfhe/src/integer/server_key/radix/tests.rs
```diff
@@ -1129,3 +1131,65 @@ fn integer_smart_scalar_sub(param: Parameters) {
         }
     }
 }
+
+fn integer_unchecked_scalar_decomposition_overflow(param: Parameters) {
```
IceTDrinker, line 1135: no test for scalar_sub that was also fixed ?

soonum: Yes it's tested starting at line 1159.

IceTDrinker: right thanks !

PR #151, opened 2023-03-17
chore(ci): CI at the speed of ligth

IceTDrinker:
- use a 128 vcpu instance
- update script to have a no compromise test run
- update Makefile to be able to run the "no compromise" CI mode

### DIFF
--- a/scripts/integer-tests.sh
+++ b/scripts/integer-tests.sh
@@ -24,12 +24,13 @@ else
     n_threads=6
 fi
 
-# block pbs are too slow for high params
-# mul_crt_4_4 is extremely flaky (~80% failure)
-# test_wopbs_bivariate_crt_wopbs_param_message generate tables that are too big at the moment
-# same for test_wopbs_bivariate_radix_wopbs_param_message_4_carry_4
-# test_integer_smart_mul_param_message_4_carry_4 is too slow
-filter_expression=''\
+if [[ "${BIG_TESTS_INSTANCE}" != TRUE ]]; then
+    # block pbs are too slow for high params
+    # mul_crt_4_4 is extremely flaky (~80% failure)
+    # test_wopbs_bivariate_crt_wopbs_param_message generate tables that are too big at the moment
+    # same for test_wopbs_bivariate_radix_wopbs_param_message_4_carry_4
+    # test_integer_smart_mul_param_message_4_carry_4 is too slow
+    filter_expression=''\
 'test(/^integer::.*$/)'\
 'and not test(/.*_block_pbs(_base)?_param_message_[34]_carry_[34]$/)'\
 'and not test(~mul_crt_param_message_4_carry_4)'\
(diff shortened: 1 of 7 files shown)


scripts/integer-tests.sh
```diff
@@ -24,35 +24,64 @@ else
     n_threads=6
 fi
 
-# block pbs are too slow for high params
-# mul_crt_4_4 is extremely flaky (~80% failure)
-# test_wopbs_bivariate_crt_wopbs_param_message generate tables that are too big at the moment
-# same for test_wopbs_bivariate_radix_wopbs_param_message_4_carry_4
-# test_integer_smart_mul_param_message_4_carry_4 is too slow
-filter_expression=''\
+if [[ "${BIG_TESTS_INSTANCE}" == FALSE ]]; then
```
IceTDrinker, line 27: this needs to be `!= TRUE` otherwise just launching the script outside the makefile will fail

scripts/shortint-tests.sh
```diff
@@ -30,7 +30,8 @@ else
     n_threads_big=13
 fi
 
-filter_expression_small_params=''\
+if [[ "${BIG_TESTS_INSTANCE}" == FALSE ]]; then
```
IceTDrinker, line 33: same thing here

scripts/integer-tests.sh
```diff
@@ -24,35 +24,64 @@ else
...
+    # block pbs are too slow for high params
```
soonum, line 28: The filter expression block seems to be duplicated in both conditional statements. Can't we get it out of the if statement and declare it once ?

scripts/integer-tests.sh
```diff
@@ -24,35 +24,64 @@ else
...
+    # test_integer_smart_mul_param_message_4_carry_4 is too slow
+    filter_expression=''\
 'test(/^integer::.*$/)'\
 'and not test(/.*_block_pbs(_base)?_param_message_[34]_carry_[34]$/)'\
 'and not test(~mul_crt_param_message_4_carry_4)'\
 'and not test(/.*test_wopbs_bivariate_crt_wopbs_param_message_[34]_carry_[34]$/)'\
 'and not test(/.*test_wopbs_bivariate_radix_wopbs_param_message_4_carry_4$/)'\
 'and not test(/.*test_integer_smart_mul_param_message_4_carry_4$/)'
 
-export RUSTFLAGS="-C target-cpu=native"
```
soonum, line 40: No more RUSTFLAGS is there a reason behind it ?

soonum (commented):
Just need few details to be explained and then we should be good to go.

scripts/integer-tests.sh
IceTDrinker: If we put the RUSTFLAGS here the Makefile may not be able to choose the RUSTFLAGS for the script, I think initially I was thinking about being able to run the script as a standalone, now that most things go through the Makefile I don't think this makes sense anymore

scripts/integer-tests.sh
IceTDrinker: they are not exactly the same, the BIT_TESTS_INSTANCE == FALSE case runs the 4_4 tests separately while the TRUE case just runs everything at once

scripts/integer-tests.sh
soonum: Thanks for the explanation.

soonum (approved):

tmontaigu (approved):

PR #154, opened 2023-03-22
feat(core): add std multi-bit bootstrapping

IceTDrinker:
An alternative way of generating the key bits for the multi-bit bootstrapping, computations are made in the standard domain meaning the resulting ciphertext contains less noise, but is more costly to compute, it is interesting to have that trade-off to better manage noise even if it means using more thread to keep the consumer thread saturated

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
@@ -1,5 +1,6 @@
 use crate::core_crypto::algorithms::extract_lwe_sample_from_glwe_ciphertext;
 use crate::core_crypto::algorithms::polynomial_algorithms::*;
+use crate::core_crypto::algorithms::slice_algorithms::*;
 use crate::core_crypto::commons::computation_buffers::ComputationBuffers;
 use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
 use crate::core_crypto::commons::parameters::*;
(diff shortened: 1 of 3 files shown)


IceTDrinker:
@slab-ci cpu_test

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -259,6 +259,108 @@ pub fn polynomial_wrapping_monic_monomial_mul_assign<Scalar, OutputCont>(
...
+    monomial_degree: MonomialDegree,
+) where
+    Scalar: UnsignedInteger,
+    OutputCont: ContainerMut<Element = Scalar>,
+    InputCont: Container<Element = Scalar>,
+{
+    let polynomial_len = output.container_len();
+    let remaining_degree = monomial_degree.0 % output.as_ref().container_len();
+
+    for (dst, &src) in output[..].iter_mut().zip(input[remaining_degree..].iter()) {
```
tmontaigu, line 293: wouldn't `output[..].copy_from_slice(input[remaining_degree..])` be applicable here ?

IceTDrinker: fairly certain it should be applicable, I either copied this from the wrapping mul or I just derped when writing it

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -9,11 +10,89 @@ use crate::core_crypto::fft_impl::common::pbs_modulus_switch;
 use crate::core_crypto::fft_impl::fft64::crypto::ggsw::{
     add_external_product_assign, add_external_product_assign_scratch, update_with_fmadd,
 };
-use crate::core_crypto::fft_impl::fft64::math::fft::Fft;
+use crate::core_crypto::fft_impl::fft64::math::fft::{Fft, FftView};
 use concrete_fft::c64;
 use std::sync::{mpsc, Condvar, Mutex};
 use std::thread;
 
+pub fn prepare_multi_bit_ggsw_mem_optimized<
```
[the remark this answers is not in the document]
IceTDrinker: well the mem optimized part indicates that we pass a properly sized ComputationBuffers, I have to say this PR is a bit old I'll need to re read the code a bit

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
IceTDrinker: not applicable directly as the two slices need to have the same size, here I believe I'm "cheating" using the fact the iterator will end on the shortest slice

IceTDrinker:
unless there is something blocking, I'll go ahead and merge once tests are green

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
IceTDrinker: updated to measure the slice and use copy from slice which should be better than iterating

PR #156, opened 2023-03-23
chore(ci): record benchmarks parameters to be stored in database

soonum:
This is done to comply with the new Zama benchmark standard. Exhaustive parameters list is stored so once it's parsed and send to database, one can easily filter results on such parameters in visualization tool.

### DIFF
--- a/tfhe/benches/core_crypto/pbs_bench.rs
+++ b/tfhe/benches/core_crypto/pbs_bench.rs
@@ -1,3 +1,7 @@
+#[path = "../utilities.rs"]
+mod utilities;
+use crate::utilities::{write_to_json, CryptoParametersRecord, OperatorType};
+
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use tfhe::boolean::parameters::{BooleanParameters, DEFAULT_PARAMETERS, TFHE_LIB_PARAMETERS};
 use tfhe::core_crypto::prelude::*;
@@ -24,39 +28,8 @@ const SHORTINT_BENCH_PARAMS: [Parameters; 15] = [
 ];
 
 const BOOLEAN_BENCH_PARAMS: [(&str, BooleanParameters); 2] = [
-    ("boolean_default_params", DEFAULT_PARAMETERS),
-    ("boolean_tfhe_lib_params", TFHE_LIB_PARAMETERS),
-];
-
-const MULTI_BIT_PARAMS: [(&str, (BenchmarkPbsParameters, LweBskGroupingFactor)); 2] = [
-    (
-        "4_bits_multi_bit_group_2",
-        (
-            BenchmarkPbsParameters {
-                input_lwe_dimension: LweDimension(724),
-                lwe_modular_std_dev: StandardDev(1.58705e-10),
-                decomp_base_log: DecompositionBaseLog(1),
-                decomp_level_count: DecompositionLevelCount(18),
-                glwe_dimension: GlweDimension(3),
-                polynomial_size: PolynomialSize(512),
-            },
-            LweBskGroupingFactor(2),
-        ),
-    ),
-    (
-        "4_bits_multi_bit_group_3",
-        (
-            BenchmarkPbsParameters {
-                input_lwe_dimension: LweDimension(732),
-                lwe_modular_std_dev: StandardDev(1.18161e-10),
-                decomp_base_log: DecompositionBaseLog(1),
-                decomp_level_count: DecompositionLevelCount(18),
-                glwe_dimension: GlweDimension(3),
-                polynomial_size: PolynomialSize(512),
-            },
-            LweBskGroupingFactor(3),
-        ),
-    ),
+    ("BOOLEAN_DEFAULT_PARAMS", DEFAULT_PARAMETERS),
+    ("BOOLEAN_TFHE_LIB_PARAMS", TFHE_LIB_PARAMETERS),
 ];
 
 criterion_group!(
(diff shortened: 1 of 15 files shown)


soonum:
Guys I need your review even if we are in draft phase since next step is pretty much modify the code in a somewhat copy/paste fashion to the rest of benched functions.

IceTDrinker (commented):
We saw together how to fix the pcc stuff

IceTDrinker (commented):
Looks generally ok, I don't know about the python script like this, lots of changes and the input data structure is not clear to me

tmontaigu:
Id also say it looks ok, maybe if the only purpose of the CryptoRecordParameters is to be created then dumped to json
```rust
        let mut params_record =
            utilities::BenchmarkParametersRecord::from_params(param, display_name);
        params_record.write(&bench_id);
```

then I would make the utilities module only expose one function
```rust
pub fn write_as_json<T: Into<CryptoRecordParameters>>(params: T, bench_id: &str, display_name) { ... } 
```

soonum:
I like the idea very much ! I've just one question what about the future benchmarks (integer to be added, as well as CRT decomposition) that requires a different BenchmarkRecordParameters ? How can I change the default values sets in the associated function `from_params()` ?

soonum:
Now I think this is done for tfhe-rs part. I need to implement handling in Slab side and then test it, that why it's not ready to merge.
Otherwise it will simply break the benchmarks.

tfhe/benches/utilities.rs
```diff
@@ -0,0 +1,171 @@
+use serde::Serialize;
+use std::fs;
+use std::path::PathBuf;
+#[cfg(feature = "boolean")]
+use tfhe::boolean::parameters::BooleanParameters;
+use tfhe::core_crypto::prelude::*;
+// #[cfg(feature = "shortint")]
+// use tfhe::shortint::keycache::NamedParam;
```
IceTDrinker, line 8: comment

tfhe/benches/utilities.rs
```diff
@@ -0,0 +1,171 @@
...
+
+#[derive(Serialize)]
+struct BenchmarkParametersRecord {
+    display_name: String,
+    crypto_parameters_alias: String,
+    crypto_parameters: CryptoParametersRecord,
+    message_modulus: Option<usize>,
+    carry_modulus: Option<usize>,
+    ciphertext_modulus: usize,
+    polynomial_multiplication: String, // TODO Can we do an Enum ?
```
IceTDrinker, line 113: we can likely do an enum yes

.github/workflows/shortint_benchmark.yml
```diff
@@ -123,13 +123,15 @@ jobs:
...
+          echo "Computing HMac on results file"
+          SIGNATURE="$(slab/scripts/hmac_calculator.sh ${{ env.RESULTS_FILENAME }}.gz '${{ secrets.JOB_SECRET }}')"
           echo "Sending results to Slab..."
           curl -v -k \
           -H "Content-Type: application/json" \
           -H "X-Slab-Repository: ${{ github.repository }}" \
           -H "X-Slab-Command: store_data" \
           -H "X-Hub-Signature-256: sha256=${SIGNATURE}" \
-          -d @${{ env.RESULTS_FILENAME }} \
+          -d @${{ env.RESULTS_FILENAME }}.gz \
```
IceTDrinker, line 136: could maybe have a variable containing the name with the .gz, would avoid errors

IceTDrinker (approved):
Looks good to my tired eyes 😪

tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -92,7 +62,51 @@ fn benchmark_parameters<Scalar: Numeric>() -> Vec<(String, BenchmarkPbsParameter
     }
 }
 
+fn multi_bit_benchmark_parameters<Scalar: Numeric>(
+) -> Vec<(String, (CryptoParametersRecord, LweBskGroupingFactor))> {
+    if Scalar::BITS == 64 {
+        vec![
+            (
+                "4_bits_multi_bit_group_2".to_string(),
```
IceTDrinker, line 70: still unsure about this, do you want to add the multi bit in the same PR ?

tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -92,7 +62,51 @@ fn benchmark_parameters<Scalar: Numeric>() -> Vec<(String, BenchmarkPbsParameter
...
+                (
+                    CryptoParametersRecord {
+                        lwe_dimension: Some(LweDimension(724)),
+                        lwe_modular_std_dev: Some(StandardDev(1.58705e-10)),
+                        pbs_base_log: Some(DecompositionBaseLog(1)),
+                        pbs_level: Some(DecompositionLevelCount(18)),
```
IceTDrinker, line 76: params are wrong, it's my bad

tfhe/benches/core_crypto/pbs_bench.rs
soonum: Yes we'll keep your initial commit so we can get two birds with one stone.

tfhe/benches/integer/bench.rs
```diff
@@ -19,15 +23,16 @@ use tfhe::shortint::parameters::{
...
 impl Default for ParamsAndNumBlocksIter {
     fn default() -> Self {
-        const PARAMS: [tfhe::shortint::Parameters; 3] = [
+        // FIXME One set of parameter is tested since we want to benchmark only quickest operations.
+        const PARAMS: [tfhe::shortint::Parameters; 1] = [
             PARAM_MESSAGE_2_CARRY_2,
-            PARAM_MESSAGE_3_CARRY_3,
-            PARAM_MESSAGE_4_CARRY_4,
+            // PARAM_MESSAGE_3_CARRY_3,
+            // PARAM_MESSAGE_4_CARRY_4,
```
IceTDrinker, line 35: remove comments ?

tfhe/benches/integer/bench.rs
```diff
@@ -396,12 +482,33 @@ criterion_group!(
...
-    unchecked_arithmetic_operation,
-    unchecked_scalar_arithmetic_operation,
-    misc,
+    fast_integer_benchmarks,
+    // smart_arithmetic_operation,
+    // smart_arithmetic_parallelized_operation,
+    // smart_scalar_arithmetic_operation,
+    // smart_scalar_arithmetic_parallel_operation,
+    // unchecked_arithmetic_operation,
+    // unchecked_scalar_arithmetic_operation,
```
IceTDrinker, line 512: comments to remove ?

PR #162, opened 2023-03-27
chore(core_crypto): enable the choice of a single fixed fft algorithm

IceTDrinker:


tmontaigu:
don't we also need the feature to appear in the cargo.toml ?

tmontaigu:
no its not optional, I tried 

```
default = ["experimental-force_fft_algo_dif4"]
```

and I get 

image


And it seems logical, how would cargo know about this feature ? it would need to parse the source code

IceTDrinker:
Ok then it's my bad, passing it as a feature in the command line does not work either ?

tmontaigu:
Same problem 
image

The only way would be to pass the feature as rustflag via the RUSTFLAG env var

IceTDrinker:
Ok, I'll add it to the Cargo.toml, I don't know why I thought it was not necessary, but yeah other packages depending on it need a simple way to determine which features they can enable

IceTDrinker:
done !

IceTDrinker:
@slab-ci cpu_test

tmontaigu (approved):

PR #163, opened 2023-03-27
wip

IceTDrinker:


IceTDrinker:
will be reworked and designed properly later

PR #165, opened 2023-03-28
feat(core): implement 128bit pbs

sarah-ek:


tfhe/src/core_crypto/fft_128_impl/math/fft/tests.rs
```diff
@@ -0,0 +1,211 @@
...
+            for (expected, actual) in izip!(
+                convolution_from_naive.as_ref().iter(),
+                convolution_from_fft.as_ref().iter()
+            ) {
+                let threshold = Scalar::ONE << (Scalar::BITS - 5);
+                let abs_diff = abs_diff(*expected, *actual);
+                assert!(
+                    abs_diff < threshold,
+                    "abs_diff: {abs_diff}, threshold: {threshold}",
+                );
```
IceTDrinker, line 179: 1 << (Scalar::BITS - 5) would be 2^{59} for u64 and 2^{123} for u128 right ? this bound is not tight at all right ? 🤔 or am  I misreading this ?

IceTDrinker (commented):
Thanks for the PR and the hard work on the 128 bits fft and PBS.

I did not review the fft_split_impl for now as it looked to be organized in the same way as the fft_128_impl.

Can you see with @soonum for trying to run the core_crypto tests with avx512 as well ? No need to run the whole stack with avx512 for now I would say

Generally it's a bit always the same comments:

- use to_fourier_polynomial_size for the polynomial_size / 2
- I think there is code duplication somewhat and we could probably find a way to have some common interfaces for some of the PBS operations
- for the types with special layouts I don't think we can make them work with the existing traits we have for entities
- could we try starting to steer away from the "the algorithm requires a view/mut view" approach and use what we described: a wrapper accepting all container types then deferring to an internal implementation only working on views/mut views ?

tfhe/src/core_crypto/fft_split_impl/mod.rs
```diff
@@ -0,0 +1,4 @@
+#![allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
sarah-ek: yeah, i'll change it

tfhe/src/core_crypto/fft_128_impl/math/fft/mod.rs
```diff
@@ -0,0 +1,524 @@
...
+        }
+    }
+}
+
+#[inline(always)]
+fn to_signed_to_f128<Scalar: UnsignedInteger>(x: Scalar) -> f128 {
+    if TypeId::of::<Scalar>() == TypeId::of::<u128>() {
+        // we use bytemuck::cast instead of pulp::cast since it doesn't assert the check at compile
+        // time
+        u128_to_signed_to_f128(bytemuck::cast(x))
```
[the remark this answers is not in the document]
sarah-ek: we can't use `std::mem::transmute` since it can't be used with generic types (it needs to assert that the sizes are equal at compile time)

tfhe/src/core_crypto/fft_128_impl/crypto/bootstrap.rs
```diff
@@ -0,0 +1,336 @@
...
+use crate::core_crypto::commons::traits::{
+    Container, ContiguousEntityContainer, ContiguousEntityContainerMut, Split,
+};
+use crate::core_crypto::commons::utils::izip;
+use crate::core_crypto::entities::*;
+use aligned_vec::{avec, ABox, CACHELINE_ALIGN};
+use dyn_stack::{PodStack, ReborrowMut, SizeOverflow, StackReq};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct Fourier128LweBootstrapKey<C> {
```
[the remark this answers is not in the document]
IceTDrinker: For the fourier structs I think it makes more sense for now

tfhe/src/core_crypto/fft_128_impl/math/fft/mod.rs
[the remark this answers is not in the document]
IceTDrinker: during monomorphization it does not do that at the last moment when the type is known ?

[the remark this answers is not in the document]
IceTDrinker: super clear thanks !

tfhe/src/core_crypto/fft_128_impl/math/fft/tests.rs
IceTDrinker: question is still up for line 176 ?

tfhe/src/core_crypto/fft_128_impl/crypto/bootstrap.rs
```diff
@@ -0,0 +1,336 @@
...
+    // Start doing the right shift
+    output >>= Scalar::BITS - poly_size.log2().0 - 2 + lut_count_log.0;
+    // Do the rounding
+    output += output & Scalar::ONE;
+    // Finish the right shift
+    output >>= 1;
+    // Apply the lsb padding
+    output <<= lut_count_log.0;
+    <Scalar as CastInto<usize>>::cast_into(output)
+}
```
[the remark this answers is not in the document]
IceTDrinker: the pbs modulus switch should probably be shared in some file somewhere this function exists in at least 2 or 3 places currently

tfhe/src/core_crypto/fft_128_split_impl/crypto/bootstrap.rs
```diff
@@ -0,0 +1,236 @@
...
+            .for_each(|(lo, hi)| (*lo, *hi) = wrapping_neg((*lo, *hi)));
+    }
+    let remaining_degree = monomial_degree.0 % output_lo.container_len();
+    output_lo.rotate_left(remaining_degree);
+    output_hi.rotate_left(remaining_degree);
+    izip!(output_lo, output_hi)
+        .rev()
+        .take(remaining_degree)
+        .for_each(|(lo, hi)| (*lo, *hi) = wrapping_neg((*lo, *hi)));
+}
```
IceTDrinker, line 54: would it make sense to provide these in the polynomial_algorithms module ?

tfhe/src/core_crypto/fft_128_split_impl/crypto/tests.rs
```diff
@@ -0,0 +1,215 @@
...
+        std_bootstrapping_key.decompress_into_lwe_bootstrap_key();
+
+    // Create the empty bootstrapping key in the Fourier domain
+    let mut fourier_bsk = Fourier128LweBootstrapKey::new(
+        std_bootstrapping_key.input_lwe_dimension(),
+        std_bootstrapping_key.polynomial_size(),
+        std_bootstrapping_key.glwe_size(),
+        std_bootstrapping_key.decomposition_base_log(),
+        std_bootstrapping_key.decomposition_level_count(),
+    );
```
IceTDrinker, line 79: no need for the seeded key, this was done mainly for example purposes

tfhe/src/core_crypto/fft_128_split_impl/crypto/bootstrap.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    pub fn bootstrap_u128<ContLweOut, ContLweIn, ContAcc>(
+        &self,
+        lwe_out: &mut LweCiphertext<ContLweOut>,
+        lwe_in: &LweCiphertext<ContLweIn>,
+        accumulator: &GlweCiphertext<ContAcc>,
+        fft: Fft128View<'_>,
+        stack: PodStack<'_>,
+    ) where
```
IceTDrinker, line 174: I think we can have a "Bootstrap" trait that is implemented for that function allowing to easily test the various implementations and can allow to abstract over it if needed in higher level APIs

Hear me out:
LweOut is always an LweCiphertext with same Scalar type as LweIn (also an LweCiphertext) and as the accumulator a GlweCiphertext

stack is always a PodStack

The only thing that may be slightly annoying is managing the Fft, but we could have a sealed FftView trait implement for FftView and FftView128, then the trait implementation is basically just calling the underlying implementation of each BSK, what do you think ?

We can even force the Fft struct we expect with an associated type to make it less error prone, and we can have a parameterization over the Scalar, worst case we just call the methods directly in the higher levels without using the trait, but at least the common trait allows to easily test Bootstrap version on the same high level code

IceTDrinker (commented):
See comments on the PR, a few things I caught, I think there is a way to have a lightweight Bootsrap trait to allow us to test the various versions more easily

tfhe/src/core_crypto/fft_128_impl/crypto/bootstrap.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// domain.
+    pub fn fill_with_forward_fourier<Scalar, ContBsk>(
+        &mut self,
+        coef_bsk: &LweBootstrapKey<ContBsk>,
+        fft: Fft128View<'_>,
+    ) where
+        Scalar: UnsignedTorus,
+        ContBsk: Container<Element = Scalar>,
+    {
+        fn implementation<Scalar: UnsignedTorus>(
```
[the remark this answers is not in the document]
sarah-ek: i don't know if this would be better. asserts don't display the function name, and stack traces show the full path

```rust
fn foo() {
    fn implementation() {
        panic!();
    }
    implementation()
}

fn main() {
    foo()
}
```

`implementation` is show as `project_name::foo::implementation`

tfhe/src/core_crypto/fft_128_impl/crypto/tests.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    let f = |x: Scalar| (x * x + x) % 16;
+    let accumulator: GlweCiphertextOwned<Scalar> = generate_accumulator(
+        polynomial_size,
+        glwe_dimension.to_glwe_size(),
+        message_modulus as usize,
+        delta,
+        f,
+    );
```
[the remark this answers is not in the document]
sarah-ek: i think i can share the test code between the two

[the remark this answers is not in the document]
sarah-ek: for line 176, i don't really know what threshold i should be using

tfhe/src/core_crypto/fft_128_split_impl/crypto/bootstrap.rs
sarah-ek: it depends on if we're gonna need them. what are the original polynomial mul/div functions used for?

tfhe/src/core_crypto/fft_128_split_impl/crypto/tests.rs
sarah-ek: what should i use instead?

tfhe/src/core_crypto/fft_128_impl/crypto/bootstrap.rs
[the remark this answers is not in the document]
IceTDrinker: Ah ok, then it's good like that !

tfhe/src/core_crypto/fft_128_split_impl/crypto/bootstrap.rs
IceTDrinker: The wrapping monic ones are mainly for the PBS IIRC

tfhe/src/core_crypto/fft_128_split_impl/crypto/tests.rs
IceTDrinker: par_allocate without the seeded part and then you can skip the decompression

tfhe/src/core_crypto/fft_128_impl/crypto/tests.rs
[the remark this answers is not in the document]
IceTDrinker: do you agree it is too big for the threshold ?

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
```diff
@@ -320,3 +321,114 @@ pub fn pbs_modulus_switch<Scalar: UnsignedTorus + CastInto<usize>>(
...
+
+    fn fill_with_forward_fourier<ContBsk>(
+        &mut self,
+        coef_bsk: &LweBootstrapKey<ContBsk>,
+        fft: &Self::Fft,
+    ) where
+        ContBsk: Container<Element = Scalar>,
+    {
+        let fft = fft.as_view();
+        let mut mem = GlobalPodBuffer::new(fill_with_forward_fourier_scratch(fft).unwrap());
```
IceTDrinker, line 398: any reason to take a GlobalPodBuffer here rather than a PodStack directly ?

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
```diff
@@ -320,3 +321,114 @@ pub fn pbs_modulus_switch<Scalar: UnsignedTorus + CastInto<usize>>(
     output <<= lut_count_log.0;
     <Scalar as CastInto<usize>>::cast_into(output)
 }
+
+pub trait FourierBootstrapKey<Scalar> {
```
IceTDrinker, line 325: can maybe put in some shared module ? Not a blocker

tfhe/src/core_crypto/fft_128_split_impl/crypto/bootstrap.rs
IceTDrinker: so those are probably too specific to make them shared right ?

tfhe/src/core_crypto/fft_128_impl/crypto/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+#[test]
+fn test_bootstrap_u32() {
+    test_bootstrap_generic::<u32, Fourier128LweBootstrapKeyOwned>(
+        StandardDev(sqr(0.000007069849454709433)),
+        StandardDev(sqr(0.00000000000000029403601535432533)),
+    );
+}
```
IceTDrinker, line 219: could we maybe test our other implementations thanks to the generic trait you created and put this test file in some shared test module ?

tfhe/benches/core_crypto/pbs128_bench.rs
```diff
@@ -0,0 +1,175 @@
...
+            // Rotate the accumulator
+            accumulator_scalar.rotate_left(half_box_size);
+
+            let accumulator_plaintext = PlaintextList::from_container(accumulator_scalar);
+
+            allocate_and_trivially_encrypt_new_glwe_ciphertext(glwe_size, &accumulator_plaintext)
+        }
+
+        let f = |x: Scalar| (x * x + x) % 16;
+        let accumulator: GlweCiphertextOwned<Scalar> = generate_accumulator(
```
IceTDrinker, line 137: I think you can jsut use an empty GlweCiphertext and remove the generate accumulator code

IceTDrinker (commented):
Thanks a lot @sarah-ek !

Just a few things to iron out (namely testing all implementation with the same basic generic test), simplifying the benchmark code and probably moving the common bootstrap trait to some shared module.

This is solid work, thank you!

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
sarah-ek: i don't take a GlobalPodBuffer as a parameter, i create it inside the function impl
i figured it might not be that bad to allocate in `fill_with_forward_fourier` directly since the allocation is small (about [f64; 512]), and only the f64 fft needs to allocate during the conversion

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
sarah-ek: yeah, though i don't know where i should put it

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
IceTDrinker: sorry, I meant that other primitives require a podstack (your original bootstrap implem for converting the key), but yes generally for the one time conversion it's ok, was just wondering why you preferred the Global buffer this time

tfhe/benches/core_crypto/pbs128_bench.rs
sarah-ek: don't we need the accumulator to be filled for the bootstrap to work?

tfhe/src/core_crypto/fft_128_impl/crypto/tests.rs
sarah-ek: i think so, how should i structure this?

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
IceTDrinker: may I suggest having something like

fft_impl
|---- f64 (then crypto, etc)
|---- f128
|---- common ?

or

fft_impl
|---- crypto
|       |---- f64
|       |---- f128
|---- common

so either one directory with the crypto/math etc folder for each kind of bootstrap and a common dir in the fft_impl

or a single crypto/math tree with both f64 and f128 variants every time

Not sure which one is better

sarah-ek:
for `polynomial_wrapping_monic_monomial_{mul,div}_assign_split`, i think what we can do is make the whole module split u128 fft module private for now. that way users can't access it and we can freely shuffle stuff around in the future

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
sarah-ek: i just forgot to add the podstack initially and went along with it :sweat_smile:

tfhe/benches/core_crypto/pbs128_bench.rs
IceTDrinker: (we skip the external product if an LWE mask element is zero, so as long as the lwe is non trivial we are executing the external product part)

sarah-ek: ah, right

tfhe/src/core_crypto/fft_impl/crypto/bootstrap.rs
IceTDrinker: ok your call ! I think I would keep the podstack as an input for consistency and to allow higher levels to have more control on that if that's ok 🙂

IceTDrinker:
> for `polynomial_wrapping_monic_monomial_{mul,div}_assign_split`, i think what we can do is make the whole module split u128 fft module private for now. that way users can't access it and we can freely shuffle stuff around in the future

Fine by me!

tfhe/src/core_crypto/fft_impl/f128/crypto/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    );
+}
+
+#[test]
+fn test_bootstrap_u32() {
+    test_bootstrap_generic::<u32, Fourier128LweBootstrapKeyOwned>(
+        StandardDev(sqr(0.000007069849454709433)),
+        StandardDev(sqr(0.00000000000000029403601535432533)),
+    );
+}
```
IceTDrinker, line 221: can we have the tests for the u64 f64 using the same interface as well as the split version of the PBS f128_u128 ?

IceTDrinker (commented):
Thanks just a comment to have the 3 PBS implementation running the same test structure (we can have parameters for each case if necessary)

tfhe/src/core_crypto/fft_impl/f128/math/fft/tests.rs
```diff
@@ -0,0 +1,213 @@
...
+                convolution_from_naive.as_mut(),
+                poly0.as_ref(),
+                poly1.as_ref(),
+            );
+
+            for (expected, actual) in izip!(
+                convolution_from_naive.as_ref().iter(),
+                convolution_from_fft.as_ref().iter()
+            ) {
+                let threshold = Scalar::ONE << (Scalar::BITS - 5);
```
IceTDrinker, line 176: I'm still a bit confused by this one, how many correct bits do we expect to recover ? About a 100 right ? so the error should probably more something like 1 << (Scalar::BITS - 100) ?

tfhe/src/core_crypto/fft_128_impl/crypto/tests.rs
IceTDrinker: can be a test module in fft_impl where we call implementation from the various sub modules implementing the PBS

tfhe/src/core_crypto/fft_impl/f128/math/fft/tests.rs
IceTDrinker: well that formula stands only for u128, but if we expect to recover the exact results for u32 and u64 then we should test it this way, and have the appropriate thershold for u128

sarah-ek: we lose a few bits from the multiplication by an integer of magnitude ~ 2^16, so i think a good threshold would be
`Scalar::ONE << (Scalar::BITS.saturating_sub(100 - integer_magnitude - 4))`

(tests pass with this value)

IceTDrinker: then fine by me !

I don't want to have a flaky tests but it felt like we were losing quite a lot from the formula, I let you be the judge for the security margin adding a comment explaining the reasoning for the threshold 🙂

sarah-ek: this seems like a more reasonable formula
```
                let threshold = Scalar::ONE
                    << (Scalar::BITS.saturating_sub(100 - integer_magnitude - (size_log + 1) / 2));
```
`size_log` is the log of the polynomial size, `(size_log + 1) / 2` is the error we expect to come from the fft

IceTDrinker: perfect!

tfhe/src/core_crypto/fft_impl/f128/crypto/tests.rs
sarah-ek: the split version is tested here, since the f128 bootstrap uses the split version when Scalar == u128

sarah-ek: i can add the test to the f64 code

tfhe/src/core_crypto/fft_impl/f128_u128/mod.rs
```diff
@@ -0,0 +1,4 @@
+#![allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
sarah-ek: i can add it on individual functions, but there's quite a few of them, since the fft functions now take 4 slices instead of 1, and the split fft module additionally splits the inputs and outputs into 2 slices.

what do you think?

tfhe/src/core_crypto/fft_impl/f128/crypto/tests.rs
IceTDrinker: ah perfect! I did not realize it was the case

tfhe/src/core_crypto/fft_impl/f128_u128/mod.rs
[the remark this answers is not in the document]
IceTDrinker: ok fine, plus it is limited to f128_u128 so ok for me in the end !

tfhe/src/core_crypto/fft_impl/f128/crypto/tests.rs
IceTDrinker: yes please for f64 :)

tfhe/src/core_crypto/fft_impl/f128/crypto/bootstrap.rs
```diff
@@ -0,0 +1,439 @@
...
+            fft: Fft128View<'_>,
+            stack: PodStack<'_>,
+        ) {
+            if TypeId::of::<Scalar>() == TypeId::of::<u128>() {
+                let mut lwe_out: LweCiphertext<&mut [u128]> = unsafe { transmute(lwe_out) };
+                let lwe_in: LweCiphertext<&[u128]> = unsafe { transmute(lwe_in) };
+                let accumulator: GlweCiphertext<&[u128]> = unsafe { transmute(accumulator) };
+
+                return this.bootstrap_u128(&mut lwe_out, &lwe_in, &accumulator, fft, stack);
+            }
```
IceTDrinker, line 346: maybe add a comment to indicate we are dispatch to the u128 specific implem here :) I did not catch it at first !

IceTDrinker (approved):
Thanks, if tests are green then I think we can merge this ! 🥳 

Thanks a lot !

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
I'll update the parameters for this test, it's unrelated to this PR

IceTDrinker:
the threshold for the tests are maybe not high enough, if you think it's a bad idea to keep those, then rever to the BITS - 5 thing but add a comment about it please

sarah-ek:
i think the issue might be from the fft part of the error. (size_log + 1) / 2 is about the expected error we can get, not the max error. the max error should be something like size_log

sarah-ek:
renamed the fft modules
f64 -> fft64
f128 -> fft128
f128_u128 -> fft128_u128

PR #166, opened 2023-03-29
Feature/default ops

IceTDrinker:
First commit for how default ops look like for add/add_assign

Question: should we go for macro as it seems most ops will look identical in terms of implem ?

I tend to dislike macros but here it looks really repetitive for the code part


tfhe/src/shortint/server_key/add.rs
```diff
@@ -5,6 +5,157 @@ use crate::shortint::server_key::CheckError::CarryFull;
...
+        if ct_right.carry_is_empty() {
+            rhs = ct_right;
+        } else {
+            tmp_rhs = self.clear_carry(ct_right);
+            rhs = &tmp_rhs;
+        }
+
+        let mut ct_res = self.unchecked_add(lhs, rhs);
+        self.clear_carry_assign(&mut ct_res);
+        ct_res
```
tmontaigu, line 79: Here we can clone lhs then call add_assign, we don"t have the same thing as smart operations where we want to clean the carries before the operation

IceTDrinker: hmm disagree, we want to be safe if for whatever reason someone used unchecked right before right ?

IceTDrinker: ah no my bad you are right, I started with the add op not add_assign, good point !

IceTDrinker: but do we want to clone if we can avoid it ?

tmontaigu: We should end up with the same number of clone. In you current case you don't clone, but use unchecked_add which will clone its lhs before calling`unchecked_add_assign` , in my proposition you clone, and then call `add_assign` which in turns call `unchecked_add_assign` which won't clone

tmontaigu: We might even save a clone by doing so (in the case lhs has non empty carries)

IceTDrinker: you are perfectly right, will update to use the default assign ops in the default one

IceTDrinker:
I have clean commits on this one, each can be reviewed independtly normally

IceTDrinker:
@slab-ci cpu_test

tfhe/src/shortint/server_key/neg.rs
```diff
@@ -5,6 +5,119 @@ use crate::shortint::server_key::CheckError::CarryFull;
...
+        ct: &CiphertextBase<OpOrder>,
+    ) -> CiphertextBase<OpOrder> {
+        let mut ct_res = ct.clone();
+        self.neg_assign(&mut ct_res);
+        ct_res
+    }
+
+    /// Compute homomorphically a negation of a ciphertext.
+    ///
+    /// This checks that the addition is possible. In the case where the carry buffers are full,
```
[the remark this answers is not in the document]
IceTDrinker: Thanks

tfhe/src/shortint/server_key/mul.rs
```diff
@@ -645,6 +645,294 @@ impl ServerKey {
...
+    /// // |      ct_res     |
+    /// // | carry | message |
+    /// // |-------|---------|
+    /// // |  0 0  |   0 0   |
+    ///
+    /// let res = cks.decrypt(&ct_res);
+    /// let modulus = sks.message_modulus.0;
+    /// assert_eq!(res, (msg1 * msg2) % modulus as u64);
+    /// ```
+    pub fn mul_lsb<OpOrder: PBSOrderMarker>(
```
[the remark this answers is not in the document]
IceTDrinker: I would guess so yes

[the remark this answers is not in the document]
IceTDrinker: done

tfhe/src/shortint/server_key/neg.rs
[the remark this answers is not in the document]
IceTDrinker: and done !

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
looks like I forgot adding the default ops to tests 🤦‍♂️

tmontaigu (approved):

PR #167, opened 2023-03-29
chore(ci): add windows as target build platform

soonum:


IceTDrinker:
c api tests will fail the windows build, it will propbably require disabling them

Makefile
```diff
@@ -226,11 +227,18 @@ format_doc_latex:
 	@printf "\n===============================\n"
 
 .PHONY: check_compile_tests # Build tests in debug without running them
+ifneq ($(filter $(OS),Linux Darwin),)
```
IceTDrinker, line 230: looks to be working well but I'm not a fan of condiftional Makefiles

could we do

```
.PHONY: check_compile_tests # Build tests in debug without running them
check_compile_tests: build_c_api
	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --no-run \
		--features=$(TARGET_ARCH_FEATURE),experimental,boolean,shortint,integer,internal-keycache \
		-p tfhe && \
		if [[ "$(OS)" == "Linux" || "$(OS)" == "Darwin" ]]; then \
			./scripts/c_api_tests.sh --build-only; \
		fi
```
instead ?

soonum: Yes we can do it at the cost of losing the `&&` logic here.

IceTDrinker: hmm why do we lose the && logic ? it's still in there

IceTDrinker: I'm not even sure the && is required in that case anyways

soonum: Keeping `&& \` as is, is a syntax error.

soonum: Correction, it's a syntax error when followed by a `@` right before the `if`

IceTDrinker: anyways as you said let's remove that &&

IceTDrinker:
@slab-ci cpu_test

PR #168, opened 2023-03-30
feat(typed_api): plug choice of big/small ciphertext

tmontaigu:


tfhe/src/typed_api/config.rs
```diff
@@ -113,6 +113,15 @@ impl ConfigBuilder {
...
 
+    #[cfg(feature = "integer")]
+    pub fn enable_default_uint8_small(mut self) -> Self {
+        let mut params = crate::typed_api::integers::FheUint8Parameters::default();
+        params.0.pbs_order = crate::shortint::PBSOrder::BootstrapKeyswitch;
+
+        self.config.integer_config.uint8_params = Some(params);
+        self
+    }
+
```
tmontaigu, line 124: every integer type needs the same kind of enable function

tfhe/src/typed_api/integers/client_key.rs
```diff
@@ -1,55 +1,150 @@
...
+            RadixCiphertextDyn::Big(ct) => self.inner.decrypt(ct),
+            RadixCiphertextDyn::Small(ct) => self.inner.decrypt(ct),
+        }
+    }
+}
+
+impl DecryptionKey<RadixCiphertextDyn, U256> for RadixClientKey {
+    fn decrypt(&self, ciphertext: &RadixCiphertextDyn) -> U256 {
+        match ciphertext {
+            // TODO integer::RadixClientKey should have a better decrypt fn
```
tmontaigu, line 133: do this TODO

tfhe/src/typed_api/integers/server_key.rs
```diff
@@ -34,6 +34,169 @@ where
...
+    ) -> crate::integer::RadixCiphertextBig {
+        wopbs_radix(self, sks, ct, f)
+    }
+
+    fn apply_bivariate_wopbs(
+        &self,
+        sks: &crate::integer::ServerKey,
+        lhs: &crate::integer::RadixCiphertextBig,
+        rhs: &crate::integer::RadixCiphertextBig,
+        f: impl Fn(u64, u64) -> u64,
```
IceTDrinker, line 173: can we perhaps already manage the RadixCtDyn here ? or do we need it given we manage the dyn thing above ?

tfhe/src/typed_api/integers/client_key.rs
```diff
@@ -1,55 +1,147 @@
...
-    fn decrypt(&self, ciphertext: &Self::Ciphertext) -> U256 {
+impl DecryptionKey<RadixCiphertextBig, u64> for RadixClientKey {
+    fn decrypt(&self, ciphertext: &RadixCiphertextBig) -> u64 {
+        assert_eq!(
+            self.pbs_order,
+            crate::shortint::PBSOrder::KeyswitchBootstrap
+        );
+        self.inner.decrypt(ciphertext)
+    }
+}
```
IceTDrinker, line 109: you can't match on the variant like above ? Or maybe I'm misreading something

IceTDrinker (commented):
Looks generally good, I'm not sure I understand yet why we have some specific implementation for some big ciphertexts AND the implementation for the Dyn ciphertexts ?

IceTDrinker:
could be worth adding some "Small" test for u256 to be sure we are delivering what we want

tfhe/src/typed_api/config.rs
```diff
@@ -113,6 +113,15 @@ impl ConfigBuilder {
         self
     }
 
+    #[cfg(feature = "integer")]
+    pub fn enable_default_uint8_small(mut self) -> Self {
+        let mut params = crate::typed_api::integers::FheUint8Parameters::default();
```
IceTDrinker, line 118: this will need to use the SMALL variant of parameters they are available in crate::shortint::parameters

like PARAMS_SMALL_MESSAGE_2_CARRY_2

IceTDrinker: same for other uint types

tfhe/src/typed_api/integers/client_key.rs
tmontaigu: We can't  here, we use a radix key to encrypt a big radix, the assert is there to check that it matches the fact
that the client key was created with small params

tfhe/src/typed_api/config.rs
tmontaigu: Didn't see they existes

IceTDrinker: no worries, they are rather discreet

tfhe/src/typed_api/integers/client_key.rs
IceTDrinker: then my question is rather : why do we have a specialized implementation for the Big variant ?

tmontaigu: it was needed to be able to incrementally makes things works, it is not needed now so can be removed

IceTDrinker:
Maybe we should add a field with a PBSOrder enum on the shortint parameters for sanity checks when calling small ops with the wrong parameters, wdyt ?

tfhe/src/typed_api/integers/client_key.rs
IceTDrinker: Ok, I think a ctrl+f for those "specialized" impls can be worth a go

IceTDrinker (approved):
Alright thanks a lot for the PR ! Ship it !

PR #169, opened 2023-03-30
Chore/quick fixes

IceTDrinker:


IceTDrinker:
not sure about the degree change

PR #172, opened 2023-04-03
chore(ci): test core_crypto with avx512

soonum:


.github/workflows/cargo_build.yml
```diff
@@ -31,6 +31,10 @@ jobs:
         run: |
           make build_core
 
+      - name: Build Release core (with AVX512)
+        run: |
+          AVX512_SUPPORT=ON make build_core
```
IceTDrinker, line 36: this won't play nice with windows I believe

Makefile
```diff
@@ -174,7 +174,7 @@ build_node_js_api: install_rs_build_toolchain
 .PHONY: test_core_crypto # Run the tests of the core_crypto module including experimental ones
 test_core_crypto: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --release \
-		--features=$(TARGET_ARCH_FEATURE),experimental -p tfhe -- core_crypto::
+		--features=$(TARGET_ARCH_FEATURE),experimental,$(AVX512_FEATURE) -p tfhe -- core_crypto::
```
IceTDrinker, line 177: I think the build_core target update is missing ?

IceTDrinker: also here I would like to have both tests, once without avx512 and once with it as it changes the code being executed

IceTDrinker: depending on whether AVX512 support was ON or not

soonum: I don't quite get it about the both tests thing. You're not happy by having two workflow jobs for testing ?

IceTDrinker: No I mean that we want to run the test once in this target without avx512 so the existing command is ok and then we run conditionally a second command with avx512 support on and nightly toolchain

As both code paths need to be correct

Makefile
```diff
@@ -126,11 +136,13 @@ gen_key_cache: install_rs_build_toolchain
 		--features=$(TARGET_ARCH_FEATURE),shortint,internal-keycache -p tfhe
 
 .PHONY: build_core # Build core_crypto with and without experimental features
-build_core: install_rs_build_toolchain
+build_core: install_rs_build_toolchain install_rs_build_toolchain_nightly
```
IceTDrinker, line 139: I think you could have used the nighlty check toolchain instead ? Or we can rename the check toolchain to nightly toolchain

Makefile
```diff
@@ -126,11 +136,13 @@ gen_key_cache: install_rs_build_toolchain
...
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) build --release \
 		--features=$(TARGET_ARCH_FEATURE) -p tfhe
-	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) build --release \
-		--features=$(TARGET_ARCH_FEATURE),experimental -p tfhe
+	@if [[ "$(AVX512_SUPPORT)" == "ON" ]]; then \
+		RUSTFLAGS="$(RUSTFLAGS)" cargo +nightly build --release \
+			--features=$(TARGET_ARCH_FEATURE),experimental,$(AVX512_FEATURE) -p tfhe; \
+	fi
```
IceTDrinker, line 145: if you look closely we build once with experimental, here if we don't have the avx512 we won't build with experimental anymore

soonum: Copy paste mistake

IceTDrinker (approved):
Thanks for the PR !

PR #174, opened 2023-04-03
doc(integer): reorganize user documentation

tmontaigu:
This mainly move files and set the SUMMARY.md so integer module documentation is properly accessible in gitbook


IceTDrinker (approved):
Re-org looks good, I did not check the content for now as it's taken from the previous integer docs

PR #178, opened 2023-04-05
chore(core): reverse the order in which we encrypt KS levels

IceTDrinker:
- allows to avoid reversing the iterator, potentially improving cache access during a keyswitch

BREAKING CHANGE: the keyswitch key level order has been reversed

TODO: fix the mismatch between DecompositionTerm and DecompositionIter for the meaning of a decomposition level see
https://github.com/zama-ai/tfhe-rs-internal/issues/72

The same could be done for the PBS but the change is not wopbs friendly (kind of expected) and is a much more impacting change than this one


tmontaigu (approved):

IceTDrinker:
I'll try to run the benches here to see if has an impact, otherwise it may not be necessary to merge this, I'm unsure

PR #180, opened 2023-04-05
refactor(tfhe): add support for power of 2 q for LWE linalg + KS + PBS

IceTDrinker:


tfhe/src/core_crypto/commons/math/random/generator.rs
```diff
@@ -184,6 +186,40 @@ impl<G: ByteRandomGenerator> RandomGenerator<G> {
...
+    {
+        self.fill_slice_with_random_uniform(output);
+
+        if !ciphertext_modulus.is_native_modulus() {
+            slice_wrapping_scalar_mul_assign(
+                output,
+                ciphertext_modulus.get_scaling_to_native_torus(),
+            );
+        }
+    }
```
IceTDrinker, line 221: @jborfila just need your opinion on this function to generate random uniform, I think it looks ok.

Random gaussian noise does not have scaling managed at the moment (i.e. it scales with the native type and not the user provided modulus)

Cargo.toml
```diff
@@ -7,3 +7,9 @@ lto = "fat"
 
 [profile.release]
 lto = "fat"
+
+# Compiles much faster for tests and allows reasonable performance for iterating
+[profile.devo]
+inherits = "dev"
+opt-level = 3
+lto = "off"
```
IceTDrinker, line 15: this allows to iterate much faster on tests and was a life saver in my case

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
TODO after last rebase : expose u128 PBS in algorithms and try it with non native modulus

IceTDrinker:
tests are actually green, the status check was not updated

tfhe/src/core_crypto/algorithms/slice_algorithms.rs
```diff
@@ -301,3 +301,11 @@ where
     lhs.iter_mut()
         .for_each(|lhs| *lhs = (*lhs).wrapping_mul(rhs));
 }
+
+pub fn slice_wrapping_scalar_div_assign<Scalar>(lhs: &mut [Scalar], rhs: Scalar)
```
sarah-ek, line 305: what is this used for? if i recall correctly, division is not a necessarily a constant-time operation, so it may be leaking info if not used properly

IceTDrinker: scaling during decryption if we use a torus smaller than the native torus

IceTDrinker: or during GGSW encryption of a given factor but it's not dependant on the secret key content normally

IceTDrinker: we have other algorithms that are not constant time (like pk encryption at the moment)

PR #181, opened 2023-04-06
chore(integer): annotate decrypt_radix type in tests

tmontaigu:
When working on the integer part of the crate,
if you introduced a compile error (as is common when working stuff out) the type inference of rust would not fully work and call to `let dec = cks.decrypt_radix(&ctxt);` would fail to deduce the type of `dec`.

This resulted on many errors in the compiler output about "type annotation needed", requiring to scroll up a certain amount to be able to see the errors messages you actually care about.

This commit adds these missing type annotation the the errors won't appear, so as to have less noise.


IceTDrinker (approved):
Thanks was getting this from time to time

PR #182, opened 2023-04-06
feat(c_api): add a C API of the high level API

tmontaigu:
One notable change is that since this C API
relies a lot on macro_rules! to be generated
we have to activate cbindgen's `expand` option,
which will use cargo-expand to expand macros.

However this means we can't call bindgen from the build.rs as it seems to lead to a infinite-loop
(build.rs calls bindgen which calls cargo-expand which calls build.rs...)

So we call the cbindgen binary via the makefile.


Makefile
```diff
@@ -166,7 +166,12 @@ build_tfhe_full: install_rs_build_toolchain
 .PHONY: build_c_api # Build the C API for boolean and shortint
 build_c_api: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) build --release \
-		--features=$(TARGET_ARCH_FEATURE),boolean-c-api,shortint-c-api -p tfhe --all-targets
+		--features=$(TARGET_ARCH_FEATURE),shortint,integer,internal-keycache,boolean,__c_api,boolean-c-api,shortint-c-api \
+		-p tfhe \
+		--all-targets
+	RUSTUP_TOOLCHAIN=nightly cbindgen -c ./tfhe/cbindgen.toml ./tfhe/ -o target/release/tfhe.h
```
IceTDrinker, line 172: use the CARGO_RS_CHECK_TOOLCHAIN here as it's guaranteed to be installed

Makefile
```diff
@@ -166,7 +166,12 @@ build_tfhe_full: install_rs_build_toolchain
 .PHONY: build_c_api # Build the C API for boolean and shortint
 build_c_api: install_rs_build_toolchain
```
IceTDrinker, line 167: add a dependency on install_rs_check_toolchain

I think we may want to rename the check toolchain to nightly toolchain at some point

Makefile
tmontaigu: In reality, It only works with the nightly toolchain since it runs `cargo rustc -Zunpretty=expanded`

error : `the option ``Z`` is only accepted on the nightly compiler`

IceTDrinker: yes sorry, just wanted to say we have a select check toolchain

tmontaigu: ok yes I need to use `RS_CHECK_TOOLCHAIN` not `RS_BUILD_TOOLCHAIN`

Makefile
tmontaigu: I added a `install_cbindgen` which depends on `install_rs_check_toolchain`

tmontaigu:
@slab-ci cpu_test

Makefile
```diff
@@ -51,6 +51,12 @@ install_cargo_nextest: install_rs_build_toolchain
 	cargo $(CARGO_RS_BUILD_TOOLCHAIN) install cargo-nextest --locked || \
 	( echo "Unable to install cargo nextest, unknown error." && exit 1 )
 
+.PHONY: install_cbindgen # Install cbindgen to generate the C API header
+install_cbindgen: install_rs_check_toolchain
+	@cbingen --version > /dev/null 2>&1 || \
+	cargo $(CARGO_RS_BUILD_TOOLCHAIN) install cbindgen --locked || \
+	( echo "Unable to install cbindgen, unknown error." && exit 1 )
```
IceTDrinker, line 58: here you do a install_rs_check_toolchain but use the build toolchain

Makefile
```diff
@@ -164,9 +170,14 @@ build_tfhe_full: install_rs_build_toolchain
 		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,integer -p tfhe --all-targets
 
 .PHONY: build_c_api # Build the C API for boolean and shortint
-build_c_api: install_rs_build_toolchain
+build_c_api: install_cbindgen
```
IceTDrinker, line 173: I would add the install_rs_build_toolchain and install_rs_check_toolchain as dependencies as you refer to both in the Make target

tfhe/src/c_api/typed_api/config.rs
```diff
@@ -0,0 +1,96 @@
...
+pub unsafe extern "C" fn config_builder_build(
+    builder: *mut ConfigBuilder,
+    result: *mut *mut Config,
+) -> c_int {
+    catch_panic(|| {
+        check_ptr_is_non_null_and_aligned(result).unwrap();
+
+        let config = Box::from_raw(builder).0.build();
+
+        *result = Box::into_raw(Box::new(Config(config)));
```
IceTDrinker, line 94: do you need to take ownership of the config builder ? otherwise you can use the get_ref_checked or get_mut_checked to get a ref/mut ref to the pointer, if you need ownership you could first get the ref then clone to avoid leaving a dangling pointer on the C side

tmontaigu:
@slab-ci cpu_test

tfhe/src/integer/u256.rs
```diff
@@ -9,6 +9,90 @@ pub const fn adc(l: u64, r: u64, c: bool) -> (u64, bool) {
...
+            // (4 * 64) bits / 8 bits = 32
+            let inner_byte_slice =
+                std::slice::from_raw_parts(inner_slice.as_ptr() as *const u8, 32);
+            let shortest_len = inner_byte_slice.len().min(bytes.len());
+            bytes[..shortest_len].copy_from_slice(&inner_byte_slice[..shortest_len]);
+        }
+        #[cfg(target_endian = "big")]
+        for sub_slice in bytes.chunks_mut(8) {
+            sub_slice.reverse();
+        }
```
IceTDrinker, line 77: bytes can be a slice that does not have a lenght multiple of 8 right ? the las bytes still need some for of inversion I would believ 🤔

tfhe/src/integer/u256.rs
```diff
@@ -9,6 +9,90 @@ pub const fn adc(l: u64, r: u64, c: bool) -> (u64, bool) {
...
```
IceTDrinker, line 93: same applies here, I don't know if I would not ask for 4 u64 on the C API side every time and be done with it for the moment, while I understand the flexibility of the bytes, it feels a bit silly (and if they want to they can do their own wrapper on the u64_words_to_u256 to manage smaller types)

I would probably see something like little_endian_u64_words_to_u256(word_0: u64, word_1: u64, word_2: u64, word_u64: 3, ...) and they can easily pass 0 where they want to pass zero ? Or was the byte slice thing a requirement from the block chain ? Also in the case of the byte slice I would probably require a 32 bytes slice or fixed size array of 32 bytes to avoid headaches here

tfhe/c_api_tests/test_high_level_boolean.c
```diff
@@ -0,0 +1,102 @@
+#include <tfhe.h>
+
+#include <stdio.h>
+#include <inttypes.h>
+#include <assert.h>
+
+#define GOTO_IF_TRUE(label, cond) if ((cond) == 1) {\
+  goto label;\
+  }
```
IceTDrinker, line 9: you are more disciplined than me 😛

tfhe/c_api_tests/test_high_level_boolean.c
```diff
@@ -0,0 +1,102 @@
...
+    bool clear;
+    ok = fhe_bool_decrypt(result, client_key, &clear);
+    GOTO_IF_TRUE(exit, ok != 0)
+
+    assert(clear == (lhs_clear & rhs_clear));
+
+exit:
+    fhe_bool_destroy(lhs);
+    fhe_bool_destroy(rhs);
+    fhe_bool_destroy(result);
```
IceTDrinker, line 38: I think we may want to add safe destroy to boolean and shortint the same way you do it here, I'll create an issue

tfhe/c_api_tests/test_high_level_integers.c
```diff
@@ -0,0 +1,201 @@
...
+    assert(clear == lhs_clear);
+    
+ exit:
+    fhe_uint8_destroy(lhs);
+    compressed_fhe_uint8_destroy(clhs);
+    fhe_uint8_destroy(result);
+  return ok;
+}
+
+int main(void)
```
IceTDrinker, line 146: I'm going to be a bit annoying could we have a small encrypt/decrypt with u256 ?

IceTDrinker (commented):
Thanks a lot for the hard work, a few comments here and there, but overall it's looking good !

tfhe/src/c_api/typed_api/config.rs
tmontaigu: I need to take ownership of the builder. So if I clone it that means that the C caller has to additionnally call `config_builder_destroy` (which does not exist yet since build is meant to be the freeing function)

tfhe/src/integer/u256.rs
tmontaigu: We could add a function that takes 4 u64, but I think as a common api to initialize from/to bytes
for example the go uint256 lib that I used, exposed a ToBigEndianBytes, which means that if we only have a function that takes 4u64, we would still need to reconstruct these u64 from the givens bytes, and this is what these function does

IceTDrinker: up to you, I think I would require a fixed size byte slice to make things easier

Makefile
tmontaigu: added

tfhe/c_api_tests/test_high_level_boolean.c
tmontaigu: Even tho its probably a good aproach for real code, I replace these with assert to make tracing error easier

tfhe/c_api_tests/test_high_level_integers.c
tmontaigu: Added

tfhe/src/integer/u256.rs
tmontaigu: added length checks

tfhe/c_api_tests/test_high_level_integers.c
tmontaigu: added one

Makefile
IceTDrinker: there still is a mismatch between the toolchain installed and the one used

IceTDrinker (commented):
Just a minor thing in the Makefile, otherwise looks good, you did not want to keep a C test with byte slice as a mean to create a U256 ?

tmontaigu:
The "hard" part of the logic is tested from the rust side in c its just coverting ptr + len to slice so there should not be any problem

Makefile
tmontaigu: 😨

IceTDrinker (approved):
All good thanks!

PR #184, opened 2023-04-06
chore(shortint): remove the Default impl for Parameters

tmontaigu:
The rationale behind this is that, `shortint::Parameters::default()` does not convey the information about how much bit of message and carry this parameter provides, and so might lead to errors/confusions.

Instead user will be forced to use the param name like `PARAM_MESSAGE_2_CARRY_2` which is less ambiguous.

This is obviously a breaking change.

Fixes zama-ai/tfhe-rs-internal#59


tmontaigu:
@slab-ci cpu_test

tmontaigu:
@slab-ci cpu_test

PR #189, opened 2023-04-07
Feat/integer default ops

IceTDrinker:
First batch of default ops for integers

### DIFF
--- a/tfhe/src/integer/server_key/comparator.rs
+++ b/tfhe/src/integer/server_key/comparator.rs
@@ -906,6 +906,246 @@ impl<'a> Comparator<'a> {
     ) -> RadixCiphertext<PBSOrder> {
         self.smart_min_or_max_parallelized(lhs, rhs, MinMaxSelector::Min)
     }
+
+    //======================================
+    // "Default" Multi-Threaded operations
+    //======================================
+    pub fn eq_parallelized<PBSOrder: PBSOrderMarker>(
+        &self,
+        lhs: &RadixCiphertext<PBSOrder>,
+        rhs: &RadixCiphertext<PBSOrder>,
+    ) -> RadixCiphertext<PBSOrder> {
+        let mut tmp_lhs: RadixCiphertext<PBSOrder>;
+        let mut tmp_rhs: RadixCiphertext<PBSOrder>;
+        let (lhs, rhs) = match (lhs.block_carries_are_empty(), rhs.block_carries_are_empty()) {
+            (true, true) => (lhs, rhs),
+            (true, false) => {
+                tmp_rhs = rhs.clone();
+                self.server_key.full_propagate_parallelized(&mut tmp_rhs);
+                (lhs, &tmp_rhs)
+            }
+            (false, true) => {
+                tmp_lhs = lhs.clone();
+                self.server_key.full_propagate_parallelized(&mut tmp_lhs);
+                (&tmp_lhs, rhs)
+            }
+            (false, false) => {
+                tmp_lhs = lhs.clone();
+                tmp_rhs = rhs.clone();
+                rayon::join(
+                    || self.server_key.full_propagate_parallelized(&mut tmp_lhs),
+                    || self.server_key.full_propagate_parallelized(&mut tmp_rhs),
+                );
+                (&tmp_lhs, &tmp_rhs)
+            }
+        };
+
+        self.unchecked_eq_parallelized(lhs, rhs)
+    }
+
+    pub fn gt_parallelized<PBSOrder: PBSOrderMarker>(
+        &self,
+        lhs: &RadixCiphertext<PBSOrder>,
+        rhs: &RadixCiphertext<PBSOrder>,
+    ) -> RadixCiphertext<PBSOrder> {
+        let mut tmp_lhs: RadixCiphertext<PBSOrder>;
+        let mut tmp_rhs: RadixCiphertext<PBSOrder>;
+        let (lhs, rhs) = match (lhs.block_carries_are_empty(), rhs.block_carries_are_empty()) {
+            (true, true) => (lhs, rhs),
+            (true, false) => {
+                tmp_rhs = rhs.clone();
+                self.server_key.full_propagate_parallelized(&mut tmp_rhs);
+                (lhs, &tmp_rhs)
+            }
+            (false, true) => {
+                tmp_lhs = lhs.clone();
+                self.server_key.full_propagate_parallelized(&mut tmp_lhs);
+                (&tmp_lhs, rhs)
+            }
+            (false, false) => {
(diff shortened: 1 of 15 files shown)


tfhe/src/integer/server_key/comparator.rs
```diff
@@ -906,6 +906,246 @@ impl<'a> Comparator<'a> {
...
+                rayon::join(
+                    || self.server_key.full_propagate_parallelized(&mut tmp_lhs),
+                    || self.server_key.full_propagate_parallelized(&mut tmp_rhs),
+                );
+                (&tmp_lhs, &tmp_rhs)
+            }
+        };
+
+        let mut res = self.unchecked_max_parallelized(lhs, rhs);
+        self.server_key.full_propagate_parallelized(&mut res);
```
tmontaigu, line 1110: I think we can do better than full_propagate.

The last step of unchecked_max/min is an unchecked_add(x,y)
where 
- if x != 0 => 0 < x < msg_mod and y == 0
- or if y != 0 => x == 0 and  0 < y < msg_mod

so the resulting ciphertext should not have any carries, 
thus the degree is in reality  message_mod - 1,
which means that is we want to reset the degree/noise we whould simply `clean_carry` in parallel

tfhe/src/integer/server_key/comparator.rs
```diff
@@ -906,6 +906,246 @@ impl<'a> Comparator<'a> {
...
+                rayon::join(
+                    || self.server_key.full_propagate_parallelized(&mut tmp_lhs),
+                    || self.server_key.full_propagate_parallelized(&mut tmp_rhs),
+                );
+                (&tmp_lhs, &tmp_rhs)
+            }
+        };
+
+        let mut res = self.unchecked_min_parallelized(lhs, rhs);
+        self.server_key.full_propagate_parallelized(&mut res);
```
tmontaigu, line 1146: same as max

tfhe/src/integer/server_key/comparator.rs
```diff
@@ -1092,11 +1427,12 @@ mod tests {
...
-    /// All our comparison function have 4 variants:
+    /// All our comparison function have 5 variants:
     /// - unchecked_$comparison_name
     /// - unchecked_$comparison_name_parallelized
     /// - smart_$comparison_name
     /// - smart_$comparison_name_parallelized
+    /// - $comparison_name_parallelized
     ///
     /// So, for example, for the `gt` comparison fn, this macro will generate the tests for
     /// the 4 variants described above
```
tmontaigu, line 1438: ~~4~~ 5 😁

tfhe/src/integer/server_key/comparator.rs
IceTDrinker: My test was failing I believe when I did not do the full propagate for min and max

IceTDrinker: ah right, you may be correct here

tmontaigu: on my side, without the pull propagate the test passes

but we might do:
`res.blocks.par_iter_mut().for_each(|block| self.server_key.key.message_extract_assign(block));`
to reset the noise

IceTDrinker: I think it failed in the CI here, I relaunched it several times locally and the full propagate was necessary (for the empty carry check I believe), I'll see for the message_extract_assign

tfhe/src/integer/server_key/comparator.rs
IceTDrinker: done locally

tfhe/src/integer/server_key/comparator.rs
tmontaigu: Since the unchecked_add updates the degree, if you check the degree of the block they will be == (2 * (message_mod - 1)) 
when in reality they should be at message_mod - 1

IceTDrinker: yep, but the noise refresh is probably necessary as degree and noise were managed together, the good news is the refresh can be threaded which was not the case of the propagate

IceTDrinker: doesn't work, will have to full propagate for 3_3 at least

tmontaigu: so it works for 2_2, 4_4 but not 3_3 ?

tfhe/src/integer/server_key/radix/tests.rs
```diff
@@ -18,7 +18,20 @@ create_parametrized_test!(integer_encrypt_decrypt_256_bits_specific_values);
...
+    integer_smart_add_128_bits(PARAM_MESSAGE_2_CARRY_2)
+}
+#[test]
+fn test_integer_smart_add_128_bits_param_message_3_carry_3() {
+    integer_smart_add_128_bits(PARAM_MESSAGE_3_CARRY_3)
+}
+#[test]
+fn test_integer_smart_add_128_bits_param_message_4_carry_4() {
+    integer_smart_add_128_bits(PARAM_MESSAGE_4_CARRY_4)
+}
```
tmontaigu, line 34: minor thing: you should be able to do
```rust
create_parametrized_test!{
    integer_smart_add_128_bits {
        // Skip the 1_1 params for the smart add 128 bits which proved to be the slowest test in our test
        // suite
        PARAM_MESSAGE_2_CARRY_2,
        PARAM_MESSAGE_3_CARRY_3,
        PARAM_MESSAGE_4_CARRY_4
    }
}
```

tfhe/src/integer/server_key/radix_parallel/scalar_mul.rs
```diff
@@ -313,4 +415,104 @@ impl ServerKey {
...
+        self.scalar_mul_assign_parallelized(&mut ct_res, scalar);
+        ct_res
+    }
+
+    pub fn scalar_mul_assign_parallelized<PBSOrder: PBSOrderMarker>(
+        &self,
+        ct: &mut RadixCiphertext<PBSOrder>,
+        scalar: u64,
+    ) {
+        let zero = self.create_trivial_zero_radix(ct.blocks.len());
```
tmontaigu, line 468: its the same code as smart_scalar_mul_assign ? could we call `smart_scalar_mul_parallelized` is has a compatible signature

IceTDrinker: not strictly the reduction is done via the default add to avoid variable performance 😭

tmontaigu: ah i missed that

tmontaigu (approved):

PR #195, opened 2023-04-11
chore(tfhe): rename typed_api to high_level_api

tmontaigu:
high_level_api makes it easier to understand
what this api brings (at least more than typed_api does) and is term used in the documenation


IceTDrinker (approved):
If tests are green let's go

tmontaigu:
yeah we can

PR #196, opened 2023-04-11
chore(c_api): allow to build in a simple cargo command, requires nightly

IceTDrinker:


tfhe/src/lib.rs
```diff
@@ -53,6 +53,7 @@ pub use js_on_wasm_api::*;
 mod test_user_docs;
 
 #[cfg(any(feature = "boolean", feature = "shortint", feature = "integer"))]
+/// cbindgen:ignore
```
tmontaigu, line 56: can't this be `//` instead ? I saw `///` where used for the same ignore but it makes them show up in doc
https://docs.rs/tfhe/0.1.12/tfhe/boolean/index.html#special-module-attributes

should probably be done later

tmontaigu (approved):

tfhe/src/lib.rs
IceTDrinker: I know, cbindgen requires a doc comment, but we have a neat workaround as you can see above

PR #207, opened 2023-04-12
fix(integer): fix scalar mul bug when representing integers > 64 bits

IceTDrinker:
- a product was overflowing, we now compute a progressive division with the same effect and stop once we reach zero to limit the number of generated tasks


IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

tfhe/src/integer/server_key/radix_parallel/scalar_mul.rs
tmontaigu, line 1: I undestand that this change will make us stop generating tasks once the scalar is zero, however I don't see what "a product was overflowing" before

tmontaigu (approved):

PR #223, opened 2023-04-18
chore(bench): more multi-bit bench params

IceTDrinker:


tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -272,7 +376,7 @@ fn multi_bit_pbs<Scalar: UnsignedTorus + CastInto<usize> + CastFrom<usize> + Syn
                     &accumulator.as_view(),
                     &multi_bit_bsk,
                     // Leave one thread to the OS and one for the ext product loop
```
soonum, line 378: Shouldn't we remove this comment then ?

tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -66,17 +66,49 @@ fn multi_bit_benchmark_parameters<Scalar: Numeric>(
...
+                "2_bits_multi_bit_group_2".to_string(),
+                (
+                    CryptoParametersRecord {
+                        lwe_dimension: Some(LweDimension(764)),
+                        lwe_modular_std_dev: Some(StandardDev(0.000006025673585415336)),
+                        pbs_base_log: Some(DecompositionBaseLog(18)),
+                        pbs_level: Some(DecompositionLevelCount(1)),
+                        glwe_dimension: Some(GlweDimension(3)),
+                        glwe_modular_std_dev: Some(StandardDev(0.0000000000039666089171633006)),
+                        polynomial_size: Some(PolynomialSize(1 << 9)),
```
soonum, line 79: I'm curious about why you're shifting bits instead of putting the numerical value?

soonum (approved):
Good to go for me expect one small comment to remove.

tfhe/benches/core_crypto/pbs_bench.rs
IceTDrinker: yes you are right

tfhe/benches/core_crypto/pbs_bench.rs
IceTDrinker: parameters are given as the log of the polynomial size, it's simpler to shift by the log(N) as 1 << log(N) == N for N a power of 2, rather than compute it in my head and make a mistake 🙂

IceTDrinker:
> Just one request, could you add the `message_modulus` in the `CryptoParameters` initialization ?

will do

PR #225, opened 2023-04-18
Fix/main mul

IceTDrinker:
same commits as for 0.2.x but for main

### DIFF
--- a/.github/workflows/boolean_benchmark.yml
+++ b/.github/workflows/boolean_benchmark.yml
@@ -5,19 +5,19 @@ on:
   workflow_dispatch:
     inputs:
       instance_id:
-        description: 'Instance ID'
+        description: "Instance ID"
         type: string
       instance_image_id:
-        description: 'Instance AMI ID'
+        description: "Instance AMI ID"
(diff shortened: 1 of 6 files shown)


.github/workflows/boolean_benchmark.yml
```diff
@@ -118,7 +102,7 @@ jobs:
       - name: Send data to Slab
         shell: bash
         env:
-          COMPRESSED_RESULTS : ${{ env.RESULTS_FILENAME }}.gz
+          COMPRESSED_RESULTS: ${{ env.RESULTS_FILENAME }}.gz
```
soonum, line 105: All occurrences of COMPRESSED_RESULTS can be removed throughout all the bench files.
It's an artefact from one of my devs.

soonum (approved):
Looks very good overall. Just remove my artifacts and we're good to resume

.github/workflows/boolean_benchmark.yml
soonum: Ok will do.

PR #228, opened 2023-04-19
fix(hlapi): use correct number of blocks for FheUint32

tmontaigu:
The FheUint32 was wrongly defined as being 32 blocks of 2 bits when it should have been 16 blocks.

> Cherry pick


IceTDrinker (approved):
Thanks!

IceTDrinker:
we'll merge without AWS tests in that case it's just a cherry-pick that passed tests on main IIRC

PR #229, opened 2023-04-19
chore(ci): fix decomposition basis and add bit size to params

soonum:
Decomposition basis wasn't correctly set to handle CRT. Now it uses a Vec that would be displayed as a string in the database. In addition the bit size has been added to ease comparison between various of them in Grafana.

### DIFF
--- a/tfhe/benches/boolean/bench.rs
+++ b/tfhe/benches/boolean/bench.rs
@@ -1,6 +1,6 @@
 #[path = "../utilities.rs"]
 mod utilities;
-use crate::utilities::{write_to_json, OperatorType};
+use crate::utilities::{write_to_json, CryptoParametersRecord, OperatorType};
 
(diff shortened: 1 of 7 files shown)


tfhe/benches/boolean/bench.rs
```diff
@@ -16,6 +16,24 @@ criterion_group!(
...
+    display_name: impl Into<String>,
+) {
+    write_to_json(
+        bench_id,
+        params,
+        params_alias,
+        display_name,
+        &OperatorType::Atomic,
+        1,
+        vec![1],
```
IceTDrinker, line 33: those are log values ?

IceTDrinker (approved):
Looks good

tfhe/benches/boolean/bench.rs
IceTDrinker: yes yes I just wanted to be sure those were bit values

PR #233, opened 2023-04-19
chore(ci): create release with semantic-release

soonum:
This will perform automatic release on GitHub. It will append a CHANGELOG file that will be versioned. It will create a new version tag and create a release entry on GitHub along with all the changes in the same format as CHANGELOG.

### DIFF
--- a/.github/workflows/make_release.yml
+++ b/.github/workflows/make_release.yml
@@ -1,18 +1,52 @@
-# Create and publish new release of tfhe-rs using semantic-release.
-name: Make release
+# Publish new release of tfhe-rs on various platform.
+name: Publish release
 
 on:
   workflow_dispatch:
-
-permissions:
-  contents: write
+    inputs:
+      dry_run:
+        description: "Dry-run"
+        type: boolean
+        default: true
 
 jobs:
-  release:
-    name: Release
+  publish_release:
+    name: Publish Release
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
         with:
           fetch-depth: 0
+
+      - name: Publish crate.io package
+        env:
+          CRATES_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+          DRY_RUN: ${{ inputs.dry_run && '--dry-run' || '' }}
+        run: |
+          cargo publish -p tfhe --token ${{ env.CRATES_TOKEN }} ${{ env.DRY_RUN }}
+
+      - name: Build web package
+        run: |
+          make build_web_js_api
+
+      - name: Publish web package
+        uses: JS-DevTools/npm-publish@0f451a94170d1699fd50710966d48fb26194d939
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
+          package: tfhe/pkg/package.json
+          dry-run: ${{ inputs.dry_run }}
+
+      - name: Build Node package
+        run: |
+          rm -rf tfhe/pkg
+
+          make build_node_js_api
+          sed -i 's/"tfhe"/"node-tfhe"/g' tfhe/pkg/package.json
+
+      - name: Publish Node package
+        uses: JS-DevTools/npm-publish@0f451a94170d1699fd50710966d48fb26194d939
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
+          package: tfhe/pkg/package.json
+          dry-run: ${{ inputs.dry_run }}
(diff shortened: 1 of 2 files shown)


.releaserc
```diff
@@ -0,0 +1,35 @@
+{
+  "tagFormat": "v${version}",
```
IceTDrinker, line 2: I don't think we put the v in the tag when we push a tag

scripts/prepare_release.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+
+CARGO_FILE="tfhe/Cargo.toml"
+DOCS_DIR="tfhe/docs"
+DOC_FILES="getting_started/installation.md
+core_crypto/tutorial.md
+high_level_api/serialization.md
+high_level_api/tutorial.md"
+
+NEWTAG=$1
```
IceTDrinker, line 13: new version rather ?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,33 @@
...
+          additional_packages: |
+            [
+              '@semantic-release/github',
+              '@semantic-release/git',
+              '@semantic-release/changelog',
+              '@semantic-release/exec',
+            ]
+          dry_run: ${{ github.event.client_payload.dry_run }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.CONCRETE_ACTIONS_TOKEN }}
```
IceTDrinker, line 33: no crates.io push ?

.releaserc
```diff
@@ -0,0 +1,35 @@
...
+  "verifyConditions": ["@semantic-release/github"],
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
+    [
+      "@semantic-release/changelog",
+      {
+        "changelogFile": "CHANGELOG.md",
```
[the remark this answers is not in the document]
soonum: Look at the root of Slab repository.
Note that the change log in the Release section can still be modified afterward to add some context for example.

scripts/prepare_release.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Update Cargo.toml package version
+sed -i "s|^version = \"[[:digit:]]*\.[[:digit:]]*\.[[:digit:]]*\"|version = \"${NEWTAG}\"|" $CARGO_FILE
+
+# Update package version in documentation
+for file in $DOC_FILES; do
+  sed -i "s|^\(tfhe = { version = \)\"[[:digit:]]*\.[[:digit:]]*\.[[:digit:]]*\"|\1\"${NEWTAG}\"|" "$DOCS_DIR/$file"
+done
```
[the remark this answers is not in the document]
soonum: As long as a line starts with `tfhe = { version = x.y.z`, no I don't think so.
Maybe we could remove the carret to force the regex to match the beginning of the line, wydt ?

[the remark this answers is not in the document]
IceTDrinker: ah yeah, we could remove the caret

IceTDrinker:
does this create commits for the changelog e.g. ?

.releaserc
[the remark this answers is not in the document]
soonum: It's has been already seen with him :wink:

soonum:
> does this create commits for the changelog e.g. ?

It creates a release commit containing all the changes (version upgrade, changelog updates, ...) and the version tag is associated to this commit.

IceTDrinker:
> > does this create commits for the changelog e.g. ?
> 
> It creates a release commit containing all the changes (version upgrade, changelog updates, ...) and the version tag is associated to this commit.

Does it mean it needs to pass the CI ? 🤔

.github/workflows/make_release.yml
```diff
@@ -3,16 +3,62 @@ name: Make release
...
+      - name: Build web package
+        run: |
+          make build_web_js_api
+
+      - name: Publish web package
+        uses: JS-DevTools/npm-publish@0f451a94170d1699fd50710966d48fb26194d939
+        with:
+            token: ${{ secrets.NPM_TOKEN }}
+            package: tfhe/pkg/package.json
+            dry-run: ${{ inputs.dry_run }}
```
IceTDrinker, line 52: I think some metadata needs to be checked with @immortal-tofu and be sure we are pushing something that corresponds to the npm package with the right name ?

.github/workflows/make_release.yml
```diff
@@ -3,16 +3,62 @@ name: Make release
...
+
+      - name: Build Node package
+        run: |
+          make build_node_js_api
```
immortal-tofu, line 56: Does build_node_js_api clear the directory from the previous build?

.github/workflows/make_release.yml
immortal-tofu: It is correct, we discuss it already with @soonum

.github/workflows/make_release.yml
soonum: I can remove `pkg` directory prior to that to ensure this is a fresh package.

immortal-tofu: AFAIK, you build web package in same directory just before. That's why I suggested it.

IceTDrinker (commented):
Small thing but otherwise I think we are there

PR #235, opened 2023-04-19
feat(boolean): add BooleanEngine::replace_thread_local

### LINKED ISSUE #177: Deterministic seeder
Hi! 

Is there a preferred way to use tfhe-rs with purely deterministic and reproducible encryption? The way I'm currently doing this is like so:

```rust
let deterministic_seed = Seed(0);

let mut boolean_engine = BooleanEngine::new_from_seeder(
    Box::new(DeterministicSeeder::<ActivatedRandomGenerator>::new(
        deterministic_seed,
    ))
    .as_mut(),
);

// generate client key
let client_key = boolean_engine.create_client_key(DEFAULT_PARAMETERS);

// generate the server key
let server_key = boolean_engine.create_server_key(&client_key);

// this is deterministic
let ct = boolean_engine.encrypt(true, &client_key);

```

However, this is error-prone. If somewhere in the code, I would accidentally call:

```rust
// this is not deterministic
let ct = client_key.encrypt(true);
```

This falls back to the default non-deterministic BooleanEngine, instead of the deterministic one I created earlier.

What would help here is a deterministic seeding strategy in [seeders.rs](https://github.com/zama-ai/tfhe-rs/blob/main/tfhe/src/core_crypto/seeders.rs). Is there any preferred solution here?

Best,

Michiel

tmontaigu:
This new associated function allows to replace
the engine used in the thread.

Should fix #177


tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test

tfhe/src/boolean/engine/mod.rs
```diff
@@ -257,6 +260,41 @@ impl Default for BooleanEngine {
...
+    /// # Example
+    ///
+    /// ```rust
+    /// use tfhe::boolean::engine::BooleanEngine;
+    /// use tfhe::core_crypto::commons::generators::DeterministicSeeder;
+    /// use tfhe::core_crypto::commons::math::random::Seed;
+    /// use tfhe::core_crypto::prelude::ActivatedRandomGenerator;
+    ///
+    /// // WARNING: Using a deterministic seed is not recommanded
+    /// // as renders the random generation insecure
```
IceTDrinker, line 277: recommended (a -> e)

as **it** renders

tfhe/src/boolean/engine/mod.rs
```diff
@@ -257,6 +260,41 @@ impl Default for BooleanEngine {
...
+    ///
+    /// let deterministic_seed = Seed(0);
+    ///
+    /// // We change the engine in the main thread
+    /// // then generate a client key, and then encrypt
+    /// // a boolean value and serialize it to compare
+    /// // it with other ciphertext
```
IceTDrinker, line 284: looks like this comment is not accurate

tmontaigu: Ah yes it good'ol copy paste from the test

IceTDrinker (approved):
Thanks !

Do we want to cherry-pick this for 0.2.x ?

tmontaigu:
not sure, its not stricly a bugfix

tmontaigu:
@slab-ci cpu_integer_test

IceTDrinker:
> not sure, its not strictly a bugfix

I agree, but this way it allows the person asking for it to be able to use the crates.io version

IceTDrinker:
it's not a breaking change either

PR #247, opened 2023-04-21
refactor(core): refactor CiphertextModulus to be less error prone

IceTDrinker:


tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -1,15 +1,24 @@
...
-/// A value of 0 is always interpreted as a native modulus, this is useful to work with u128 using
-/// the native modulus as $2^{128}$ cannot be stored in a u128 value.
+/// Private enum to avoid end user mis-instantiating a CiphertextModulus
 ///
-/// This also allows to not rely on an enum which would require a discriminant field adding 8 bytes
-/// to store 1 bit of information (native variant vs custom variant with u128 payload).
-pub struct CiphertextModulus<Scalar: UnsignedInteger>(u128, PhantomData<Scalar>);
+/// NonZeroU128 allows to always have a correct modulus and to have an enum that is no bigger than a
+/// u128 with the 0 optimization as the tag then corresponds to the Native variant.
+enum CiphertextModulusInner<Scalar: UnsignedInteger> {
```
IceTDrinker, line 12: This does not need Scalar PhantomData

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -1,15 +1,24 @@
...
+    Native(PhantomData<Scalar>),
+    Custom(NonZeroU128, PhantomData<Scalar>),
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+/// Structure representing a [`CiphertextModulus`] often noted $q$.
+pub struct CiphertextModulus<Scalar: UnsignedInteger> {
+    inner: CiphertextModulusInner<Scalar>,
+}
```
IceTDrinker, line 21: Will need a PhantomData here to remove it from the inner enum

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -49,52 +64,101 @@ impl<'de, Scalar: UnsignedInteger> serde::Deserialize<'de> for CiphertextModulus
...
+            let res = match 1u128.checked_shl(exponent as u32) {
+                Some(modulus) => {
+                    let non_zero_modulus = match NonZeroU128::new(modulus) {
+                        Some(val) => val,
+                        None => {
+                            return Err(
+                                "Got zero modulus for CiphertextModulusInner::Custom variant",
+                            )
+                        }
+                    };
```
tmontaigu, line 107: If I'm correct, here, If we got `Some(modulus)` that means modulus is 100% guaranteed not te be zero,
So we could use the unchecked_new (not in favor of thats) or `.expect(msg)` the `NonZeroU128::new` result
instead of matching on it

but can we panic in const fn ? 🤔

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -103,32 +167,41 @@ impl<Scalar: UnsignedInteger> CiphertextModulus<Scalar> {
     /// implementations than can rely on wrapping arithmetic operations behavior to compute the
     /// modulus.
     pub const fn is_native_modulus(&self) -> bool {
-        if self.0 == 0 {
-            return true;
+        match self.inner {
```
tmontaigu, line 170: could use `matches!(self.inner, Self::Native)`
https://doc.rust-lang.org/std/macro.matches.html

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -145,6 +218,13 @@ mod tests {
 
     #[test]
     fn test_modulus_struct() {
+        assert!(std::mem::size_of::<CiphertextModulus<u32>>() == std::mem::size_of::<u128>());
+        assert!(std::mem::size_of::<CiphertextModulus<u64>>() == std::mem::size_of::<u128>());
+        assert!(std::mem::size_of::<CiphertextModulus<u128>>() == std::mem::size_of::<u128>());
+        assert!(std::mem::align_of::<CiphertextModulus<u32>>() == std::mem::align_of::<u128>());
+        assert!(std::mem::align_of::<CiphertextModulus<u64>>() == std::mem::align_of::<u128>());
+        assert!(std::mem::align_of::<CiphertextModulus<u128>>() == std::mem::align_of::<u128>());
```
tmontaigu, line 226: 👍

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
IceTDrinker: I don't think we can panic and that's why I wrote it this way

IceTDrinker: I'll recheck

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
IceTDrinker: ah right, I never think about that matches macro

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
IceTDrinker: looks like here the const panic is fine if we don't format a string, I would expect the compiler to strip it anyway as we are not supposed to get there, I just went conservative this time as my previous CiphertextModulus struct was a bit error prone (and likely has some UB as M1 tests did not run properly)

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -49,52 +64,99 @@ impl<'de, Scalar: UnsignedInteger> serde::Deserialize<'de> for CiphertextModulus
...
     pub const fn try_new_power_of_2(exponent: usize) -> Result<Self, &'static str> {
         if exponent > Scalar::BITS {
             Err("Modulus is bigger than the maximum value of the associated Scalar type")
         } else {
-            let modulus = match 1u128.checked_shl(exponent as u32) {
-                Some(modulus) => modulus,
-                None => 0,
+            let res = match 1u128.checked_shl(exponent as u32) {
+                Some(modulus) => {
+                    let non_zero_modulus = match NonZeroU128::new(modulus) {
```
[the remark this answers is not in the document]
IceTDrinker: got it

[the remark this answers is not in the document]
IceTDrinker: requires a `#![feature(const_option)]` feature so I think we are stuck with the match statement for now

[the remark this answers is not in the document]
tmontaigu: :sadface:

tmontaigu (approved):

PR #248, opened 2023-04-21
feat(integer): improve scalar shifts performances

tmontaigu:
This reworks the left/right scalar shift method.

This new implementation takes more advantage of the
radix representation property and use a combo
of shifting/moving blocks via a rotate_left/right
optionnaly followed by in-block shifting and propagation
to other blocks.

This new implementation requires the carries to be empty
(not sure what the preconditions were for the previous implementation)
and the output will also have clean carries.
Requiring empty carries allows to do the shift in a way
that scales well with the number of blocks as we can use truly parallel
operations.

This means that the time required to shift is dependent
in the shift value, which should not be a security problem
as its a clear value. (The previous implementation also
had timing that depended on the shift value)

There are two scenarios possible:
- The shift only require to move blocks -> its fast
- The shift requires moving + in-block shifting -> slower,
  but still faster than the previous implementation.

Worst case is when shift is less than the number of bits
in a block.

This also changes the type of the `shift` parameter from
`usize` to `u64` to be consistent with other scalar operations.

The follwing pseudo bench code on 64 bits
gives the following time ranges:

With changes:
```
unchecked_left: BenchStat {
    min: 16.743µs
    max: 526.518563ms
    mean: 150.77871ms
}
unchecked_left_parallelized: BenchStat {
    min: 17.291µs
    max: 94.408455ms
    mean: 30.092279ms
}
unchecked_right: BenchStat {
    min: 16.723µs
    max: 548.417332ms
    mean: 160.345234ms
}
unchecked_right_parallelized: BenchStat {
    min: 16.978µs
    max: 97.955322ms
    mean: 33.500562ms
}
Measured in 37.890296743s
```

Previous code:
```
unchecked_left: BenchStat {
    min: 1.055401595s
    max: 1.156574075s
    mean: 1.085648592s
}
unchecked_left_parallelized: BenchStat {
    min: 559.636545ms
    max: 630.83338ms
    mean: 584.747893ms
}
unchecked_right: BenchStat {
    min: 1.055041354s
    max: 2.314891255s
    mean: 1.644513996s
}
unchecked_right_parallelized: BenchStat {
    min: 562.017144ms
    max: 1.275945891s
    mean: 894.812286ms
}
Measured in 421.412913883s
```

```rust
use rand::Rng;
use std::time::Instant;
use tfhe::integer::{gen_keys, RadixClientKey};
use tfhe::shortint::parameters::PARAM_MESSAGE_2_CARRY_2;

const NB_CTXT: usize = 32;
const NB_TEST: usize = 100;

struct BenchStat {
    min: Option<std::time::Duration>,
    max: Option<std::time::Duration>,
    sum: std::time::Duration,
    count: u32,
}

impl BenchStat {
    fn update(&mut self, elapsed: std::time::Duration) {
        if self.min.is_none() {
            self.min = Some(elapsed);
        } else {
            self.min = self.min.map(|l| l.min(elapsed));
        }
        if self.max.is_none() {
            self.max = Some(elapsed);
        } else {
            self.max = self.max.map(|l| l.max(elapsed));
        }
        self.sum += elapsed;
        self.count += 1;
    }

    fn print(&self) {
        println!("BenchStat {{");
        println!("    min: {:?}", self.min.unwrap());
        println!("    max: {:?}", self.max.unwrap());
        println!("    mean: {:?}", self.sum / self.count);
        println!("}}");
    }
}

type ShiftType = u64;

fn main() {
    let mut unchecked_left_timing = BenchStat::default();
    let mut unchecked_left_parallelized_timing = BenchStat::default();
    let mut unchecked_right_timing = BenchStat::default();
    let mut unchecked_right_parallelized_timing = BenchStat::default();

    let total = Instant::now();

    let param = PARAM_MESSAGE_2_CARRY_2;
    let (cks, sks) = gen_keys(&param);
    let cks = RadixClientKey::from((cks, NB_CTXT));

    //Nb of bits to shift
    let tmp_f64 = param.message_modulus.0 as f64;
    let nb_bits = tmp_f64.log2().floor() as usize * NB_CTXT;
    let modulus = (param.message_modulus.0 as u128).pow(NB_CTXT as u32);
    assert_eq!(nb_bits, 64);

    for i in 0..NB_TEST {
        println!("{} / {NB_TEST}", i + 1);
        let clear = rng.gen::<u128>() % modulus;
        let scalar = rng.gen::<u128>() % nb_bits as u128;

        println!("clear: {clear}, scalar: {scalar}");

        let ct = cks.encrypt(clear);

        {
            let before = Instant::now();
            let ct_res = sks.unchecked_scalar_left_shift(&ct, scalar as ShiftType);
            unchecked_left_timing.update(before.elapsed());
            //assert!(ct_res.block_carries_are_empty());
            let dec_res: u128 = cks.decrypt(&ct_res);
            assert_eq!((clear << scalar) % modulus, dec_res);

            let before = Instant::now();
            let ct_res = sks.unchecked_scalar_left_shift_parallelized(&ct, scalar as ShiftType);
            unchecked_left_parallelized_timing.update(before.elapsed());
            //assert!(ct_res.block_carries_are_empty());
            let dec_res: u128 = cks.decrypt(&ct_res);
            assert_eq!((clear << scalar) % modulus, dec_res);
        }

        {
            let before = Instant::now();
            let ct_res = sks.unchecked_scalar_right_shift(&ct, scalar as ShiftType);
            unchecked_right_timing.update(before.elapsed());
            // assert!(ct_res.block_carries_are_empty());
            let dec_res: u128 = cks.decrypt(&ct_res);
            assert_eq!((clear >> scalar) % modulus, dec_res);

            let before = Instant::now();
            let ct_res = sks.unchecked_scalar_right_shift_parallelized(&ct, scalar as ShiftType);
            unchecked_right_parallelized_timing.update(before.elapsed());
            //assert!(ct_res.block_carries_are_empty());
            let dec_res: u128 = cks.decrypt(&ct_res);
            assert_eq!((clear >> scalar) % modulus, dec_res);
        }
    }

    print!("unchecked_left: ");
    unchecked_left_timing.print();
    print!("unchecked_left_parallelized: ");
    unchecked_left_parallelized_timing.print();

    print!("unchecked_right: ");
    unchecked_right_timing.print();
    print!("unchecked_right_parallelized: ");
    unchecked_right_parallelized_timing.print();

    println!("Measured in {:?}", total.elapsed());
}
```

BREAKING CHANGE: parameter type changed from usize to u64

### DIFF
--- a/tfhe/benches/integer/bench.rs
+++ b/tfhe/benches/integer/bench.rs
@@ -585,6 +585,14 @@ define_server_key_bench_scalar_fn!(
 define_server_key_bench_scalar_default_fn!(method_name: scalar_add_parallelized, display_name: add);
 define_server_key_bench_scalar_default_fn!(method_name: scalar_sub_parallelized, display_name: sub);
 define_server_key_bench_scalar_default_fn!(method_name: scalar_mul_parallelized, display_name: mul);
+define_server_key_bench_scalar_default_fn!(
+    method_name: scalar_left_shift_parallelized,
+    display_name: left_shift
+);
+define_server_key_bench_scalar_default_fn!(
+    method_name: scalar_right_shift_parallelized,
+    display_name: right_shift
+);
 
 define_server_key_bench_scalar_fn!(method_name: unchecked_scalar_add, display_name: add);
 define_server_key_bench_scalar_fn!(method_name: unchecked_scalar_sub, display_name: sub);
@@ -731,6 +739,8 @@ criterion_group!(
     scalar_add_parallelized,
     scalar_sub_parallelized,
     scalar_mul_parallelized,
+    scalar_left_shift_parallelized,
+    scalar_right_shift_parallelized,
 );
 
 criterion_group!(
@@ -790,6 +800,8 @@ criterion_group!(
     scalar_add_parallelized,
     scalar_sub_parallelized,
     scalar_mul_parallelized,
+    scalar_left_shift_parallelized,
+    scalar_right_shift_parallelized,
 );
 
 criterion_main!(
--- a/tfhe/src/integer/server_key/radix/mod.rs
+++ b/tfhe/src/integer/server_key/radix/mod.rs
@@ -157,8 +157,21 @@ impl ServerKey {
     /// assert_eq!(msg + msg, res);
     /// ```
     pub fn full_propagate<PBSOrder: PBSOrderMarker>(&self, ctxt: &mut RadixCiphertext<PBSOrder>) {
+        self.partial_propagate(ctxt, 0);
+    }
+
+    /// Propagates carries from
+    /// start_index to then end.
+    ///
+    /// Last carry is not propagated as
+    /// it has nothing to propagate to.
+    fn partial_propagate<PBSOrder: PBSOrderMarker>(
+        &self,
+        ctxt: &mut RadixCiphertext<PBSOrder>,
+        start_index: usize,
+    ) {
         let len = ctxt.blocks.len();
-        for i in 0..len {
+        for i in start_index..len {
             self.propagate(ctxt, i);
         }
     }
--- a/tfhe/src/integer/server_key/radix/shift.rs
+++ b/tfhe/src/integer/server_key/radix/shift.rs
@@ -1,3 +1,4 @@
+use crate::core_crypto::commons::utils::izip;
 use crate::integer::ciphertext::RadixCiphertext;
 use crate::integer::ServerKey;
 use crate::shortint::PBSOrderMarker;
@@ -28,7 +29,7 @@ impl ServerKey {
     /// let div = cks.parameters().message_modulus.0.pow(shift as u32) as u64;
     ///
     /// // Decrypt:
-    /// let clear: u64 = cks.decrypt(&ct_res);
+    /// let clear = cks.decrypt(&ct_res);
     /// assert_eq!(msg / div, clear);
     /// ```
     pub fn blockshift_right<PBSOrder: PBSOrderMarker>(
@@ -72,7 +73,7 @@ impl ServerKey {
     /// let num_blocks = 4;
     /// let (cks, sks) = gen_keys_radix(&PARAM_MESSAGE_2_CARRY_2, num_blocks);
     ///
-    /// let msg = 128;
+    /// let msg = 128u64;
     /// let shift = 2;
     ///
     /// let ct = cks.encrypt(msg);
@@ -81,13 +82,13 @@ impl ServerKey {
     /// let ct_res = sks.unchecked_scalar_right_shift(&ct, shift);
     ///
     /// // Decrypt:
-    /// let dec: u64 = cks.decrypt(&ct_res);
+    /// let dec = cks.decrypt(&ct_res);
     /// assert_eq!(msg >> shift, dec);
     /// ```
     pub fn unchecked_scalar_right_shift<PBSOrder: PBSOrderMarker>(
         &self,
         ct: &RadixCiphertext<PBSOrder>,
-        shift: usize,
+        shift: u64,
     ) -> RadixCiphertext<PBSOrder> {
         let mut result = ct.clone();
         self.unchecked_scalar_right_shift_assign(&mut result, shift);
@@ -108,7 +109,7 @@ impl ServerKey {
     /// let num_blocks = 4;
     /// let (cks, sks) = gen_keys_radix(&PARAM_MESSAGE_2_CARRY_2, num_blocks);
     ///
-    /// let msg = 18;
+    /// let msg = 18u64;
     /// let shift = 4;
     ///
     /// let mut ct = cks.encrypt(msg);
@@ -117,50 +118,81 @@ impl ServerKey {
     /// sks.unchecked_scalar_right_shift_assign(&mut ct, shift);
     ///
     /// // Decrypt:
-    /// let dec: u64 = cks.decrypt(&ct);
+    /// let dec = cks.decrypt(&ct);
     /// assert_eq!(msg >> shift, dec);
     /// ```
     pub fn unchecked_scalar_right_shift_assign<PBSOrder: PBSOrderMarker>(
         &self,
         ct: &mut RadixCiphertext<PBSOrder>,
-        shift: usize,
+        shift: u64,
     ) {
-        let tmp = self.key.message_modulus.0 as f64;
+        // see parallel implementation for a bit more details
 
-        //number of bits of message
-        let nb_bits = tmp.log2() as usize;
+        debug_assert!(ct.block_carries_are_empty());
+        debug_assert!(self.key.carry_modulus.0 >= self.key.message_modulus.0 / 2);
 
-        // 2^u = 2^{p*q+r} = 2^{p*(q+1)}*2^{r-p}
-        let quotient = shift / nb_bits;
-
-        //p-r
-        let modified_remainder = nb_bits - (shift % nb_bits);
-
-        //if r == 0
-        if modified_remainder == nb_bits {
-            self.full_propagate(ct);
-            self.blockshift_right_assign(ct, quotient);
-        } else {
-            // B/2^u = (B*2^{p-r}) / (2^{p*(q+1)})
-            self.unchecked_scalar_left_shift_assign(ct, modified_remainder);
-
-            // We partially propagate in order to not lose information
-            self.partial_propagate(ct);
-            self.blockshift_right_assign(ct, 1_usize);
-
-            // We propagate the last block in order to not lose information
-            self.propagate(ct, ct.blocks.len() - 2);
-            self.blockshift_right_assign(ct, quotient);
+        if shift == 0 {
+            return;
         }
-    }
 
-    /// Propagates all carries except the last one.
-    /// For development purpose only.
-    fn partial_propagate<PBSOrder: PBSOrderMarker>(&self, ctxt: &mut RadixCiphertext<PBSOrder>) {
-        let len = ctxt.blocks.len() - 1;
-        for i in 0..len {
-            self.propagate(ctxt, i);
+        let num_bits_in_block = self.key.message_modulus.0.ilog2() as u64;
+        let rotations = ((shift / num_bits_in_block) as usize).min(ct.blocks.len());
+        let shift_within_block = shift % num_bits_in_block;
+        let num_blocks = ct.blocks.len();
+
+        // rotate left as the blocks are from LSB to MSB
+        ct.blocks.rotate_left(rotations);
+        for block in &mut ct.blocks[num_blocks - rotations..] {
+            self.key.create_trivial_assign(block, 0)
         }
+
+        if shift_within_block == 0 || rotations == ct.blocks.len() {
+            return;
+        }
+
+        let message_modulus = self.key.message_modulus.0 as u64;
+        let lut = self
+            .key
+            .generate_accumulator_bivariate(|current_block, mut previous_block| {
+                // left shift not to lose
+                // bits when shifting right afterwards
+                previous_block <<= num_bits_in_block;
+                previous_block >>= shift_within_block;
+
+                // The way of getting carry / message is reversed compared
+                // to the usual way but its normal
+                let message_of_current_block = current_block >> shift_within_block;
+                let carry_of_previous_block = previous_block % message_modulus;
+
+                message_of_current_block + carry_of_previous_block
+            });
+        let partial_blocks = ct.blocks[..num_blocks - rotations]
+            .windows(2)
+            .map(|blocks| {
+                // We are right-shifting, which in our representation
+                // means the previous_block (the one with the carry to be
+                // progatated to the current_block) is the next block in the Vec
+                let (current_block, previous_block) = (&blocks[0], &blocks[1]);
+                self.key
+                    .unchecked_apply_lookup_table_bivariate(current_block, previous_block, &lut)
+            })
+            .collect::<Vec<_>>();
+
+        // We do this block separately as this one does not
+        // need to get incoming bits from it neighbour
+        self.key.scalar_right_shift_assign(
+            &mut ct.blocks[num_blocks - rotations - 1],
+            shift_within_block as u8,
+        );
+
+        // We started with num_blocks, discarded 'rotations' blocks
+        // and did the last one separately
+        let blocks_to_replace = &mut ct.blocks[..num_blocks - rotations - 1];
+        assert_eq!(partial_blocks.len(), blocks_to_replace.len());
+        for (block, shifted_block) in izip!(blocks_to_replace, partial_blocks) {
+            *block = shifted_block
+        }
+        debug_assert!(ct.block_carries_are_empty());
     }
 
     /// Computes homomorphically a left shift by a scalar.
@@ -177,7 +209,7 @@ impl ServerKey {
     /// let num_blocks = 4;
     /// let (cks, sks) = gen_keys_radix(&PARAM_MESSAGE_2_CARRY_2, num_blocks);
     ///
-    /// let msg = 21;
+    /// let msg = 21u64;
     /// let shift = 2;
     ///
     /// let ct1 = cks.encrypt(msg);
@@ -186,13 +218,13 @@ impl ServerKey {
     /// let ct_res = sks.unchecked_scalar_left_shift(&ct1, shift);
     ///
     /// // Decrypt:
-    /// let dec: u64 = cks.decrypt(&ct_res);
+    /// let dec = cks.decrypt(&ct_res);
     /// assert_eq!(msg << shift, dec);
     /// ```
     pub fn unchecked_scalar_left_shift<PBSOrder: PBSOrderMarker>(
         &self,
         ct_left: &RadixCiphertext<PBSOrder>,
-        shift: usize,
+        shift: u64,
     ) -> RadixCiphertext<PBSOrder> {
         let mut result = ct_left.clone();
         self.unchecked_scalar_left_shift_assign(&mut result, shift);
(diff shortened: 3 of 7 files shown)


tfhe/src/integer/server_key/radix_parallel/shift.rs
```diff
@@ -2,11 +2,18 @@ use crate::integer::ciphertext::RadixCiphertext;
...
 impl ServerKey {
     /// Computes homomorphically a right shift.
     ///
     /// The result is returned as a new ciphertext.
     ///
+    /// # Requirements
+    ///
+    /// - The blocks parameter's carry space have at least one more bit than message space
+    /// - The input ciphertext carry buffer is emtpy / clean
+    ///
```
tmontaigu, line 16: Add the #output

tfhe/benches/integer/bench.rs
```diff
@@ -731,6 +739,9 @@ criterion_group!(
     scalar_add_parallelized,
     scalar_sub_parallelized,
     scalar_mul_parallelized,
+    scalar_mul_parallelized,
```
IceTDrinker, line 742: twice saclar_mul_parallelized

tfhe/src/integer/server_key/radix/mod.rs
```diff
@@ -157,8 +157,18 @@ impl ServerKey {
...
+        &self,
+        ctxt: &mut RadixCiphertext<PBSOrder>,
+        start_index: usize,
+    ) {
         let len = ctxt.blocks.len();
-        for i in 0..len {
+        for i in start_index..len {
             self.propagate(ctxt, i);
         }
     }
```
IceTDrinker, line 174: is partial propagate really dev only then ?

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -117,50 +117,82 @@ impl ServerKey {
...
+        }
 
-        //if r == 0
-        if modified_remainder == nb_bits {
-            self.full_propagate(ct);
-            self.blockshift_right_assign(ct, quotient);
-        } else {
-            // B/2^u = (B*2^{p-r}) / (2^{p*(q+1)})
-            self.unchecked_scalar_left_shift_assign(ct, modified_remainder);
+        let num_bits_in_block = (self.key.message_modulus.0 as f64).log2() as u64;
```
IceTDrinker, line 137: can self.key.message_modulus.0.ilog2() be enough ? it's the floor of the log2 though

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -117,50 +117,82 @@ impl ServerKey {
...
+        let num_blocks = ct.blocks.len();
 
-            // We partially propagate in order to not lose information
-            self.partial_propagate(ct);
-            self.blockshift_right_assign(ct, 1_usize);
+        // rotate left as the blocks are from LSB to MSB
+        ct.blocks.rotate_left(rotations);
+        for block in &mut ct.blocks[num_blocks - rotations..] {
+            *block = self.key.create_trivial(0);
+        }
```
IceTDrinker, line 146: I was wondering at some point, would be interesting to have a "fill_with_trivial(val: u64)" function on the blocks ?

tfhe/src/integer/server_key/radix_parallel/tests.rs
```diff
@@ -834,25 +834,19 @@ fn integer_unchecked_scalar_left_shift(param: Parameters) {
     // message_modulus^vec_length
     let modulus = param.message_modulus.0.pow(NB_CTXT as u32) as u64;
 
-    //Nb of bits to shift
     let tmp_f64 = param.message_modulus.0 as f64;
     let nb_bits = tmp_f64.log2().floor() as usize * NB_CTXT;
 
     for _ in 0..NB_TEST {
         let clear = rng.gen::<u64>() % modulus;
+        let scalar = rng.gen::<u64>() % nb_bits as u64;
```
IceTDrinker, line 842: don't we want to check the overflow behavior ? We can do checked shift on the clear values

same thing for all tests on shifts that have the same pattern

IceTDrinker (commented):
Got a call, I'll finish the review of the core of the algorithm after it

tfhe/src/integer/server_key/radix/shift.rs
IceTDrinker: to avoid some allocations/assignments and re-use the existing memory

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -117,50 +117,82 @@ impl ServerKey {
...
+                    .unchecked_apply_lookup_table_bivariate(current_block, previous_block, &lut)
+            })
+            .collect::<Vec<_>>();
+        self.key.unchecked_scalar_right_shift_assign(
+            &mut ct.blocks[num_blocks - rotations - 1],
+            shift_within_block as u8,
+        );
+        self.key
+            .message_extract_assign(&mut ct.blocks[num_blocks - rotations - 1]);
+        assert_eq!(partial_blocks.len(), ct.blocks.len() - rotations - 1);
```
IceTDrinker, line 188: could we maybe create the mutable slice first and then compare its length to the partial_blocks len ? Would make the comparison a bit less confusing, or maybe use the izip! macro from core crypto I think it does some bound check (or zip checked) but I'm sure we have a zip that verifies iterators have the same len

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -222,15 +263,76 @@ impl ServerKey {
...
     /// ```
     pub fn unchecked_scalar_left_shift_assign<PBSOrder: PBSOrderMarker>(
         &self,
         ct: &mut RadixCiphertext<PBSOrder>,
-        shift: usize,
+        shift: u64,
     ) {
-        let tmp = 1_u64 << shift;
-        self.smart_scalar_mul_assign(ct, tmp);
+        // see parallel implementation for a bit more details
```
IceTDrinker, line 274: comments from right shift may apply here as well

tfhe/src/integer/server_key/radix_parallel/shift.rs
```diff
@@ -68,35 +89,105 @@ impl ServerKey {
...
+        let (partial_blocks, block) =
+            rayon::join(create_blocks_using_bivariate_pbs, shift_last_block);
+
+        debug_assert_eq!(partial_blocks.len(), ct.blocks.len() - rotations - 1);
+        for (block, shifted_block) in ct.blocks[..num_blocks - rotations - 1]
+            .iter_mut()
+            .zip(partial_blocks.into_iter())
+        {
+            *block = shifted_block
         }
```
IceTDrinker, line 188: could be parallelized ? not sure we win a lot

tfhe/src/integer/server_key/radix_parallel/shift.rs
```diff
@@ -268,10 +365,99 @@ impl ServerKey {
...
+        // We now have to put the blocks back togeter
+        debug_assert_eq!(partial_blocks.len(), ct.blocks.len() - rotations - 1);
+        ct.blocks[rotations] = block;
+        for (block, shifted_block) in ct.blocks[rotations + 1..]
+            .iter_mut()
+            .zip(partial_blocks.into_iter())
+        {
+            *block = shifted_block
+        }
+
```
IceTDrinker, line 459: could be parallelized potentially, not sure we would win much

tfhe/src/integer/server_key/radix/shift.rs
tmontaigu: Yes that would be cool, that would have much impact on performances, but would make the code look nicer i think

tfhe/src/integer/server_key/radix/mod.rs
tmontaigu: I've left the comment as it was originally, but yes its not really a dev only (as in you use it to debug) function

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -117,50 +117,82 @@ impl ServerKey {
...
+                    // bits when shifting right afterwards
+                    current_block <<= num_bits_in_block;
+                    previous_block <<= num_bits_in_block;
+
+                    current_block >>= shift_within_block;
+                    previous_block >>= shift_within_block;
+
+                    // The way of gettint carry / message is reversed compared
+                    // to the usual way but its normal
+                    let message_of_current_block = current_block >> num_bits_in_block;
```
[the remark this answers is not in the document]
tmontaigu: no it wont overflow the full code is

```                    // left shift not to lose
                    // bits when shifting right afterwards
                    current_block <<= num_bits_in_block;
                    previous_block <<= num_bits_in_block;

                    current_block >>= shift_within_block;
                    previous_block >>= shift_within_block;
```

Where current_block and previous_block are in [0...message_modulus-1] 
and shift_within_block < num_bits_in_block, because shift_within_block  = shift % num_bits_in_block

The we first left shift by a value that is strictly greater than the value we are right shifting afterwards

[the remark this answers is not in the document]
IceTDrinker: yeah but here you re shift by num_bits_in_block right ?

tfhe/src/integer/server_key/radix_parallel/shift.rs
tmontaigu: Entering a thread/switching context to a thread just to replace (replace + implicit drop) a value won't provide any

tfhe/src/integer/server_key/radix_parallel/shift.rs
tmontaigu: Entering a thread/switching context to a thread just to replace (replace + implicit drop) a value won't provide any gains

tfhe/src/integer/server_key/radix_parallel/shift.rs
IceTDrinker: right 🙄

tfhe/src/integer/server_key/radix_parallel/tests.rs
tmontaigu: Yes we should have test for when the input shift is > nb_bits.

the current behavior is that implemented is (should be)  equivalent to `x.checked_shl(y).unwrap_or(0)`, that is, if the shift is bigger than nb_bits the bits are discarded. Which is what I'm accustomed to.

default for rust is a bit different, something like `3u8 << 232` won't compile in debug mode with 
```
 attempt to shift left by `232_i32`, which would overflow
```

In release mode it compiles and gets replaced with something like `3u8 << (232 & 7)` which as en quivialent
`3u8.wrapping_shl(232)`.

What default should be chose ?

IceTDrinker: I would expect an overflowing shift to output 0 I think, I think that as long as it's well documented it should be fine

tfhe/src/integer/server_key/radix/shift.rs
tmontaigu: It should work, but it would bump the rust requirement to 1.67

IceTDrinker: ah damn, I think I put some ilog2 somewhere 🤦‍♂️

IceTDrinker: we need to upgrade our rust version -_-

tmontaigu: So i can put it

tfhe/src/integer/server_key/radix/mod.rs
tmontaigu: changed the comment

tmontaigu:
@slab-ci cpu_test

tfhe/src/integer/server_key/radix/shift.rs
tmontaigu: actually we have a `create_trivial_assign` I should use that intead

tfhe/src/integer/server_key/radix/shift.rs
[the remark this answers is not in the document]
IceTDrinker: I agree on the good result, I just wonder if we should prefer a shift which underflows or the division the same way you did for the other scalar shift :)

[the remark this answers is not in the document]
tmontaigu: Visualizing shift works better in my brain than division, and the underflow here is defined behaviour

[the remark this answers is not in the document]
IceTDrinker: then all good

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -222,15 +267,85 @@ impl ServerKey {
...
+            .collect::<Vec<_>>();
+
+        // We do this block separately as this one does not
+        // need to get incoming bits from it neighbour
+        self.key.unchecked_scalar_left_shift_assign(
+            &mut ct.blocks[rotations],
+            shift_within_block as u8,
+        );
+        // as left shift is not done via PBS we message extract
+        self.key.message_extract_assign(&mut ct.blocks[rotations]);
```
IceTDrinker, line 337: let's perform a single PBS here maybe ?

tfhe/src/integer/server_key/radix_parallel/shift.rs
```diff
@@ -268,10 +367,103 @@ impl ServerKey {
...
+        let (partial_blocks, block) =
+            rayon::join(create_blocks_using_bivariate_pbs, shift_last_block);
+
+        // We started with num_blocks, discarded 'rotations' blocks
+        // and did the last one separately
+        let expected_num_partial_block = ct.blocks.len() - rotations - 1;
+        assert_eq!(partial_blocks.len(), expected_num_partial_block);
+        ct.blocks[rotations] = block;
+        for (block, shifted_block) in ct.blocks[rotations + 1..]
```
IceTDrinker, line 459: I still believe something like

let dest_blocks = &mut ct.blocks[rotations + 1..];
assert_eq!(partial_blocks.ln(), dest_blocks.len());
for ... in dest_blocks.iter_mut()...

or using izip!/zip checked would be more readable 🤔

tfhe/src/integer/server_key/radix_parallel/tests.rs
```diff
@@ -866,98 +858,71 @@ fn integer_default_scalar_left_shift(param: Parameters) {
...
     // message_modulus^vec_length
     let modulus = param.message_modulus.0.pow(NB_CTXT as u32) as u64;
 
-    //Nb of bits to shift
-    let tmp_f64 = param.message_modulus.0 as f64;
-    let nb_bits = tmp_f64.log2().floor() as usize * NB_CTXT;
-
     for _ in 0..NB_TEST {
         let clear = rng.gen::<u64>() % modulus;
+        let scalar = rng.gen::<u32>();
```
IceTDrinker, line 914: sorry to be annoying but here most of the time we would be testing the oveflowing scalar shift, can we have two halves to this test ? one where we know we are not overflowing/underflowing and the other where we force the overflow/underflow ?

potentially being sure we test "in ciphertext shifts" by taking shift in 0..message_modulus would be good here

IceTDrinker: I think we have seen in the past that corner cases are an important part of our testing and here I don't think we are covering the different shift behaviors

tfhe/src/integer/server_key/radix/shift.rs
IceTDrinker: I mean perform shift + message extract in one go

tmontaigu: yeah we can

tfhe/src/integer/server_key/radix_parallel/tests.rs
tmontaigu: should this be in to different test functions or in the same one ?

IceTDrinker: it could be the same test function as long as we have the 3/4 sub cases well defined

IceTDrinker: could be in their own blocks like
//Generic case
{

}

// Overlow case
{

}

// All small shifts
{
for shift in 0..message_modulus or carry_modulus {
}

tmontaigu: testing for all small shift for each test will increase time required to test, is that ok ?

IceTDrinker: hmm maybe we won't loop for those (like no need to repeat the test 10 times) ? I let you be the judge of that

tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -117,50 +117,82 @@ impl ServerKey {
...
+            .key
+            .generate_accumulator(|x| (x % message_modulus) >> shift_within_block);
+        self.key
+            .apply_lookup_table_assign(&mut ct.blocks[num_blocks - rotations - 1], &lut);
+
+        // We started with num_blocks, discarded 'rotations' blocks
+        // and did the last one separately
+        let blocks_to_replace = &mut ct.blocks[..num_blocks - rotations - 1];
+        assert_eq!(partial_blocks.len(), blocks_to_replace.len());
+        for (block, shifted_block) in itertools::izip!(blocks_to_replace, partial_blocks) {
```
IceTDrinker, line 192: sorry if I was unclear, izip was needed only if it was checking the interator lengths itself, I was fine with the way it was written before, I just found the length check hard to read

btw we have our own izip reimported in core to avoid the itertools dependency

tmontaigu: ah so what should i do 😁?

IceTDrinker: if you like the izip! look then I woudl use the one hidden in core_crypto

otherwise the previous way was fine

tmontaigu: ok gonna import the izip from core_crypto

tfhe/src/integer/server_key/radix_parallel/tests.rs
tmontaigu: I put the small shift outside the main test loop

tfhe/src/integer/server_key/radix/shift.rs
IceTDrinker: looks like this one is not fused in a single PBS?

IceTDrinker (approved):
Pending the last missing PBS "merge" I think this looks good !

Thanks a lot 🙏

PR #258, opened 2023-04-25
refactor(integer): remove usage of Mutex for determinism

IceTDrinker:
cherry-pick you know the drill


IceTDrinker:
or maybe we don't need it for 0.2 unsure

PR #261, opened 2023-04-27
refactor(shortint): add encryption key choice in parameters

IceTDrinker:
BREAKING CHANGE:
- Parameters layout change
- C API removal of SHORTINT_NATIVE_MODULUS which was a leftover from a refactor

Add checks to verify the encryption is using the proper kind of parameters


tfhe/src/shortint/parameters/mod.rs
```diff
@@ -10,12 +10,28 @@ pub use crate::core_crypto::commons::parameters::{
...
 
 pub mod parameters_wopbs;
 pub mod parameters_wopbs_message_carry;
 pub(crate) mod parameters_wopbs_prime_moduli;
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Serialize, Deserialize)]
+pub enum EncryptionKeyChoice {
+    Big,
+    Small,
+}
```
tmontaigu, line 24: This could use a small docstring, mainly to quicly explain the tradeoff between the possible choices

tfhe/src/shortint/engine/client_side.rs
tmontaigu, line 1: The panic could be replaced with Err(msg). But its not like it really matters as the resulting engine error would get unwraped by the ClientKey method so no need to change

tmontaigu (commented):
Its cool because in the HLAPI, in my wip refactor branch I need an enum EncryptionKeyChoice

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: agreed !

tfhe/src/shortint/engine/client_side.rs
IceTDrinker: (as algorithms would panic on mismatched LWE dim in KS or PBS)

IceTDrinker: Err(msg) would be easier to test, I think I'll use an err and add a small test case to see we are panicking with the right error message

IceTDrinker: more work than I would have liked, let me know what you think about it

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: feedback wanted on that docstring

tfhe/src/shortint/engine/client_side.rs
tmontaigu: looks good to me

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

IceTDrinker:
@slab-ci cpu_test

PR #263, opened 2023-04-27
chore(ci): start benches only on our repo

IceTDrinker:


.github/workflows/start_benchmarks.yml
```diff
@@ -4,11 +4,12 @@ name: Start all benchmarks
 on:
   push:
     branches:
-      - 'main'
+      - "main"
   workflow_dispatch:
 
 jobs:
   start-benchmarks:
+    if: ${{ (github.event_name == 'push' && github.repository == 'zama-ai/tfhe-rs') ||  github.event_name == 'workflow_dispatch' }}
```
soonum, line 12: I don't quite get it. Does this mean that push on other repo (forked ones ?) can trigger the CI ?

IceTDrinker: no it's to avoid double launching with internal

soonum: Ok makes sense now.

soonum (approved):

PR #264, opened 2023-04-27
chore(tfhe): as seen there are uses of ilog2 which come from rust 1.67

IceTDrinker:


scripts/check_cargo_min_ver.sh
tmontaigu, line 1: I tried  so that we don't need to update this file when min version changes.
However, it won't work with 'stock' macOs

```bash
ABS_SCRIPT_PATH=$(readlink -f "$0")
ABS_SCRIPT_DIR=$(dirname "$ABS_SCRIPT_PATH")
REL_CARGO_TOML_PATH="../tfhe/Cargo.toml"
ABS_CARGO_TOML_PATH=$(readlink -e "$ABS_SCRIPT_DIR/$REL_CARGO_TOML_PATH")

RUST_TOOLCHAIN=""

MIN_RUST_VERSION=$(grep rust-version $ABS_CARGO_TOML_PATH | cut -d '=' -f 2 | xargs)

function usage() {
    echo "$0: check minimum cargo version"
    echo
    echo "--help                    Print this message"
    echo "--rust-toolchain          The toolchain to check the version for with leading"
    echo "--min-rust-version        Check toolchain version is >= to this version, default is $MIN_RUST_VERSION"
    echo
}
```

tmontaigu (approved):

scripts/check_cargo_min_ver.sh
IceTDrinker: what's the issue with mac OS ? I have the office M1 to check and try to adapt

tmontaigu: the `readlink` shipped with macos is likely to have a different behaviour from the GNU one shipped in GNU/Linux (allow me to interject...)

to can be "fixed" by brew installing gnucoreutils

IceTDrinker: I'll see what I can come up with

IceTDrinker: got something that worked ok on the M1 mac

scripts/check_cargo_min_ver.sh
```diff
@@ -7,13 +7,14 @@ function usage() {
     echo
     echo "--help                    Print this message"
     echo "--rust-toolchain          The toolchain to check the version for with leading"
-    echo "--min-rust-version        Check toolchain version is >= to this version, default is 1.65"
+    echo "--min-rust-version        Check toolchain version is >= to this version, default is 1.67"
     echo
 }
```
tmontaigu, line 12: the usage function should be defined below so the help string can read the `MIN_RUST_VERSION` variable

IceTDrinker: oh yeah, that's a mistake on my end

tmontaigu (approved):

PR #266, opened 2023-04-27
feat(integer): add parallelized scalar rotate_left/right

### LINKED ISSUE #145: Add rotation (left/right) operation to integer digits
At the moment, no rotate (left/right) operation is available for the shortint/integer digits.

Adding to the shortint/integers API two methods:
- rotate_left
- rotate_right

Only manual implementation can offer an alternative at the moment.

tmontaigu:
Like shifts, rotates are implemented by combining
a rotation of the block and bivariate PBSs in case the rotation number `n` is not a multiple of the number of bits in a block.

Since the behaviour of rotations is to 'cycle' bits back the end/beginning of the 'bit slice' (i.e. no bits is ever lost like it can with shifts), the performance is always the same when (n % nb_bits_in block) != 0. However the implementation is simpler.

So assuming a machine where the number of threads
is >= to the ciphertext's number of block, the operation cost one bivariate PBS.

closes #145

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/rotate.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/rotate.rs
@@ -1,0 +1,665 @@
+use crate::integer::ciphertext::RadixCiphertext;
+use crate::integer::ServerKey;
+use crate::shortint::PBSOrderMarker;
+
+use rayon::prelude::*;
+
+impl ServerKey {
+    //======================================================================
+    //                Rotate Right
+    //======================================================================
+
+    /// Computes homomorphically a rotation of bits.
+    ///
+    /// Shifts the bits to the right by a specified amount,
+    /// `n`, wrapping the truncated bits to the beginning of the resulting integer.
+    ///
+    /// If necessary the carries of the input will be cleaned beforehand,
+    /// but its value won't change, the result is returned in a new ciphertext
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use tfhe::integer::gen_keys_radix;
+    /// use tfhe::shortint::parameters::PARAM_MESSAGE_2_CARRY_2;
+    ///
+    /// // We have 4 * 2 = 8 bits of message
+    /// let size = 4;
+    /// let (cks, sks) = gen_keys_radix(&PARAM_MESSAGE_2_CARRY_2, size);
+    ///
+    /// let msg = 128u8;
+    /// let n = 2;
+    ///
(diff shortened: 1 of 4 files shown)


tfhe/src/integer/server_key/radix_parallel/rotate.rs
```diff
@@ -0,0 +1,665 @@
...
+            return;
+        }
+
+        let rotations = (n / num_bits_in_block) as usize;
+        let shift_within_block = n % num_bits_in_block;
+        let num_blocks = ct.blocks.len();
+
+        dbg!(rotations);
+
+        // rotate left as the blocks are from LSB to MSB
```
IceTDrinker, line 623: rotate right, there may be the same typo in the previous shift code

tmontaigu:
> Looks generally good, I'm wondering if we maybe should have a special test case to be sure we handle the rotation part when bits "overflow/underflow" ?

We could but the random tests do not 'modularize' the random picked rotatation number, so that case il fairly certain to be tested i would believe

IceTDrinker:
alright, it's just as seen before our statistical testing was not covering some edge cases enough, it can be a few hardcoded tests, or  even do as you did and lower the number of loops but with specialized test cases

tmontaigu:
So what we can do is add a third case un the loop similar to the `// Force case where n is not multiple of block size`
and add `nb_bits` to the rotation number so we are guaranteed to have a overlow case

IceTDrinker:
> So what we can do is add a third case in the loop similar to the `// Force case where n is not multiple of block size` and add `nb_bits` to the rotation number so we are guaranteed to have a overflow case

sounds like it should work well yes

tmontaigu:
> > So what we can do is add a third case in the loop similar to the `// Force case where n is not multiple of block size` and add `nb_bits` to the rotation number so we are guaranteed to have a overflow case
> 
> sounds like it should work well yes

thinking about it, the thing i proposed is useless,
in rotations, bit alway circle around, if you rotate by 1, the then 1 bit circles back to the end/beginnin

IceTDrinker:
well there is a chance the rotation is just a shift, I just want to be sure we cover the rotation case

tfhe/src/integer/server_key/radix_parallel/rotate.rs
tmontaigu: I checked, shifts did not have this typo

IceTDrinker:
@slab-ci cpu_test

IceTDrinker (approved):
Thanks a lot for the added test cases !

PR #267, opened 2023-04-28
chore(shortint): fix some wopbs function signature

IceTDrinker:


IceTDrinker:
will need a full green on shortint (will re run if wopbs flaky shows up)

tmontaigu (approved):

PR #273, opened 2023-05-05
refactor(hlapi): remove RefCell used to wrap integers

tmontaigu:
Our integer types are based on tfhe::integer.

Originally, operators (+, -, *, <<, etc) were mapped to "smart" operations until commit "ee96a0ff185fedb9c4467a5b0c8195798c30b19f" where we swapped to "default" ops.

The motivation of swapping from smart to default was that default ops timing was always the same, so its easier to predict and reason about when comparing with available benchmarks.
However, they may give worse performance depending on the computations being done (addition/subtraction heavy or not).

In the High level API, we overloaded operators on const ref e.g. `&a + &b` but as we initially mapped to smart operations we needed interior mutability. RefCell was chosen, mainly because using Mutexes would have allowed users to write "fake" parallel code, that is, the code compiles but is not truly parallel due to the mutexes. RefCell makes writting parallel code harder but If you manage to do it, its truly parallel.

After moving to default ops for the hlapi, we kept the inner RefCells to allow time to decide to retract the change.

The final choice is to keep default ops as the default, so that means we can remove the RefCells.

Entry points to smart operation will be added later to enable 'power users' to explicitely try them and see if they bring improvement(s) for them.
Letting advanced user explicitely handle mutability of smart operations and let them choose their synchronisation is likely a better choice than making it for them as its an important choice.

### DIFF
--- a/tfhe/src/c_api/high_level_api/integers.rs
+++ b/tfhe/src/c_api/high_level_api/integers.rs
@@ -16,12 +16,12 @@ macro_rules! impl_operations_for_integer_type {
         name: $name:ident,
         clear_scalar_type: $clear_scalar_type:ty
     ) => {
-        impl_binary_fn_on_type_mut!($name => add, sub, mul, bitand, bitor, bitxor, eq, ge, gt, le, lt, min, max);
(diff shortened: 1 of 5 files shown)


IceTDrinker (approved):
Kill it with fire! 🔥

PR #275, opened 2023-05-05
hlapi casting

tmontaigu:
requires:
 - [x] #273 

This adds the casting of integer types.

Downcasting truncates blocks.
Upcasting appends 0s

Casting is done via the introduced `cast_from` associated
function and the `cast_into` method. They are the equivalent of
the `From` and `Into` traits.

It was not possible to implement casting by implementing the
standard `From` and `Into` traits as initially planned because:

```rust
impl<P1, P2> From<GenericInteger<P1>> for GenericInteger<P2>
where P1: IntegerParameter,
      P2: Integer Parameter, {
    fn from(_: GenericInteger<P1>) -> Self {
        todo!()
    }
}
```

As it conflicts the blanket impl found in the stdlib
`impl<T> From<T> for T;` as P1, P2 may be the same and we have no way
of telling the compiler to consider this impl only when P1 != P2.
So we had to create our own stuff

### DIFF
--- a/tfhe/c_api_tests/test_high_level_256_bits.c
+++ b/tfhe/c_api_tests/test_high_level_256_bits.c
@@ -9,6 +9,7 @@ int uint256_client_key(const ClientKey *client_key) {
   FheUint256 *lhs = NULL;
   FheUint256 *rhs = NULL;
   FheUint256 *result = NULL;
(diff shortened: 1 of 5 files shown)


IceTDrinker (approved):
🔥 🔥 awesome stuff!

PR #278, opened 2023-05-07
FHE Regex Pattern Matching Engine

RKlompUU:
Most important locations:

- the tutorial is in `tfhe/docs/regex/tutorial.md`
- the example implementation is in `tfhe/src/regex/..`

Not quite sure if you want the example implementation to be under `tfhe/src` directly (how it is in this PR atm), or to have it be more of an external dependency (like how I initially had set it up in a separate repo). Let me know.

I might still want to add a `Further work` section at the end of the tutorial, and to add some comments to the code. But I think it's best to open this PR now to get the final review process started.

---

To run the included tests run: `cargo test --release --features=x86_64,integer --example regex_engine`.

### DIFF
--- a/tfhe/Cargo.toml
+++ b/tfhe/Cargo.toml
@@ -30,6 +30,11 @@ itertools = "0.10.5"
 num_cpus = "1.15"
 # For erf and normality test
 libm = "0.2.6"
+test-case = "*"
+combine = "*"
+anyhow = "*"
+env_logger = "*"
+log = "*"
 
 [build-dependencies]
 cbindgen = { version = "0.24.3", optional = true }
@@ -191,6 +196,10 @@ required-features = ["shortint", "internal-keycache"]
 [[example]]
 name = "micro_bench_and"
 required-features = ["boolean"]
+
+[[example]]
+name = "regex_engine"
+required-features = ["integer"]
 
 [lib]
 crate-type = ["lib", "staticlib", "cdylib"]
--- a/tfhe/docs/regex/tutorial.md
+++ b/tfhe/docs/regex/tutorial.md
@@ -1,0 +1,466 @@
+# FHE Regex Pattern Matching Tutorial
+
+This tutorial explains how to build a Regex pattern matching engine where the
+content that is matched against is ciphertext.
+
+A regex Pattern Matching Engine (PME) is an essential tool for programmers. It
+allows to perform complex searches on a content. The less powerful simple
+search on string can only find matches of the exact given sequence of
+characters (eg, your Browser's default search function does this). Regex PME
+are more powerful, it allows to search on certain structures of text (where a
+structure may take form in multiple possible sequences of characters). The
+structure to be searched for is defined with the regex, a very concise
+language. Here are some example regexes to give some idea of what is possible:
+
+Regex | Semantics
+--- | ---
+/abc/ | Searches for the sequence `abc` (equivalent to the simple text search)
+/^abc/ | Searches for the sequence `abc` at the beginning of the content
+/a?bc/ | Searches for sequences `abc`, `bc`
+/ab\|c+d/ | Searches for sequences of `ab`, `c` repeated 1 or more times followed by `d`
+
+Regexes are powerful enough to be able to express structures like email address
+formats. Capability like this is what makes regexes useful for many programming
+solutions.
+
+There are two main components identifiable in a PME:
+1. the pattern that is to be matched has to be parsed, this translates from a
+   textual representation into a recursively structured object (an Abstract
+   Syntax Tree, or AST)
+2. this AST has to then be applied to the text that is to be matched against,
+   resulting in a yes or no to whether the pattern matched (and in the case of
+   our FHE implementation, this result is an encrypted yes or an encrypted no)
+
+Parsing is a well understood problem. There are a couple of different
+approaches possible here.  Regardless of the approach chosen, it starts with
+figuring out what the language is that we want to support. That is, what are
+the kinds of sentences that we want our regex language to include?  A few
+example sentences we definitely want to support are for example: `/a/`,
+`/a?bc/`, `/^ab$/`, `/ab|cd/`, however example sentences don't suffice here as
+a specification because they can never be exhaustive (they're endless). We need
+something to specify _exactly_ the full set of sentences our language supports.
+There exists a language that can help us describe exactly what our own
+language's structure is: Grammars.
+
+## The Grammar and Datastructure
+
+It is useful to start with defining the Grammar before starting to write
+the code for the parser. Because the code structure follows directly from the
+Grammar. A Grammar consists of a (generally small) set of rules. For example,
+a very basic Grammar could look like this:
+```
+Start := 'a'
+```
+this describes a language that only contains the sentence "a". Not a very interesting language.
+
+We can make it more interesting though by introducing choice into the grammar
+with \| (called a 'pipe') operators. If we want the above grammar to accept
+either "a" or "b":
+```
+Start := 'a' | 'b'
+```
+
+So far only Grammars with a single rule have been shown. However, a Grammar can
+consist of multiple rules. And in fact, most languages require to be defined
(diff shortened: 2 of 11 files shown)


IceTDrinker:
first request for this PR: squash in a single commit and follow the commit message format, in that case it should be

```
docs(tfhe): add FHE Regex Pattern Matching Engine

(... more details)
```

tfhe/Cargo.toml
```diff
@@ -61,15 +62,19 @@ console_error_panic_hook = { version = "0.1.7", optional = true }
...
 boolean = []
 shortint = []
 integer = ["shortint"]
+regex = ["integer", "dep:combine", "dep:anyhow"]
 internal-keycache = ["lazy_static", "fs2", "bincode"]
 
 # Experimental section
-experimental = []
+experimental = ["experimental-multi_bit_pbs"]
+experimental-multi_bit_pbs = []
```
tmontaigu, line 77: These features do not exist anymore, i guess this comes from a rebasing

tfhe/test_data/client_key
tmontaigu, line 1: We want to avoid committing binary files to the repo

Also as you are using param_message_2_carry_2, we can afford regerating the key on each run

tfhe/Cargo.toml
```diff
@@ -92,6 +97,8 @@ shortint-client-js-wasm-api = ["shortint", "__wasm_api"]
 
 nightly-avx512 = ["concrete-fft/nightly", "pulp/nightly"]
 
+gen_test_keys = []
+
```
tmontaigu, line 101: We should be able to afford regenerating the key on each run for this example

tmontaigu (commented):
I still need to review a bit more deeply I looked good from what i've seen so far, I've made some comments.

> Not quite sure if you want the example implementation to be under tfhe/src directly (how it is in this PR atm), or to have it be more of an external dependency (like how I initially had set it up in a separate repo). Let me know.

Yes we would prefer the implementation to be more separated from the lib.
Normally you should be able to have all your code under a sub folder of the `examples` folder
You could take at look at this other bounty where its organized in the example folder https://github.com/zama-ai/tfhe-rs/pull/283/files

tfhe/src/regex/ciphertext.rs
```diff
@@ -0,0 +1,45 @@
...
+            if msg & (1 << (block * block_size + bit)) != 0 {
+                let base: usize = 2;
+                block_value += base.pow(bit as u32);
+            }
+        }
+        vec_res.push(shortkey.create_trivial(block_value as u64));
+    }
+
+    RadixCiphertextBig::from(vec_res)
+}
```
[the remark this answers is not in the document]
RKlompUU: Yeah it wasn't available yet when I was working on the bounty. I'll replace it with the create_trivial_radix function provided by tfhe.

tfhe/Cargo.toml
RKlompUU: Yeah you're right, I think I added this feature in when I hadn't realized yet that re-generation of the key is already prevented with `lazy_static` .

tfhe/test_data/client_key
RKlompUU: Makes sense! Removing this `gen_test_keys` feature.

tfhe/Cargo.toml
RKlompUU: correct, went wrong during a rebase merge. removing

tfhe/docs/regex/tutorial.md
```diff
@@ -0,0 +1,514 @@
...
+`/[^ab]/` | `RegExpr::Not { not_re: Box::new(RegExpr::Range { cs: vec!['a', 'b'] }) }`
+`/av\|d?/` | `RegExpr::Either { l_re: Box::new(RegExpr::Seq { re_xs: vec![RegExpr::Char { c: 'a' }, RegExpr::Char { c: 'v' }] }), r_re: Box::new(RegExpr::Optional { opt_re: Box::new(RegExpr::Char { c: 'd' }) }) }`
+`/(av\|d)?/` | `RegExpr::Optional { opt_re: Box::new(RegExpr::Either { l_re: Box::new(RegExpr::Seq { re_xs: vec![RegExpr::Char { c: 'a' }, RegExpr::Char { c: 'v' }] }), r_re: Box::new(RegExpr::Char { c: 'd' }) }) }`
+
+With both the Grammar and the datastructure to parse into defined, we can now
+start implementing the actual parsing logic. There are multiple ways this can
+be done. For example there exist tools that can automatically generate parser
+code by giving it the Grammar definition (these are called parser generators).
+However, I prefer to write parsers myself with a parser combinator library, as
+in my opinion the behavior in runtime is better understandable of these than of
```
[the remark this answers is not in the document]
RKlompUU: I think the real issue is that the sentence is too long. Broke it into 2 sentences, WDYT?

RKlompUU:
I addressed each comment that's been left here so far.

I moved the example implementation from `src/regex` to `examples/regex_engine`. This did mean that the final section of the tutorial made no sense anymore as far as I can tell. Instead I replaced that with a section on how to run the example implementation.

tfhe/examples/regex_engine/engine.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn test_has_match(content: &str, pattern: &str, exp: u64) {
+        // it would be more correct to encrypt the content properly with the
+        // client key, however in order to significantly speed up these test
+        // cases we're encoding them trivially into ciphertext
+        let ct_content: StringCiphertext = content
+            .as_bytes()
+            .iter()
+            .map(|byte| KEYS.1.create_trivial_radix(*byte as u64, 4))
+            .collect();
+        let ct_res = has_match(&KEYS.1, &ct_content, pattern).unwrap();
```
tmontaigu, line 263: Here, we want a real encryption using the client key, so that if we change parameters in PARAM_MESSAGE_2_CARRY_2 would would know it that breaks the regex example.

And using real encryption does not seem to be that slower, well its 2x slower, but it only goes from 2sec to 4sec so its fine

tfhe/examples/regex_engine/parser.rs
```diff
@@ -0,0 +1,685 @@
...
+
+fn atom<Input>() -> impl Parser<Input, Output = RegExpr>
+where
+    Input: Stream<Token = u8>,
+    Input::Error: ParseError<Input::Token, Input::Range, Input::Position>,
+{
+    choice((
+        byte(b'.').map(|_| RegExpr::AnyChar),
+        attempt(byte(b'\\').with(parser::token::any())).map(|c| RegExpr::Char { c }),
+        choice((byte::letter(), parser::token::one_of(NON_ESCAPABLE_SYMBOLS)))
```
tmontaigu, line 264: instead of `byte::letter()` I think it would be better to use `byte::alpha_num` to be able to do regex with numbers, the range rule (and maybe others) would also  need to be update
(and add tests)

tfhe/test_data/client_key
tmontaigu, line 1: This file should be removed

tfhe/examples/regex_engine/execution.rs
```diff
@@ -0,0 +1,275 @@
...
+        )
+    }
+
+    fn with_cache(&mut self, ctx: Executed, f: LazyExecution) -> ExecutedResult {
+        if let Some(res) = self.cache.get(&ctx) {
+            //trace!("cache hit: {:?}", &ctx);
+            self.cache_hits += 1;
+            return (res.clone(), ctx);
+        }
+        //debug!("evaluation for: {:?}", &ctx);
```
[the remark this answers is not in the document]
RKlompUU: Ah, it should be uncommented. I wanted to uncomment them (now that this is again a standalone solution as a `example/..`).

tfhe/examples/regex_engine/parser.rs
RKlompUU: Other comments addressed. This comment I will address later today after work, as it requires a bit work.

tmontaigu (approved):

PR #281, opened 2023-05-09
refactor(shortint): make a difference between PBS and Wopbs parameters

IceTDrinker:
- preparatory work to manage several PBS implementations and harmonize parameters management

BREAKING CHANGE:
parameters structures changed

### DIFF
--- a/tfhe/src/shortint/client_key/mod.rs
+++ b/tfhe/src/shortint/client_key/mod.rs
@@ -6,7 +6,7 @@ use crate::shortint::ciphertext::{
     CompressedCiphertextSmall, PBSOrderMarker,
 };
 use crate::shortint::engine::ShortintEngine;
-use crate::shortint::parameters::{MessageModulus, Parameters};
+use crate::shortint::parameters::{MessageModulus, ShortintParameterSet};
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
 
@@ -25,7 +25,7 @@ pub struct ClientKey {
     pub(crate) glwe_secret_key: GlweSecretKeyOwned<u64>,
     /// Key used as the output of the keyswitch operation
     pub(crate) small_lwe_secret_key: LweSecretKeyOwned<u64>,
-    pub parameters: Parameters,
+    pub parameters: ShortintParameterSet,
 }
 
 impl ClientKey {
@@ -40,8 +40,16 @@ impl ClientKey {
     /// // Generate the client key:
     /// let cks = ClientKey::new(PARAM_MESSAGE_2_CARRY_2);
     /// ```
-    pub fn new(parameters: Parameters) -> ClientKey {
-        ShortintEngine::with_thread_local_mut(|engine| engine.new_client_key(parameters).unwrap())
+    pub fn new<P>(parameters: P) -> ClientKey
+    where
+        P: TryInto<ShortintParameterSet>,
+        <P as TryInto<ShortintParameterSet>>::Error: Debug,
+    {
+        ShortintEngine::with_thread_local_mut(|engine| {
+            engine
+                .new_client_key(parameters.try_into().unwrap())
+                .unwrap()
+        })
     }
 
     /// Encrypt a small integer message using the client key.
@@ -68,7 +76,7 @@ impl ClientKey {
     /// let ct = cks.encrypt(msg);
     ///
     /// let dec = cks.decrypt(&ct);
-    /// let modulus = cks.parameters.message_modulus.0 as u64;
+    /// let modulus = cks.parameters.message_modulus().0 as u64;
     /// assert_eq!(msg % modulus, dec);
     /// ```
     pub fn encrypt(&self, message: u64) -> CiphertextBig {
@@ -99,7 +107,7 @@ impl ClientKey {
     /// let ct = cks.encrypt_small(msg);
     ///
     /// let dec = cks.decrypt(&ct);
-    /// let modulus = cks.parameters.message_modulus.0 as u64;
+    /// let modulus = cks.parameters.message_modulus().0 as u64;
     /// assert_eq!(msg % modulus, dec);
     /// ```
     pub fn encrypt_small(&self, message: u64) -> CiphertextSmall {
@@ -134,7 +142,7 @@ impl ClientKey {
     /// let ct = ct.decompress();
     ///
     /// let dec = cks.decrypt(&ct);
-    /// let modulus = cks.parameters.message_modulus.0 as u64;
+    /// let modulus = cks.parameters.message_modulus().0 as u64;
     /// assert_eq!(msg % modulus, dec);
     /// ```
     pub fn encrypt_compressed(&self, message: u64) -> CompressedCiphertextBig {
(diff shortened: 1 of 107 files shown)


tfhe/src/shortint/parameters/mod.rs
```diff
@@ -131,11 +118,195 @@ impl Parameters {
     }
 }
 
+#[derive(Serialize, Copy, Clone, Deserialize, Debug, PartialEq)]
+pub enum ShortintParameterSetInner {
```
tmontaigu, line 122: I don't think this needs to be public, especially since the `inner` member in  `ShortintParameterSet` is not

I think either both are public or none are, unlesss i'm missing something ?

tfhe/src/shortint/parameters/mod.rs
```diff
@@ -131,11 +118,195 @@ impl Parameters {
...
+    PBSAndWopbs(PBSParameters, WopbsParameters),
+}
+
+#[derive(Serialize, Copy, Clone, Deserialize, Debug, PartialEq)]
+pub struct ShortintParameterSet {
+    inner: ShortintParameterSetInner,
+}
+
+impl ShortintParameterSet {
+    pub const fn new_pbs_param_set(params: &PBSParameters) -> Self {
```
tmontaigu, line 134: I would take input params by ownership and not ref here

tfhe/src/shortint/parameters/parameters_wopbs.rs
```diff
@@ -7,9 +7,33 @@ pub use crate::core_crypto::commons::parameters::{
 use crate::shortint::parameters::{
     CarryModulus, CiphertextModulus, EncryptionKeyChoice, MessageModulus,
 };
-use crate::shortint::Parameters;
+use serde::{Deserialize, Serialize};
 
-pub const ALL_PARAMETER_VEC_WOPBS_NORM2: [Parameters; 31] = [
+/// A structure defining the set of cryptographic parameters for homomorphic integer circuit
+/// evaluation.
```
tmontaigu, line 13: Maybe this doctring needs a little update, mainly to emphasize that these are meant for WopPBS

tfhe/src/shortint/mod.rs
```diff
@@ -85,8 +86,12 @@ pub use server_key::{CheckError, CompressedServerKey, ServerKey};
...
 /// ```
-pub fn gen_keys(parameters_set: Parameters) -> (ClientKey, ServerKey) {
-    let cks = ClientKey::new(parameters_set);
+pub fn gen_keys<P>(parameters_set: P) -> (ClientKey, ServerKey)
+where
+    P: TryInto<ShortintParameterSet>,
+    <P as TryInto<ShortintParameterSet>>::Error: std::fmt::Debug,
+{
+    let cks = ClientKey::new(parameters_set.try_into().unwrap());
     let sks = ServerKey::new(&cks);
```
tmontaigu, line 95: So this is where later down the refactoring and feature introduction (or in ServerKey::new) we would match the in parameter set to create the ServeKey that holds a classical pbs key, a multi bit pbs  with or without and associated Wopbs Key ?

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -40,8 +40,16 @@ impl ClientKey {
     /// // Generate the client key:
     /// let cks = ClientKey::new(PARAM_MESSAGE_2_CARRY_2);
     /// ```
-    pub fn new(parameters: Parameters) -> ClientKey {
-        ShortintEngine::with_thread_local_mut(|engine| engine.new_client_key(parameters).unwrap())
+    pub fn new<P>(parameters: P) -> ClientKey
```
tmontaigu, line 43: Shouldn't this still only accept `PBSParameters` ?

Are WopbsParam meant to be used / compatible with classical PBS ?

Meaning if the parameter set is `WopbsOnly` do we expect the params to work for both the classical pbs key and wopbs key ? 

Also If I remember correctly, the way ou wopbs key is, is that it contains a keyswitch key to go from PBS parameters to Wopbs Parameters, which means that is `WopbsOnly(WopbsParameters)` is meant to be compatible with classical pbs then we'd be doing unecessary keyswitch when calling wopbs

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: yeah I realized that after pushing

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: any reason for that ? I think "large" structs are better passed by ref ?

tfhe/src/shortint/parameters/parameters_wopbs.rs
IceTDrinker: definitely

tfhe/src/shortint/mod.rs
IceTDrinker: either that or matching on cks.parameters un ServerKey::new

tfhe/src/shortint/client_key/mod.rs
IceTDrinker: I think this is due to the fact we have a "new_wopbs_key_only_for_wopbs" which requires to be able to create a CKS for a WopbsOnly, I'm having a hard time figuring how to organize that differently at the moment as a WopbsOnly key would still need to have a CKS to encrypt stuff even if it won't perform any PBS (with proper dispatch it would just panic on any classic PBS call)

tfhe/src/shortint/parameters/mod.rs
tmontaigu: Yes but I would say its only if you don't copy them, here we take a ref, to then copy what it points to, why not take a copy directly ?

tfhe/src/shortint/client_key/mod.rs
tmontaigu: wopbs integration, true nightmare stuff

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: fair enough

tfhe/src/shortint/client_key/mod.rs
IceTDrinker: should improve once there is only a ServerKey dispatching stuff underneath, as long as you call functions that are legal for a the parameter set used to generate it

tfhe/src/shortint/parameters/mod.rs
IceTDrinker: gen_keys from integer takes params by ref while gen_keys from shortint takes params by value, could be the occasion to chose one over the other

IceTDrinker: normally done and uniformized integers

tfhe/src/shortint/parameters/parameters_wopbs.rs
IceTDrinker: done

tfhe/src/shortint/parameters/mod.rs
tmontaigu: I'd vote to take by value :sweat_smile: but its not the end of the world if the other way is chosen

IceTDrinker: by value is what I ended up using 🙂

tmontaigu (approved):

IceTDrinker:
😭 conflicts again

PR #282, opened 2023-05-09
feat(integer): improve parallel algorithms for add/sub

tmontaigu:
This adds fully parallel algorithms for the addition and subtraction

These algorithms take ciphertexts with clean carries and return creates a sum ciphertext that also has clean carries.

The carries are propagated in parallel, using
parallel algorithms for prefix sum / cumulative sum.

The is one based on Hillis and Steele, it the fastest but uses a lot of threads.

The other on Blelloch, it requires less threads but is a bit slower.

256bits addition using param_2_2 goes down from ~2.7s to:
- 364ms using Hillis and Steele
- 474ms using Blelloch

The commit also adds bitwise not, as it it necessary for the subtraction.


tfhe/src/integer/server_key/radix_parallel/tests.rs
```diff
@@ -251,6 +263,37 @@ fn integer_default_add(param: Parameters) {
...
+    let (cks, sks) = KEY_CACHE.get_from_params(param);
+    let cks = RadixClientKey::from((cks, NB_CTXT));
+
+    //RNG
+    let mut rng = rand::thread_rng();
+
+    // message_modulus^vec_length
+    let modulus = param.message_modulus.0.pow(NB_CTXT as u32) as u64;
+
+    for _ in 0..(NB_TEST_SMALLER / 5).max(1) {
```
IceTDrinker, line 277: divided by 5 ? we only compute the heavy operation twice I would think

tfhe/src/integer/server_key/radix_parallel/tests.rs
```diff
@@ -1466,6 +1509,38 @@ fn integer_default_sub(param: Parameters) {
...
```
IceTDrinker, line 1522: same thing for the / 5 as above, heavy computation done twice only

tfhe/src/integer/u256.rs
```diff
@@ -134,6 +135,80 @@ impl std::ops::Add<Self> for U256 {
...
+
+impl std::ops::BitAnd<Self> for U256 {
+    type Output = Self;
+
+    fn bitand(mut self, rhs: Self) -> Self::Output {
+        self &= rhs;
+        self
+    }
+}
+
```
IceTDrinker, line 211: we would need unit tests on the u256 ops, if not now, then later with a follow-up issue

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+
+        self.unchecked_add_assign_parallelized_low_latency(lhs, rhs, AddExtraOne::No);
+    }
+
+    pub(crate) fn is_eligible_for_parallel_carryless_add(&self, num_blocks: usize) -> bool {
+        // having 3-bits is a hard requirement
+        let total_modulus = self.key.message_modulus.0 * self.key.carry_modulus.0;
+        let has_at_least_3_bits = total_modulus >= (1 << 3);
+        // This threshold is arbitrary
+        let has_enough_threads = rayon::current_num_threads() >= (num_blocks / 2);
```
IceTDrinker, line 213: not sure about this if it's not tied to some measurements, as the more thread heavy version may not trigger for "obscure reasons" while we want to alway trigger it in the default path (if parameters are compatible) and let users use the other version if they want to

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+
+    /// This add_assign two numbers
+    ///
+    /// It uses the Hillis and Steele algorithm to do
+    /// prefix sum / cumulative sum in parallel.
+    ///
+    /// It it not "work efficient" as in, it adds a lot
+    /// of work compared to the single threaded aproach,
+    /// however it is higly parallelized and so is the fastest
+    /// assuming egough threads are available.
```
IceTDrinker, line 225: egouhg -> enough

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+                tmp_rhs = ct_right.clone();
+                rayon::join(
+                    || self.full_propagate_parallelized(ct_left),
+                    || self.full_propagate_parallelized(&mut tmp_rhs),
+                );
+                (ct_left, &tmp_rhs)
+            }
+        };
+
+        self.unchecked_add_assign_parallelized_low_latency(lhs, rhs, AddExtraOne::No);
```
IceTDrinker, line 205: I seemed to understand work efficient does not mean the parallelized low latency ? so here we have the work efficient version calling the non efficient version if I'm reading this right ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+        add_extra_one: AddExtraOne,
+    ) {
+        debug_assert!(lhs.block_carries_are_empty());
+        debug_assert!(rhs.block_carries_are_empty());
+        debug_assert!(self.key.message_modulus.0 * self.key.carry_modulus.0 >= (1 << 3));
+
+        let mut carry_out = self.add_and_generate_init_carry_array(lhs, rhs, add_extra_one);
+
+        let num_blocks = carry_out.len();
+        let num_steps = (carry_out.len() as f64).log2() as usize;
```
IceTDrinker, line 250: ilog2 is legal now right ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+                self.key.message_extract_assign(block);
+            });
+    }
+
+    /// Initialization function for parallal carryless sum
+    ///
+    /// This function adds rhs into lhs
+    ///
+    /// It Returns the array that tells whether the
+    /// block resulting from the sum will propagat or generate
```
IceTDrinker, line 439: propagat -> propagate

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+            (0..num_blocks)
+                .into_par_iter()
+                .step_by(two_pow_i_plus_1)
+                .for_each(|k| {
+                    let current_index = k + two_pow_i_plus_1 - 1;
+                    let previous_index = k + two_pow_i - 1;
+
+                    unsafe {
+                        // SAFETY
+                        // We know none are the threads
```
IceTDrinker, line 377: typo: none of the threads

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+    ) -> Vec<crate::shortint::CiphertextBase<PBSOrder>> {
+        let modulus = self.key.message_modulus.0 as u64;
+
+        let lut_does_block_generate_carry =
+            self.key.generate_accumulator(|x| (x >= modulus) as u64);
+
+        let lut_does_block_generate_or_propagate = self.key.generate_accumulator(|x| {
+            if x >= modulus {
+                OuputCarry::Generated as u64
+            } else if x == (modulus - 1) {
```
IceTDrinker, line 456: this could happen even without a propagation I think ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+        lhs.blocks
+            .par_iter_mut()
+            .zip(rhs.blocks.par_iter())
+            .enumerate()
+            .map(|(i, (ct_left_i, ct_right_i))| {
+                self.key.unchecked_add_assign(ct_left_i, ct_right_i);
+                if i == 0 {
+                    if matches!(add_extra_one, AddExtraOne::Yes) {
+                        self.key.unchecked_scalar_add_assign(ct_left_i, 1);
+                    }
```
IceTDrinker, line 473: I'm guessing this is ok to do as lhs and rhs are assumed to have empty carries, correct ? If so worth a small comment here (and also indicate it's to simplify the sub algorithm if possible)

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+                    let prev_block_carry = &carry_out[i];
+                    self.key.unchecked_apply_lookup_table_bivariate_assign(
+                        block,
+                        prev_block_carry,
+                        &lut_carry_propagation_sum,
+                    )
+                });
+            for i in space..num_blocks {
+                carry_out[i].copy_from(&step_output[i]);
+            }
```
IceTDrinker, line 276: carry_out[space..num_blocks].copy_from_slice(step_output[space..num_blocks])

maybe ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+        debug_assert!(lhs.block_carries_are_empty());
+        debug_assert!(rhs.block_carries_are_empty());
+        debug_assert!(self.key.message_modulus.0 * self.key.carry_modulus.0 >= (1 << 3));
+
+        let mut carry_out = self.add_and_generate_init_carry_array(lhs, rhs, add_extra_one);
+
+        let num_blocks = carry_out.len();
+        let num_steps = (carry_out.len() as f64).log2() as usize;
+
+        let lut_carry_propagation_sum = self.key.generate_accumulator_bivariate(|y, x| {
```
IceTDrinker, line 332: the reversed y, x makes it harder to read, is there some more descriptive name that could be used here ? like

previous_block_carry, current_block_value or something ?

IceTDrinker (commented):
I think I'm starting to understand, you perform the first stage of the addition, get the carries out of that operation and then perform the carry propagate with the parallel prefix sum algorithm ?

tfhe/src/integer/server_key/radix_parallel/add.rs
IceTDrinker: well it's not that arbitrary as the algorithm will at most use num_blocks / 2 threads

tfhe/src/integer/server_key/radix_parallel/add.rs
IceTDrinker: ok it propagates if the sum is at the modulus capacity

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+            .par_iter_mut()
+            .zip(rhs.blocks.par_iter())
+            .enumerate()
+            .map(|(i, (ct_left_i, ct_right_i))| {
+                self.key.unchecked_add_assign(ct_left_i, ct_right_i);
+                if i == 0 {
+                    if matches!(add_extra_one, AddExtraOne::Yes) {
+                        self.key.unchecked_scalar_add_assign(ct_left_i, 1);
+                    }
+                    // The first block can only ouput a carry
```
IceTDrinker, line 474: do we really need the differentiated luts here ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+    fn add_and_generate_init_carry_array<PBSOrder: PBSOrderMarker>(
+        &self,
+        lhs: &mut RadixCiphertext<PBSOrder>,
+        rhs: &RadixCiphertext<PBSOrder>,
+        add_extra_one: AddExtraOne,
+    ) -> Vec<crate::shortint::CiphertextBase<PBSOrder>> {
+        let modulus = self.key.message_modulus.0 as u64;
+
+        let lut_does_block_generate_carry =
+            self.key.generate_accumulator(|x| (x >= modulus) as u64);
```
IceTDrinker, line 451: kind of dangerous here to not do some matching to return the same values as the OutputCarry enum, here they match but it would be easy to make a mistake it feels like

tfhe/src/integer/server_key/radix_parallel/add.rs
IceTDrinker: ah right, damn I forgot those were ciphertexts

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: yeah that is a typo :sweat_smile: :hot_face:

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+            }
+        };
+
+        self.unchecked_add_assign_parallelized_low_latency(lhs, rhs, AddExtraOne::No);
+    }
+
+    pub(crate) fn is_eligible_for_parallel_carryless_add(&self, num_blocks: usize) -> bool {
+        // having 3-bits is a hard requirement
+        let total_modulus = self.key.message_modulus.0 * self.key.carry_modulus.0;
+        let has_at_least_3_bits = total_modulus >= (1 << 3);
```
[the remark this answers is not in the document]
tmontaigu: Actually its 4 bits, the LUT used in the parallel prefix sum requires 4 bit as value (0, 1, 2) are encoded in the block and we do a bivariate PBS

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+        lhs.blocks
+            .par_iter_mut()
+            .zip(carry_out.par_iter())
+            .enumerate()
+            .for_each(|(i, (block, input_carry))| {
+                if i != 0 {
+                    self.key.unchecked_add_assign(block, input_carry);
+                }
+                self.key.message_extract_assign(block);
+            });
```
IceTDrinker, line 293: here could probably zero out the least significant ciphertext/trivial_encrypt_assign 0 and then get rid of the if i != 0 in the loop below allowing to remove the enumerate from the iterator and making the code a tiny bit less complicated with the removed if test

tfhe/src/integer/server_key/radix_parallel/add.rs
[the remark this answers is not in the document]
IceTDrinker: ok so this needs updating with a comment as to why the limit is 4 bits minimum

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+        let num_blocks = carry_out.len();
+        let num_steps = (carry_out.len() as f64).log2() as usize;
+
+        let lut_carry_propagation_sum = self.key.generate_accumulator_bivariate(|y, x| {
+            if y == OuputCarry::Propagated as u64 {
+                x
+            } else {
+                y
+            }
+        });
```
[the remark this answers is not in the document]
IceTDrinker: ok I think I can see it

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +153,335 @@ impl ServerKey {
...
+                let b = drainer.next().unwrap();
+                carry_out.swap(k + i_pow_2 - 1, k + i_plus_1_pow_2 - 1);
+                carry_out[k + i_plus_1_pow_2 - 1] = b;
+            }
+            drop(drainer);
+            assert!(buffer.is_empty());
+        }
+
+        // The first step of the Down-Sweep phase sets the
+        // first block to 0
```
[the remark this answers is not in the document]
IceTDrinker: nevermind, I now know why, all good !

tfhe/src/integer/server_key/radix_parallel/add.rs
IceTDrinker: ok now I see how the Propagate variant is a smart trick that is not a true carry value, still I feel like there should be consistency between both accumulators potentially :/

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: yes

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: Honestly initing the carry_out[0] to a trivial zero to not have a if won't be mesureable in the perf so if you think its more readble lets do it this way

IceTDrinker: It would match the other implem but it's not necessary

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: If we always enable the parallel version, we are likely to slowdown computations on traditionnal laptops that have less threads, that was the idea behind that check

IceTDrinker: I think it's worth discussing, but I understand where it comes from

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: Yes because the first block may only generate or not, but never propagate

IceTDrinker: thx

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -2,6 +2,20 @@ use crate::integer::ciphertext::RadixCiphertext;
 use crate::integer::ServerKey;
 use crate::shortint::PBSOrderMarker;
 
+use rayon::prelude::*;
+
+pub(crate) enum AddExtraOne {
```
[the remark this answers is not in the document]
tmontaigu: added

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: corrected

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: fixed

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: changed that part to take (msb ,lsb) and added comment

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: yeah I use enum variant to have more readability

I updated this part to use the variant

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -2,6 +2,50 @@ use crate::integer::ciphertext::RadixCiphertext;
 use crate::integer::ServerKey;
 use crate::shortint::PBSOrderMarker;
 
+use rayon::prelude::*;
+
+/// Enum to tell parallel additions whether they should
+/// add an extra one to the first pait of block being added
```
IceTDrinker, line 8: pait ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -139,8 +183,332 @@ impl ServerKey {
...
+        lhs: &mut RadixCiphertext<PBSOrder>,
+        rhs: &RadixCiphertext<PBSOrder>,
+        add_extra_one: AddExtraOne,
+    ) -> Vec<crate::shortint::CiphertextBase<PBSOrder>> {
+        let modulus = self.key.message_modulus.0 as u64;
+
+        // This used for the first pair of blocks
+        // as this pair can only generate or not but never propagate
+        let lut_does_block_generate_carry = self.key.generate_accumulator(|x| {
+            if x >= modulus - 1 {
```
IceTDrinker, line 473: is this correct the previous code was x >= modulus for generated

tmontaigu: :hot_face: trying to be too fast

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: pair*

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: I removed the check on threads for now

tfhe/src/integer/u256.rs
tmontaigu: I will add them later

IceTDrinker: please create a follow-up issue

tmontaigu: https://github.com/zama-ai/tfhe-rs-internal/issues/120

IceTDrinker (approved):
Looks good like this thanks ! let's plan a PR to update the benchmarks once this runs after the merge

PR #283, opened 2023-05-09
docs(tfhe): add boolean sha256 tutorial

JoseSK999:
This is the updated PR with the squashed commits.

### DIFF
--- a/tfhe/Cargo.toml
+++ b/tfhe/Cargo.toml
@@ -23,6 +23,7 @@ lazy_static = { version = "1.4.0" }
 criterion = "0.4.0"
 doc-comment = "0.3.3"
 serde_json = "1.0.94"
+clap = "4.2.7"
 # Used in user documentation
 bincode = "1.3.3"
 fs2 = { version = "0.4.3" }
@@ -204,6 +205,10 @@ required-features = ["boolean"]
 [[example]]
 name = "regex_engine"
 required-features = ["integer"]
+
+[[example]]
+name = "sha256_bool"
+required-features = ["boolean"]
 
 [lib]
 crate-type = ["lib", "staticlib", "cdylib"]
--- a/tfhe/docs/tutorial/sha256_bool.md
+++ b/tfhe/docs/tutorial/sha256_bool.md
@@ -1,0 +1,322 @@
+# Tutorial
+
+## Intro
+
+In this tutorial we will go through the steps to turn a regular sha256 implementation into its homomorphic version. We explain the basics of the sha256 function first, and then how to implement it homomorphically with performance considerations.
+
+## Sha256
+
+The first step in this experiment is actually implementing the sha256 function. We can find the specification [here](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf), but let's summarize the three main sections of the document.
+
+#### Padding
+
+The sha256 function processes the input data in blocks or chunks of 512 bits. Before actually performing the hash computations we have to pad the input in the following way:
+* Append a single "1" bit
+* Append a number of "0" bits such that exactly 64 bits are left to make the message length a multiple of 512
+* Append the last 64 bits as a binary encoding of the original input length
+
+Or visually:
+
+```
+0                                   L   L+1                              L+1+k                  L+1+k+64
+|-----------------------------------|---|--------------------------------|----------------------|
+    Original input (L bits)        "1" bit          "0" bits             Encoding of the number L
+```
+Where the numbers on the top represent the length of the padded input at each position, and L+1+k+64 is a multiple of 512 (the length of the padded input).
+
+#### Operations and functions
+
+Let's take a look at the operations that we will use as building blocks for functions inside the sha256 computation. These are bitwise AND, XOR, NOT, addition modulo 2^32 and the Rotate Right (ROTR) and Shift Right (SHR) operations, all working with 32-bit words and producing a new word.
+
+We combine these operations inside the sigma (with 4 variations), Ch and Maj functions. At the end of the day, when we change the sha256 to be computed homomorphically, we will mainly change the isolated code of each operation.
+
+Here is the definition of each function:
+```
+Ch(x, y, z) = (x AND y) XOR ((NOT x) AND z)
+Maj(x, y, z) = (x AND y) XOR (x AND z) XOR (y AND z)
+
+Σ0(x) = ROTR-2(x) XOR ROTR-13(x) XOR ROTR-22(x)
+Σ1(x) = ROTR-6(x) XOR ROTR-11(x) XOR ROTR-25(x)
+σ0(x) = ROTR-7(x) XOR ROTR-18(x) XOR SHR-3(x)
+σ1(x) = ROTR-17(x) XOR ROTR-19(x) XOR SHR-10(x)
+```
+There are some things to note about the functions. Firstly we see that Maj can be simplified by applying the boolean distributive law (x AND y) XOR (x AND z) = x AND (y XOR z). So the new Maj function looks like this:
+
+```
+Maj(x, y, z) = (x AND (y XOR z)) XOR (y AND z)
+```
+Next we can also see that Ch can be simplified by using a single bitwise multiplexer. Let's take a look at the truth table of the Ch expression.
+| x | y | z | Result |
+| - | - | - | ------ |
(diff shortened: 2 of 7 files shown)


IceTDrinker:
Hey @JoseSK999 no need to open new PRs for the changes we request, force pushing to the PR branch is fine by us (and makes less noise with additional PRs)

IceTDrinker:
Let's keep this one for the other reviews

tfhe/Cargo.toml
```diff
@@ -116,6 +116,9 @@ x86_64-unix = ["x86_64", "seeder_unix"]
 aarch64 = ["generator_aarch64_aes"]
 aarch64-unix = ["aarch64", "seeder_unix"]
 
+# Example-specific features
+sha256_bool_ladner_fischer = []
+
```
IceTDrinker, line 121: let's avoid adding a feature, please add clap as a dev dependency and update the example to parse a flag from the command line to chose one carry propagation version over the other, thanks

JoseSK999: Done!

IceTDrinker: thanks !

tfhe/examples/sha256_bool/boolean_ops.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    let mut carry = trivial_bools(&[false; 32], sk);
+    for bit in 0..31 {
+        carry[bit] = generate[bit + 1].clone();
+    }
+
+    carry
+}
+
+// 2 (homomorphic) bitwise ops
```
IceTDrinker, line 171: what does this mean ?

tfhe/examples/sha256_bool/boolean_ops.rs
```diff
@@ -0,0 +1,394 @@
...
+    xor(&lhs, &rhs, sk)
+}
+
+// Parallelized homomorphic bitwise ops
+// Building block for most of the previous functions
+fn xor(a: &[Ciphertext; 32], b: &[Ciphertext; 32], sk: &ServerKey) -> [Ciphertext; 32] {
+    let result: Vec<Ciphertext> = (0..32)
+        .into_par_iter()
+        .map(|i| sk.xor(&a[i], &b[i]))
+        .collect();
```
IceTDrinker, line 238: how about something like

```rust
let mut result = a.clone();
result.par_iter_mut().zip(a.par_iter().zip(b.par_iter())).map(|(dst, (lhs, rhs))| *dst = sk.xor(lhs, rhs));
result
```

?

same thing for similar funcs, not sure if it has an impact on perfs

tfhe/examples/sha256_bool/boolean_ops.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    let mut array = trivial_bools(&[false; 32], sk);
+    for (i, elem) in result.into_iter().enumerate() {
+        array[i] = elem;
+    }
```
IceTDrinker, line 243: if the above is not relevant, something like the following avoids indexing which has range checks for each indexing op :)

```rust
array.as_mut().copy.from_slice(&result);
```

to be applied where relevant in your code

tfhe/examples/sha256_bool/boolean_ops.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    array
+}
+
+fn mux(condition: &[Ciphertext; 32], then: &[Ciphertext; 32], otherwise: &[Ciphertext; 32], sk: &ServerKey) -> [Ciphertext; 32] {
```
IceTDrinker, line 263: see above

tfhe/docs/tutorial/sha256_bool.md
```diff
@@ -0,0 +1,332 @@
...
+fn shift_right(x: &[Ciphertext; 32], n: usize, sk: &ServerKey) -> [Ciphertext; 32] {
+    let mut result = trivial_bools(&[false; 32], sk);
+    for i in 0..(32 - n) {
+        result[i + n] = x[i].clone();
+    }
+    result
+}
+```
+We see a function called ``trivial_bools`` that we use along our implementation to initialize an array of 32 Ciphertexts. This is because we cannot copy the same Ciphertext for each position of the array as we do with simple bools, so we create 32 different trivial encryptions. We will also use this function in order to trivially encrypt constant values to operate homomorphically with them.
+
```
IceTDrinker, line 166: not sure I understand here, you could clone x to be the mutable result and the call rotate/shift ops from rust right ?

https://doc.rust-lang.org/std/primitive.slice.html#method.rotate_right for the shift may need to do some manual management at the end

I guess the code could be updated as well then :)

IceTDrinker (commented):
Hello @JoseSK999 thanks a lot for this work, it is very good indeed.

Several suggestions/questions that you could fix, for the last few modifications (like location of the tutorial in the docs) we'll probably do it ourselves.

In any case this contribution is very much appreciated.

Thanks!

tfhe/examples/sha256_bool/boolean_ops.rs
JoseSK999: I quantify the cost of each function using homomorphic bitwise operations as unit (kinda)

tfhe/examples/sha256_bool/boolean_ops.rs
JoseSK999: It seems a bit more efficient, thanks! We have to use for_each to consume the iterator

tfhe/examples/sha256_bool/padding.rs
```diff
@@ -0,0 +1,67 @@
...
+    let data_len_bits = (data.len() as u64) * 8;
+    bits.extend((0..64).rev().map(|i| (data_len_bits >> i) & 1 == 1));
+
+    bits
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::sha256::bools_to_hex;
```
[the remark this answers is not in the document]
JoseSK999: Fixed. I forgot to add "_function" in the test

tfhe/docs/tutorial/sha256_bool.md
JoseSK999: That's a good point. About the highlighted text, I meant that we cannot initialize arrays of ciphertexts like this:
```let a = [sk.trivial_encrypt(false); 32];```

**Comment by JoseSK999** [comment]
I have updated everything, thanks for the feedback and the nice words!

File: tfhe/docs/tutorial/sha256_bool.md (hunk 6 above)
Comment (IceTDrinker) [reply tfhe/docs/tutorial/sha256_bool.md]: ok yes

File: tfhe/examples/sha256_bool/boolean_ops.rs (hunk 3 above)
Comment (IceTDrinker) [reply tfhe/examples/sha256_bool/boolean_ops.rs]: oh yeah my code snippet was slightly wrong

**Comment by JoseSK999** [comment]
Btw @IceTDrinker I just found out that perhaps it's better to implement ```trivial_bools``` and ```initialize_w``` in this way:
```rust
fn trivial_bools(bools: &[bool; 32], sk: &ServerKey) -> [Ciphertext; 32] {
    let array = array::from_fn(|i| sk.trivial_encrypt(bools[i]));
    array
}

fn initialize_w(sk: &ServerKey) -> [[Ciphertext; 32]; 64] {
    let array = array::from_fn(|_| trivial_bools(&[false; 32], sk));
    array
}
```

tmontaigu:
> Btw @IceTDrinker I just found out that perhaps it's better to implement `trivial_bools` and `initialize_w` in this way:
> 
> ```rust
> fn trivial_bools(bools: &[bool; 32], sk: &ServerKey) -> [Ciphertext; 32] {
>     let array = array::from_fn(|i| sk.trivial_encrypt(bools[i]));
>     array
> }
> 
> fn initialize_w(sk: &ServerKey) -> [[Ciphertext; 32]; 64] {
>     let array = array::from_fn(|_| trivial_bools(&[false; 32], sk));
>     array
> }
> ```

It looks better/shorter than the current code, you should be able to drop the temporary to make it even shorter
```rust
fn trivial_bools(bools: &[bool; 32], sk: &ServerKey) -> [Ciphertext; 32] {
    array::from_fn(|i| sk.trivial_encrypt(bools[i]))
}
```

aquint-zama:
relates https://github.com/zama-ai/bounty-program/issues/39

tmontaigu (approved):

PR #287, opened 2023-05-10
chore(ci): compute throughput as operations per second

soonum:
Since most of the operations are over 1 ms, there is no point to compute the number of operation per millisecond.


IceTDrinker (approved):
Looks good let's wait to merge to leave capacity for bench marchines

PR #288, opened 2023-05-12
bench(multi-bit-pbs): hard set the number of threads to 10 for the multi-bit PBS

agnesLeroy:
It's the optimal value measured on an m6i.metal instance where we run the benchmarks.
See: https://zama.grafana.net/d/a60a669c-5199-4e26-bcd6-8a69d7bc97e5/misc-panels-to-help-team?orgId=1&from=1683712319833&to=1683885119833&var-branch=ci%2Fbench_multi_bits&var-messagemodulus=All&var-hardware=m6i.metal


IceTDrinker:
commit first line probably too long, thanks for the patch !

IceTDrinker (approved):
Just need to fix that commit message first line it seems, otherwise thanks a lot 🙏

PR #290, opened 2023-05-12
chore(ci): handle aws tests in pull request from forked repository

soonum:


.github/workflows/aws_tfhe_integer_tests.yml
```diff
@@ -4,6 +4,7 @@ env:
   CARGO_TERM_COLOR: always
   ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
   RUSTFLAGS: "-C target-cpu=native"
+  REPOSITORY: ${{ inputs.fork_user || github.repository_owner}}/tfhe-rs
```
IceTDrinker, line 7: not compatible with internal repo, not sure how big of a problem it is

.github/workflows/aws_tfhe_integer_tests.yml
```diff
@@ -25,26 +26,35 @@ on:
...
+        description: 'User login of forked repo'
+        type: string
+      fork_git_sha:
+        description: 'Git SHA to checkout from fork'
         type: string
 
 jobs:
   integer-tests:
     concurrency:
-      group: ${{ github.workflow }}_${{ github.ref }}_${{ github.event.inputs.instance_image_id }}_${{ github.event.inputs.instance_type }}
```
IceTDrinker, line 35: what's the reason for the github.event removal ?

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -4,6 +4,7 @@ env:
   CARGO_TERM_COLOR: always
   ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
   RUSTFLAGS: "-C target-cpu=native"
+  REPOSITORY: ${{ inputs.fork_user || github.repository_owner}}/tfhe-rs
```
IceTDrinker, line 7: incompatible with internal repo

.github/workflows/aws_tfhe_integer_tests.yml
soonum: I think the best and cleanest way to deal with it is to pass `fork_repo` instead of `fork_user`. It would imply to modify the code on Slab side a little bit, but it would avoid complicated way to deal with internal repo.

IceTDrinker: up to you to chose the best solution

.github/workflows/aws_tfhe_integer_tests.yml
soonum: It's just a shorthand. `inputs` context is directly available, so prefixing it by `github.event` is redundant.

IceTDrinker (approved):
Looks good thanks !

PR #294, opened 2023-05-15
refactor(integer): improve decomposition/recomposition into blocks

tmontaigu:
This refactors the code that decompose / recompose an integer into smaller blocks.

This new implementation should hopefully be a little bit easier to understand.

But more importantly it is more general/generic,
the previous implementation required the input type to be able to be described as u64 words, the new one works for any type as it directly iterates over the bytes which also makes the big-endian support easier. And allows to natively support types smaller than u64.

Also the new implementation is separted from the encryption code meaning it will be usable by scalar operation, which will allow us to deduplicate code and start making scalar ops support scalar values that are on more than 64-bits.


tmontaigu:
It was a first atempt that worked but felt too complex

PR #296, opened 2023-05-15
refactor(integer): disable WopPBS by default in hlapi

tmontaigu:
In the HLAPI, the WopPBS is enabled by default,
meaning the WopPBS key is generated when integers
are enabled.

This is not really good as the wopbs key is huge
(~700MB with PARAM_2_2) and only used for function evaluation which does not scale for all types exposed by the halpi and is still a bit experimental so not really advertised in the docs.

Also keys for wopbs are not compressible yet
(that is why the HLAPI does not yet have a CompressedServerKey).

So disabling wopbs by default will enable to have a compressed server key that actually compresse things.


tmontaigu:
commit should be `refactor(hlapi): ...`

IceTDrinker (approved):
Good once green, thanks!

PR #298, opened 2023-05-16
feat(hlapi): add CompressedServerKey

tmontaigu:
Now that WopPBS key are optional in the hlapi
we can have a CompressedServerKey.
If a user tries to create a CompressedServerKey
but has enabled function evaluation on integers
(WopPBS) then it will panic as WopPBS are not yet compressible. And 'stuffing' the non-compressed wopbs-key in the compressed server key, would defeat the purpose of compressed server key, as WopPBS key makes of for
the vast majority of the space used.

Also having CompressedServerKey is required to
be able to have wasm API of the hlapi
as wasm cannot generate normal server key.

### DIFF
--- a/tfhe/src/high_level_api/integers/keys.rs
+++ b/tfhe/src/high_level_api/integers/keys.rs
@@ -131,3 +131,43 @@ impl IntegerServerKey {
             .expect("Integer ServerKey is not initialized")
     }
 }
+
+#[derive(Clone, serde::Serialize, serde::Deserialize)]
+pub struct IntegerCompressedServerKey {
+    pub(crate) key: Option<crate::integer::CompressedServerKey>,
+    pub(crate) encryption_type: crate::shortint::EncryptionKeyChoice,
+}
+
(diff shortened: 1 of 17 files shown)


tfhe/src/high_level_api/integers/keys.rs
```diff
@@ -131,3 +131,36 @@ impl IntegerServerKey {
...
+    pub(in crate::high_level_api) fn new(client_key: &IntegerClientKey) -> Self {
+        let Some(integer_key) = &client_key.key else {
+            return Self {
+                key: None,
+                encryption_type: EncryptionKeyChoice::Big,
+            };
+        };
+        if client_key.wopbs_block_parameters.is_some() {
+            panic!("Cannot compress the ServerKey as it requires a function evaluation key");
+        }
```
IceTDrinker, line 151: as it requires a function evaluation key, which is not currently compressible.

Otherwise the error message does not make much sense

PR #299, opened 2023-05-16
feat(tfhe): add glwe keyswitch, glwe tensor product, trace packing ke…

Carl-Zama:
…yswitch and public functional packing keyswitch

This PR adds the following functionalities which work for native ciphertext moduli only at the moment:

1) GLWE keyswitch operation, this uses the ability to decompose a slice which is also added
2) GLWE tensor product and relinearisation operations to compute the product of two polynomials homomorphically
3) LWE trace packing keyswitch operation to pack LWE ciphertexts into a single GLWE ciphertext (the LWE plaintexts are placed in user specified coefficients of the GLWE plaintext polynomial)
4) LWE public functional packing keyswitch to pack LWE ciphertexts into a single GLWE ciphertext while applying a public scalar function to the vector of inputs.


IceTDrinker:
this will need a rebase, I'll do the review on the parts that don't concern decomposition !

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -406,6 +406,33 @@ pub fn polynomial_wrapping_mul<Scalar, OutputCont, LhsCont, RhsCont>(
...
+/// assert_eq!(pol.as_ref(), &[127u8, 254, 125, 252, 123, 250]);
+/// ```
+pub fn polynomial_wrapping_scalar_mul_assign<Scalar, PolyCont>(
+    output: &mut Polynomial<PolyCont>,
+    scalar: Scalar,
+) where
+    Scalar: UnsignedInteger,
+    PolyCont: ContainerMut<Element = Scalar>,
+{
+    slice_wrapping_scalar_mul_assign(output, scalar)
```
IceTDrinker, line 433: though there is a deref implementation (considered an anti pattern in rust as I discovered) could we call output.as_mut() here ?

tfhe/src/core_crypto/entities/seeded_glwe_keyswitch_key.rs
```diff
@@ -0,0 +1,379 @@
+/*
```
IceTDrinker, line 1: why is this in a comment at the moment ? if need be we can add seeded entities in a separate PR

IceTDrinker (commented):
First batch of comments on GlweKeyswitchKey entity

IceTDrinker:
closing as this was rebased elsewhere

PR #305, opened 2023-05-22
chore(ci): create fast feedback unit test profile

soonum:
This is done to get quick feedback to developpers in a Pull Request.

It tests shorint level with only three sets of parameters. Integer level is tested with only the default operations with two sets of parameters.

This profile will be automatically triggered on each push in a pull request.
Conversely the full suite of test will also be triggered automatically but once the review is approved.

### DIFF
--- a/.github/workflows/aws_tfhe_fast_tests.yml
+++ b/.github/workflows/aws_tfhe_fast_tests.yml
@@ -1,0 +1,119 @@
+# Run a small subset of shortint and integer tests to ensure quick feedback.
+name: Fast AWS Tests on CPU
+
+env:
+  CARGO_TERM_COLOR: always
+  ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  RUSTFLAGS: "-C target-cpu=native"
+
+on:
+  # Allows you to run this workflow manually from the Actions tab as an alternative.
+  workflow_dispatch:
+    # All the inputs are provided by Slab
+    inputs:
+      instance_id:
+        description: "AWS instance ID"
+        type: string
+      instance_image_id:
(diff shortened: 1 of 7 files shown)


IceTDrinker (approved):
Test approve

.github/workflows/aws_tfhe_fast_tests.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: |
+          BIG_TESTS_INSTANCE=TRUE FAST_TESTS=TRUE make test_integer_ci
+
+      - name: Run shortint multi-bit tests
+        run: |
+          BIG_TESTS_INSTANCE=TRUE FAST_TESTS=TRUE make test_shortint_multi_bit_ci
+
+      - name: Run integer multi-bit tests
+        run: |
+          BIG_TESTS_INSTANCE=TRUE FAST_TESTS=TRUE make test_integer_multi_bit_ci
```
IceTDrinker, line 103: we'll see if BIG_TESTS_INSTANCE thing holds for multi bit, but as we run less parameters it should be fine

IceTDrinker (approved):
Looks good !

IceTDrinker:
hmm looks like it relaunches the fast tests ?

soonum:
> hmm looks like it relaunches the fast tests ?

Yes it's expected since the fast tests is the test suite that is automatically launched on each push in a PR.

IceTDrinker:
what I mean is what's the point of relaunching the fast test on approval since it's contained in the full test suite ?

soonum:
You're right, I'll fix that right away.

soonum:
That was a false impression. I've checked the Actions runs and the fast tests is not launched upon approval.
I'm merging as is.

IceTDrinker (approved):

PR #308, opened 2023-05-26
feat(integer): make eq (==) faster and add ne (!=)

tmontaigu:


tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -39,7 +39,93 @@ impl ServerKey {
...
+
+        while block_comparisons.len() > 1 {
+            block_comparisons = block_comparisons
+                .chunks(max_value)
+                .map(|blocks| {
+                    let mut sum = blocks[0].clone();
+                    for other_block in &blocks[1..] {
+                        self.key.unchecked_add_assign(&mut sum, other_block);
+                    }
+                    let lut = self.key.generate_accumulator(|x| {
```
jborfila, line 70: Wouldn't be worth to generate these LUTs outside of the loop? The length for all but the last should be the same

tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -39,7 +39,93 @@ impl ServerKey {
...
+        block_comparisons
+            .iter_mut()
+            .zip(rhs.blocks.iter())
+            .for_each(|(lhs_block, rhs_block)| {
+                self.key
+                    .unchecked_apply_lookup_table_bivariate_assign(lhs_block, rhs_block, &lut);
+            });
+
+        let message_modulus = self.key.message_modulus.0;
+        let carry_modulus = self.key.message_modulus.0;
```
jborfila, line 106: Same comment here

tfhe/src/integer/server_key/radix_parallel/comparison.rs
```diff
@@ -4,13 +4,109 @@ use crate::integer::ciphertext::RadixCiphertext;
...
+        block_comparisons
+            .par_iter_mut()
+            .zip(rhs.blocks.par_iter())
+            .for_each(|(lhs_block, rhs_block)| {
+                self.key
+                    .unchecked_apply_lookup_table_bivariate_assign(lhs_block, rhs_block, &lut);
+            });
+
+        let message_modulus = self.key.message_modulus.0;
+        let carry_modulus = self.key.message_modulus.0;
```
jborfila, line 32: same here

tfhe/src/integer/server_key/radix_parallel/comparison.rs
```diff
@@ -4,13 +4,109 @@ use crate::integer::ciphertext::RadixCiphertext;
...
```
jborfila, line 84: same here

jborfila (changes requested):
Nice work! Few comments to be addressed before merging :)

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: We could but to a small changed has to be made, as one the the chunk of blocks won't have the same lentgh as the others 
(we use par_chunk not par_chunk_exact)

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: yes

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: should be fixed, only the lut for the last chunk of blocks will be allocated

tmontaigu:
@slab-ci cpu_test

jborfila (approved):

PR #312, opened 2023-05-31
Scalar add sub neg

tmontaigu:
- scalar_add now uses the same parallel carry propagation algorithm
  as the add function.

- scalar_sub now uses the same parallel carry propagation algorithm
  as the sub function.

- the 'default' negation function uses the now improved scalar_add
  to be faster

- unchecked_scalar_add, smart_scalar_add, checked_scalar_add, scalar_add
  have been updated to work on generic scalar type so it should work
  on u32, u64, u128, U256, etc

- unchecked_scalar_sub, smart_scalar_sub, checked_scalar_sub, scalar_sub
  have been updated to work on generic scalar type so it should work
  on u32, u64, u128.
  As U256 does not yet implement the UnsignedInteger trait, its not
  usable yet as a scalar type for the sub operation.

- The HLAPI is still locked to u64 scalars, it will be updated
  when most / all scalar ops are ready

# Requires
- #306

### DIFF
--- a/tfhe/src/integer/server_key/radix/scalar_add.rs
+++ b/tfhe/src/integer/server_key/radix/scalar_add.rs
@@ -1,3 +1,4 @@
+use crate::integer::block_decomposition::{BlockDecomposer, DecomposableInto};
 use crate::integer::ciphertext::RadixCiphertext;
 use crate::integer::server_key::CheckError;
 use crate::integer::server_key::CheckError::CarryFull;
@@ -34,11 +35,15 @@ impl ServerKey {
     /// let dec: u64 = cks.decrypt(&ct_res);
     /// assert_eq!(msg + scalar, dec);
     /// ```
-    pub fn unchecked_scalar_add<PBSOrder: PBSOrderMarker>(
+    pub fn unchecked_scalar_add<PBSOrder, T>(
         &self,
         ct: &RadixCiphertext<PBSOrder>,
-        scalar: u64,
-    ) -> RadixCiphertext<PBSOrder> {
+        scalar: T,
+    ) -> RadixCiphertext<PBSOrder>
+    where
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
         let mut result = ct.clone();
         self.unchecked_scalar_add_assign(&mut result, scalar);
         result
@@ -50,25 +55,20 @@ impl ServerKey {
     /// ciphertext.
     ///
     /// The result is assigned to the `ct_left` ciphertext.
-    pub fn unchecked_scalar_add_assign<PBSOrder: PBSOrderMarker>(
+    pub fn unchecked_scalar_add_assign<PBSOrder, T>(
         &self,
         ct: &mut RadixCiphertext<PBSOrder>,
-        scalar: u64,
-    ) {
-        // Bits of message put to 1
-        let mask = (self.key.message_modulus.0 - 1) as u64;
-
-        let mut power = 1_u64;
-        // Put each decomposition into a new ciphertext
-        for ct_i in ct.blocks.iter_mut() {
-            let mut decomp = scalar & (mask * power);
-            decomp /= power;
-
-            self.key.unchecked_scalar_add_assign(ct_i, decomp as u8);
-
-            //modulus to the power i
-            let Some(new_power) = power.checked_mul(self.key.message_modulus.0 as u64) else {break};
-            power = new_power;
+        scalar: T,
+    ) where
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
+        let bits_in_message = self.key.message_modulus.0.ilog2();
+        let decomposer =
+            BlockDecomposer::with_early_stop_at_zero(scalar, bits_in_message).iter_as::<u8>();
+        for (ciphertext_block, scalar_block) in ct.blocks.iter_mut().zip(decomposer) {
+            self.key
+                .unchecked_scalar_add_assign(ciphertext_block, scalar_block);
         }
     }
 
@@ -96,29 +96,26 @@ impl ServerKey {
     ///
     /// assert_eq!(true, res);
     /// ```
-    pub fn is_scalar_add_possible<PBSOrder: PBSOrderMarker>(
+    pub fn is_scalar_add_possible<PBSOrder, T>(
         &self,
         ct: &RadixCiphertext<PBSOrder>,
-        scalar: u64,
-    ) -> bool {
-        //Bits of message put to 1
-        let mask = (self.key.message_modulus.0 - 1) as u64;
+        scalar: T,
+    ) -> bool
+    where
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
+        let bits_in_message = self.key.message_modulus.0.ilog2();
+        let decomposer =
+            BlockDecomposer::with_early_stop_at_zero(scalar, bits_in_message).iter_as::<u8>();
 
-        let mut power = 1_u64;
-
-        for ct_i in ct.blocks.iter() {
-            let mut decomp = scalar & (mask * power);
-            decomp /= power;
-
-            if !self.key.is_scalar_add_possible(ct_i, decomp as u8) {
-                return false;
-            }
-
-            //modulus to the power i
-            let Some(new_power) = power.checked_mul(self.key.message_modulus.0 as u64) else {break};
-            power = new_power;
-        }
-        true
+        ct.blocks
+            .iter()
+            .zip(decomposer)
+            .all(|(ciphertext_block, scalar_block)| {
+                self.key
+                    .is_scalar_add_possible(ciphertext_block, scalar_block)
+            })
     }
 
     /// Computes homomorphically an addition between a scalar and a ciphertext.
@@ -151,11 +148,15 @@ impl ServerKey {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn checked_scalar_add<PBSOrder: PBSOrderMarker>(
+    pub fn checked_scalar_add<PBSOrder, T>(
         &self,
         ct: &RadixCiphertext<PBSOrder>,
-        scalar: u64,
-    ) -> Result<RadixCiphertext<PBSOrder>, CheckError> {
+        scalar: T,
+    ) -> Result<RadixCiphertext<PBSOrder>, CheckError>
+    where
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
         if self.is_scalar_add_possible(ct, scalar) {
             Ok(self.unchecked_scalar_add(ct, scalar))
         } else {
@@ -167,11 +168,15 @@ impl ServerKey {
     ///
     /// If the operation can be performed, the result is stored in the `ct_left` ciphertext.
     /// Otherwise [CheckError::CarryFull] is returned, and `ct_left` is not modified.
-    pub fn checked_scalar_add_assign<PBSOrder: PBSOrderMarker>(
+    pub fn checked_scalar_add_assign<PBSOrder, T>(
         &self,
         ct: &mut RadixCiphertext<PBSOrder>,
-        scalar: u64,
-    ) -> Result<(), CheckError> {
+        scalar: T,
+    ) -> Result<(), CheckError>
+    where
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
         if self.is_scalar_add_possible(ct, scalar) {
             self.unchecked_scalar_add_assign(ct, scalar);
             Ok(())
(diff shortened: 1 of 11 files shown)


tfhe/src/integer/u256.rs
```diff
@@ -269,6 +269,23 @@ impl std::ops::AddAssign<Self> for U256 {
...
+}
+
+impl std::ops::BitOrAssign<Self> for U256 {
+    fn bitor_assign(&mut self, rhs: Self) {
+        for (self_word, rhs_word) in self.0.iter_mut().zip(rhs.0) {
+            *self_word |= rhs_word;
+        }
+    }
+}
+
```
IceTDrinker, line 288: unit test when you get a chance or backlog issue

tfhe/src/integer/server_key/radix_parallel/neg.rs
```diff
@@ -78,16 +78,8 @@ impl ServerKey {
...
-        } else {
-            tmp_ctxt = ctxt.clone();
-            self.full_propagate_parallelized(&mut tmp_ctxt);
-            self.unchecked_neg(&tmp_ctxt)
-        };
-        self.full_propagate_parallelized(&mut ctxt);
-        ctxt
+        let mut flipped = self.bitnot_parallelized(ctxt);
+        self.scalar_add_assign_parallelized(&mut flipped, 1);
+        flipped
```
IceTDrinker, line 83: two default calls, could there be a more efficient combination of ops ?

tfhe/src/integer/server_key/radix/tests.rs
```diff
@@ -1165,12 +1176,12 @@ fn integer_smart_scalar_sub(param: ClassicPBSParameters) {
     }
 }
 
-fn integer_unchecked_scalar_decomposition_overflow(param: ClassicPBSParameters) {
-    // This is a regression test. The purpose here is to check if the number of decomposition
-    // blocks doesn't exceed 64 bits. This is why we test only 128 bits size.
-    // If overflow occurs the test case will panic.
+fn integer_unchecked_scalar_decomposition_overflow(param: PBSParameters) {
```
IceTDrinker, line 1179: this will break the CI as is, unless you want to add multi bit params (which IMO does not make sense for radix non parallelized) here you need to keep ClassicPBSParameters

IceTDrinker (commented):
High level comments not related to core logic

tfhe/src/integer/server_key/radix/tests.rs
tmontaigu: I rebased that too quickly :D

tfhe/src/integer/server_key/radix_parallel/neg.rs
tmontaigu: Its not that inefficient, since bitnot is just the tipical potentiel pre-emptive carry propagation as all defaults ops do, followed by parallel pbs on each block.

However we can indeed try to do

```
 let mut tmp_ctxt: RadixCiphertext<PBSOrder>;

        let ct = if !ctxt.block_carries_are_empty() {
            tmp_ctxt = ctxt.clone();
            self.full_propagate_parallelized(&mut tmp_ctxt);
            &tmp_ctxt
        } else {
            ctxt
        };

        if self.is_eligible_for_parallel_carryless_add() {
            let mut ct = self.unchecked_neg(ct);
            let generates_or_propagates = self.generate_init_carry_array(&ct);
            let input_carries =
                self.compute_carry_propagation_parallelized_low_latency(generates_or_propagates);
            ct.blocks
                .par_iter_mut()
                .zip(input_carries.par_iter())
                .for_each(|(block, carry_in)| {
                    self.key.unchecked_add_assign(block, carry_in);
                    self.key.message_extract_assign(block);
                });

            ct
        } else {
            let mut ct = self.unchecked_neg(ct);
            self.full_propagate_parallelized(&mut ct);
            ct
        }
```

which basicaly replaces the parallel bps via sequential lwe_additions

Tests seems ok, I'll run benches to see the difference

tfhe/src/integer/u256.rs
tmontaigu: I am creating a backlog issue where i'll keep track of every ops we need (and test) to be able to have U256 an implementor of the UnsignedInteger Trait (which sadly will be blocked by the fact that we don't yet have a I256)

tfhe/src/integer/server_key/radix_parallel/neg.rs
tmontaigu: Benches on m6i seems to give:

- 400ms for default_bitnot + default_add(1)
- 350ms for unchecked_neg + parallel carry propagation

tfhe/src/integer/server_key/radix/scalar_sub.rs
```diff
@@ -97,37 +126,23 @@ impl ServerKey {
...
+        PBSOrder: PBSOrderMarker,
+        T: UnsignedInteger + DecomposableInto<u8>,
+    {
+        let decomposer = self.create_negated_block_decomposer(scalar);
+        ct.blocks
+            .iter()
+            .zip(decomposer)
+            .all(|(ciphertext_block, scalar_block)| {
+                self.key
+                    .is_scalar_sub_possible(ciphertext_block, scalar_block)
```
IceTDrinker, line 144: scalar_add if you took the scalar's 2's complement ?

IceTDrinker (commented):
Few comments, it would be good to have a sort of comment/explanation of the logic behind the block decomposer like there used to be for the previous decomposition logic at encryption, it's not necessarily super easy to follow what's happening.

Not required to go ahead with this PR, a backlog issue (yay) could be good to not forget

tfhe/src/integer/server_key/radix/scalar_add.rs
```diff
@@ -50,25 +55,20 @@ impl ServerKey {
...
+        PBSOrder: PBSOrderMarker,
+        T: DecomposableInto<u8>,
+    {
+        let bits_in_message = self.key.message_modulus.0.ilog2();
+        let decomposer =
+            BlockDecomposer::with_early_stop_at_zero(scalar, bits_in_message).iter_as::<u8>();
+        for (ciphertext_block, scalar_block) in ct.blocks.iter_mut().zip(decomposer) {
+            self.key
+                .unchecked_scalar_add_assign(ciphertext_block, scalar_block);
         }
```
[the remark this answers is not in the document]
tmontaigu: Yes but I prefer to still write algorithm that rely on the fact that all blocks share the same message modulus,
then once we really need to add support for multiple message mod blocks we will work on all algorithms.

Just like to add signed integers support we will work again on all algorithms.

I think its going to be easier to manage: as of now no signed integer support at all, and no support for mixed message modulus

tfhe/src/integer/server_key/radix/scalar_sub.rs
```diff
@@ -34,44 +36,71 @@ impl ServerKey {
...
-        //Put each decomposition into a new ciphertext
-        for ct_i in ct.blocks.iter_mut() {
-            let mut decomp = neg_scalar & (mask * power);
-            decomp /= power;
+        // If we had upcasted the scalar, its msb would be zeros (0)
+        // then they would become ones (1) after the bitwise_not (!).
+        // We need to know if after the `+1` these padded bits are 0s or 1s.
+        // And we get this information by looking at if addition overflowed
+        // meaning there is a carry.
+        let padding_bit = if neg_scalar < flipped { 0u32 } else { 1u32 };
```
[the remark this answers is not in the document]
tmontaigu: overflowing_add is not in unsigned integers, adding it would also probably mean to add all of the other operations

tmontaigu:
> Few comments, it would be good to have a sort of comment/explanation of the logic behind the block decomposer like there used to be for the previous decomposition logic at encryption, it's not necessarily super easy to follow what's happening.

I did not write one because I felt like it was easier to understand, its just: bit mask + shift left, the more complex part maybe is the padding bit stuff added in this pr thanks to 3_3 existing and making stuff more complex

tfhe/src/integer/server_key/radix/scalar_add.rs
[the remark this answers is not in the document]
tmontaigu: Its an OKR on its own

tfhe/src/integer/server_key/radix/scalar_sub.rs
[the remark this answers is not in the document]
IceTDrinker: Yes I would think it’s ok to expand progressively for these operations

tfhe/src/integer/server_key/radix/scalar_add.rs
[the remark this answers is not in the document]
IceTDrinker: Fine by me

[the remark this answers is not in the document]
IceTDrinker: Just means a few algorithms are broken for parameters we provide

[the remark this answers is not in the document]
tmontaigu: Hmm, I would not say that
While true, it is not an intended use case of the current API of the crate

- As we don't provide functions that allow to encrypt while mixing message moduluses at the integer level, that means to have an integer ciphertext with mixed message modulus, the user would have to fiddle with the internals of the integer ciphertext after encryption or create its own encryption function using shortint API, construct a RadixCiphertext from these shortint, so then expecting the integer level to work is something else.
And of course by fidling the internals of the ciphertext I can break almost all functions available in the api.

- There is no mention of that possibility in the doc

- And more importantly, in my opinion, mixed message modulus was always broken since there are no tests for it, and with how easy it is to miss a corner case or simply not write a good algorithm or break existing stuff while trying to improve it, no tests == does not work == not supported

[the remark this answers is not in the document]
IceTDrinker: All I’m saying is anybody using 3_3 parameters with integers may encounter issues while not doing anything fancy :) and yes it won’t be fixed overnight and will require a big chunk of work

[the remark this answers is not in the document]
tmontaigu: I think only comparisons are known to be truly broken on 3_3, other op have tests which passes, maybe i'm more optimist on this than you 😁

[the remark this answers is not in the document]
IceTDrinker: Just a glass half empty kind of guy 😄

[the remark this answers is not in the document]
tmontaigu: To be fair, 3_3 is fancy, with the way we encrypt by using 3_3 you can't represent any of the common integer sizes (u8, u16, etc) you always end up with more bits, so a different modulus than if you compare it the the closes `uX` type, but our algorithms still works with 3_3, you just can't do arithmetic on (3 * 43 = 129 bits) and expects results to be the same as arithemtic over u128 its just not the same thing

tfhe/src/integer/server_key/radix/scalar_sub.rs
```diff
@@ -34,44 +36,70 @@ impl ServerKey {
...
-        //Put each decomposition into a new ciphertext
-        for ct_i in ct.blocks.iter_mut() {
-            let mut decomp = neg_scalar & (mask * power);
-            decomp /= power;
+        // If we had upcasted the scalar, its msb would be zeros (0)
+        // then they would become ones (1) after the bitwise_not (!).
+        // We need to know if after the `+1` these padded bits are 0s or 1s.
+        // And we get this information by looking at if addition overflowed
+        // meaning there is a carry.
+        let padding_bit = if has_overflowed { 0u32 } else { 1u32 };
```
IceTDrinker, line 77: the only case this can happen is for the value 0 right ? do we need to handle that case ? I'm not sure quite yet

tfhe/src/integer/server_key/radix/scalar_sub.rs
```diff
@@ -97,37 +125,23 @@ impl ServerKey {
...
+        PBSOrder: PBSOrderMarker,
+        T: UnsignedInteger + DecomposableInto<u8>,
+    {
+        let decomposer = self.create_negated_block_decomposer(scalar);
+        ct.blocks
+            .iter()
+            .zip(decomposer)
+            .all(|(ciphertext_block, scalar_block)| {
+                self.key
+                    .is_scalar_add_possible(ciphertext_block, scalar_block)
```
IceTDrinker, line 143: perhaps here a small comment explaining we are using 2's complement and so the add is the operation we want (?)

tfhe/src/integer/server_key/radix/scalar_sub.rs
tmontaigu: Yes I think its the only case it can happen, we could use `  let padding_bit = if scalar == T::ZERO { 0u32 } else { 1u32 };
` or `let padding_bit = u32::from(scalar != T::ZERO)`

IceTDrinker: my question is more do we have to have a special case for the padding bits in case the scalar is 0 ?

tmontaigu: Well, This handles the special case of when the scalar is zero, but its less obvious the way it is currently written

tmontaigu: if scalar is zero we could return an empty iterator as subtractring 0 means subtractring nothing

IceTDrinker: agreed, I think skipping computations for scalar == 0 is ok

tmontaigu: Opted for returning an Option<impl Iterator> and handling that case in the 2 function that needed to handle it

PR #316, opened 2023-06-02
fix(u256): align with rust for shift behaviours

tmontaigu:
Address comments made in https://github.com/zama-ai/tfhe-rs/pull/306#discussion_r1213409221

### DIFF
--- a/tfhe/src/integer/u256.rs
+++ b/tfhe/src/integer/u256.rs
@@ -157,9 +157,7 @@ impl std::ops::Sub<Self> for U256 {
 impl std::ops::ShrAssign<u32> for U256 {
     // move bits from MSB to LSB
     fn shr_assign(&mut self, shift: u32) {
-        if shift > 256 {
-            self.0.as_mut_slice().fill(0);
-        }
+        let shift = shift % Self::BITS;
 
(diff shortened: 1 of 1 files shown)


tfhe/src/integer/u256.rs
```diff
@@ -349,6 +345,26 @@ mod tests {
...
+
+        // We aim to have same behaviour as rust native types
+        assert_eq!(1u128.wrapping_shl(128), 1u128 << (128 % u128::BITS));
+        assert_eq!(1u128.wrapping_shl(129), 1u128 << (129 % u128::BITS));
+    }
+
+    #[test]
+    fn test_shr_limits() {
+        assert_eq!(U256::MAX >> 256, U256::MAX >> 0);
+        assert_eq!(U256::MAX >> 257, U256::MAX >> 1);
```
IceTDrinker, line 361: you don't reuse 256 % U256::BITS ?

IceTDrinker (approved):
Looks ok just a question about a pattern you don't re-use for u256 shift right

tfhe/src/integer/u256.rs
IceTDrinker: thanks !

PR #317, opened 2023-06-02
chore(wasm): remove serialization in web test

tmontaigu:
In the web wasm test we serialize the public key
to print its size (38_931_6265 bytes) this
means we hold the public key twice in ram.

I suspect this causes frequent out of
memory errors which then result in the
test timing out.

So we remove that hoping it has a positive impact


IceTDrinker (approved):
🔥

PR #324, opened 2023-06-06
CiphertextBase can convert from generic type to explicit type

twiby:
This PR provides CiphertextBase with functions to convert from a generic type OpOrder to a specific struct.

This allows removing all calls to std::mem::transmute in shortint/engine/server_side/mod.rs, isolating unsafe blocks in the conversion functions. This makes the code safer and more likely to panic! in case of an error.

This means when you have a function `my_fun` that takes a `CiphertextBase<KeyswitchBootstrap>` and you only have a `CiphertextBase<OpOrder: PBSO0rderMarker>` that you know to be the right type, you can just do `my_fun(ct.to_concrete_type())`.

### DIFF
--- a/tfhe/src/shortint/ciphertext/mod.rs
+++ b/tfhe/src/shortint/ciphertext/mod.rs
@@ -131,6 +131,34 @@ pub struct CiphertextBase<OpOrder: PBSOrderMarker> {
 pub type CiphertextBig = CiphertextBase<KeyswitchBootstrap>;
 pub type CiphertextSmall = CiphertextBase<BootstrapKeyswitch>;
 
+pub struct CiphertextTypeError<InputOpOrder, OutputOpOrder>
+where
+    InputOpOrder: PBSOrderMarker,
+    OutputOpOrder: PBSOrderMarker,
+{
+    input_type: PhantomData<InputOpOrder>,
+    output_type: PhantomData<OutputOpOrder>,
+}
+
+impl<InputOpOrder, OutputOpOrder> std::fmt::Debug
+    for CiphertextTypeError<InputOpOrder, OutputOpOrder>
+where
+    InputOpOrder: PBSOrderMarker,
+    OutputOpOrder: PBSOrderMarker,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Expected PBSOrder: {:?}, got {:?}, \
+            did you mix CiphertextBig or CompressedCiphertextBig ({:?}) with CiphertextSmall or CompressedCiphertextSmall ({:?})?",
+            InputOpOrder::pbs_order(),
+            OutputOpOrder::pbs_order(),
+            PBSOrder::KeyswitchBootstrap,
+            PBSOrder::BootstrapKeyswitch
+        )
+    }
+}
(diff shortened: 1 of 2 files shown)


IceTDrinker:
@cla-bot check

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -131,6 +131,9 @@ pub struct CiphertextBase<OpOrder: PBSOrderMarker> {
 pub type CiphertextBig = CiphertextBase<KeyswitchBootstrap>;
 pub type CiphertextSmall = CiphertextBase<BootstrapKeyswitch>;
 
+#[derive(Debug)]
+pub struct CiphertextExplicitConversionError {}
```
IceTDrinker, line 135: let's carry the input ciphertext op order and the output op order and implement format with a nice error message please !

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -142,6 +145,38 @@ impl<OpOrder: PBSOrderMarker> CiphertextBase<OpOrder> {
         self.carry_modulus = other.carry_modulus;
         self._order_marker = other._order_marker;
     }
+
+    pub fn to_explicit<OtherOpOrder: PBSOrderMarker>(&self) -> &CiphertextBase<OtherOpOrder> {
+        self.try_to_explicit().expect("Wrong ciphertext type.")
+    }
```
IceTDrinker, line 151: unsure on the naming here, maybe to_concrete_type ? or whatever the naming is when resolving a generic type to the actual type ? Maybe to_explicit is the right wording

With the implementation of fmt as proposed above you can unwrap on try explicit and get a nice error message

IceTDrinker (commented):
Thanks, a question on naming, and a request on better error message + unwrap in the to_explicit method :)

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: or even use generics here if you want to and make sure everything gets inlined/monomorphized at compile time

twiby: How about now ?

tfhe/src/shortint/ciphertext/mod.rs
twiby: Yes I've included better error handling.
About the naming: I'm not decided on anything. The most blatant use case is the `my_fun(ct.to_explicit())` call, which seemed clear enough, but `to_concrete_type` is fine as well. Since it's not technically a conversion we could just do `explicit` which will take less room in code

wyt ?

IceTDrinker (approved):
Looks good, just the naming thing, it's not critical but it feels like we can do a bit better than to_explicit :)

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: Looks good thanks !

tfhe/src/shortint/ciphertext/mod.rs
twiby: I've changed to `to_concrete_type`

twiby:
@slab-ci cpu_test

PR #325, opened 2023-06-06
chore(ci): put wasm tests in their own workflow

soonum:
This is mostly done to avoid failure on AWS tests (core, boolean, shortint, ...) workflow due to flaky tests in WASM.

### DIFF
--- a/ci/slab.toml
+++ b/ci/slab.toml
@@ -3,6 +3,11 @@ region = "eu-west-3"
 image_id = "ami-04deffe45b5b236fd"
 instance_type = "m6i.32xlarge"
 
+[profile.cpu-small]
+region = "eu-west-3"
+image_id = "ami-04deffe45b5b236fd"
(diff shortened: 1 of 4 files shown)


ci/slab.toml
```diff
@@ -23,6 +23,11 @@ workflow = "aws_tfhe_multi_bit_tests.yml"
 profile = "cpu-big"
 check_run_name = "CPU AWS Multi Bit Tests"
 
+[command.cpu_wasm_test]
+workflow = "aws_tfhe_wasm_tests.yml"
+profile = "cpu-big"
+check_run_name = "CPU AWS WASM Tests"
```
IceTDrinker, line 29: Can we use smaller than cpu big ? Like an 8 core machine to match a client ?

IceTDrinker (approved):

IceTDrinker (approved):
Thanks a lot !

PR #328, opened 2023-06-07
feat(integer): scalar_mul generic over UnsignedInteger

tmontaigu:
This makes the scalar_mul family of operation
accepts any scalar of type T that implements
the UnsignedInteger traits.

This unlocks scalar_multiplication with
scalar being a u128.

### DIFF
--- a/tfhe/src/integer/block_decomposition.rs
+++ b/tfhe/src/integer/block_decomposition.rs
@@ -130,6 +130,17 @@ where
 
         Some(masked)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        // In the case self we constructed with an early stop value
+        // the upper bound might be higher than the actual number of iteration.
+        //
+        // The size_hint docs states that it is ok (not best thing
+        // but won't break code)
+        let max_remaining_iter = self.num_bits_valid / self.num_bits_in_mask;
+        let min_remaining_iter = if max_remaining_iter != 0 { 1 } else { 0 };
+        (min_remaining_iter, Some(max_remaining_iter as usize))
+    }
 }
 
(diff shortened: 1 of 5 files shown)


tfhe/src/integer/block_decomposition.rs
```diff
@@ -130,6 +130,17 @@ where
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        // In the case self we constructed with an early stop value
+        // the upper bound might be higher than the actual number of iteration.
+        //
+        // The size_hint docs states that it is ok (not best thing
+        // but won't break code)
+        let max_remaining_iter = self.num_bits_valid / self.num_bits_in_mask;
+        let min_remaining_iter = if max_remaining_iter != 0 { 1 } else { 0 };
+        (min_remaining_iter, Some(max_remaining_iter as usize))
+    }
```
IceTDrinker, line 143: can we maybe do better with ilog2 of the scalar ? it would give a measure of leading zeros

IceTDrinker (approved):
One small thing with size_hints where we might be able to do a bit better, otherwise looks good thanks !

tfhe/src/integer/block_decomposition.rs
IceTDrinker: ilog2 should not be too bad to write for u256 as you can take the ilog 2 of all 64 blocks and adapt the value of the first MSB non zero value you get by adding the right multiple of 64

I don't know how interesting having a more precise size hints is, if it's too annoying let's just open an issue or progressively make u256 an UnsignedInteger by marking some operations todo!()

IceTDrinker: true, not a blocker, can be merged as is

PR #332, opened 2023-06-08
chore(ci): add a parser to output csv files for integer benchmarks

IceTDrinker:
- will simplify "just to see" benchmarks output parsing to share when iterating on performance work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ MIN_RUST_VERSION:=$(shell grep rust-version tfhe/Cargo.toml | cut -d '=' -f 2 | 
 AVX512_SUPPORT?=OFF
 WASM_RUSTFLAGS:=
 BIG_TESTS_INSTANCE?=FALSE
+GEN_KEY_CACHE_MULTI_BIT_ONLY?=FALSE
+PARSE_INTEGER_BENCH_CSV_FILE?=tfhe_rs_integer_benches.csv
 # This is done to avoid forgetting it, we still precise the RUSTFLAGS in the commands to be able to
 # copy paste the command in the terminal and change them if required without forgetting the flags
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -389,6 +389,15 @@ bench_pbs: install_rs_check_toolchain
 	--bench pbs-bench \
 	--features=$(TARGET_ARCH_FEATURE),boolean,shortint,internal-keycache,$(AVX512_FEATURE) -p tfhe
 
+.PHONY: parse_integer_benches # Run python parser to output a csv containing integer benches data
+parse_integer_benches:
+	if [[ "$${BENCH_CSV_FILE}" == "" ]]; then \
```
soonum, line 394: Does this double $ implies that you fetch this variable value from the environment ?

soonum (approved):
Just need a detail about the Makefile but the rest looks good.

Makefile
IceTDrinker: yes, you need to escape the dollar to have the "true" dollar sign for the script, not sure if environment is better or not 🤔 wdyt ?

soonum: It just seems odd since all other variable are handled at the beginning of the Makefile.
I would go for the classic way `BENCH_CSV_FILE="foo.csv" make parse_integer_benches`

IceTDrinker: I set the variable at the start of the workflow like the other file required for slab to have the env var available for parsing and upload

soonum: Yes but you can still do in the workflow something like `BENCH_CSV_FILE=${{ env.BENCH_CSV_FILE }} make parse_integer_benches`

IceTDrinker: alright will take a second look

soonum (approved):

PR #340, opened 2023-06-12
Am/c api/multi bit keygen

IceTDrinker:
@agnesLeroy no C header as discussed with @tmontaigu you will need to incorporate a build step for the C API for TFHE-rs, if you need infos on feature configs let us know !

### DIFF
--- a/tfhe/src/c_api/core_crypto/mod.rs
+++ b/tfhe/src/c_api/core_crypto/mod.rs
@@ -1,0 +1,122 @@
+use super::utils::*;
+use std::os::raw::c_int;
+
+#[no_mangle]
+pub unsafe extern "C" fn core_crypto_lwe_multi_bit_bootstrapping_key_element_size(
+    input_lwe_sk_dim: usize,
+    output_glwe_sk_dim: usize,
+    output_glwe_sk_poly_size: usize,
+    lwe_multi_bit_level_count: usize,
+    lwe_multi_bit_grouping_factor: usize,
+    result: *mut usize,
+) -> c_int {
+    catch_panic(|| {
+        use crate::core_crypto::prelude::*;
+
+        let result = get_mut_checked(result).unwrap();
+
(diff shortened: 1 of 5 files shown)


tfhe/src/c_api/core_crypto/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+            output_glwe_sk_dim.to_glwe_size(),
+            output_glwe_sk_poly_size,
+            lwe_multi_bit_base_log,
+            lwe_multi_bit_level_count,
+            input_lwe_sk_dim,
+            lwe_multi_bit_grouping_factor,
+            CiphertextModulus::new_native(),
+        );
+
+        *result = bsk.into_container().len();
```
IceTDrinker, line 39: can definitely avoid an allocation here, it was done quick and dirty (yikes)

agnesLeroy: multi-bit pbs key has this size:

```
int multi_bit_bsk_size = lwe_dimension * pbs_level * (glwe_dimension + 1) *
                 (glwe_dimension + 1) * polynomial_size *
                 (1 << grouping_factor) / grouping_factor;
```

agnesLeroy (approved):
I've written the multi-bit bsk size in comment in case you want to remove the unnecessary alloc, otherwise looks good to me. Thanks a lot!

tfhe/src/c_api/core_crypto/mod.rs
IceTDrinker: yeah, I'll re-use some primitives from concrete-core

IceTDrinker: @agnesLeroy be careful there is an API break, if you can just check the new entry point works ok for you ?

IceTDrinker: the formula looks good on your end anyways, but you can use the entry point from the C API to have consistency if you want

PR #341, opened 2023-06-12
chore(ci): notify slack channel in case of benchmarks failure

soonum:
resolves zama-ai/tfhe-rs-internal#155


IceTDrinker (approved):
Thanks a lot !

PR #344, opened 2023-06-12
Feature/casting keys

twiby:
This introduces casting keys to low levels of the API (switching a cipher from a pair of keys to another, for boolean, shortint, and integers, but not HL, which will be the subject of a future PR).

This is a draft for 2 reasons:

 - There is an uncertainty whether the present way actually respects an acceptable noise level. Waiting on JB and others to come back to me.
 - I feel the construction API is a bit clunky for now, though workable. I would love inputs on this.


tfhe/src/boolean/casting_key/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+impl CastingKey {
+    pub fn new(ck1: &ClientKey, ck2: &ClientKey) -> Self {
+        Self {
+            key_switching_key: BooleanEngine::with_thread_local_mut(|engine| {
+                engine.new_key_switching_key(ck1, ck2)
+            }),
+        }
+    }
+
+    pub fn cast_assign(&self, ct: &Ciphertext, ct_dest: &mut Ciphertext) {
```
tmontaigu, line 24: Generally our `_assign` functions takes the receiver (`&mut` argument) as the first argument

tfhe/src/shortint/casting_key/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    ///
+    /// let cipher = ck1.encrypt(1);
+    /// let cipher_2 = sk2.create_trivial(0);
+    /// ksk.cast_assign(&cipher, &mut cipher_2);
+    /// ```
+    pub fn cast_assign<OpOrder: PBSOrderMarker>(
+        &self,
+        ct: &CiphertextBase<OpOrder>,
+        ct_dest: &mut CiphertextBase<OpOrder>,
+    ) {
```
tmontaigu, line 102: Same as in boolean, generaly we take the dest as first argument in our assign functions

tmontaigu (commented):
Quick review, looks mostly good.

Still I'd lean towards naming them KeyswitchKey / KeyswitchingKey  rather than casting but thats to be discussed 🤔 

The construction API seems ok, what is it that you don't like about it?

Maybe in `pub fn new(ck1: &ClientKey, ck2: &ClientKey) -> Self` we could use `from` and `to` names instead of ck1, ck2.

We could go further by doing

```rust
pub struct ClientKey;

pub struct CastingDirection<'a> {
    from: &'a ClientKey,
    to: &'a ClientKey,
}

fn new_casting_key(CastingDirection { from, to }: CastingDirection<'_>) {
    todo!()
}

fn main() {
    let cks1 = ClientKey;
    let cks2 = ClientKey;
    new_casting_key(CastingDirection {
        from: &cks1,
        to: &cks2,
    });
}
```

but is it worth ?

tfhe/src/boolean/casting_key/mod.rs
tmontaigu: I would name it `cast_into` then

twiby: Ok let's do this thx

twiby:
I've renamed everything to "KeySwitchingKey" and similar things when needed. It think it's gonna be clearer in the future.

For the construction, your suggestion is nice ! I was wondering whether it was wise to pass the ClientKey around other stuff, like one mistake and pouf it is sent somewhere (the alternative is that everything using the client key happens in `generate_keys`, but that's even clunkier). Maybe that's too much worry, and I should keep it simple as it is

tmontaigu:
Yes that 's too much worry, the api in itself is good, the produced switching key can be shared, then its a matter of the developper making sure it uses the feature correctly.

We cannot completely control how the developper makes his application use/share the ClientKey, we can only tell them things like "the client key is meant to be private as it allows decryption", and if the developpers really need to share that key then the responsibility of doing that safely is on him

twiby:
I've pushed the last step before this is done: handling of parameters.

Building a KeySwitchingKey requires 2 parameters: the B and l of the decomposition during the key switch, and these are gonna be hyper specific depending on (but not deducible from) the 2 parameter systems between which we switch. Even in the case when the user just changes keys without changing the parameters, one can't just compute at runtime the correct parameters.

To handle this situation, I took a hint from the current handling of parameters: providing an `unsafe` `new` function, specifying in the documentation that you should know what you're doing. Only one set of parameters are actually provided : `PARAM_KEYSWITCH_1_1_TO_2_2`. The constructor of the `KeySwitchingKey` needs an instance of these parameters now.

Most of the unit tests use unsafe parameters that should kinda always work, but we are not sure these actually respect the 2 to the -40 probability of error.

PR #347, opened 2023-06-12
feat(c_api): safer destroys

tmontaigu:
This adds a null check in the different destroy function of the C API. The performance impact of this should be negligible / inexistant but should make usage more ergonomic and safer.

Also this renames the detroy functions from being named `destroy_shortint_*` | `destroy_boolean_*` to
`shortint_destroy_*` | `boolean_destroy_*` as it is more coherent with the rest of the API where functions starts with `shortint` or `boolean`.

Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/76

### DIFF
--- a/tfhe/src/c_api/shortint/destroy.rs
+++ b/tfhe/src/c_api/shortint/destroy.rs
@@ -8,7 +8,10 @@ use super::{
 };
 
 #[no_mangle]
-pub unsafe extern "C" fn destroy_shortint_client_key(client_key: *mut ShortintClientKey) -> c_int {
+pub unsafe extern "C" fn shortint_destroy_client_key(client_key: *mut ShortintClientKey) -> c_int {
+    if client_key.is_null() {
+        return 0;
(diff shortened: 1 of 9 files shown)


tfhe/src/c_api/shortint/destroy.rs
```diff
@@ -8,7 +8,10 @@ use super::{
 };
 
 #[no_mangle]
-pub unsafe extern "C" fn destroy_shortint_client_key(client_key: *mut ShortintClientKey) -> c_int {
+pub unsafe extern "C" fn shortint_destroy_client_key(client_key: *mut ShortintClientKey) -> c_int {
+    if client_key.is_null() {
+        return 0;
+    }
```
IceTDrinker, line 14: if at some point it makes sense to start macro-izing these modules feel free to do so !

IceTDrinker (approved):
Thanks a lot !

PR #353, opened 2023-06-14
cb/feat/support_large_custom_modulus_multiplication

Carl-Zama:
Fixes an issue where if we have two polynomials p_1 mod q and p_2 mod q with q a custom 64-bit modulus and we want to compute p_1 * p_2 mod q^2 then the result could be incorrect.

This is due to the polynomial multiplication base case (when using a custom modulus) using a modified schoolbook algorithm which first adds two polynomials together (giving coefficients > 2^64 in general) first before multiplying two such polynomials. This can result in trying to multiply two integers > 2^64 modulo a 2^128 bit custom modulus which fails to be correct due to a wrap around modulo 2^128.

The fix now checks for a wrap around and if it occurs tries to compute the correct result using a simple divide and conquer approach and also fails instead of outputting an incorrect result.

### DIFF
--- a/tfhe/src/core_crypto/commons/numeric/unsigned.rs
+++ b/tfhe/src/core_crypto/commons/numeric/unsigned.rs
@@ -157,10 +157,52 @@ macro_rules! implement {
                 let self_u128: u128 = self.cast_into();
                 let other_u128: u128 = other.cast_into();
                 let custom_modulus_u128: u128 = custom_modulus.cast_into();
-                self_u128
-                    .wrapping_mul(other_u128)
-                    .wrapping_rem(custom_modulus_u128)
-                    .cast_into()
+                let (prod, wrong) = self_u128.overflowing_mul(other_u128);
+                // if we are not able to multiply directly without wrapping around
+                if wrong {
+                    // we try to do the multiplication as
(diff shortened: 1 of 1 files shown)


tfhe/src/core_crypto/commons/numeric/unsigned.rs
```diff
@@ -157,10 +157,63 @@ macro_rules! implement {
...
+                    );
+                    let top = top.wrapping_rem(custom_modulus_u128);
+                    let out = top.wrapping_add(middle).wrapping_rem(custom_modulus_u128);
+                    out.wrapping_add(bottom)
+                        .wrapping_rem(custom_modulus_u128)
+                        .cast_into()
+                } else {
+                    prod.wrapping_rem(custom_modulus_u128).cast_into()
+                }
             }
```
IceTDrinker, line 217: For performance I would probably prefer something like:

```rust
// Case for unsigned integers smaller than u128, e.g. u64
if Self::BITS < 128
    self_u128
                    .wrapping_mul(other_u128)
                    .wrapping_rem(custom_modulus_u128)
                    .cast_into()
else {
    // New code goes here and will be used for u128 only and hopefully compiled only for u128
};
```

IceTDrinker:
As discussed, closing this as it's not yet in main and you can therefore use the code in your work branch 🙂

PR #354, opened 2023-06-14
Add KeySwitchingKey in the high-level API

twiby:
This is the follow-up PR from #344
This is allowing HL users to swich keys for boolean, shortint, or integer ciphertexts.

What's missing before merging: the actual key switching parameters. In the HL API we only allow switching from one set of parameters to the same, but we would still need to provide key switching parameters, which are not available right now. In the current state of the PR I use a workaround that kinda works but don't have the guarantee of "below 2 to the -40th proba of error", but this should be fixed before merging.

### DIFF
--- a/tfhe/src/high_level_api/details.rs
+++ b/tfhe/src/high_level_api/details.rs
@@ -18,6 +18,8 @@ macro_rules! define_key_structs {
                     [<$base_ty_name CompressedPublicKey>],
                     [<$base_ty_name ServerKey>],
                     [<$base_ty_name CompressedServerKey>],
+                    [<$base_ty_name KeySwitchingKey>],
+                    [<$base_ty_name KeySwitchingParameters>],
                 };
             )*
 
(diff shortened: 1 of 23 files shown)


aquint-zama:
@cla-bot check

IceTDrinker:
what's the status on this @tmontaigu ?

tmontaigu:
waiting for parameters i think

IceTDrinker:
let's reset the commit author to avoid the cla bot freaking out

IceTDrinker:
and add a co-autored line in the commit to Render unto Caesar the things that are Caesar's

tmontaigu:
@slab-ci cpu_fast_test

IceTDrinker:
I see conflicts

tmontaigu:
Yes it was caused by the merge of https://github.com/zama-ai/tfhe-rs/commit/238359135189a8630ddf4ef1de0712d2874c1f4a

tfhe/src/high_level_api/shortints/types/base.rs
```diff
@@ -13,16 +13,16 @@ use crate::high_level_api::errors::OutOfRangeError;
...
-    FheBootstrap, FheDecrypt, FheEq, FheNumberConstant, FheOrd, FheTrivialEncrypt, FheTryEncrypt,
-    FheTryTrivialEncrypt,
+    FheBootstrap, FheDecrypt, FheEq, FheKeySwitch, FheNumberConstant, FheOrd, FheTrivialEncrypt,
+    FheTryEncrypt, FheTryTrivialEncrypt,
 };
-use crate::high_level_api::PublicKey;
+use crate::high_level_api::{KeySwitchingKey, PublicKey};
 
-use super::{GenericShortIntClientKey, GenericShortIntServerKey};
+use super::{GenericShortIntClientKey, GenericShortIntKeySwitchingKey, GenericShortIntServerKey};
```
IceTDrinker, line 25: is it "limited" to shortint for now ?

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -663,6 +663,26 @@ where
...
+            .unwrap_display();
+
+        let ciphertext = integer_casting_key.cast(&self.ciphertext);
+        Self::new(ciphertext, id)
+    }
+}
+
 impl<P> GenericInteger<P>
 where
     P: IntegerParameter,
```
IceTDrinker, line 688: casting file below is emtpy ?

tfhe/src/high_level_api/details.rs
```diff
@@ -212,6 +214,82 @@ macro_rules! define_key_structs {
...
+                            [<$name _key>]: match (
+                                key_pair_1.0.[<$name _key>].as_ref(),
+                                key_pair_1.1.[<$name _key>].as_ref(),
+                                key_pair_2.0.[<$name _key>].as_ref(),
+                                key_pair_2.1.[<$name _key>].as_ref())
+                            {
+                                (Some(ck1), Some(sk1), Some(ck2), Some(sk2)) => {
+                                    // Forward error if parameters are not compatible
+                                    let key = <[<$base_ty_name KeySwitchingKey>]>::for_same_parameters((ck1, sk1), (ck2, sk2))?;
+                                    Some(key)
```
IceTDrinker, line 274: no need to forward ? so need for the `?` and the let binding here I would say as it already returns an option

tfhe/src/high_level_api/integers/tests.rs
```diff
@@ -817,3 +817,71 @@ fn test_if_then_else() {
...
+    let a = a.keyswitch(&ksk);
+    let da: u16 = a.decrypt(&client_key_2);
+    assert_eq!(da, clear);
+}
+
+#[test]
+fn test_integer_key_switching_2() {
+    let mut parameters = crate::shortint::parameters::PARAM_MESSAGE_1_CARRY_1_KS_PBS;
+    parameters.message_modulus.0 = 4;
+    parameters.carry_modulus.0 = 4;
```
IceTDrinker, line 846: this is beyond weird and likely incorrect ? or are we getting away with "murder" by just encrypting ? plus integer keyswitching between different params would require resizing blocks and shuffling stuff around right in the general case ?

IceTDrinker (commented):
First comments will finish the larger files in a second pass

tfhe/src/high_level_api/shortints/types/base.rs
tmontaigu: No keyswitchting exists for all types

tfhe/src/high_level_api/integers/tests.rs
tmontaigu: This is what JB told me to do (or at least what I understand I should do for the test)

IceTDrinker: yes ok, I'm just not clear on what the consequence is for correctness of the encryption KS (PBS ?) decryption we are testing here

does this tests pass let's say 1000 iterations with different inputs/keys ?

tfhe/src/high_level_api/details.rs
tmontaigu: ermf, here its a bit more complex than that I think:

If we don't have a `Some(_)` for each key in the pairs the match forwards a None the be placed at the member level 
```rust
e.g: Self {
uint2: None
}
```

If we do have all Some(_), then we try to create a ksk using the `for_same_param`, however that function will return None if the keys in the pair do not actually use the same params.

In the case they indeed do not use the same params, we want the whole function to return None, and not just the key member (eg uint2_key) to be None.

IceTDrinker: if it works like this fine, was just thinking it looked odd

tmontaigu: One potential thing to make it clearer would be to create an error type

```rust
struct IncompatibleParameters(type); // type is an enum variant, either FheBool, FheUint2, Integer

and have the `for_same_params` functions return a `Result<Self, IncompatibleParameter>`
```

IceTDrinker: I think it derserves a comment though potentially

tmontaigu: Ok i'll add the comment, and not the error struct as its les tedious

tmontaigu:
Closing as the internals of the high level api are going to change (even more than it already did), and adding this feature will be be easier to redo from 'scratch'

PR #359, opened 2023-06-16
refactor(all): remove PBSOrder generic marker

tmontaigu:
The motivation for this change is that having the Big/Small being a generic parameter of types, makes the code more complex than it should be.

Also, It was not complete, not all structs had this generic parameter making the whole thing a bit more clunky.


twiby (approved):
Seems good to me, though of course this is really big, and something might have slipped me. I tried to scan everything (except the c/wasm/js bindings, because I don't know the code well enough), and it looks alright.
The change looks great, it's gonna be much simpler for users, and future maintainers as well.

tfhe/docs/js_on_wasm_api/tutorial.md
tmontaigu, line 1: These should still be BigInt

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

tfhe/js_on_wasm_tests/test.js
```diff
@@ -1,233 +1,233 @@
-const crypto = require('crypto');
```
IceTDrinker, line 1: missed that 🤦‍♂️

PR #361, opened 2023-06-16
chore(core): remove Deref and DerefMut for Polynomial

IceTDrinker:
- add indexing and iteration primitives

### DIFF
--- a/tfhe/src/core_crypto/entities/polynomial.rs
+++ b/tfhe/src/core_crypto/entities/polynomial.rs
@@ -2,7 +2,7 @@
 
 use crate::core_crypto::commons::parameters::*;
 use crate::core_crypto::commons::traits::*;
-use std::ops::{Deref, DerefMut};
+use std::ops::{Index, IndexMut};
 
 /// A [`polynomial`](`Polynomial`).
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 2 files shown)


tfhe/src/core_crypto/entities/polynomial.rs
```diff
@@ -22,6 +24,104 @@ impl<T, C: ContainerMut<Element = T>> AsMut<[T]> for Polynomial<C> {
...
+        self.as_ref().index(index)
+    }
+}
+
+impl<C: ContainerMut> IndexMut<RangeToInclusive<usize>> for Polynomial<C> {
+    fn index_mut(&mut self, index: RangeToInclusive<usize>) -> &mut Self::Output {
+        self.as_mut().index_mut(index)
+    }
+}
+
```
tmontaigu, line 124: Maybe it is possible to shorten these with 

```rust
impl<C, I> Index<I> for Polynomial<C>
where C: Container + Index<I> {
    type Output = <C as Index<I>>::Output;

    fn index(&self, index: I) -> &Self::Output {
        self.as_ref().index(index)
    }
}
```

and Same for IndexMut

IceTDrinker: I'll check, but I tried a generic type name without bound it did not seem to work ok, maybe I wrote it wrong

IceTDrinker: actually, I was trying to go through the self.as_ref() that returns a slice and the indexing seemed to not work ok the first time around, now looks to be fine, must have been a bit tired Friday, updating

tmontaigu: Yeah if you use .as_ref()/.as_mut the bounds would have to be `for <'a> &'a [C::Element]: Index<I>`

IceTDrinker: looks like it was a bit easier than that, but this time I managed to get it working 🤷‍♂️

tmontaigu (approved):

PR #364, opened 2023-06-16
Chore/m1 tests without integer

IceTDrinker:


IceTDrinker:
closing this again, as it's still suffering from issues from lto for now and other weird failures happening during CI but not in a "normal" terminal

PR #368, opened 2023-06-20
feat(hlapi): bind shift by encrypted amount

tmontaigu:
### PR content/description

To keep things easy we have to drop the part
where the macro generated the docs.

### Check-list:

### DIFF
--- a/tfhe/src/high_level_api/integers/tests.rs
+++ b/tfhe/src/high_level_api/integers/tests.rs
@@ -148,7 +148,7 @@ fn test_uint32() {
 }
 
 #[test]
-fn test_uint32_shift_and_rotate() {
+fn test_uint32_shift() {
     let config = ConfigBuilder::all_disabled()
         .enable_default_integers()
         .build();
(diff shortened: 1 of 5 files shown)


tfhe/src/high_level_api/integers/tests.rs
```diff
@@ -147,6 +147,130 @@ fn test_uint32() {
...
+        .enable_default_integers()
+        .build();
+
+    let (cks, sks) = generate_keys(config);
+
+    use rand::prelude::*;
+
+    let mut rng = rand::thread_rng();
+    let clear_a = rng.gen::<u32>();
+    let clear_b = rng.gen_range(0u32..32u32);
```
IceTDrinker, line 162: no overflow case ? not sure what the semantics are for rust in that case

tfhe/src/high_level_api/integers/tests.rs
```diff
@@ -147,6 +147,130 @@ fn test_uint32() {
...
```
IceTDrinker, line 224: no overflow case ? not sure what the semantics are for rust in that case

IceTDrinker (commented):
Thanks a lot for this ! A few questions here and there, nothing major/truly blocking

tfhe/src/high_level_api/integers/tests.rs
tmontaigu: semantics are same as masking shift value with the maximum num bits (shift % 32 in 32 bits), overflow values are tested in integer tests normally

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -382,60 +383,8 @@ where
 }
 
 macro_rules! generic_integer_impl_operation (
-    ($rust_trait_name:ident($rust_trait_method:ident,$op:tt) => $key_method:ident) => {
-        #[doc = concat!(" Allows using the `", stringify!($op), "` operator between a")]
-        #[doc = " `GenericInteger` and a `GenericInteger` or a `&GenericInteger`"]
```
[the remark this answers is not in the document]
tmontaigu: u8 does not impl shift by a u8 amount nor Wrapping<u8> only by u32/usize, the macros generated docs for binary ops
where both lhs and rhs were Wrapping<u8>. So i prefer removing it and rely on the user docs to showcase operations. 

If having this kind of docstring is that important then I think its better to re-craft a better one, that would also be able to
generate docs for assign and scalar ops

tfhe/src/high_level_api/integers/tests.rs
IceTDrinker: ah right, sorry forgot

tfhe/src/high_level_api/integers/types/base.rs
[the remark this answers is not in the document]
IceTDrinker: Understood

Care to open a follow-up issue so that we add this back in down the line ? Or maybe have some sort of "stub" documentation pointing to the rust documentation for the expected behavior ?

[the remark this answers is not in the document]
IceTDrinker: @tmontaigu maybe repurpose 97 ? up to you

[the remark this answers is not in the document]
IceTDrinker: agreed

tfhe/src/high_level_api/integers/tests.rs
IceTDrinker: if they are in integer feel free to close

IceTDrinker (approved):
Modulo the small house keeping tasks good to merge then 👍 thanks !

PR #369, opened 2023-06-20
Add Trivium/Kreyvium application of tfhe-rs

twiby:
This PR adds an application of the-rs, in a separate Cargo workspace. This allows using Trivium and Kreyvium as transciphering schemes.

For this application to be possible, I added 2 things in tfhe-rs itself:
 - Constructors of GenericInteger from a RadixCiphertext or a Vec<Ciphertext>: this allows users to make api levels interact, via building a FheUint8 from the individual blocks.
 - AsRef for ClientKey and ServerKey, to be able to extract the underlying integer key. I don't think this is done very gracefully (in particular, this approach wouldn't translate well to shortint and boolean keys).

Unit tests have been added in the Trivium/Kreyvium workspace, but some are quite slow (several minutes). Should they be reduced ? Are these tests going to run in CI though they aren't in the same workspace ?


apps/trivium/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+Example of a Rust main below:
+```rust
+use tfhe::{ConfigBuilder, generate_keys, FheBool};
+use tfhe::prelude::*;
+
+use tfhe_trivium::TriviumStream;
+
+fn get_hexagonal_string_from_lsb_first_stream(a: Vec<bool>) -> String {
```
tmontaigu, line 21: hexagonal or hexadecimal ?

apps/trivium/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dev-dependencies]
+criterion = { version = "0.3", features = [ "html_reports" ]}
+
+[[bench]]
+name = "trivium"
+harness = false
+[[bench]]
+name = "casting"
+harness = false
```
IceTDrinker, line 20: missing newline at the end of the file to be Unix compliant, we have an issue to add a line linter we haven't had time for now to add it

apps/trivium/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tfhe = { path = "../../tfhe", features = [ "boolean", "shortint", "integer" ] }
+rayon = { version = "1.7.0"}
+
+[dev-dependencies]
+criterion = { version = "0.3", features = [ "html_reports" ]}
```
IceTDrinker, line 13: there is a 0.4 cirterion unless you specifically need some behavior of 0.3 I think it would be good to update

apps/trivium/README.md
```diff
@@ -0,0 +1,204 @@
+# FHE boolean Trivium implementation using tfhe-rs crate
```
IceTDrinker, line 1: I think I would go for the simpler "using TFHE-rs"

apps/trivium/src/trivium/test.rs
```diff
@@ -0,0 +1,411 @@
+use tfhe::prelude::*;
```
IceTDrinker, line 1: can you add a Makefile target to run the tests of trivium please ? :)

apps/trivium/src/trivium/test.rs
```diff
@@ -0,0 +1,411 @@
...
+use tfhe::{generate_keys, ConfigBuilder, FheBool, FheUint64, FheUint8};
+
+use crate::{TransCiphering, TriviumStream, TriviumStreamByte, TriviumStreamShortint};
+
+///! Values for these tests come from: https://github.com/cantora/avr-crypto-lib/blob/master/testvectors/trivium-80.80.test-vectors
```
IceTDrinker, line 6: I think ///! is not a doc comment you would need to use //! maybe at the start of the file, can you maybe select a specific commit from that repo if the authors make updates and move stuff (unlikely I agree) it would break this link

apps/trivium/src/kreyvium/test.rs
```diff
@@ -0,0 +1,377 @@
+use tfhe::prelude::*;
+use tfhe::{generate_keys, ConfigBuilder, FheBool, FheUint64, FheUint8};
+
+use crate::{KreyviumStream, KreyviumStreamByte, KreyviumStreamShortint, TransCiphering};
+
+///! Values for these tests come from: https://github.com/renaud1239/Kreyvium
```
IceTDrinker, line 6: same remark for the doc comment, and same thing to get a given commit to avoid having that link break if possible

apps/trivium/benches/casting.rs
```diff
@@ -0,0 +1,81 @@
+use tfhe::shortint::prelude::*;
+
```
IceTDrinker, line 2: we may want to bench the casting in shortint actually

IceTDrinker (commented):
First bach of comment haven't had time to read everything yet

apps/trivium/benches/kreyvium_bool.rs
```diff
@@ -0,0 +1,77 @@
...
+            val >>= 1;
+        }
+    }
+
+    c.bench_function("kreyvium bool warmup", |b| {
+        b.iter(|| {
+            set_server_key(server_key.clone());
+            let cipher_key = key.map(|x| FheBool::encrypt(x, &client_key));
+            unset_server_key();
+            let _kreyvium = KreyviumStream::<FheBool>::new(cipher_key, iv, &server_key);
```
IceTDrinker, line 74: this looks a bit odd to include the server key part in the warmup benchmark ?

Also the unset server key looks misplaced

apps/trivium/benches/kreyvium_byte.rs
```diff
@@ -0,0 +1,98 @@
...
+    let iv_string = "0D74DB42A91077DE45AC000000000000".to_string();
+    let mut iv = [0u8; 16];
+
+    for i in (0..iv_string.len()).step_by(2) {
+        iv[i >> 1] = u8::from_str_radix(&iv_string[i..i + 2], 16).unwrap();
+    }
+
+    c.bench_function("kreyvium byte warmup", |b| {
+        b.iter(|| {
+            set_server_key(server_key.clone());
```
IceTDrinker, line 92: same question for warmup and set server key

apps/trivium/benches/trivium_bool.rs
```diff
@@ -0,0 +1,77 @@
...
+        let mut val: u8 = u8::from_str_radix(&iv_string[i..i + 2], 16).unwrap();
+        for j in 0..8 {
+            iv[8 * (i >> 1) + j] = val % 2 == 1;
+            val >>= 1;
+        }
+    }
+
+    c.bench_function("trivium bool warmup", |b| {
+        b.iter(|| {
+            set_server_key(server_key.clone());
```
IceTDrinker, line 71: ditto bench warmup

apps/trivium/benches/trivium_byte.rs
```diff
@@ -0,0 +1,95 @@
...
+    let iv_string = "0D74DB42A91077DE45AC".to_string();
+    let mut iv = [0u8; 10];
+
+    for i in (0..iv_string.len()).step_by(2) {
+        iv[i >> 1] = u8::from_str_radix(&iv_string[i..i + 2], 16).unwrap();
+    }
+
+    c.bench_function("trivium byte warmup", |b| {
+        b.iter(|| {
+            set_server_key(server_key.clone());
```
IceTDrinker, line 89: ditto bench warmup

apps/trivium/src/trivium/trivium_shortint.rs
```diff
@@ -0,0 +1,190 @@
+use crate::static_deque::StaticDeque;
+
+// use tfhe::prelude::*;
```
IceTDrinker, line 3: comment can be removed

apps/trivium/src/trivium/trivium_byte.rs
```diff
@@ -0,0 +1,258 @@
...
+        let (ret5, ((ret1, ret2), (ret3, ret4))) = rayon::join(
+            move || reg5.byte(offsets[4]),
+            || {
+                rayon::join(
+                    || rayon::join(move || reg1.byte(offsets[0]), move || reg2.byte(offsets[1])),
+                    || rayon::join(move || reg3.byte(offsets[2]), move || reg4.byte(offsets[3])),
+                )
+            },
+        );
+        (ret1, ret2, ret3, ret4, ret5)
```
IceTDrinker, line 250: par_iter_mut + enumerate and a match on the index maybe to fill with the correct offset ?

I'm not a fan of nested rayon join

apps/trivium/src/trivium/trivium_byte.rs
```diff
@@ -0,0 +1,258 @@
...
+            }
+            None => (),
+        }
+        let values = self.get_64_output_and_values();
+        match &self.fhe_key {
+            Some(_) => {
+                rayon::broadcast(|_| unset_server_key());
+            }
+            None => (),
+        }
```
IceTDrinker, line 216: is it necessary to unset ? can't we assume computations will set their keys if required ?

The only usage where I think this could make sense is with a "stack" of server keys where we restore the previous key that was there, in that case I don't think I see the point of unsetting our key ? unless it allows to drop memory more easily, but I guess we don't have an actual copy of the BSK in each thread right ?

It's a general remark, I won't pin point all unset sk calls

apps/trivium/src/trivium/trivium_byte.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let ((a1, a2, a3, a4, a5), ((b1, b2, b3, b4, b5), (c1, c2, c3, c4, c5))) = rayon::join(
+            || Self::get_bytes(a_byte, [91 - n, 90 - n, 68 - n, 65 - n, 92 - n]),
+            || {
+                rayon::join(
+                    || Self::get_bytes(b_byte, [82 - n, 81 - n, 77 - n, 68 - n, 83 - n]),
+                    || Self::get_bytes(c_byte, [109 - n, 108 - n, 86 - n, 65 - n, 110 - n]),
+                )
+            },
+        );
```
IceTDrinker, line 157: [a_byte, b_bytes, c_byte].par_iter().zip([array of indices ordered properly].par_iter()) ?

Not sure if nested joins are discouraged or not, I just find it awkard, unless you benched it and perf looks better with this rather than parallel iterators

general remak for nested joins, I let you check if there is a way to update the nested join pattern to parallel iterators, and if it has some performance implications, hopefully the code is not less readable with parallel iterators, I let you be the judge of that, performance being criterion number 1 to select one pattern over the other

apps/trivium/src/static_deque/static_bit_deque.rs
```diff
@@ -0,0 +1,136 @@
+//! This module implements the StaticBitDeque struct: a deque of bytes. The idea
```
IceTDrinker, line 1: StaticBitDeque or StaticByteDeque as it seems to store bytes ?

Really a detail

apps/trivium/src/static_deque/static_bit_deque.rs
```diff
@@ -0,0 +1,136 @@
...
+    T: StaticBitDequeInput<T>,
+    for<'a> &'a T: StaticBitDequeInput<T>,
+{
+    /// Constructor always uses a fully initialized array, the first element of
+    /// which is oldest, the last is newest
+    pub fn new(_arr: [T; N]) -> Self {
+        Self {
+            deque: StaticDeque::<N, T>::new(_arr),
+        }
+    }
```
IceTDrinker, line 40: put an assert on the size of T here ? as it is required for it to be a byte-like object with an explicit error message

apps/trivium/src/static_deque/static_deque.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn index_mut(&mut self, i: usize) -> &mut T {
+        if i >= N {
+            panic!("Index {:?} too high for size {:?}", i, N);
+        }
+        &mut self.arr[(N + self.cursor - i - 1) % N]
+    }
+}
+
+#[test]
+fn test_static_deque() {
```
IceTDrinker, line 76: this should probably be in a test module otherwise the final binary will be shipping test functions

I let you check the tests are all behind a `#[cfg(test)]`

apps/trivium/src/static_deque/static_deque.rs
```diff
@@ -0,0 +1,130 @@
...
+    let a = [1, 2, 3, 4, 5, 6];
+
+    let mut static_deque = StaticDeque::new(a);
+    for i in 7..11 {
+        static_deque.push(i);
+    }
+    assert_eq!(static_deque.arr, [7, 8, 9, 10, 5, 6]);
```
IceTDrinker, line 83: most likely will need to use the getter

apps/trivium/src/static_deque/static_bit_deque.rs
```diff
@@ -0,0 +1,136 @@
...
+        return (byte << bit_idx) | (byte_next >> (8 - bit_idx as u8));
+    }
+}
+
+#[test]
+fn bit_deque_test() {
+    let mut deque = StaticBitDeque::<3, u8>::new([2, 64, 128]);
+    deque.push(4);
+
+    // Yougest: 4
```
IceTDrinker, line 89: youngest ?

apps/trivium/src/kreyvium/kreyvium_shortint.rs
```diff
@@ -0,0 +1,206 @@
+use crate::static_deque::StaticDeque;
+
+// use tfhe::prelude::*;
```
IceTDrinker, line 3: check for comments like this to be removed in other files, I think I saw it in a few places

IceTDrinker (commented):
Another batch of comments with a question around rayon::join vs parallel iterators not sure it has an impact nor that we want to change the pattern, I just find nested joins a bit harder to read and am unsure on a potential performance advantage of one over the other.

Question around unset_server_key

thanks a lot for all the work !

apps/trivium/benches/casting.rs
```diff
@@ -0,0 +1,81 @@
...
+                    let pair_idx = i % 4;
+
+                    let b0 = &vec_ct[8 * byte_idx + 2 * pair_idx];
+                    let b1 = &vec_ct[8 * byte_idx + 2 * pair_idx + 1];
+
+                    ksk.cast(
+                        &server_key_1.unchecked_add(b0, &server_key_1.unchecked_scalar_mul(b1, 2)),
+                    )
+                })
+                .collect::<Vec<_>>();
```
tmontaigu, line 34: Feels strange to redefine the function in the bench, when its already in transcipher_from_1_1_stream

apps/trivium/benches/kreyvium_bool.rs
```diff
@@ -0,0 +1,77 @@
...
+    let iv_string = "0D74DB42A91077DE45AC000000000000".to_string();
+    let mut iv = [false; 128];
+
+    for i in (0..iv_string.len()).step_by(2) {
+        let mut val: u8 = u8::from_str_radix(&iv_string[i..i + 2], 16).unwrap();
+        for j in 0..8 {
+            iv[8 * (i >> 1) + j] = val % 2 == 1;
+            val >>= 1;
+        }
+    }
```
tmontaigu, line 32: Seems like this piece of code is often copy pasted, could it be moved in a function ?

apps/trivium/src/trivium/test.rs
twiby: I'm adding one for Trivium and one for Kreyvium

apps/trivium/benches/casting.rs
twiby: Do you suggest moving this file in `tfhe/benches/shortint` ?

apps/trivium/src/static_deque/static_bit_deque.rs
twiby: But if T can be a FheUint8, it can have any size, doesn't it ?

Also the byte-like object is a soft requirement: in practice you only need a type that implements some bitshift operations

apps/trivium/src/trivium/trivium_byte.rs
twiby: My thinking was that a potential user of this implementation might have other things going on in their life (other computations with other server keys). I didn't want for the implem to have as a side effect that every thread of the rayon pool will have this key set. (Imagine if they forget to set their server key: I'd rather see them have a runtime error than incorrect computations)

There is one alternative: that the user is responsible for setting the key before calling this. But with rayon/multithreading this is clunky because they will have to use rayon::broadcast or equivalent.

This seemed (and seems) best to me for now

apps/trivium/benches/casting.rs
twiby: It's close, but here you don't do the Fheuint64::xor
I thought about factorizing this part in a function, but it would make the main code a little weirder, to only improve a bit the bench code, it didn't seem worth it

apps/trivium/src/static_deque/static_bit_deque.rs
IceTDrinker: Indeed was a bit too optimistic on that assert haha

apps/trivium/src/static_deque/static_bit_deque.rs
twiby: This is much better, I don't like StaticBitDeque at all

apps/trivium/src/trivium/trivium_byte.rs
```diff
@@ -0,0 +1,258 @@
...
+        reg: StaticBitDeque<N, T>,
+        offsets: [usize; 5],
+    ) -> (T, T, T, T, T) {
+        let [reg1, reg2, reg3, reg4, reg5] = [
+            reg.clone(),
+            reg.clone(),
+            reg.clone(),
+            reg.clone(),
+            reg.clone(),
+        ];
```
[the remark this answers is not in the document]
twiby: Maybe no change in performance, but in readability ! I'll do that

apps/trivium/benches/kreyvium_bool.rs
twiby: I agree ! It's only used in tests and benches though, I'm open to suggestions to where I should put it (hoping it doesn't get compiled into the prod binary). Like a `utils.rs` file in a test folder ?

apps/trivium/src/trivium/trivium_byte.rs
twiby: I agree it's ugly. I'm looking for ways to remove the nested join pattern as much as possible ! I think par_iter has the same performance and readability, when applicable. What is discourage by the Rayon doc is using `scope` (the ideal API here): they clearly state it's more convenient but less efficient

IceTDrinker: if it makes sense, no need to waste time if it's already ok and there is no real benefit

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -98,6 +98,27 @@ where
...
+where
+    P: IntegerParameter,
+    P::Id: Default,
+    RadixCiphertext: From<Vec<T>>,
+{
+    fn from(blocks: Vec<T>) -> GenericInteger<P> {
+        GenericInteger::from(RadixCiphertext::from(blocks))
+    }
+}
+
```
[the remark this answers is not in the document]
twiby: I agree ! I can check the number of blocks and change these From to TryFrom. Should I also add a check in GenericInteger::new ?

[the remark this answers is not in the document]
twiby: This requires using information from the ServerKey by the way, so a user must have called `set_server_key` before using these try_from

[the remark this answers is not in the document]
twiby: I've pushed changes, I'm waiting on your comment before resolving

apps/trivium/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "tfhe-trivium"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tfhe = { path = "../../tfhe", features = [ "boolean", "shortint", "integer" ] }
+rayon = { version = "1.7.0"}
+
```
IceTDrinker, line 11: can you take inspiration from https://github.com/zama-ai/concrete_internal/blob/359d4c59190f0f0cb70ec6b6dba77b1c24dfea76/concrete-integer/Cargo.toml#L23 to select the hardware feature depending on platform ? :)

apps/trivium/benches/kreyvium_bool.rs
tmontaigu: We could do something like in the tfhe benches https://github.com/zama-ai/tfhe-rs/blob/7ea13715eeaf006cf39849b4fd22226d2deb6b1a/tfhe/benches/integer/bench.rs#L3C1-L3C28

But we can probably leave with how the current code is

PR #377, opened 2023-06-22
feat(integer): implement more U256 operators

tmontaigu:
### PR content/description

This implements the following operators for U256
- BitXor
- Mul
- is_power_of_two
- ilog2
- SubAssign

This makes us progress on https://github.com/zama-ai/tfhe-rs-internal/issues/120

### Check-list:

### DIFF
--- a/tfhe/src/integer/u256.rs
+++ b/tfhe/src/integer/u256.rs
@@ -1,10 +1,57 @@
-use crate::core_crypto::prelude::{CastFrom, Numeric};
+use crate::core_crypto::prelude::{CastFrom, Numeric, UnsignedInteger};
 
 #[inline(always)]
-pub const fn adc(l: u64, r: u64, c: bool) -> (u64, bool) {
+pub fn add_with_carry<T: UnsignedInteger>(l: T, r: T, c: bool) -> (T, bool) {
     let (lr, o0) = l.overflowing_add(r);
-    let (lrc, o1) = lr.overflowing_add(c as u64);
+    let (lrc, o1) = lr.overflowing_add(T::cast_from(c));
     (lrc, o0 | o1)
+}
+
+pub fn add_assign_words<T: UnsignedInteger>(lhs: &mut [T], rhs: &[T]) {
+    let iter = lhs
+        .iter_mut()
+        .zip(rhs.iter().copied().chain(std::iter::repeat(T::ZERO)));
+
+    let mut carry = false;
+    for (lhs_block, rhs_block) in iter {
+        let (result, out_carry) = add_with_carry(*lhs_block, rhs_block, carry);
+        *lhs_block = result;
+        carry = out_carry;
+    }
+}
+
+/// lhs and rhs are slice of words.
+///
+/// They must be in lsb -> msb order
+pub(crate) fn schoolbook_mul_assign(lhs: &mut [u64], rhs: &[u64]) {
+    assert!(lhs.len() >= rhs.len());
+    let mut terms = Vec::with_capacity(rhs.len());
+
+    for (i, rhs_block) in rhs.iter().copied().enumerate() {
+        let mut blocks = Vec::with_capacity(lhs.len() + i);
+        blocks.resize(i, 0u64); // pad with 0
+
+        let mut carry = 0;
+        for lhs_block in lhs.iter().copied() {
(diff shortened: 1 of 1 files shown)


tfhe/src/integer/u256.rs
```diff
@@ -344,10 +454,189 @@ impl From<u128> for U256 {
...
+    fn test_is_power_of_two() {
+        assert!(!U256::ZERO.is_power_of_two());
+        assert!(!U256::MAX.is_power_of_two());
+        assert!(!U256::from(8329842348123u64).is_power_of_two());
+
+        assert!(U256::ONE.is_power_of_two());
+        assert!(U256::from(2u64).is_power_of_two());
+        assert!(U256::from(1u64 << 32).is_power_of_two());
+        assert!(U256::from(1u64 << 47).is_power_of_two());
+        assert!(U256::from(1u128 << 84).is_power_of_two());
```
IceTDrinker, line 603: can we do them all in a few loops maybe ?

tfhe/src/integer/u256.rs
```diff
@@ -344,10 +454,189 @@ impl From<u128> for U256 {
...
+
+        let mut rng = rand::thread_rng();
+        for _ in 0..5 {
+            let a = rng.gen::<u64>();
+            let b = rng.gen::<u64>();
+
+            let res = U256::from(a) * U256::from(b);
+            let expected = a as u128 * b as u128;
+            assert_eq!(res, U256::from(expected));
+        }
```
IceTDrinker, line 637: could we have some cases where high bits are set ? maybe with some power of 2 cases ?

Maybe we can re-use some test cases from libs wih some u256 implementations

tfhe/src/integer/u256.rs
```diff
@@ -1,12 +1,56 @@
-use crate::core_crypto::prelude::{CastFrom, Numeric};
+use crate::core_crypto::prelude::{CastFrom, Numeric, UnsignedInteger};
 
 #[inline(always)]
-pub const fn adc(l: u64, r: u64, c: bool) -> (u64, bool) {
+pub fn adc<T: UnsignedInteger>(l: T, r: T, c: bool) -> (T, bool) {
```
IceTDrinker, line 4: what does adc mean ?

tfhe/src/integer/u256.rs
```diff
@@ -344,10 +454,189 @@ impl From<u128> for U256 {
...
+
+    #[test]
+    fn test_ilog2() {
+        assert!(catch_unwind(|| { U256::ZERO.ilog2() }).is_err());
+
+        assert_eq!(U256::MAX.ilog2(), 255);
+        assert_eq!(
+            U256::from(8329842348123u64).ilog2(),
+            8329842348123u64.ilog2()
+        );
```
IceTDrinker, line 614: could we do some more values here ?

IceTDrinker (commented):
Thanks for this, I don't know if it's the most urgent thing, I know it's annoying but some doc work would go a long way !

I would like some add/mul tes cases with high bits set that we could find from the internet

tfhe/src/integer/u256.rs
tmontaigu: add carry or something like that 😅

tfhe/src/integer/u256.rs
tmontaigu: Will need to find examples we can hardcode then

tfhe/src/integer/u256.rs
tmontaigu: I think yes

tmontaigu:
> I know it's annoying but some doc work would go a long way !

What kind of docs ? User oriented or more algorithms description ?

IceTDrinker:
> > I know it's annoying but some doc work would go a long way !
> 
> What kind of docs ? User oriented or more algorithms description ?

Sorry I meant outside of this u256 PR, I think more user oriented at the moment

tfhe/src/integer/u256.rs
tmontaigu: Added some hardcoded examples, checked using python

IceTDrinker: awesome thanks !

IceTDrinker (approved):
Thanks for the additional tests mul coming from python !

PR #380, opened 2023-06-23
chore(ci): add pbs throughput benchmarks

soonum:
closes: zama-ai/tfhe-rs-internal#164

### PR content/description

This implies to add a conversion method to CiphertextModulus in order to create the CryptoParametersRecord struct used as utils.

### Check-list:

* [ ] ~~Docs have been added / updated (for bug fixes / features)~~
* [ ] ~~Check for breaking changes (including serialization changes) and add them to commit message following the conventional commit [specification][conventional-breaking]~~

### DIFF
--- a/ci/benchmark_parser.py
+++ b/ci/benchmark_parser.py
@@ -95,6 +95,7 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
 
             for stat_name, value in parse_estimate_file(subdir).items():
                 test_name_parts = list(filter(None, [test_name, stat_name, name_suffix]))
+
                 result_values.append(
                     _create_point(
                         value,
--- a/tfhe/benches/boolean/bench.rs
+++ b/tfhe/benches/boolean/bench.rs
@@ -17,7 +17,7 @@ criterion_group!(
 criterion_main!(gates_benches);
 
 /// Helper function to write boolean benchmarks parameters to disk in JSON format.
-pub fn write_to_json_boolean<T: Into<CryptoParametersRecord>>(
+pub fn write_to_json_boolean<T: Into<CryptoParametersRecord<u32>>>(
     bench_id: &str,
     params: T,
     params_alias: impl Into<String>,
--- a/tfhe/benches/core_crypto/pbs_bench.rs
+++ b/tfhe/benches/core_crypto/pbs_bench.rs
@@ -1,8 +1,10 @@
 #[path = "../utilities.rs"]
 mod utilities;
 use crate::utilities::{write_to_json, CryptoParametersRecord, OperatorType};
+use rayon::prelude::*;
 
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use serde::Serialize;
 use tfhe::boolean::parameters::{BooleanParameters, DEFAULT_PARAMETERS, TFHE_LIB_PARAMETERS};
 use tfhe::core_crypto::prelude::*;
 use tfhe::shortint::keycache::NamedParam;
@@ -47,9 +49,16 @@ criterion_group!(
                 multi_bit_deterministic_pbs::<u32>,
 );
 
-criterion_main!(pbs_group, multi_bit_pbs_group);
+criterion_group!(
+    name = pbs_throughput_groug;
+    config = Criterion::default().sample_size(100);
+    targets = pbs_throughput::<u64>, pbs_throughput::<u32>
+);
 
-fn benchmark_parameters<Scalar: Numeric>() -> Vec<(String, CryptoParametersRecord)> {
+criterion_main!(pbs_group, multi_bit_pbs_group, pbs_throughput_groug);
+
+fn benchmark_parameters<Scalar: UnsignedInteger>() -> Vec<(String, CryptoParametersRecord<Scalar>)>
+{
     if Scalar::BITS == 64 {
         SHORTINT_BENCH_PARAMS
             .iter()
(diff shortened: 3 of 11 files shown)


ci/benchmark_parser.py
```diff
@@ -95,19 +95,25 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
...
-                        "latency",
+                        bench_type,
                         operator,
                         params,
                         display_name=display_name
                     )
                 )
 
-                if stat_name == "mean" and compute_throughput:
+                if stat_name == "mean" and "throughput" not in test_name and compute_throughput:
```
IceTDrinker, line 116: is the throughput not in test name required ? that would be weird if compute_throughput is True but we don't compute it ?

tfhe/benches/boolean/bench.rs
```diff
@@ -17,7 +17,7 @@ criterion_group!(
 criterion_main!(gates_benches);
 
 /// Helper function to write boolean benchmarks parameters to disk in JSON format.
-pub fn write_to_json_boolean<T: Into<CryptoParametersRecord>>(
+pub fn write_to_json_boolean<T: Into<CryptoParametersRecord<u64>>>(
```
IceTDrinker, line 20: what's the u64 here ? if it's for the ciphertext modulus I believe boolean is u32 ?

tfhe/examples/utilities/boolean_key_sizes.rs
```diff
@@ -42,7 +42,7 @@ fn client_server_key_sizes(results_file: &Path) {
         let test_name = format!("boolean_key_sizes_{params_name}_ksk");
 
         write_result(&mut file, &test_name, ksk_size);
-        write_to_json(
+        write_to_json::<u64, _>(
```
IceTDrinker, line 45: u32 for boolean, doesn't the crypto param record already have the u32 set somewhere for Booleans ?

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -212,6 +222,37 @@ impl<Scalar: UnsignedInteger> CiphertextModulus<Scalar> {
             CiphertextModulusInner::Custom(modulus) => modulus.is_power_of_two(),
         }
     }
+
+    pub fn try_to<ScalarTo: UnsignedInteger + CastInto<u128>>(
+        &self,
+    ) -> Option<CiphertextModulus<ScalarTo>> {
+        let new_inner = match self.inner {
```
IceTDrinker, line 229: why do you need this ?

tfhe/benches/utilities.rs
```diff
@@ -25,10 +25,11 @@ pub struct CryptoParametersRecord {
     pub cbs_base_log: Option<DecompositionBaseLog>,
     pub message_modulus: Option<usize>,
     pub carry_modulus: Option<usize>,
+    pub ciphertext_modulus: Option<CiphertextModulus<Scalar>>,
 }
 
 #[cfg(feature = "boolean")]
-impl From<BooleanParameters> for CryptoParametersRecord {
+impl<Scalar: UnsignedInteger> From<BooleanParameters> for CryptoParametersRecord<Scalar> {
```
IceTDrinker, line 32: I think we may fix Scalar = u32 for Boolean

tfhe/benches/utilities.rs
```diff
@@ -47,12 +48,16 @@ impl From<BooleanParameters> for CryptoParametersRecord {
...
         }
     }
 }
 
 #[cfg(feature = "shortint")]
-impl From<ClassicPBSParameters> for CryptoParametersRecord {
+impl<Scalar> From<ClassicPBSParameters> for CryptoParametersRecord<Scalar>
+where
+    Scalar: UnsignedInteger + CastInto<u128>,
+{
```
IceTDrinker, line 60: and here probably we can fix Scalar = u64 for shortint params

IceTDrinker:
I think you could create an enum for crypto records which has boolean/shortint variants to avoid the annoyance of the scalar generic with hardcoded scalars for each variant

ci/benchmark_parser.py
soonum: Oh you're right ! The output from criterion is a latency even if we set up the bench to be parallel. We still have to compute the throughput while parsing.

IceTDrinker: Does not look like it is solved (sorry if it was in the other commit you were working on)

tfhe/benches/core_crypto/pbs_bench.rs
```diff
@@ -504,3 +540,126 @@ fn multi_bit_deterministic_pbs<Scalar: UnsignedTorus + CastInto<usize> + CastFro
...
+
+        let glwe_secret_key = GlweSecretKey::new_empty_key(
+            Scalar::ZERO,
+            params.glwe_dimension.unwrap(),
+            params.polynomial_size.unwrap(),
+        );
+        let big_lwe_sk = glwe_secret_key.into_lwe_secret_key();
+        let big_lwe_dimension = big_lwe_sk.lwe_dimension();
+
+        const NUM_CTS: usize = 16384;
```
IceTDrinker, line 572: if we only test up to 512, we can lower this to 512

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -212,6 +222,40 @@ impl<Scalar: UnsignedInteger> CiphertextModulus<Scalar> {
             CiphertextModulusInner::Custom(modulus) => modulus.is_power_of_two(),
         }
     }
+
+    pub fn try_to<ScalarTo: UnsignedInteger + CastInto<u128>>(
+        &self,
+    ) -> Option<CiphertextModulus<ScalarTo>> {
```
IceTDrinker, line 228: A result I think makes more sense here

PR #382, opened 2023-06-23
feat(shortint): add encrypt_message_and_carry

jborfila:
Function to choose message and carry modulus.


tfhe/src/shortint/client_key/mod.rs
```diff
@@ -149,6 +150,49 @@ impl ClientKey {
         })
     }
 
+    /// Encrypt a small integer message using the client key with a specific message and carry
+    /// moduli
+    ///
+    /// # Warning
+    /// Defining specific message AND carry moduli might lead to undefined behavior.
```
IceTDrinker, line 157: undefined behavior or crash ? You have an assert in the engine implem.

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -149,6 +150,49 @@ impl ClientKey {
         })
     }
 
+    /// Encrypt a small integer message using the client key with a specific message and carry
+    /// moduli
```
IceTDrinker, line 154: missing final `.` at the end of the sentence

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -149,6 +150,49 @@ impl ClientKey {
...
+    /// use tfhe::shortint::parameters::{MessageModulus, PARAM_MESSAGE_2_CARRY_2};
+    /// use tfhe::shortint::{CarryModulus, ClientKey};
+    ///
+    /// // Generate the client key
+    /// let cks = ClientKey::new(PARAM_MESSAGE_2_CARRY_2);
+    ///
+    /// let msg = 3;
+    ///
+    /// // Encryption of one message with MessageModulus = 2 and CarryModulus = 2
+    /// // so that 6*3 < 2^(2 + 2) from the parameter set
```
IceTDrinker, line 171: not sure I understand the 6*3 here ? especially since 6 * 3 = 18 and 2 ^ 4 = 16

tfhe/src/shortint/client_key/mod.rs
tmontaigu: Undefined behaviour as a specific meaning in C/C++/Rust,etc.

And In Rust, under no circumstance ever, a safe function (so a function not marked as `unsafe`) should
be able to invoke undefined behavior.

So either the use of 'undefined behavior' wording is not correct, or codes needs changes

tfhe/src/shortint/server_key/tests.rs
```diff
@@ -3128,6 +3131,61 @@ where
...
+
+    for _ in 0..NB_TEST {
+        let mut msg_modulus = rng.gen::<u64>() % param_msg_mod as u64;
+        while msg_modulus <= 1 {
+            msg_modulus = rng.gen::<u64>() % param_msg_mod as u64;
+        }
+        let mut carry_modulus = rng.gen::<u64>() % param_carry_mod as u64;
+        while carry_modulus <= 1 {
+            carry_modulus = rng.gen::<u64>() % param_carry_mod as u64;
+        }
```
tmontaigu, line 3155: might want to do
```rust
let msg_modulus = rns.gen_range(2u64..param_msg_mod);
let carry_modulus = rns.gen_range(2u64..param_carry_mod);
```

IceTDrinker (approved):
Suggestion from Thomas is good, we could do it ourselves if you are too busy

tmontaigu (approved):

PR #385, opened 2023-06-26
fix(integer): fix strict assert in add_paralellized

tmontaigu:
### PR content/description
Should fix https://github.com/zama-ai/tfhe-rs-internal/issues/167

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -272,7 +272,7 @@ impl ServerKey {
     /// # Requirements
     ///
     /// - The parameters have 4 bits in total
-    /// - The input carries of both lhs and rhs must be empty
+    /// - Adding rhs to lhs must not consume more than one carry
     ///
     /// # Output
     ///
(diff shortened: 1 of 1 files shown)


tmontaigu:
@slab-ci cpu_integer_test

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -282,8 +282,12 @@ impl ServerKey {
         lhs: &mut RadixCiphertext,
         rhs: &RadixCiphertext,
     ) {
-        debug_assert!(lhs.block_carries_are_empty());
-        debug_assert!(rhs.block_carries_are_empty());
+        let degree_after_add_does_not_go_beyond_first_carry =
+            lhs.blocks.iter().zip(rhs.blocks.iter()).all(|(bl, br)| {
+                let degree_after_add = bl.degree.0 + br.degree.0;
+                degree_after_add < (self.key.message_modulus.0 * 2)
```
IceTDrinker, line 288: should be message_modulus * carry_modulus ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -382,8 +386,12 @@ impl ServerKey {
...
```
IceTDrinker, line 392: same thing ?

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: No, it would not be the same `message_modulus * carry_modulus` is the whole plaintext space, here we want to not go beyond first carry

IceTDrinker: ah yes indeed ! thanks

IceTDrinker (approved):

PR #394, opened 2023-06-29
feat(core): use ntt-based polynomial multiplication for large polynomials

sarah-ek:
### PR content/description
use concrete-ntt for negacyclic polynomial multiplication. this speeds up keygen by 2x for MESSAGE_2_CARRY_2 and 6x for MESSAGE_4_CARRY_4 

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
+++ b/tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
@@ -5,6 +5,8 @@ use crate::core_crypto::commons::numeric::UnsignedInteger;
 use crate::core_crypto::commons::parameters::MonomialDegree;
 use crate::core_crypto::commons::traits::*;
 use crate::core_crypto::entities::*;
+use concrete_ntt::{native128, native32, native64};
+use core::any::TypeId;
 
 /// Add a polynomial to the output polynomial.
 ///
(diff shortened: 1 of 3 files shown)


IceTDrinker (commented):
Could you check the tests of the polynomial mul and cases to be sure we cover all the threshold domains ?

i.e. N < 64
64 <= N <= 256
256 < N

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -712,7 +778,7 @@ mod test {
             let mut rng = rand::thread_rng();
 
             // random settings settings
-            let polynomial_log = (rng.gen::<usize>() % 7) + 6;
+            let polynomial_log = (rng.gen::<usize>() % 7) + 4;
```
IceTDrinker, line 781: so the range changes from 6..12 to 4..10 ?

can we maybe expand it or keep it the same ? as we now have params with polynomial_log up to 16 IIRC ?

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -605,6 +649,7 @@ pub fn polynomial_karatsuba_wrapping_mul<Scalar, OutputCont, LhsCont, RhsCont>(
 }
 
 const KARATUSBA_STOP: usize = 64;
+const NTT_THRESHOLD: usize = 256;
```
IceTDrinker, line 652: I don't see the specialized/hard coded test cases for the various ranges of possible Polynomial sizes ? To be sure each domain is tested

IceTDrinker (commented):
Thanks for the schoolbook test function, still the question about specific tests for each threshold range

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -703,40 +748,58 @@ mod test {
         assert_eq!(&poly, &ground_truth);
     }
 
+    fn negacyclic_multiply_schoolbook<Scalar: UnsignedInteger>(
```
mayeul-zama, line 751: Could be used in the `else` block of `polynomial_wrapping_add_mul_assign`

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -703,40 +834,71 @@ mod test {
...
+                // test
+                assert_eq!(&sb_mul, &ka_mul);
+            }
+        }
+    }
+
+    /// test if we have the same result when using schoolbook or ntt/karatsuba
+    /// for random polynomial multiplication
+    fn test_sub_mul<T: UnsignedTorus>() {
+        // 50 times the test
```
IceTDrinker, line 874: comment is out of date

IceTDrinker (commented):
Two small comments, did you clear up the NTT overflow you mentioned at some point ?

I seem to recall you were worried some computations might overflow

tfhe/src/core_crypto/algorithms/polynomial_algorithms.rs
```diff
@@ -150,29 +152,54 @@ pub fn polynomial_wrapping_add_mul_assign<Scalar, OutputCont, InputCont1, InputC
...
+                    let lhs: &[u32] = unsafe { transmute(lhs) };
+                    let rhs: &[u32] = unsafe { transmute(rhs) };
+                    let out: &mut [u32] = unsafe { transmute(out) };
+                    let mut tmp = Polynomial::new(0u32, polynomial_size);
+                    plan.negacyclic_polymul(tmp.as_mut(), lhs, rhs);
+                    polynomial_wrapping_add_assign(&mut Polynomial::from_container(out), &tmp);
+                    return;
+                }
+            } else if id == TypeId::of::<u64>() {
+                if let Some(plan) = native64::Plan32::try_new(polynomial_size.0) {
```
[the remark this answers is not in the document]
sarah-ek: yeah, internally the plan decomposes the input into several `u32` since `u64` doesn't have efficient simd multiplication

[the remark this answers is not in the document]
IceTDrinker: ok thanks !

sarah-ek:
> Two small comments, did you clear up the NTT overflow you mentioned at some point ?
> 
> I seem to recall you were worried some computations might overflow

i still haven't determined the overflow limit. i'll need to compute it from the prime values in concrete-ntt. just to be safe. i'll add a test case for a polynomial size of 2^16

IceTDrinker:
> > Two small comments, did you clear up the NTT overflow you mentioned at some point ?
> > I seem to recall you were worried some computations might overflow
> 
> i still haven't determined the overflow limit. i'll need to compute it from the prime values in concrete-ntt. just to be safe. i'll add a test case for a polynomial size of 2^16

do we want to wait for you to have the overflow limit ? So that we are safe here before merging ?

sarah-ek:
> > > Two small comments, did you clear up the NTT overflow you mentioned at some point ?
> > > I seem to recall you were worried some computations might overflow
> > 
> > 
> > i still haven't determined the overflow limit. i'll need to compute it from the prime values in concrete-ntt. just to be safe. i'll add a test case for a polynomial size of 2^16
> 
> do we want to wait for you to have the overflow limit ? So that we are safe here before merging ?

as long as one of the polynomial operands contains only small values (positive, or negative), then we should be safe
i believe this is the case since for keygen, one of the polynomials has only 0/1 values

IceTDrinker (changes requested):
For now as a safety, requesting changes until we have the thresholds for NTT potential overflows and a counter measure in place

IceTDrinker:
> > > > Two small comments, did you clear up the NTT overflow you mentioned at some point ?
> > > > I seem to recall you were worried some computations might overflow
> > > 
> > > 
> > > i still haven't determined the overflow limit. i'll need to compute it from the prime values in concrete-ntt. just to be safe. i'll add a test case for a polynomial size of 2^16
> > 
> > 
> > do we want to wait for you to have the overflow limit ? So that we are safe here before merging ?
> 
> as long as one of the polynomial operands contains only small values (positive, or negative), then we should be safe i believe this is the case since for keygen, one of the polynomials has only 0/1 values

Indeed for the small values, I know for research they used some very big polynomials when testing something we may want to check with JB to see what's preferred in that case, I just would like to avoid having silent failures in the NTT polynomial code

IceTDrinker:
Maybe with unit tests with worst cases and not using NTT for the larger cases where there may be overflows but we need to be rock solid on that

IceTDrinker:
@sarah-ek could we maybe separate the test overhaul in a different PR and merge it ? so that the improvements are in main and the ntt-only problem can be investigated without blocking the fixes you did ? 🙂

PR #397, opened 2023-06-29
chore(ci): use multi-bit params in shortint for pbs benchmarks

soonum:
Use up-to-date crypto parameters for PBS benchmarks with multi-bit instead of hardcoded ones.


IceTDrinker (approved):
Thanks !

PR #399, opened 2023-06-29
chore(bench): proper param name fix for WASM bench

IceTDrinker:
previous fix had a typo with an extra ks_pbs which was in the wrong place, rework naming system to have constants available


soonum (commented):
I'm not sure about the outcome of this, does this mean that every shortint parameters name will be suffixed by `_NAME` now ?

tmontaigu (approved):

IceTDrinker:
> I'm not sure about the outcome of this, does this mean that every shortint parameters name will be suffixed by `_NAME` now ?

No, we create pub const &'static str names $param_NAME containing the name of the parameter, I use the static name in the bench to be able to match

PR #407, opened 2023-06-30
chore(ci): bench multiple scalar bit sizes and add operators

soonum:
Scalar operations benchmark is done against scalar bit sizes.
Scalar division and modulo are added to the benchmark set.

### DIFF
--- a/.github/workflows/integer_full_benchmark.yml
+++ b/.github/workflows/integer_full_benchmark.yml
@@ -38,16 +38,27 @@ jobs:
           echo "Type: ${{ inputs.instance_type }}"
           echo "Request ID: ${{ inputs.request_id }}"
 
-      - name: Checkout tfhe-rs repo with tags
-        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
-        with:
-          fetch-depth: 0
-
       - name: Get benchmark details
         run: |
           echo "BENCH_DATE=$(date --iso-8601=seconds)" >> "${GITHUB_ENV}"
           echo "COMMIT_DATE=$(git --no-pager show -s --format=%cd --date=iso8601-strict ${{ github.sha }})" >> "${GITHUB_ENV}"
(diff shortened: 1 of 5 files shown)


soonum:
@slab-ci integer_bench

soonum:
@slab-ci integer_bench

IceTDrinker:
potential issues : the parse integer to csv would have a broken format in that case (not hard to update) but would also need to handle scalar formatted benches (to get the scalar bit size) and the non scalar variant, I have an updated piece of code somewhere but I think it begs the question of the format in general, do we need the ServerKey part for example ? as it's common between shortint and integer I think I would replace it by the module name so something like shortint::add_parallelized instead of ServerKey::add_parallelized or similar

soonum:
@slab-ci integer_full_bench

soonum:
@slab-ci integer_full_bench

soonum:
@slab-ci integer_full_bench

soonum:
@slab-ci integer_full_bench

soonum:
@slab-ci shortint_full_bench

soonum:
@slab-ci integer_full_bench

soonum:
@slab-ci shortint_full_bench

soonum:
@slab-ci shortint_full_bench

.github/workflows/integer_full_benchmark.yml
```diff
@@ -41,7 +46,7 @@ jobs:
       - name: Checkout tfhe-rs repo with tags
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
         with:
-          fetch-depth: 0
+            fetch-depth: 0
```
IceTDrinker, line 49: indent looks wrong

IceTDrinker (commented):
Just a nitpick and then go

IceTDrinker (approved):

IceTDrinker:
Thanks !

PR #411, opened 2023-07-06
fix(integer): fix U256::copy_to_be_byte_slice

tmontaigu:
### PR content/description

There was a bug in to_be_bytes_slice, it was missing a `slice.reverse()` (from_be_bytes correctly has it)

The from/to functions have been refactored to used from_be_bytes / to_be_bytes, etc from the stdlib to only have one layer of endianess to manage.

The test value used did not catch that so we change the value used to expose the problem.

### Check-list:

### DIFF
--- a/tfhe/src/integer/u256.rs
+++ b/tfhe/src/integer/u256.rs
@@ -85,6 +85,8 @@ where
     (quotient, remainder)
 }
 
+const BYTES_PER_U64: usize = std::mem::size_of::<u64>() / std::mem::size_of::<u8>();
+
 // Little endian order
 #[derive(Default, Copy, Clone, Debug, PartialEq, Eq)]
 #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


tfhe/src/integer/u256.rs
```diff
@@ -101,88 +101,38 @@ impl U256 {
...
-                std::slice::from_raw_parts_mut(inner_slice.as_mut_ptr() as *mut u8, 32);
-            let shortest_len = inner_byte_slice.len().min(bytes.len());
-            inner_byte_slice[..shortest_len].copy_from_slice(&bytes[..shortest_len]);
-            inner_byte_slice.reverse();
-        }
-        #[cfg(target_endian = "big")]
-        for word in self.0.iter_mut() {
-            *word = word.swap_bytes();
+
+        let mut array = [0u8; 8];
```
IceTDrinker, line 105: can we maybe even use std::mem::size_of::<u64>() instead of 8 ?

IceTDrinker (commented):
thx a small comment, let me know if you want to take it into account, otherwise I'll approve

IceTDrinker (approved):
Thanks !

Issue #410, opened 2023-07-04
The `r` resulted from integer `division` when multiplied together always return `0`.


y-pakorn:
**Describe the bug**
The `remainer` or `r` resulted from integer `division` when multiplied together always return `0`.

**To Reproduce**
Steps to reproduce the behaviour

1. Install latest `tfhe`

```toml
tfhe = { git = "https://github.com/zama-ai/tfhe-rs", features = [ "boolean", "shortint", "integer", "internal-keycache"] }
```

2. Encrypt some number with integer radix key
3. Do two division that results in `q` and `r`, In this case `ServerKey::div_rem_parallelized`.
4. Multiply `r` from those division together.
5. The result from that multiplication will always be **zero**.

```rust
fn main() {
    let (client_key, server_key) = IntegerKeyCache.get_from_params(PARAM_MESSAGE_2_CARRY_2);
    const NUM_BLOCK: usize = 4;
    let a: u128 = 248;
    let b: u128 = 249;
    let c: u128 = 250;
    let d: u128 = 251;
    let enc_a = client_key.encrypt_radix(a, NUM_BLOCK);
    let enc_b = client_key.encrypt_radix(b, NUM_BLOCK);
    let enc_c = client_key.encrypt_radix(c, NUM_BLOCK);
    let enc_d = client_key.encrypt_radix(d, NUM_BLOCK);

    let (mut q1, mut r1) = server_key.div_rem_parallelized(&enc_b, &enc_a);
    let (mut q2, mut r2) = server_key.div_rem_parallelized(&enc_d, &enc_c);

    println!("r1: {:?}", client_key.decrypt_radix::<u8>(&r1));
    println!("r2: {:?}", client_key.decrypt_radix::<u8>(&r2));
    println!("q1: {:?}", client_key.decrypt_radix::<u8>(&q1));
    println!("q2: {:?}", client_key.decrypt_radix::<u8>(&q2));

    let r1r2 = server_key.smart_mul_parallelized(&mut r1, &mut r2);
    println!("r1r2: {:?}", client_key.decrypt_radix::<u8>(&r1r2));
    let q1q2 = server_key.smart_mul_parallelized(&mut q1, &mut q2);
    println!("q1q2: {:?}", client_key.decrypt_radix::<u8>(&q1q2));

    let r1_dec = client_key.decrypt_radix::<u8>(&r1);
    let r2_dec = client_key.decrypt_radix::<u8>(&r2);
[...]

PR #412, opened 2023-07-06
fix(shortint): Ciphertext::copy_from

tmontaigu:
### PR content/description

Ciphertext::copy_from did not copy the degree
resulting in potential bad results for some operations.

This fixes it, and rewrites to use destructuring
in order to prevent such thing from happenning again
(with destructuring, if a member is not destructured,
a compile error is emited)

Also we move the implementation of copy_from into
clone_from.

Fixes https://github.com/zama-ai/tfhe-rs/issues/410

### Check-list:

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -346,7 +346,7 @@ impl ServerKey {
                     )
                 });
             for i in space..num_blocks {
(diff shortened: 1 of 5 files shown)


IceTDrinker (commented):
Should we take the repro case from the bug report as a non regression test maybe ? Otherwise looks good

tmontaigu:
Added the example from issue as test

PR #414, opened 2023-07-07
chore(ci): select benches to run

IceTDrinker:


IceTDrinker:
not sure if this works on push, wdyt @soonum ?

I needed something similar to this to run only select benchmarks for multi bit, thought I might as well try to open a PR

IceTDrinker:
example run with this approach https://github.com/zama-ai/tfhe-rs/actions/runs/5485024318/jobs/9993301864

IceTDrinker:
alright looks like input is not given to the push event

soonum:
> alright looks like input is not given to the push event

Yes it makes sense, since `push` event and `workflow_dispatch` are two different beasts.
Maybe add another step that does the original call to Slab (i.e using the matrix build). This step would be run only if the event is `push`. Conversely the one you've modified would only be run in case of `workflow_dispatch`.

IceTDrinker:
test run : https://github.com/zama-ai/tfhe-rs/actions/runs/5486276852/jobs/9996168151 with only boolean

tmontaigu (approved):

PR #444, opened 2023-07-17
Scalar div u256

tmontaigu:
requires 
- [ ] #442


tfhe/src/core_crypto/commons/numeric/unsigned.rs
```diff
@@ -28,8 +28,12 @@ pub trait UnsignedInteger:
     + Not<Output = Self>
     + Shl<usize, Output = Self>
     + ShlAssign<usize>
+    + Shl<u32, Output = Self>
+    + ShlAssign<u32>
     + Shr<usize, Output = Self>
     + ShrAssign<usize>
+    + Shr<u32, Output = Self>
+    + ShrAssign<u32>
```
IceTDrinker, line 36: looks to create annoying situations, any reason for the u32 additional requirement ?

tfhe/src/integer/server_key/mod.rs
```diff
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 
 /// Error returned when the carry buffer is full.
 pub use crate::shortint::CheckError;
+pub use radix_parallel::{MiniUnsignedInteger, Reciprocable};
```
IceTDrinker, line 17: we should we do the same for the other traits introduced recently ?

tfhe/src/integer/bigint/u512.rs
```diff
@@ -0,0 +1,362 @@
...
+    const BITS: usize = Self::BITS as usize;
+
+    const ZERO: Self = Self::ZERO;
+
+    const ONE: Self = Self::ONE;
+
+    const TWO: Self = Self::TWO;
+
+    const MAX: Self = Self::MAX;
+}
```
IceTDrinker, line 362: no tests for u512 inspired of u256 module ?

tfhe/src/integer/bigint/algorithms.rs
```diff
@@ -0,0 +1,240 @@
+use crate::core_crypto::prelude::{Numeric, UnsignedInteger};
+use std::ops::{BitAnd, BitOrAssign, Not, Shl, ShlAssign, Shr, SubAssign};
+
+const BYTES_PER_U64: usize = std::mem::size_of::<u64>() / std::mem::size_of::<u8>();
+
+pub(crate) fn copy_from_le_byte_slice(lhs: &mut [u64], bytes: &[u8]) {
+    assert_eq!(bytes.len(), BYTES_PER_U64 * 4);
```
IceTDrinker, line 7: in this form this is u256 specific, I would either prefer a form of const generic or keep this in the u256 file if it's not shared with u512

Same comment for all copy from slice, for the other algorithms you know better than me if they are shared or not between u256 and u512 :)

IceTDrinker (commented):
Looks good generally, woud like unit tests for u512 on the model of u256 to be rock solid on those primitives !

tfhe/src/integer/server_key/mod.rs
tmontaigu: We could

tfhe/src/integer/bigint/u512.rs
```diff
@@ -0,0 +1,362 @@
...
+        );
+        (self.0.len() as u32 * u64::BITS) - self.leading_zeros() - 1
+    }
+
+    pub fn ceil_ilog2(self) -> u32 {
+        self.ilog2() + u32::from(!self.is_power_of_two())
+    }
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
tmontaigu: We are forced to have this cfg here as rand is a dev-dependency, so only available in tests

tfhe/src/integer/bigint/algorithms.rs
tmontaigu: hum, the assert is some left over from moving the code, this thing will work for u512 we just have to change the assert to `assert_eq(lhs.len(), bytes.len())`

IceTDrinker: indeed

tfhe/src/integer/bigint/u512.rs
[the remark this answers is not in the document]
IceTDrinker: ah right too bad

tfhe/src/integer/bigint/u512.rs
tmontaigu: I added some tests for u512

tfhe/src/integer/bigint/algorithms.rs
tmontaigu: fixed + test added for u512

IceTDrinker:
stupid question are we sure we run the unit tests in shortint/integer and others ? given our filters in the shortint and integer scripts ?

tmontaigu:
It seems that yes https://github.com/zama-ai/tfhe-rs/actions/runs/5585239221/jobs/10207707992#step:8:1282

IceTDrinker:
> It seems that yes https://github.com/zama-ai/tfhe-rs/actions/runs/5585239221/jobs/10207707992#step:8:1282

alright good for us, a bit surprised it gets tested but I'm not going to complain

tfhe/src/integer/bigint/u512.rs
```diff
@@ -0,0 +1,729 @@
...
+            0,
+            0,
+        ])
+    }
+}
+
+impl From<(u128, u128, u128, u128)> for U512 {
+    fn from((value0, value1, value2, value3): (u128, u128, u128, u128)) -> Self {
+        Self([
+            (value0 & u128::from(u64::MAX)) as u64,
```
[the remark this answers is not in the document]
tmontaigu: hum yeah it should work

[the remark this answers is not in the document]
IceTDrinker: you do the same for the value you shift down 🙂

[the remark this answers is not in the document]
tmontaigu: Its written like to to be explicit i think:

`(value0 & u128::from(u64::MAX)) as u64` -> We first mask with u64 max to I get the 64 lower bits on a u128 so casting to u64 does not truncate anything,  `value0 as u64` should give the same thing but might be less explicit ?

`(value0 >> 64) as u64` -> we shift right 64 bits, so the u128 only contains 64 bits, that previously were the upper ones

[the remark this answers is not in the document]
IceTDrinker: it's not a blocker

IceTDrinker (approved):
Looks good, thanks !

PR #450, opened 2023-07-20
fix(integer): remove incorrect bounds

tmontaigu:
### PR content/description

In 35c6aea84b7ee89e0ded748fe4f822d9e3a10a13 the bounds for the scalar_div family of functions were changed.

However, the a few bounds `u64: From<T>` were
not removed meaning the functions which still
had these were still stuck with u64 as the max scalar value.


IceTDrinker (approved):

tmontaigu:
@slab-ci cpu_fast_test

PR #451, opened 2023-07-20
feat(integer): allow scalar shift/rotate with more unsigned types

tmontaigu:
### PR content/description

This is mainly for convenience.

Also, rust implements shift by u8, u16..u128 for each types.
(even shift by i8...i128 are implemented).

### DIFF
--- a/tfhe/src/integer/server_key/radix/shift.rs
+++ b/tfhe/src/integer/server_key/radix/shift.rs
@@ -1,4 +1,7 @@
+use std::ops::Rem;
+
 use crate::core_crypto::commons::utils::izip;
+use crate::core_crypto::prelude::CastFrom;
 use crate::integer::ciphertext::RadixCiphertext;
 use crate::integer::ServerKey;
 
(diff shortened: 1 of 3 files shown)


tfhe/src/integer/server_key/radix/shift.rs
```diff
@@ -76,11 +77,10 @@ impl ServerKey {
...
     /// assert_eq!(msg >> shift, dec);
     /// ```
-    pub fn unchecked_scalar_right_shift(
-        &self,
-        ct: &RadixCiphertext,
-        shift: u64,
-    ) -> RadixCiphertext {
+    pub fn unchecked_scalar_right_shift<T>(&self, ct: &RadixCiphertext, shift: T) -> RadixCiphertext
+    where
+        u64: CastFrom<T>,
```
IceTDrinker, line 82: this works even for u128 because the high bits would be multiples of the bit sizes ?

IceTDrinker: I’m going to be annoying and say that if the objective is to match rust then we need to apply the modulo operator before the u64 conversion 🙂 the bound should not be too bad I believe ?

IceTDrinker: though rust does not have weird non power of two N uN types

IceTDrinker: still feels like we can be consistent without it being too costly so I'd still go for the additional bound I think ?

IceTDrinker (approved):
Thanks a lot !

tfhe/src/integer/server_key/radix/shift.rs
tmontaigu: I changed the % to be before the casting

IceTDrinker: yep thanks a lot !

PR #452, opened 2023-07-21
feat(hlapi): allow scalar ops on values up to U256 

tmontaigu:
### PR content/description

This enables to use u128 and U256 as operands to
operations in the high level api.

BREAKING CHANGE: a breaking change in the C API for scalar operations
for FheUint128 and FheUint256 as they previously required
a u64 and now a U218 / U256 respectively.

### Check-list:

### DIFF
--- a/tfhe/src/c_api/high_level_api/utils.rs
+++ b/tfhe/src/c_api/high_level_api/utils.rs
@@ -274,6 +274,53 @@ macro_rules! impl_binary_assign_fn_on_type {
     };
 }
 
+/// The C Standard only define integers from u8 to u64.
+/// So to support u128 and u256 we had to create our own C friendly
+/// data types.
+///
+/// This trait exists to be able to easily write wrapper function
+/// by allowing to generically go from a C API scalar type to a rust one
+pub(in crate::c_api::high_level_api) trait ToRustScalarType {
+    type RustScalarType;
+
+    fn to_rust_scalar_type(self) -> Self::RustScalarType;
+}
+
+/// Implements the trait for when the C API type is the same
+/// as the Rust API type (eg u64)
+macro_rules! impl_to_rust_scalar_type(
+    ($type:ty) => {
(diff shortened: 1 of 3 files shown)


tfhe/src/c_api/high_level_api/utils.rs
```diff
@@ -274,6 +274,53 @@ macro_rules! impl_binary_assign_fn_on_type {
...
+    }
+}
+
+impl ToRustScalarType for crate::c_api::high_level_api::u256::U256 {
+    type RustScalarType = crate::integer::U256;
+
+    fn to_rust_scalar_type(self) -> Self::RustScalarType {
+        crate::integer::U256::from(self)
+    }
+}
```
IceTDrinker, line 322: I think I'm missing something here, how does the conversion happen ?

tmontaigu: which one ? 

In the macro we do
```
let rhs = <$scalar_type as $crate::c_api::high_level_api::utils::ToRustScalarType>::to_rust_scalar_type(rhs);
```

IceTDrinker: I mean there must be a From implementation between the C-friendly type and the rust type ?

tmontaigu: Yes they are in the C API module

- https://github.com/zama-ai/tfhe-rs/blob/6a56af0b074bfe479221857968ed769a65509a8b/tfhe/src/c_api/high_level_api/u256.rs#L24
- https://github.com/zama-ai/tfhe-rs/blob/main/tfhe/src/c_api/high_level_api/u128.rs

IceTDrinker: thanks !

IceTDrinker:
apart for the two typos you noted I think it's ok !

PR #457, opened 2023-07-24
chore(ci): trigger benchmarks only if layers have changed

soonum:
For example, if only shortint layer related files have changed, only the shortint benchmarks would be run on push. However, if any files changed in the common_benches group then all the benchmarks would be run.

### DIFF
--- a/.github/workflows/start_benchmarks.yml
+++ b/.github/workflows/start_benchmarks.yml
@@ -51,16 +51,38 @@ jobs:
         uses: tj-actions/changed-files@de0eba32790fb9bf87471b32855a30fc8f9d5fc6
         with:
           files_yaml: |
-            benches:
+            common_benches:
               - toolchain.txt
               - Makefile
               - ci/slab.toml
               - tfhe/Cargo.toml
-              - tfhe/src/**
-              - tfhe/benches/**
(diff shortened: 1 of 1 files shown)


.github/workflows/start_benchmarks.yml
```diff
@@ -51,16 +51,39 @@ jobs:
...
+            pbs_bench:
+              - tfhe/src/core_crypto/**
+              - tfhe/benches/core_crypto/**
+              - .github/workflows/pbs_benchmark.yml
+            wasm_client_bench:
+              - tfhe/web_wasm_parallel_tests/**
+              - .github/workflows/wasm_client_benchmark.yml
+
+      - name: Process changed files
+        run: |
```
IceTDrinker, line 86: this can be removed

IceTDrinker (commented):
Just a thing you forgot to remove I think, otherwise should be good hopefully, did you get a chance to check this works ?

soonum:
> Just a thing you forgot to remove I think, otherwise should be good hopefully, did you get a chance to check this works ?

Yes I've tested it with WIP commit in this branch by adding it to the branches triggered on `push`event. It work like a charm :+1:

IceTDrinker (approved):

PR #465, opened 2023-07-26
chore(ci): create full benchmarks workflow that runs weekly

soonum:


IceTDrinker (commented):
Main question is why not re-use the existing benchmark files ?

I think the only benchmark needing an adaptation or a dedicated workflow are shortint (to make things simpler) and integer + integer multi bit ?

IceTDrinker:
@slab-ci cpu_fast_test

soonum:
> Main question is why not re-use the existing benchmark files ?
> 
> I think the only benchmark needing an adaptation or a dedicated workflow are shortint (to make things simpler) and integer + integer multi bit ?

Yes you're right but it shortint and integer adaptation would require careful thinking to be able to handle both case (fast bench VS full bench)

IceTDrinker:
> > Main question is why not re-use the existing benchmark files ?
> > I think the only benchmark needing an adaptation or a dedicated workflow are shortint (to make things simpler) and integer + integer multi bit ?
> 
> Yes you're right but it shortint and integer adaptation would require careful thinking to be able to handle both case (fast bench VS full bench)

worst case we can have a dedicated full bench one if it's too annoying to adapt, separate workflows have the advantage to parallelize the benchmarks (even if they run during the week-end)

soonum:
Here what I propose, keeping the new file `start_full_benchmark.yml`. Inside it we'll call the exisiting short benchmarks (boolean, pbs and wasm). In addition to that, I'd break the `full_benchmark.yml` into two separate files `full_shortint_benchmarks.yml` and `full_integer_benchmarks.yml`.
That way we could benefit from parallelism, with not too much duplication. Moreover, handling matrix benchmarks into existing workflows for shortint and integer would be a nightmare.

What do you think about that design?

IceTDrinker:
@soonum perfect 🙂 that's what I had in mind as well !

IceTDrinker:
one last detail it feels like it would be _super_ expensive to run, maybe let's not enable the schedule right now until we have one full run to assess the runtime and cost wdyt ?

.github/workflows/integer_full_benchmark.yml
```diff
@@ -0,0 +1,133 @@
...
+  integer-benchmarks:
+    name: Execute integer benchmarks for all operations flavor
+    runs-on: ${{ github.event.inputs.runner_name }}
+    if: ${{ !cancelled() }}
+    needs: prepare-benchmarks
+    strategy:
+      max-parallel: 1
+      matrix:
+        command: [ integer, integer_multi_bit]
+        op_flavor: [ default, default_scalar, smart, smart_scalar, smart_parallelized, smart_scalar_parallelized, unchecked, unchecked_scalar, misc ]
```
IceTDrinker, line 79: 18 combinations in total is that right ? single bench machine with up to 6 hours per bench

108 hours of bench ? 😱

IceTDrinker (approved):
Looks good but I'm worried for bench timings

.github/workflows/integer_full_benchmark.yml
soonum: Yes I know but we'll see what is the actual number.

PR #468, opened 2023-07-27
chore(ci): add a command to just build docs to see how it looks like

IceTDrinker:


soonum (approved):
This is done to warn us about dead links in the doc right? So it's not just "to see how it looks like", isn't it ?

IceTDrinker:
> This is done to warn us about dead links in the doc right? So it's not just "to see how it looks like", isn't it ?

Previously you could not build the doc if it was in a semi broken state so if you were in a hurry to see if the docs looked good the compilation would just fail because of the deny, with the warn we have the warning but the doc still builds, I added the lint target to be sure we are still rejecting commits where the doc fails to build

PR #471, opened 2023-08-01
refactor(keycache): refactor the named_params_impl macro

IceTDrinker:
- allows to easily create named parameters from non standard shortint parameter structs
- allows to reate such named parameters easily outside of the keycache mod the way the macro was written before expanded to more macro calls that may not have been in scope and rendered the macro virtually useless


IceTDrinker:
@slab-ci cpu_fast_test

IceTDrinker:
@slab-ci cpu_test

PR #473, opened 2023-08-01
feat(integer): make full_propagate_parallelized more parallel

tmontaigu:
Using the functions that were introduced recently, it is possible to make the full_propagate_parallelized method more parallel than it was, resulting in faster computations.

The new carry propapagation should now be the cost of a default add + one PBS, so ~400ms in 256 instead of ~3s.

However its probably slower for smaller number of blocks (eg 4 blocks)

This is done by extracting carry and messages in parallel, then adding the carries to the correct message, the final step is to use the single carry propapagation function.


tmontaigu:
@slab-ci cpu_integer_test

tmontaigu:
@slab-ci cpu_integer_test

jborfila (approved):

PR #488, opened 2023-08-11
chore(ci): ignore tfhe_lib_parameters check in lattice esitmator

soonum:
This third-party parameters set is known to be insecure, so there is no need to make the CI fail because of this set.

### DIFF
--- a/ci/lattice_estimator.sage
+++ b/ci/lattice_estimator.sage
@@ -27,6 +27,10 @@ def check_security(filename):
     to_update = []
 
     for param in all_params:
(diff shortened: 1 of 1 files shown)


tmontaigu (approved):

ci/lattice_estimator.sage
```diff
@@ -27,6 +27,10 @@ def check_security(filename):
     to_update = []
 
     for param in all_params:
+        if param.tag.startswith("TFHE_LIB_PARAMETERS"):
+            # This third-party parameters set is known to be insecure, just skip the analysis.
```
tmontaigu, line 31: "Less secure" ?

PR #470, opened 2023-07-31
chore(ci): update to account for new lints

IceTDrinker:
closes: _please link all relevant issues_

### Check-list:

* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Relevant issues are marked as resolved/closed, related issues are linked in the description
* [ ] Check for breaking changes (including serialization changes) and add them to commit message following the conventional commit [specification][conventional-breaking]

[conventional-breaking]: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer

PR #494, opened 2023-08-22
Update toolchain

mayeul-zama:
### PR content/description

Update rust toolchain


tfhe/src/shortint/engine/server_side/add.rs
```diff
@@ -27,7 +27,7 @@ impl ShortintEngine {
     pub(crate) fn smart_add(
         &mut self,
         server_key: &ServerKey,
-        ct_left: &mut Ciphertext,
```
tmontaigu, line 30: Normally all the smart functions should take &mut for both input to clean carries if needed,
but these one do not seems to be correctly written

apps/trivium/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+#![allow(clippy::module_inception)]
+#![allow(clippy::should_implement_trait)]
```
tmontaigu, line 2: what method was it complaining about ? 'next' ?

tmontaigu (approved):

tfhe/src/shortint/engine/server_side/add.rs
mayeul-zama: Ok
Do you think we should fix this code by cleaning carries:
- in this PR (not sure how big this would be)?
- in another fix PR and in the meantime:
  - keep `mut` and silence this warning for the smart functions?
  - remove `mut` (as in this PR) and reintroduce it in the fix PR?

If we only clean carries of outputs, is it not enough (as all inputs will then be clean)? Or maybe we want to do it only when really needed to avoid it in some cases

tmontaigu: smart_ functions are meant to clean carries of input only when needed

I would say we keep the changes in the PR and try to find some time later to fix the shortint smart functions that
needs to be fixed

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

tmontaigu:
Well I saw that Arthur did the same thing as you but  also corrected the smart function in this PR: https://github.com/zama-ai/tfhe-rs/pull/470 maybe we should just update Arthur's PR

IceTDrinker:
if this is working we could probably port the smart changes from my PR to this one I remember mine not passing tests IIRC
last commit is broken however in this PR

apps/trivium/src/lib.rs
IceTDrinker, line 2: why is this needed ? 🤔

tfhe/src/c_api/shortint/server_key/scalar_add.rs
```diff
@@ -6,17 +6,17 @@ use super::{ShortintCiphertext, ShortintServerKey};
 #[no_mangle]
 pub unsafe extern "C" fn shortint_server_key_smart_scalar_add(
     server_key: *const ShortintServerKey,
-    ct_left: *mut ShortintCiphertext,
+    ct_left: *const ShortintCiphertext,
```
IceTDrinker, line 9: this should still be mut right ? as discussed for the smart stuff

code below should probably be updated as well

tfhe/src/c_api/shortint/server_key/scalar_mul.rs
```diff
@@ -6,17 +6,17 @@ use super::{ShortintCiphertext, ShortintServerKey};
 #[no_mangle]
 pub unsafe extern "C" fn shortint_server_key_smart_scalar_mul(
     server_key: *const ShortintServerKey,
-    ct_left: *mut ShortintCiphertext,
+    ct_left: *const ShortintCiphertext,
```
IceTDrinker, line 9: idem

tfhe/src/c_api/shortint/server_key/scalar_sub.rs
```diff
@@ -6,17 +6,17 @@ use super::{ShortintCiphertext, ShortintServerKey};
 #[no_mangle]
 pub unsafe extern "C" fn shortint_server_key_smart_scalar_sub(
     server_key: *const ShortintServerKey,
-    ct_left: *mut ShortintCiphertext,
+    ct_left: *const ShortintCiphertext,
```
IceTDrinker, line 9: idem

tfhe/src/c_api/shortint/server_key/shift.rs
```diff
@@ -6,17 +6,17 @@ use super::{ShortintCiphertext, ShortintServerKey};
 #[no_mangle]
 pub unsafe extern "C" fn shortint_server_key_smart_scalar_left_shift(
     server_key: *const ShortintServerKey,
-    ct: *mut ShortintCiphertext,
+    ct: *const ShortintCiphertext,
```
IceTDrinker, line 9: idem

tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::redundant_locals)]
```
IceTDrinker, line 1: why is this needed ?

tfhe/src/js_on_wasm_api/mod.rs
```diff
@@ -1,15 +1,15 @@
...
 #[cfg(feature = "boolean-client-js-wasm-api")]
 pub use boolean::*;
 
 #[cfg(feature = "parallel-wasm-api")]
 pub use wasm_bindgen_rayon::init_thread_pool;
 
-pub mod high_level_api;
-pub use high_level_api::*;
+mod js_high_level_api;
+pub use js_high_level_api::*;
```
IceTDrinker, line 15: how is this legal for the mod and pub use ?

tfhe/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub mod shortint;
 
 #[cfg(feature = "__wasm_api")]
 /// cbindgen:ignore
-pub mod js_on_wasm_api;
+mod js_on_wasm_api;
 #[cfg(feature = "__wasm_api")]
 pub use js_on_wasm_api::*;
```
IceTDrinker, line 45: same question

IceTDrinker (commented):
First batch of comments looks like some smart ops are still taking some inputs by non mutable reference

tfhe/src/c_api/shortint/server_key/scalar_add.rs
IceTDrinker: the first version you describe sounds like a default operation, i.e. output is always clean
the second is a smart operation as far as I can tell

tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw.rs
mayeul-zama: Otherwise, the line `let output_fft_buffer_re0 = &mut *output_fft_buffer_re0;` and the 3 next ones are generating a warning whihc I believe is a false positive

IceTDrinker: removing the &mut is not enough ? looks like deref returns a mutable ref to a slice

IceTDrinker: basically I'm all against relaxing lints at file level, hence why I'm aksing this :)

it's an easy way to over disable stuff

IceTDrinker: if this is really a false positive then I would prefer a function level disable or even a block disable as these rust attributes can be put in a lot of places

tfhe/src/js_on_wasm_api/mod.rs
mayeul-zama: If a private item is reexported publicly (with a pub use), it can be accessed (only) under the new name (https://doc.rust-lang.org/reference/visibility-and-privacy.html#re-exporting-and-visibility)

I removed the `pub` in front of `mod` to have a simpler public interface

IceTDrinker: all good then !

tfhe/src/c_api/shortint/server_key/scalar_add.rs
mayeul-zama: I see what you mean
But it seems to me that the first version can give better performance than the second one if the input is never reused
So with this definition of smart operation, using smart operations can be more costly than using default ones (which is not intuitive)

I think the second version is more consistent with the rest of the codebase so I'll use it

IceTDrinker: indeed, though smart ops have this approach of cleaning inputs to hopefully gain time on later ops if re-used

that's why default ops have been made default, easier to reason about and not that bad performance characteristics ! 🙂

tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw.rs
mayeul-zama: > removing the &mut is not enough ? looks like deref returns a mutable ref to a slice
No, apparently it returns a plain `[f64]`


> if this is really a false positive then I would prefer a function level disable or even a block disable as these rust attributes can be put in a lot of places

Sure, I had the same idea after seeing your comment :)

IceTDrinker: could the lines be removed maybe ? maybe the compiler can deref things on its own

apps/trivium/src/lib.rs
IceTDrinker: which next function is impacted ?

tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw.rs
mayeul-zama: This does not work :(
I think that's the first thing I tried

mayeul-zama: Removing the variables and adding `&mut` where they were used works (the `*` is not even needed)

IceTDrinker: perfect

apps/trivium/src/lib.rs
mayeul-zama: apps/trivium/src/kreyvium/kreyvium_shortint.rs, line 78
apps/trivium/src/kreyvium/kreyvium.rs, line 152
apps/trivium/src/trivium/trivium_shortint.rs, line 66
apps/trivium/src/trivium/trivium.rs, line 124

I'll put the lint on each function

IceTDrinker: otherwise rename the function to "next_byte" or whatever name fits the returned value ?

or implement iterator for those ? I'm really not sure as I don't really know the trivium/kreyvium apps

mayeul-zama: Renamed `next_ct` and `next_bool`

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -603,9 +603,29 @@ impl ShortintEngine {
     where
         F: Fn(u64, u64) -> u64,
     {
+        // TODO this looks wrong for the Accumulator generation
```
mayeul-zama, line 606: Should his be fixed now (this commit comes from @IceTDrinker PR) ?

IceTDrinker: it should have been fixed by Thomas IIRC check the main branch

mayeul-zama: In commit a0946ac509a481614d58ea6c9dc3fad70b54fc64?

IceTDrinker: think so yes

IceTDrinker: if you are confident with the current version launch all the test suites and we'll redo a thorough review of the new code

mayeul-zama: All the tests (except CPU AWS Fast Tests) ran fine just before adding the last commit so I don't think rerunning them is necessary

IceTDrinker: disagree

IceTDrinker: they will re run anyways once we review 😄

apps/trivium/src/kreyvium/mod.rs
```diff
@@ -1,3 +1,4 @@
+#[allow(clippy::module_inception)]
```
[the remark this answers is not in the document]
IceTDrinker: ah yes my bad it's due to the module structure, https://github.com/zama-ai/tfhe-rs-internal/issues/216

tfhe/src/shortint/engine/server_side/scalar_bitwise_op.rs
```diff
@@ -30,7 +30,7 @@ impl ShortintEngine {
     pub(crate) fn smart_scalar_bitand(
         &mut self,
         server_key: &ServerKey,
-        lhs: &mut Ciphertext,
+        lhs: &Ciphertext,
```
IceTDrinker, line 33: needs to be mutable

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: I will remove it

IceTDrinker: is the change you made still required ?

tfhe/src/shortint/engine/server_side/scalar_bitwise_op.rs
mayeul-zama: It seems to me that this operations never modifies the input and always applies a LUT
Which condition do you suggest we modify this input under?

IceTDrinker: same for the other ops of the module

IceTDrinker: ah this is tricky should we clean the inputs if we are just applying a lut

IceTDrinker: as we never need to clean it in that case, no should be fine I guess in this case my bad

tfhe/src/shortint/engine/server_side/shift.rs
```diff
@@ -53,6 +53,9 @@ impl ShortintEngine {
         ct: &mut Ciphertext,
         shift: u8,
     ) -> EngineResult<Ciphertext> {
+        if !server_key.is_scalar_left_shift_possible(ct, shift) {
+            self.message_extract_assign(server_key, ct)?;
+        }
```
IceTDrinker, line 58: below it can be a unchecked call ?

tfhe/src/shortint/engine/server_side/scalar_sub.rs
```diff
@@ -37,6 +37,9 @@ impl ShortintEngine {
         ct: &mut Ciphertext,
         scalar: u8,
     ) -> EngineResult<Ciphertext> {
+        if !server_key.is_scalar_sub_possible(ct, scalar) {
+            self.message_extract_assign(server_key, ct)?;
+        }
         let mut ct_result = ct.clone();
         self.smart_scalar_sub_assign(server_key, &mut ct_result, scalar)?;
```
IceTDrinker, line 44: unchecked call ?

tfhe/src/shortint/engine/server_side/scalar_mul.rs
```diff
@@ -47,6 +47,9 @@ impl ShortintEngine {
         ctxt: &mut Ciphertext,
         scalar: u8,
     ) -> EngineResult<Ciphertext> {
+        if !server_key.is_scalar_mul_possible(ctxt, scalar) {
+            self.message_extract_assign(server_key, ctxt)?;
+        }
         let mut ct_result = ctxt.clone();
         self.smart_scalar_mul_assign(server_key, &mut ct_result, scalar)?;
```
IceTDrinker, line 54: unchecked call ? not sure if the function does other funky stuff

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: I took it from your PR
But I think so as it cleans the input if necessary instead of cleaning the result (cloned from the input)

tfhe/src/shortint/engine/server_side/scalar_mul.rs
mayeul-zama: Even with a clean ct, the multiplication might not pass, is that right?

Anyway, the cost of calling smart instead of unchecked should be negligible

tfhe/src/shortint/engine/server_side/scalar_sub.rs
mayeul-zama: If we always have at least one bit of carry, I guess this is safe.
Is it always the case?

tfhe/src/shortint/engine/server_side/scalar_mul.rs
IceTDrinker: fair enough

tfhe/src/shortint/engine/server_side/scalar_sub.rs
IceTDrinker: nope good point

tfhe/src/shortint/engine/server_side/mod.rs
IceTDrinker: my PR predates the fix from Thomas, i.e. the fix from Thomas should be the one to keep

tfhe/src/shortint/engine/server_side/shift.rs
mayeul-zama: I believe even with a clean ct, the shift might not pass

IceTDrinker: ok then

IceTDrinker:
> I think it would be better to split this PR in 2, to handle behavioral changes in a separate PR (2 commits `feat(shortint): cleanup input if necessary in **`) since I think of similar changes in other files WDYT?

fine by me yes

IceTDrinker (approved):
Thanks a lot @mayeul-zama for this tedious work 🙏

PR #495, opened 2023-08-22
chore(bench): add division and modulo ops to integer benchmarks

soonum:


tmontaigu:
This will increase bench time by a lot (2+ hours easily)

PR #496, opened 2023-08-23
chore(bench): bench integer scalar ops up to 256 bits

tmontaigu:


tfhe/benches/integer/bench.rs
```diff
@@ -810,9 +819,9 @@ define_server_key_bench_scalar_default_fn!(
     rng_func: default_scalar
 );
 define_server_key_bench_scalar_default_fn!(
-    method_name: unchecked_small_scalar_mul,
+    method_name: unchecked_scalar_mul_parallelized,
     display_name: mul,
-    rng_func: mul_scalar
+    rng_func: default_scalar
```
soonum, line 824: Shouldn't we also add `unchecked_scalar_add` and unchecked_scalar_sub` with this new  random generator ?

soonum (changes requested):
Nice job, see my comment about the two scalar operations left.

tfhe/benches/integer/bench.rs
tmontaigu: Well here I guess it sould be `rng_func: mul_scalar`but we could add unchecked_scalar_sub/add to the benches if they are not benched

soonum: They are already in the benchmarks. I'm just saying that they don't use the `rng_func` in their macro call. Look some lines above.

soonum: But you're totally right about using `rng_func: mul_scalar` for this multiplication.

tmontaigu: hum  I see that they are already using it 🤔

soonum: You're right my bad, I was checking the file on my laptop on a branch that is not up to date.
So just change `default_scalar` to `mul_scalar` for this multiplication and then we're good to go.

soonum (approved):

PR #511, opened 2023-08-29
Measure compressed key size & benchmark decompression

soonum:


tfhe/benches/shortint/bench.rs
```diff
@@ -304,6 +306,39 @@ fn programmable_bootstrapping(c: &mut Criterion) {
     bench_group.finish();
 }
 
+fn server_key_from_compressed_key(c: &mut Criterion) {
+    let mut bench_group = c.benchmark_group("uncompress_key");
+    bench_group
+        .sample_size(10)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    for param in SERVER_KEY_BENCH_PARAMS {
```
IceTDrinker, line 315: no multi bit ? don't remember how we manage benchmarks params

tfhe/benches/shortint/bench.rs
```diff
@@ -304,6 +306,39 @@ fn programmable_bootstrapping(c: &mut Criterion) {
...
+        let param: PBSParameters = param.into();
+        let keys = KEY_CACHE.get_from_param(param);
+        let sks_compressed = CompressedServerKey::new(&keys.client_key());
+
+        let bench_id = format!("shortint::uncompress_key::{}", param.name());
+
+        bench_group.bench_function(&bench_id, |b| {
+            b.iter(|| {
+                let _ = ServerKey::from(sks_compressed.clone());
```
IceTDrinker, line 324: the clone is going to give wrong results as the clone operation is going to appear in the global time of the benchmark
we need an iter with input preparation I believe, integer benchmarks have it IIRC or maybe it's also in that file

IceTDrinker (changes requested):
The benchmarks timings will be wrong because of the clone in the iter closure, we need to use the same patterns as for integer benchmarks where the input is prepared beforehand

tfhe/benches/shortint/bench.rs
soonum: Thing is we don't have any multi-bit parameters in shortint. Should I move this benchmark to integer layer then ?

IceTDrinker: There should be multi bits params in shortint, check parameters::multi_bit or similar

tfhe/benches/shortint/bench.rs
IceTDrinker: perfect

soonum:
@slab-ci shortint_bench

.github/workflows/shortint_benchmark.yml
```diff
@@ -77,6 +77,20 @@ jobs:
           --name-suffix avx512 \
           --throughput
 
+      - name: Run multi-bit benchmarks with AVX512
+        run: |
+          rm -rf target/criterion
+          make AVX512_SUPPORT=ON bench_shortint_multi_bit
```
IceTDrinker, line 83: I don't know if this is the right PR for this, as there were some slow downs/some runs took longer as we did not have enough cores

tfhe/benches/shortint/bench.rs
```diff
@@ -304,6 +340,47 @@ fn programmable_bootstrapping(c: &mut Criterion) {
...
+        let keys = KEY_CACHE.get_from_param(*param);
+        let sks_compressed = CompressedServerKey::new(&keys.client_key());
+
+        let bench_id = format!("shortint::uncompress_key::{}", param.name());
+
+        bench_group.bench_function(&bench_id, |b| {
+            let clone_compressed_key = || {
+                let sks_cloned = sks_compressed.clone();
+                sks_cloned
+            };
```
IceTDrinker, line 359: let clone_compressed_key = || sks_compressed.clone(); should be enough I think (clippy will likely yell at you)

.github/workflows/shortint_benchmark.yml
soonum: What do you mean? I've made several clear commits in this PR , you want me to move the last commit to a new PR ?

IceTDrinker (commented):
Thanks, my suggestion for the multi bit was mainly for the decompression of the keys, I'm not sure we want to bench all ops with all sizes for multi bit ? unless it was ready but I don't remember to be honest

.github/workflows/shortint_benchmark.yml
IceTDrinker: I think so, adding the decompression is easy, while there may be issues with the multi bit benchmarks ? Ah but you added them only to shortint, well if the timings are not crazy on shortint then should be fine !

tfhe/benches/shortint/bench.rs
```diff
@@ -333,309 +410,333 @@ fn _bench_wopbs_param_message_8_norm2_5(c: &mut Criterion) {
...
+define_custom_bench_fn!(function_name: carry_extract, BenchParamsSet::Standard);
+
+define_custom_bench_fn!(
+    function_name: programmable_bootstrapping,
+    BenchParamsSet::Standard
+);
+
+define_custom_bench_fn!(
+    function_name: server_key_from_compressed_key,
+    BenchParamsSet::Standard
```
IceTDrinker, line 739: do we want the extended set for this maybe ?

tfhe/benches/shortint/bench.rs
soonum: Ok but how I get my hand on the value? `clone_compressed_key` is a closure not the value returned from the cloning.

tfhe/benches/shortint/bench.rs
soonum: depends on how long the bench would be.

tfhe/benches/shortint/bench.rs
IceTDrinker: clone_compressed_key()

one liner closures have the right to omit curly braces

soonum:
> Thanks, my suggestion for the multi bit was mainly for the decompression of the keys, I'm not sure we want to bench all ops with all sizes for multi bit ? unless it was ready but I don't remember to be honest

Maybe we could do it only for uncompress server key bench and leave the other.

soonum:
@slab-ci shortint_bench

soonum:
@slab-ci shortint_bench

tfhe/benches/shortint/bench.rs
soonum: Extended in both classic and multi-bit PBS is now implemented.

.github/workflows/shortint_benchmark.yml
soonum: I remove the multi-bit shortint in the benchmark for now.

IceTDrinker (approved):
Thanks a lot !

PR #513, opened 2023-08-29
fix(integer): add assert on scalar bits in scalar_div

tmontaigu:
### PR content/description

To work the scalar div requires that the scalar type (u32, u64, etc) has at least as many bits of precision that what the ciphertext encrypts (e.g. if the ciphertext encrypt 32bits, the scalar type must have at least 32 bits)

We were missing the assert which lead to misuse, so we add it.

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/scalar_div_mod.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/scalar_div_mod.rs
@@ -220,6 +220,15 @@ impl ServerKey {
         // Rust has a check on all division, so we shall also have one
         assert_ne!(divisor, T::ZERO, "Cannot divide by zero");
 
+        assert!(
+            T::BITS >= numerator_bits as usize,
+            "The scalar divisor type must have a number of bits that is \
+            >= to the number of bits encrypted in the ciphertext: \n\
+            encrypted bits: {numerator_bits}, scalar bits: {}
+            ",
+            T::BITS
+        );
+
         if divisor.is_power_of_two() {
             // Even in FHE, shifting is faster than multiplying / dividing
             return self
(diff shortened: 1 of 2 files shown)


tfhe/src/integer/server_key/radix_parallel/tests_unsigned.rs
```diff
@@ -2135,7 +2135,7 @@ where
     // hard-coded tests
     // 10, 7, 14 are from the paper and should trigger different branches
     // 16 is a power of two and should trigger the corresponding branch
-    let hard_coded_divisors: [u32; 4] = [10, 7, 14, 16];
+    let hard_coded_divisors: [u64; 4] = [10, 7, 14, 16];
```
IceTDrinker, line 2138: can you comment on this change ?

tfhe/src/integer/server_key/radix_parallel/scalar_div_mod.rs
```diff
@@ -220,6 +220,12 @@ impl ServerKey {
         // Rust has a check on all division, so we shall also have one
         assert_ne!(divisor, T::ZERO, "Cannot divide by zero");
 
+        assert!(
+            T::BITS >= numerator_bits as usize,
+            "The scalar divisor type must have a number of bits that is \
+            >= to the number of bits encrypted in the ciphertext"
```
IceTDrinker, line 226: can we maybe give the current number of bits encrypted in the ciphertext and the bits we got for the scalar ?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned.rs
tmontaigu: We encrrypt 32 bits integers, however for parameters 3_3 we actually encrypt more than 32 bits, which will trigger the new assert so we have to use a u64 scalar

IceTDrinker: I'm fine with the change anyways as I saw we were casting to u64

IceTDrinker (approved):

PR #519, opened 2023-08-31
chore(ci): add multiplier to get effective pbs throughput value

soonum:


ci/benchmark_parser.py
```diff
@@ -108,12 +108,17 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
                     )
                 )
 
+                # This is a special case where PBS are blasted as vector of 512 LWE ciphertext
+                # to saturate the machine. To get the actual throughput we need to multiply by the
+                # length of the vector.
+                multiplier = 512 if "PBS_throughput" in test_name else 1
+
```
IceTDrinker, line 115: some benchmarks have different chunk sizes though ?

soonum: This is the only one we've set up to saturate the machine.

IceTDrinker: What I mean is that we have other throughput stuff, and this looks like a hack more than actual solution to compute throughput for all cases we are interested in

soonum: Indeed, this is clearly a hack. That said, I haven't found a way at criterion level to pre-process the value.

IceTDrinker: Can we detect « chunk » in the bench name ? I think it’s the only benchmarks with chunk in the name and then parse the chunk size from there

soonum:
@slab-ci pbs_bench

ci/benchmark_parser.py
```diff
@@ -108,12 +108,26 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
                     )
                 )
 
+                # This is a special case where PBS are blasted as vector LWE ciphertext with
+                # variable length to saturate the machine. To get the actual throughput we need to
+                # multiply by the length of the vector.
+                if "PBS_throughput" in test_name and ("chunk") in test_name:
```
IceTDrinker, line 114: ("chunk") -> "chunk"

ci/benchmark_parser.py
```diff
@@ -108,12 +108,26 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
...
+                    try:
+                        multiplier = int(test_name.split("chunk")[0].split("_")[-1])
+                        print("Name:", test_name, "Multi:", multiplier)
```
IceTDrinker, line 117: remove the print which should not be needed

ci/benchmark_parser.py
```diff
@@ -108,12 +108,26 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
...
+                    except ValueError:
+                        parsing_failures.append((full_name,
+                                                 f"failed to extract throughput multiplier"))
```
IceTDrinker, line 120: f string without {}

IceTDrinker (approved):
Looks good thanks !

PR #521, opened 2023-08-31
feat(integer): be smarter about parallel carry propagation usage

tmontaigu:
The parallel carry propagation algorithm is a lot faster than the non parallel one, however to actually be faster it requires the CPU to have enough threads.

This adds checks so that we use the parallel algorithm only if this condition is met.

This should improve performance on smaller CPUs
typically found in standard laptops/desktops.

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -191,7 +191,7 @@ impl ServerKey {
             }
         };
 
-        if self.is_eligible_for_parallel_carryless_add() {
+        if self.is_eligible_for_parallel_single_carry_propagation(lhs) {
             self.unchecked_add_assign_parallelized_low_latency(lhs, rhs);
         } else {
             self.unchecked_add_assign(lhs, rhs);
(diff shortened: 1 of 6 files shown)


tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -241,12 +241,34 @@ impl ServerKey {
...
+        let has_enough_blocks = ct.blocks().len() >= MIN_NUM_BLOCKS;
+        if !has_enough_blocks {
+            return false;
+        }
+
+        // Use rayon to get that number as the implementation uses rayon for parallelism
+        let has_enough_threads = rayon::current_num_threads() >= ct.blocks().len();
+        if !has_enough_threads {
+            return false;
+        }
```
IceTDrinker, line 269: do we use a single thread per block in the implementation ?

tmontaigu: The parallel carry propagation is done in log2(num_block) steps, where the first step start wth num_block threads
then each following iteration divides the requires number of blocks by 2.

On my laptop, this checks makes me use the non parallel version of the algorithms for all bit sizes because I do not have enough threads, and this makes the operations (add/sub/neg) between 25% and 45% faster

IceTDrinker: Ok ! I did not recall how many threads were required for each step of the algo, thanks

IceTDrinker (approved):
Thanks a lot !

tmontaigu:
should improve the situation described in https://community.zama.ai/t/tfhe-rs-v0-2-5-vs-v0-3-1-release-question/731/8
there's still the mul the maybe potentially be slower if the cpu does not have enough threads, but it requires benches and probably more important changes

PR #524, opened 2023-09-07
chore(hlapi): move integrations tests into src dir

tmontaigu:
Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/220


tfhe/src/high_level_api/shortints/tests.rs
```diff
@@ -1,8 +1,9 @@
+#![allow(clippy::assign_op_pattern)]
```
IceTDrinker, line 1: what is this for ?

tmontaigu: clippy conplains about doing `let c = a + b`, it would like us to do `a+=b;`

IceTDrinker: thanks

IceTDrinker (approved):
Thanks !

PR #525, opened 2023-09-07
chore(tfhe): remove unused deps, drop once_cell, enable paste when needed

IceTDrinker:


tmontaigu (approved):

IceTDrinker:
arf OnceLock is available from 1.70 onwards, we would need to update the min rust version I believe

IceTDrinker:
yeah we were at 1.67

IceTDrinker:
thoughts on this for upping the min version to 1.70 ?

tmontaigu:
Seems ok to me, to end-users this upgrade will happen in the release of 0.4.0.

I think we can afford requiring recent version of the compiler, don't think its as restrictive as requiring nightly for example

IceTDrinker:
I was wondering as some people wanting to compute ZK pocs or something had issues as their tool supported older versions of rust exclusively

IceTDrinker:
> > I was wondering as some people wanting to compute ZK pocs or something had issues as their tool supported older versions of rust exclusively
> 
> I see, but without the numver of actual maximum version they were able to use, we can't guess 🤔 I did not find any issue in the public repo, was it reported somewhere else ?

It was an obscure repo spotted by the BC team IIRC, do super well known package enforce a particular policy wrt compiler version support ?

tmontaigu:
> do super well known package enforce a particular policy wrt compiler version support ?



No idea about that, I guess some will constrain into supporting older rust version themselves more than others, maybe its a thing more relevant for major versions of crates where more stability is expected

PR #528, opened 2023-09-08
feat(core): add par_fill_forward_fourier on bsk

mayeul-zama:
Add parallel bootstrap key conversion (standard -> fft)

Running a benchmark could be useful

Edit by @IceTDrinker :

closes https://github.com/zama-ai/tfhe-rs-internal/issues/210

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_bootstrap_key_conversion.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_bootstrap_key_conversion.rs
@@ -96,6 +96,60 @@ pub fn convert_standard_lwe_bootstrap_key_to_fourier_mem_optimized<Scalar, Input
         .fill_with_forward_fourier(input_bsk.as_view(), fft, stack);
 }
 
+pub fn par_convert_standard_lwe_bootstrap_key_to_fourier<Scalar, InputCont, OutputCont>(
+    input_bsk: &LweBootstrapKey<InputCont>,
+    output_bsk: &mut FourierLweBootstrapKey<OutputCont>,
+) where
+    Scalar: UnsignedTorus,
+    InputCont: Container<Element = Scalar>,
+    OutputCont: ContainerMut<Element = c64>,
+{
+    assert_eq!(
+        input_bsk.polynomial_size(),
+        output_bsk.polynomial_size(),
+        "Mismatched PolynomialSize between input_bsk {:?} and output_bsk {:?}",
+        input_bsk.polynomial_size(),
+        output_bsk.polynomial_size(),
+    );
+
+    assert_eq!(
+        input_bsk.glwe_size(),
+        output_bsk.glwe_size(),
+        "Mismatched GlweSize"
+    );
+
+    assert_eq!(
+        input_bsk.decomposition_base_log(),
+        output_bsk.decomposition_base_log(),
+        "Mismatched DecompositionBaseLog between input_bsk {:?} and output_bsk {:?}",
+        input_bsk.glwe_size(),
+        output_bsk.glwe_size(),
+    );
+
+    assert_eq!(
+        input_bsk.decomposition_level_count(),
+        output_bsk.decomposition_level_count(),
+        "Mismatched DecompositionLevelCount between input_bsk {:?} and output_bsk {:?}",
+        input_bsk.decomposition_level_count(),
+        output_bsk.decomposition_level_count(),
+    );
+
+    assert_eq!(
+        input_bsk.input_lwe_dimension(),
+        output_bsk.input_lwe_dimension(),
+        "Mismatched input LweDimension between input_bsk {:?} and output_bsk {:?}",
+        input_bsk.input_lwe_dimension(),
+        output_bsk.input_lwe_dimension(),
+    );
+
+    let fft = Fft::new(input_bsk.polynomial_size());
+    let fft = fft.as_view();
+
+    output_bsk
+        .as_mut_view()
+        .par_fill_with_forward_fourier(input_bsk.as_view(), fft);
+}
+
 /// Return the required memory for [`convert_standard_lwe_bootstrap_key_to_fourier_mem_optimized`].
 pub fn convert_standard_lwe_bootstrap_key_to_fourier_mem_optimized_requirement(
     fft: FftView<'_>,
--- a/tfhe/src/core_crypto/algorithms/lwe_multi_bit_bootstrap_key_conversion.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_multi_bit_bootstrap_key_conversion.rs
@@ -5,7 +5,9 @@
 use crate::core_crypto::commons::computation_buffers::ComputationBuffers;
 use crate::core_crypto::commons::traits::*;
 use crate::core_crypto::entities::*;
-use crate::core_crypto::fft_impl::fft64::math::fft::{Fft, FftView};
+use crate::core_crypto::fft_impl::fft64::math::fft::{
+    par_convert_polynomials_list_to_fourier, Fft, FftView,
+};
 use concrete_fft::c64;
 use dyn_stack::{PodStack, ReborrowMut, SizeOverflow, StackReq};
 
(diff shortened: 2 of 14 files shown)


tfhe/src/core_crypto/commons/math/torus/mod.rs
```diff
@@ -97,6 +97,8 @@ pub trait UnsignedTorus:
     + RandomGenerable<Uniform, CustomModulus = Self>
     + Display
     + Debug
+    + Sync
+    + Send
```
IceTDrinker, line 101: @tmontaigu opinion on the Sync + Send bound on UnsignedTorus ?

btw @mayeul-zama maybe better on the UnsignedInteger trait, but it's something I really haven't thought about enough

mayeul-zama:
@slab-ci uncompress_key

mayeul-zama:
@slab-ci shortint_bench

tfhe/src/core_crypto/commons/math/torus/mod.rs
tmontaigu: Seems ok to me, UnsignedTorus requires UnsignedInteger which require UnsignedNumeric
which finally require Copy, and I don't know if many Copy types are non-Send or non-Sync,
so i don't think adding this will be too restrictive

IceTDrinker: But for big ints this can be an issue right ? like if the UnsignedInteger were marked Sync and Send ?

IceTDrinker: what I mean is UnsignedTorus is related to the mathematical torus we manage and has nothing to do with "programming properties" of the type

tmontaigu: For 'static' bigint (bigints backed by a `[u64; N]`  which we have in the signed integers PR) this wont be a problem, for 'dynamic' bigint (backed by a `Vec<u64>` which we don't have (but we have all building block)) then UnsignedInteger and Numeric familly of type are not implementatble as they are not `Copy`. Send + Sync should not be a problem
since they are just going to be Vec<u64> without interior mutability thing (Rc, Arc)

If you prefer them to be on the UnsignedInteger trait then its ok as well

I don't have strong opininion on whether these Send + Sync should be UnsignedInteger trait or directly on the function trait bounds that actually need them

IceTDrinker: are we going to need dynamic big ints ?

also if Numeric is marked Copy then maybe the Send + Sync ought to be on the Numeric trait directly ?

tmontaigu: > are we going to need dynamic big ints ?

It depends, for now we don't, but it could happen

Right now users can't encrypt-decrypt more than U256, with the static big int types they will be able to.
However StaticBigInts might create problems due to their stack size required.

Another example is the scalar div, we added and assert that ensures the scalar type has enough bits of precision compared to the ciphertexts number of bits. If we wanted to do the upcasting ourselves we would require dynamic bigint.

Also if we wanted to have the boolean like enum between real and trivial ciphertext at the `tfhe::integer` level and not at the hlapi level, then we would also need them.

> also if Numeric is marked Copy then maybe the Send + Sync ought to be on the Numeric trait directly ?

seems fair

IceTDrinker: @mayeul-zama if you can add the bound on Numeric rather than UnsignedTorus please :) the existing Send/Sync bounds will stay and disappear over time I guess

mayeul-zama: Done

tfhe/src/core_crypto/fft_impl/fft64/math/fft/mod.rs
```diff
@@ -685,5 +686,54 @@ impl<'de, C: IntoContainerOwned<Element = c64>> serde::Deserialize<'de>
     }
 }
 
+pub fn convert_polynomials_list_to_fft<Scalar: UnsignedTorus>(
```
tmontaigu, line 689: rename to `par_convert_...`

tfhe/src/core_crypto/algorithms/lwe_bootstrap_key_conversion.rs
```diff
@@ -96,6 +110,62 @@ pub fn convert_standard_lwe_bootstrap_key_to_fourier_mem_optimized<Scalar, Input
...
 }
 
+pub fn par_convert_standard_lwe_bootstrap_key_to_fourier_mem_optimized<
+    Scalar,
+    InputCont,
+    OutputCont,
+>(
+    input_bsk: &LweBootstrapKey<InputCont>,
+    output_bsk: &mut FourierLweBootstrapKey<OutputCont>,
+    fft: FftView<'_>,
```
IceTDrinker, line 120: this function is not required as it's not taking buffers, no need for a mem optimized version here

tfhe/src/core_crypto/algorithms/lwe_multi_bit_bootstrap_key_conversion.rs
```diff
@@ -78,3 +80,40 @@ pub fn convert_standard_lwe_multi_bit_bootstrap_key_to_fourier_mem_optimized_req
...
+{
+    let fft = Fft::new(input_bsk.polynomial_size());
+    let fft = fft.as_view();
+
+    par_convert_standard_lwe_multi_bit_bootstrap_key_to_fourier_mem_optimized(
+        input_bsk, output_bsk, fft,
+    );
+}
+
+pub fn par_convert_standard_lwe_multi_bit_bootstrap_key_to_fourier_mem_optimized<
```
IceTDrinker, line 100: same thing

tfhe/src/core_crypto/entities/polynomial.rs
```diff
@@ -7,7 +7,7 @@ use std::ops::{Index, IndexMut};
 /// A [`polynomial`](`Polynomial`).
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct Polynomial<C: Container> {
-    data: C,
+    pub data: C,
```
IceTDrinker, line 10: why ? this is not something we allow

IceTDrinker (commented):
Small remarks, but I don't see why you need data to be public ?

If you need access to the data use .as_ref() which returns a slice to the data or .as_mut to have mutable access to it (not recommended for high level algorithms but ok for low level algos potentially)

tfhe/src/core_crypto/entities/polynomial.rs
mayeul-zama: `FourierPolynomial` has a public `data` field
I'll try make both private

mayeul-zama: Seems to complicated to make `FourierPolynomial.data` private
But I'll keep `Polynomial.data` private

IceTDrinker: yep, that's what I was going to say for now fft module plays by different rules

tfhe/src/core_crypto/algorithms/lwe_bootstrap_key_conversion.rs
mayeul-zama: Removed

tfhe/src/core_crypto/algorithms/lwe_multi_bit_bootstrap_key_conversion.rs
mayeul-zama: Removed

tfhe/src/core_crypto/entities/polynomial.rs
mayeul-zama: Done

tfhe/src/core_crypto/fft_impl/fft64/math/fft/mod.rs
```diff
@@ -685,5 +686,52 @@ impl<'de, C: IntoContainerOwned<Element = c64>> serde::Deserialize<'de>
     }
 }
 
+pub fn par_convert_polynomials_list_to_fft<Scalar: UnsignedTorus>(
```
IceTDrinker, line 689: last nitpick to_fourier ?

mayeul-zama: Done

IceTDrinker (approved):
Looks good :) thanks a lot !

mayeul-zama:
@slab-ci shortint_bench

mayeul-zama:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

PR #531, opened 2023-09-08
Am/feat/parallel decompression - part 1

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/139

### PR content/description

- add a MaskRandomGenerator and NoiseRandomGenerator to be able to use them separately without an EncryptionRandomGenerator (allows for parallel decompression to use the same primitives easily)
- updated code for multi bit BSK generation which was half assed to get it done faster
- this adds code to keep current interfaces, it makes adding stuff more tedious, but I currently don't see a way to have a consistent code base around this part of the code while granting the possibility to do parallel decompression properly, propositions welcome

### Check-list:


IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

IceTDrinker:
this was made to be reviewable commit by commit, first one is a rename/split next ones progressively take the code out of the EncryptionRandomGenerator which then becomes a convenience structure

tfhe/src/core_crypto/commons/generators/encryption/mod.rs
```diff
@@ -297,8 +294,7 @@ impl<G: ByteRandomGenerator> EncryptionRandomGenerator<G> {
     ) where
         (Scalar, Scalar): RandomGenerable<Gaussian<f64>>,
     {
-        self.noise
-            .fill_slice_with_random_gaussian(output, 0., std.get_standard_dev());
+        self.noise.fill_slice_with_random_noise(output, std)
```
mayeul-zama, line 297: I would rename `fill_slice_with_random_noise` to something like `fill_slice_with_(random)_gaussian_(noise)`
Same comment for `random_noise_custom_mod` to `(random)_gaussian_(noise)_custom_mod`

Having a name meaning that the mean is 0 would be even better but I don't have an idea of such a short name

tfhe/src/core_crypto/commons/generators/encryption/mod.rs
```diff
@@ -524,86 +526,21 @@ impl<G: ParallelByteRandomGenerator> EncryptionRandomGenerator<G> {
...
 
     // Forks both generators into a parallel iterator.
-    fn par_try_fork(
+    fn par_map_to_encryption_generator(
         &mut self,
-        n_child: usize,
         mask_iter: impl IndexedParallelIterator<Item = MaskRandomGenerator<G>>,
-        noise_bytes: usize,
+        noise_iter: impl IndexedParallelIterator<Item = NoiseRandomGenerator<G>>,
     ) -> Result<impl IndexedParallelIterator<Item = EncryptionRandomGenerator<G>>, ForkError> {
```
mayeul-zama, line 540: Result could be removed from signature
Not sure if this is useful as it forces callers to wrap in an `Ok` themselves

IceTDrinker: Disagree, when wrong number of bytes/children are computed we can and will fail the fork, it's a very useful tool when debugging

tfhe/src/core_crypto/commons/generators/encryption/mod.rs
IceTDrinker: I re-used the previous EncryptionRandomGenerator notation as it now "just" wraps a mask and noise generator

IceTDrinker: but I can see your point

IceTDrinker: I'll think about it as for now our noise are always gaussian as far as the user is concerned

IceTDrinker: Issue https://github.com/zama-ai/tfhe-rs-internal/issues/222

IceTDrinker:
@slab-ci cpu_wasm_test

PR #535, opened 2023-09-11
smart ops always take mut inputs

mayeul-zama:


IceTDrinker:
can we keep the smart always mut as a separate commit ? I'm not sure Thomas agreed on what was preferable, we thought API consitency seemed better but in practice smart clean only if needed (plus mutable refs are a pain for concurrency)

mayeul-zama:
> can we keep the smart always mut as a separate commit ? I'm not sure Thomas agreed on what was preferable, we thought API consitency seemed better but in practice smart clean only if needed (plus mutable refs are a pain for concurrency)

Created PR https://github.com/zama-ai/tfhe-rs/pull/539 with the 3 first commit, and changed this PR base to it so it only contains the smart ops mut inputs change

This avoids conflicts with others PRs based on this one

IceTDrinker:
@tmontaigu ok to use the mut by default on scalar ops ? some ops have updated coming where the clippy lint gets removed

tmontaigu:
🤔, I think it could be better not use use &mut for smart which don't actually need, yes it makes the API not 100% the same, however not having &mut might be more convenient (less typing, easier multhithreading handling) (although its minor as the number of affected ops is small and its the shortint API which is propably the less used of the ones we have)

IceTDrinker:
I can see both sides of this I'm a bit torn

IceTDrinker:
alright then let's rebase this @mayeul-zama and go for the always mut approach

IceTDrinker (approved):
Looks good, thanks !

mayeul-zama:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

PR #536, opened 2023-09-11
shortint/small optim

mayeul-zama:
- cleanup inputs of smart ops instead of cleanup clones
- do not bootstrap when not necessary
- add asserts on smart ops

### DIFF
--- a/tfhe/src/integer/server_key/mod.rs
+++ b/tfhe/src/integer/server_key/mod.rs
@@ -33,13 +33,17 @@ impl From<ServerKey> for crate::shortint::ServerKey {
     }
 }
 
-/// Compute the [`MaxDegree`] for an integer server key (compressed or uncompressed). This formula
-/// provisions a free carry bit. This allows carry propagation between shortint blocks in a
-/// [`RadixCiphertext`](`crate::integer::RadixCiphertext`), as that process requires adding a bit of
-/// carry from one shortint block to the next, which would overflow and lead to wrong results if we
-/// did not provision that carry bit.
+/// Compute the [`MaxDegree`] for an integer server key (compressed or uncompressed).
+/// To allow carry propagation between shortint blocks in a
+/// [`RadixCiphertext`](`crate::integer::RadixCiphertext`) (which includes adding the extracted
+/// carry from one shortint block to the next block), this formula provisions space to add a carry.
 fn integer_server_key_max_degree(parameters: crate::shortint::ShortintParameterSet) -> MaxDegree {
-    MaxDegree((parameters.message_modulus().0 - 1) * parameters.carry_modulus().0 - 1)
+    let full_max_degree = parameters.message_modulus().0 * parameters.carry_modulus().0 - 1;
+
+    let carry_max_degree = parameters.carry_modulus().0 - 1;
+
+    // We want to be have a margin to add a carry from another block
+    MaxDegree(full_max_degree - carry_max_degree)
 }
 
 impl ServerKey {
@@ -144,8 +148,8 @@ mod test {
     #[test]
     fn test_compressed_server_key_max_degree() {
         let cks = ClientKey::new(crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS);
-        // msg_mod = 4, carry_mod = 4, (msg_mod - 1) * carry_mod = 12; minus 1 => 11
-        let expected_max_degree = MaxDegree(11);
+        // msg_mod = 4, carry_mod = 4, (msg_mod * carry_mod - 1) - (carry_mod - 1) = 12
+        let expected_max_degree = MaxDegree(12);
 
         let sks = ServerKey::new(&cks);
         assert_eq!(sks.key.max_degree, expected_max_degree);
--- a/tfhe/src/integer/server_key/radix/scalar_mul.rs
+++ b/tfhe/src/integer/server_key/radix/scalar_mul.rs
@@ -263,7 +263,7 @@ impl ServerKey {
     /// let (cks, sks) = gen_keys_radix(PARAM_MESSAGE_2_CARRY_2_KS_PBS, size);
     ///
     /// let msg = 13;
-    /// let scalar = 5;
+    /// let scalar = 2;
     ///
     /// let mut ct = cks.encrypt(msg);
     ///
--- a/tfhe/src/shortint/engine/server_side/div_mod.rs
+++ b/tfhe/src/shortint/engine/server_side/div_mod.rs
@@ -39,17 +39,27 @@ impl ShortintEngine {
         Ok(())
     }
 
-    // by convention smart operations take mut refs to their inputs, even if they do not modify them
-    #[allow(clippy::needless_pass_by_ref_mut)]
     pub(crate) fn smart_div(
         &mut self,
         server_key: &ServerKey,
         ct_left: &mut Ciphertext,
         ct_right: &mut Ciphertext,
     ) -> EngineResult<Ciphertext> {
-        let mut result = ct_left.clone();
-        self.smart_div_assign(server_key, &mut result, ct_right)?;
-        Ok(result)
+        if !server_key.is_functional_bivariate_pbs_possible(ct_left, ct_right) {
+            if ct_left.message_modulus.0 + ct_right.degree.0 <= server_key.max_degree.0 {
+                self.message_extract_assign(server_key, ct_left)?;
+            } else if ct_right.message_modulus.0 + (ct_left.degree.0 + 1) <= server_key.max_degree.0
+            {
+                self.message_extract_assign(server_key, ct_right)?;
+            } else {
+                self.message_extract_assign(server_key, ct_left)?;
+                self.message_extract_assign(server_key, ct_right)?;
+            }
+        }
+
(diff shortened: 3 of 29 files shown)


tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -681,11 +681,15 @@ impl ShortintEngine {
         server_key: &ServerKey,
         ct: &mut Ciphertext,
     ) -> EngineResult<()> {
-        let modulus = ct.message_modulus.0 as u64;
+        if ct.carry_is_empty() {
+            self.create_trivial_assign(server_key, ct, 0)?;
```
IceTDrinker, line 718: This change is dangerous as it currently is also used for noise management, with a degree/noise rework yes, but I would not make that change/check currently

tfhe/src/shortint/engine/server_side/shift.rs
```diff
@@ -47,14 +47,15 @@ impl ShortintEngine {
...
         server_key: &ServerKey,
         ct: &mut Ciphertext,
         shift: u8,
     ) -> EngineResult<Ciphertext> {
+        if !server_key.is_scalar_left_shift_possible(ct, shift) {
+            self.message_extract_assign(server_key, ct)?;
+        }
         let mut result = ct.clone();
         self.smart_scalar_left_shift_assign(server_key, &mut result, shift)?;
         Ok(result)
```
tmontaigu, line 61: I think this is possibly a bit sub optimal here.

Image a shift value that makes `is_scalar_left_shift` possible return false.
We do a message_extract_assign.

Then we call the `assign` version. however, inside the assign version the same check `is_scalar_left_shift_possible` may also return false because because maybe the shift value is too big which will in turn force the shift to be done with a PBS.

So we did 2 PBS, where we could only do one

mayeul-zama: I guess the best way would be to `message_extract_assign` if it would make `scalar_left_shift` possible
But it's probably not worth the effort
Should I remove the whole commit then ?

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: I agree
I'll keep this commit in another branch for later

mayeul-zama: Done

tfhe/src/shortint/engine/server_side/shift.rs
mayeul-zama: Removed

mayeul-zama:
@slab-ci cpu_fast_test

tfhe/src/integer/server_key/radix/scalar_mul.rs
```diff
@@ -255,6 +255,7 @@ impl ServerKey {
         if !self.is_small_scalar_mul_possible(ctxt, scalar) {
             self.full_propagate(ctxt);
         }
+        // assert!(self.is_small_scalar_mul_possible(ctxt, scalar));
```
IceTDrinker, line 258: same question everywhere, why is it commented here and in this case even more so given it's not commented below

IceTDrinker (commented):
Not sure about the pattern where we check if the operation is possible, clean the ciphertexts and then assert right after 🤔

mayeul-zama:
@slab-ci cpu_fast_test

IceTDrinker (approved):
All right let's see what the tests have to say about this !

Needs a second pair of eyes though, I may have missed some important details wrt to how ops are used in Integer

tfhe/src/integer/server_key/mod.rs
```diff
@@ -33,13 +33,17 @@ impl From<ServerKey> for crate::shortint::ServerKey {
...
-/// did not provision that carry bit.
+/// Compute the [`MaxDegree`] for an integer server key (compressed or uncompressed).
+/// To allow carry propagation between shortint blocks in a
+/// [`RadixCiphertext`](`crate::integer::RadixCiphertext`) (which includes adding the extracted
+/// carry from one shortint block to the next block), this formula provisions space to add a carry.
 fn integer_server_key_max_degree(parameters: crate::shortint::ShortintParameterSet) -> MaxDegree {
-    MaxDegree((parameters.message_modulus().0 - 1) * parameters.carry_modulus().0 - 1)
+    let full_max_degree = parameters.message_modulus().0 * parameters.carry_modulus().0 - 1;
+
+    let carry_max_degree = parameters.carry_modulus().0 - 1;
```
IceTDrinker, line 43: I think this is not correct, the value to consider here is `parameters.carry_modulus().0 - 1` while for most parameters sets the two moduli are equal.

@tmontaigu to confirm

tfhe/src/shortint/engine/server_side/div_mod.rs
```diff
@@ -39,17 +39,27 @@ impl ShortintEngine {
...
+            if ct_left.message_modulus.0 + ct_right.degree.0 <= server_key.max_degree.0 {
+                self.message_extract_assign(server_key, ct_left)?;
+            } else if ct_right.message_modulus.0 + (ct_left.degree.0 + 1) <= server_key.max_degree.0
+            {
+                self.message_extract_assign(server_key, ct_right)?;
+            } else {
+                self.message_extract_assign(server_key, ct_left)?;
+                self.message_extract_assign(server_key, ct_right)?;
+            }
+        }
```
tmontaigu, line 58: Why did you do something more complex here than for bitand/bitor which are also bivariate PBSes

Also the correlations of the ifs to the message extract are weird:

- `if ct_left.message_modulus.0 + ct_right.degree.0 <= server_key.max_degree.0`
 -> you message extract ct_left, but you checked ct_right degree not ct_left
    and you used ct_left message modulus, but after message_extract modulus, ct_right degree would not have changed, and ct_left degreee would be message_modulus - 1
to me, the potential issue is not fixed 

same idea for the `else if`

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -482,10 +482,22 @@ impl ShortintEngine {
...
 
-        self.smart_evaluate_bivariate_function_assign(server_key, &mut ct_res, ct_right, f)?;
-        Ok(ct_res)
+        if !server_key.is_functional_bivariate_pbs_possible(ct_left, ct_right) {
+            // After the message_extract, we'll have ct_left, ct_right in [0, message_modulus[
+            // so the factor has to be message_modulus
+            assert_eq!(ct_right.message_modulus.0, lookup_table.ct_right_modulus.0);
+            self.message_extract_assign(server_key, ct_left)?;
+            self.message_extract_assign(server_key, ct_right)?;
+        }
```
tmontaigu, line 496: I think it would be easier to generate the Lut *after* the potential message extracts have been done, so that the assert is not needed.

Ie why chose a factor if it may be wrong, when you could choose the factor and have it be always valid ?

tfhe/src/integer/server_key/mod.rs
IceTDrinker: carry modulus makes more sense actually, apologies, I was focused on the formula we wrote in the issue, all good !

tfhe/src/shortint/engine/server_side/div_mod.rs
mayeul-zama: I just copied what is done in `smart_div_assign` to `smart_div`  so that inputs are cleaned instead of their clones

The behavior (except the potential cleaning of the inputs) should be the same as before
I suggest we fix this in another PR

WDYT?

tfhe/src/shortint/engine/server_side/mod.rs
IceTDrinker: that I did it in a rush and to not take this for the best solution, if there is a better piece of code use it, I would not know if mine makes much sense

mayeul-zama: Fixed

IceTDrinker:
I let Thomas approve for the latest changes linked to his requests if that's fine with you

tfhe/src/shortint/engine/server_side/div_mod.rs
tmontaigu: Ok if it comes from smart_div_assign then we can merge and think more about this strange thing

tmontaigu (approved):

tfhe/src/shortint/engine/server_side/div_mod.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/250

IceTDrinker:
that one https://github.com/zama-ai/tfhe-rs/actions/runs/6275470013/job/17043013270#step:8:3202

is a hard coded test that can be updated safely and was a sanity check with the previous accpeted degree value (though we knew it was broken, and needed updating later)

tmontaigu (approved):

IceTDrinker:

mayeul-zama:
Moved add assert commit to PR https://github.com/zama-ai/tfhe-rs/pull/590

tmontaigu (approved):

PR #537, opened 2023-09-11
shortint/fix degree

mayeul-zama:
In shortint, change the meaning of degree to be a measure to guarantee both:
- no padding-bit overflow
- not breaking PBS noise limit

### DIFF
--- a/tfhe/src/shortint/server_key/mod.rs
+++ b/tfhe/src/shortint/server_key/mod.rs
@@ -28,7 +28,7 @@ use crate::core_crypto::commons::parameters::{
 };
 use crate::core_crypto::commons::traits::*;
 use crate::core_crypto::entities::*;
-use crate::shortint::ciphertext::{Ciphertext, Degree};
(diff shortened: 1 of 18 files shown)


mayeul-zama:
@slab-ci cpu_fast_test

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -274,7 +274,7 @@ fn ciphertexts_can_be_packed_without_exceeding_space(
 #[must_use]
 pub struct LookupTable<C: Container<Element = u64>> {
     pub acc: GlweCiphertext<C>,
-    pub degree: Degree,
+    pub max_value: u64,
```
IceTDrinker, line 277: disagree strongly with this change

IceTDrinker (commented):
As discussed I believe this change would be better in conjunction with the noise update

IceTDrinker:
can this be put in draft or closed for now as we prefer having the noise rework at the same time ? 🙂

PR #538, opened 2023-09-11
shortint/fix: take modular inputs on operations 

mayeul-zama:


tfhe/src/shortint/engine/server_side/scalar_sub.rs
```diff
@@ -58,7 +58,11 @@ impl ShortintEngine {
         } else {
             let scalar = u64::from(scalar);
             // If the scalar is too large, PBS is used to compute the scalar mul
-            let acc = self.generate_lookup_table(server_key, |x| (x - scalar) % modulus)?;
+            let acc = self.generate_msg_lookup_table(
+                server_key,
+                |x| (((x - scalar) % modulus) + modulus) % modulus,
```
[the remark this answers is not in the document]
mayeul-zama: Yes, indeed, when overflow checks are disabled and the modulus is a power of 2, it's correct

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -315,6 +315,18 @@ impl ShortintEngine {
         })
     }
 
+    pub(crate) fn generate_msg_lookup_table<F>(
+        &mut self,
+        server_key: &ServerKey,
+        f: F,
+        modulus: u64,
```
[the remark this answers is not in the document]
mayeul-zama: Done

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -363,6 +363,15 @@ impl ServerKey {
         })
     }
 
+    pub fn generate_msg_lookup_table<F>(&self, f: F, modulus: u64) -> LookupTableOwned
```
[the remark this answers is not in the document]
mayeul-zama: Added

tfhe/src/shortint/engine/server_side/scalar_sub.rs
```diff
@@ -58,7 +58,11 @@ impl ShortintEngine {
         } else {
             let scalar = u64::from(scalar);
             // If the scalar is too large, PBS is used to compute the scalar mul
```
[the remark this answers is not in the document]
mayeul-zama: Fixed

tfhe/src/shortint/engine/server_side/scalar_sub.rs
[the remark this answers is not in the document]
mayeul-zama: Use `wrapping_sub`, removed double modulus

tfhe/src/shortint/engine/server_side/div_mod.rs
```diff
@@ -97,7 +97,12 @@ impl ShortintEngine {
         scalar: u8,
     ) -> EngineResult<()> {
         assert_ne!(scalar, 0, "attempt to divide by zero");
-        let lookup_table = self.generate_lookup_table(server_key, |x| x / (scalar as u64))?;
+
+        let lookup_table = self.generate_msg_lookup_table(
```
IceTDrinker, line 101: not sure I like this style of lookup generation, it hides the parts where there are moduli especially for "low level" usage as we do here 🤔 

is there still a raw primitive without automated modulus management ?

mayeul-zama: Sure, `generate_lookup_table` still exists and is used in many places, especially in the integer API

IceTDrinker: ok, I'll check more thorougly but the changes seemed straightforward otherwise

tfhe/src/shortint/engine/server_side/scalar_sub.rs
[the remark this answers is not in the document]
IceTDrinker: when in doubt use .rem_euclid() for the "mathematical" modulo operator, only is an issue for signed integer types

IceTDrinker:
please fix the commit messages not passing conformance before we trigger test suites

tfhe/src/shortint/engine/server_side/shift.rs
```diff
@@ -20,7 +20,7 @@ impl ShortintEngine {
         ct: &mut Ciphertext,
         shift: u8,
     ) -> EngineResult<()> {
-        let acc = self.generate_lookup_table(server_key, |x| x >> shift)?;
+        let acc = self.generate_msg_lookup_table(server_key, |x| x >> shift, ct.message_modulus)?;
```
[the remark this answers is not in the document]
mayeul-zama: Yes
If this function was used to build the integer API, this would break things
But I also guess that this function is supposed to be used ignoring the carry bits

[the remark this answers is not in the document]
IceTDrinker: could be we'll know once you have pushed the updated commit respecting the conventional notaion for approval test runs

mayeul-zama:
> please fix the commit messages not passing conformance before we trigger test suites

Actually I'm not sure the change in the last commit is correct which is why I separated it for someone to give their opinion on it

I can remove or squash the commit once I know whether the changes should be kept or not

IceTDrinker:
are there some tests associated with those ?

mayeul-zama:
> are there some tests associated with those ?

Yes, but I guess they are not part of the fast tests
I added asserts in the associated doc tests

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -363,6 +363,41 @@ impl ServerKey {
...
+    ///
+    /// // Generate the client key and the server key:
+    /// let (cks, sks) = gen_keys(PARAM_MESSAGE_2_CARRY_2_KS_PBS);
+    ///
+    /// let msg = 3;
+    ///
+    /// let ct = cks.encrypt(msg);
+    ///
+    /// // Generate the lookup table for the function f: x -> x^2 mod 2^2
+    /// let f = |x| x ^ 2 % 4;
```
IceTDrinker, line 383: use the msg mod from parameters directly ?

tfhe/src/shortint/key_switching_key/mod.rs
```diff
@@ -102,9 +102,12 @@ impl KeySwitchingKey {
     ///     PARAM_KEYSWITCH_1_1_KS_PBS_TO_2_2_KS_PBS,
     /// );
     ///
-    /// let cipher = ck1.encrypt(1);
+    /// let plaintext = 1;
```
IceTDrinker, line 105: that's the cleartext as it's not encoded yet

tfhe/src/shortint/key_switching_key/mod.rs
```diff
@@ -159,8 +162,12 @@ impl KeySwitchingKey {
...
```
IceTDrinker, line 165: same thing

IceTDrinker (commented):
A few last comments before an approval for test runs, I would like a second pair of eyes on this before we merge (though tests should be a pretty good indicator already)

tfhe/src/shortint/engine/mod.rs
```diff
@@ -114,7 +114,7 @@ where
             .iter_mut()
             .for_each(|a| {
                 let f_eval = f(i as u64);
-                *a = f_eval * delta;
+                *a = f_eval.wrapping_mul(delta);
```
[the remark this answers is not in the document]
mayeul-zama: `shortint::key_switching_key::test::gen_multi_keys_test_truncate` fails on main with `attempt to multiply with overflow` (with overflow checks enabled)

But it's because of a bug which I'll push a fix for
I'll remove this wrapping_mul and add an assert `f_eval < message * carry`

I think we should enable overflow checks on CI tests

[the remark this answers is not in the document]
IceTDrinker: yeah for the overflow checks it's in the backlog

what's the bug ?

but I think the code itself should do a checked mul so that we can warn/crash on overflow here as it means the table is likely broken

tfhe/src/shortint/server_key/mod.rs
mayeul-zama: Done

tfhe/src/shortint/key_switching_key/mod.rs
mayeul-zama: Done

tfhe/src/shortint/key_switching_key/mod.rs
mayeul-zama: Done

tfhe/src/shortint/engine/mod.rs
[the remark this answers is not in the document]
mayeul-zama: Yes, the assert should be equivalent to a checked mul

tfhe/src/shortint/server_key/tests/shortint.rs
```diff
@@ -532,7 +532,7 @@ where
 {
     let keys = KEY_CACHE.get_from_param(param);
     let (cks, sks) = (keys.client_key(), keys.server_key());
-    let double = |x| 2 * x;
+    let double = |x| (2 * x) % sks.message_modulus.0 as u64;
```
IceTDrinker, line 535: was this test failing ?
otherwise I agree that this change could destroy the padding bit

tfhe/src/shortint/key_switching_key/mod.rs
```diff
@@ -122,7 +125,10 @@ impl KeySwitchingKey {
 
             // Cast to smaller bit length: left shift, then keyswitch
             i if i < 0 => {
-                let acc = self.src_server_key.generate_lookup_table(|n| n << -i);
+                // We want to avoid the padding bit to be dirty, hence the modulus
+                let acc = self.src_server_key.generate_lookup_table(|n| {
+                    (n << -i) % (ct.carry_modulus.0 * ct.message_modulus.0) as u64
```
IceTDrinker, line 130: does this make sense functionally ? I'm not sure about what the casting was doing, if we are trashing the padding bit we may want to check what's going on

tfhe/src/shortint/engine/client_side.rs
```diff
@@ -300,7 +300,13 @@ impl ShortintEngine {
 
     pub fn decrypt(&mut self, client_key: &ClientKey, ct: &Ciphertext) -> EngineResult<u64> {
         self.decrypt_message_and_carry(client_key, ct)
-            .map(|message_and_carry| message_and_carry % ct.message_modulus.0 as u64)
+            .map(|message_and_carry| {
+                let padding_bit =
+                    message_and_carry / (ct.carry_modulus.0 * ct.message_modulus.0) as u64;
+
+                assert_eq!(padding_bit, 0);
```
IceTDrinker, line 307: so that's an interesting idea but I would probably see something like an Ok(val) if padding bit is 0 and an Err(val) with a nice error formatting for the Err variant

I don't know if asserting here is a good idea, decrypting is not failing per se

tfhe/src/shortint/engine/mod.rs
```diff
@@ -114,6 +114,9 @@ where
             .iter_mut()
             .for_each(|a| {
                 let f_eval = f(i as u64);
+
+                assert!(f_eval < modulus_sup as u64);
+
```
IceTDrinker, line 119: so same thing having a bit on the padding bit is not illegal per se, but we may want to return the information as an Enum/Result down the line as this may lead to invalid computations later, don't know what the best here would be

@tmontaigu if you have an opinion on this ?

also I wonder if it may not break some stuff like Parmesan by Jakub which relied on some tricks to have negacyclic functions

IceTDrinker (commented):
A few comments

Generally speaking when I'm making comments/asking questions I'm not always expecting a fix in the commits in the next iteration it's more to open the conversation, especially when adding asserts for some behavior as they may require a bit more of planning on our end :) (talking about the decrypt assert and the TLU generation)

tfhe/src/shortint/server_key/tests/shortint.rs
mayeul-zama: With the new `assert!(f_eval < modulus_sup as u64)` the `shortint_generate_lookup_table` test was failing with all the parameters I think

tfhe/src/high_level_api/shortints/tests.rs
```diff
@@ -169,11 +169,11 @@ fn test_programmable_bootstrap_fhe_uint2() -> Result<(), Box<dyn std::error::Err
 
     let mut a = FheUint2::try_encrypt(2, &keys)?;
 
-    let c = a.map(|value| value * value);
+    let c = a.map(|value| (value * value) % 4);
     let decrypted = c.decrypt(&keys);
     assert_eq!(decrypted, (2 * 2) % 2);
 
-    a.apply(|value| value * value);
+    a.apply(|value| (value * value) % 4);
```
tmontaigu, line 176: This should not be needed here, the modulus should be applied internally, not at the user code level

mayeul-zama: You mean `map` and `apply` should apply the modulus on the output of the closure itself?

If so, would you do the modulus in `GenericShortIntServerKey::bootstrap_(inplace)_with`?

Should the modulo be over the message space, the message and the carry space or the message space, carry space and padding bit? It seems to me that it depends on the user. We could have a function for each or leave it to the caller as it is now

tmontaigu: To me in the HLP API, we expose things like FheUint2, FheUint3 where the modulus is 2**2, 2**3
the idea is that the carry system is hidden, so users only see the 2/3 bits of message

mayeul-zama: Makes sense, will fix

mayeul-zama: Done

tmontaigu: Also the modulus used in the assert is wrong : it does `%2` but it should be "%4" 

(The 'bug' was there before your modifications)

tfhe/src/shortint/engine/mod.rs
tmontaigu: There is a place in the integers comparision where we do a PBS on values where the bit of padding could be 1

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -290,6 +290,18 @@ impl ShortintEngine {
...
+        &mut self,
+        server_key: &ServerKey,
+        f: F,
+        modulus: MessageModulus,
+    ) -> EngineResult<LookupTableOwned>
+    where
+        F: Fn(u64) -> u64,
+    {
+        Self::generate_lookup_table_with_engine(server_key, |x| f(x % modulus.0 as u64))
+    }
```
tmontaigu, line 303: I wonder, if this applies modulus to the input, would it be coherent for it to also apply the modulus on the output ?

mayeul-zama: I guess if we want to ignore the carry bits of the input, we also want to ignore the ones of the output
So IMO it makes sense to apply the same modulus on the ouput here

mayeul-zama: Done

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -367,7 +367,7 @@ impl ServerKey {
     /// let ct = cks.encrypt(msg);
     ///
     /// // Generate the lookup table for the function f: x -> x^2 mod 2^2
-    /// let f = |x| x ^ 2 % 4;
+    /// let f = |x| x ^ 2;
```
IceTDrinker, line 370: this is most likely a wrong usage of ^, this should be a power instead what's the status on https://github.com/zama-ai/tfhe-rs/pull/638 @tmontaigu ?

tfhe/src/shortint/engine/server_side/shift.rs
```diff
@@ -70,7 +70,8 @@ impl ShortintEngine {
             self.unchecked_scalar_left_shift_assign(ct, shift)?;
         } else {
             let modulus = server_key.message_modulus.0 as u64;
-            let acc = self.generate_lookup_table(server_key, |x| (x << shift) % modulus)?;
+            let acc =
+                self.generate_msg_lookup_table(server_key, |x| x << shift, ct.message_modulus)?;
```
IceTDrinker, line 74: this changes the modulus used here and mismatches the one in the call to after_left_shift below

tfhe/src/high_level_api/shortints/server_key.rs
```diff
@@ -429,7 +429,9 @@ where
     where
         F: Fn(u64) -> u64,
     {
-        let lookup_table = self.key.generate_lookup_table(func);
+        let lookup_table = self
+            .key
+            .generate_lookup_table(|a| func(a) % self.key.message_modulus.0 as u64);
```
IceTDrinker, line 434: should this be the message_modulus or the max modulus including the carry ? as long as we are not trashing the padding bit ? same bellow

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -290,6 +290,20 @@ impl ShortintEngine {
...
+        server_key: &ServerKey,
+        f: F,
+        modulus: MessageModulus,
+    ) -> EngineResult<LookupTableOwned>
+    where
+        F: Fn(u64) -> u64,
+    {
+        Self::generate_lookup_table_with_engine(server_key, |x| {
+            f(x % modulus.0 as u64) % modulus.0 as u64
+        })
```
IceTDrinker, line 304: I think I disagree with that % modulus at the end, someone may want to have bits in the carries even if only applying a LUT on the message part of x

tfhe/src/shortint/engine/server_side/scalar_sub.rs
```diff
@@ -51,14 +51,17 @@ impl ShortintEngine {
...
         } else {
             let scalar = u64::from(scalar);
-            // If the scalar is too large, PBS is used to compute the scalar mul
-            let acc = self.generate_lookup_table(server_key, |x| (x - scalar) % modulus)?;
+            // If the scalar is too large, PBS is used to compute the scalar sub
+            let acc = self.generate_msg_lookup_table(
+                server_key,
+                |x| x.wrapping_sub(scalar),
+                server_key.message_modulus,
+            )?;
```
IceTDrinker, line 64: is this still correct when x is replaced by x % modulus ?

tfhe/src/shortint/server_key/mod.rs
tmontaigu: Its ready to be reviewed

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -385,6 +385,41 @@ impl ServerKey {
...
+    ///
+    /// // Generate the lookup table for the function f: x -> x^2 mod 2^2
+    /// let f = |x| x ^ 2;
+    ///
+    /// let acc = sks.generate_msg_lookup_table(f, ct.message_modulus);
+    /// let ct_res = sks.apply_lookup_table(&ct, &acc);
+    ///
+    /// let dec = cks.decrypt(&ct_res);
+    /// // 3^2 mod 4 = 1
+    /// assert_eq!(dec, f(msg));
```
tmontaigu, line 412: x^2 in the lut means xor(x, 2) not power so this might confuse readers

tfhe/src/high_level_api/shortints/server_key.rs
tmontaigu: I think it might be better to have it include the carry space also since using the carry space is inherent to the proper use of shortint :thinking:

tfhe/src/shortint/engine/server_side/shift.rs
mayeul-zama: Do we support applying LUT with `sk.message_modulus != ct.message_modulus`?

tfhe/src/high_level_api/shortints/server_key.rs
mayeul-zama: I guess it's moslty an interpretation of what users should expect from this function

But I agree, I could make more sense to include the carry modulus

IceTDrinker: I guess an unchecked_generated_lut would be nice ? and maybe a smart_generate_lut which would give the msg and carry values basically would do

```rust
for plaintext in 0..plaintext_mod {
    let msg = plaintext % msg_mod;
    let carry = plaintext / msg_mod;
    let lut_val = user_provided_fn(msg, carry);
    // store in lut
}
```

this way you can chose if you are manipulating the message and/or the carry

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: See https://github.com/zama-ai/tfhe-rs/pull/538#discussion_r1334440315

IMO, if someone wants to play with the input or output carry, they should use `generate_lookup_table` and handle the input and output carry themselves

IceTDrinker: that's a design question, what is the meaning of the various functions we have in there, could you create a quick list of those function and their current behavior ? this way we can discuss and then agree on what we want, as those are mostly used by us (given shortint is already fairly low level) it's mainly a QoL for us so we can chose what we prefer as implementers

tfhe/src/shortint/engine/server_side/scalar_sub.rs
mayeul-zama: Yes, I think so

But there may be a problem with wrapping if `ct_modulus % message_modulus != 0`
I think `x + msg_modulus - (scalar % msg_modulus)` is what we want. Does it make sense to you?

tfhe/src/shortint/server_key/mod.rs
mayeul-zama: Yes, I thought it was a pow :upside_down_face:

mayeul-zama: Should I change that to `x * x` ?

IceTDrinker: the corresponding PR from Thomas has been accepted so should get merged fairly soon, so if you rebase and update should be ok 👍

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: - `generate_lookup_table` does not deal with modulus
- `generate_msg_lookup_table` applies `% msg_modulus` to the input and to the output of the function
- `generate_lookup_table_bivariate_with_factor`  which wrap `f` this way
```
    let lhs = (input / factor_u64) % message_modulus;
    let rhs = (input % factor_u64) % message_modulus;

    f(lhs, rhs)
};
```
- `generate_lookup_table_bivariate` same as `generate_lookup_table_bivariate_with_factor` with `factor = server_key.message_modulus`

tfhe/src/shortint/engine/server_side/scalar_sub.rs
IceTDrinker: indeed for non power of two it's probably better the way you just wrote it

IceTDrinker: let's go with that formula then I'd say

tfhe/src/shortint/engine/server_side/mod.rs
tmontaigu: If I take the integers implementations, most of them actually play with the carry space whether its the one of the input or output, I even have a lut that ouputs a block with padding bit set.

So as long as we have `generate_lookup_table` for maximum control and `generate_lookup_table_bivariate` its fine. 

`generate_msg_lookup_table` could be used in some places to replace `generate_lookup_table` where we manually apply msg_mod to both input/output but i don't think that is very common

tfhe/src/shortint/engine/server_side/scalar_sub.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/mod.rs
mayeul-zama: Done

tfhe/src/shortint/engine/server_side/mod.rs
IceTDrinker: all right thanks for the detailed infos, will do a last round of reviews

tfhe/src/high_level_api/shortints/server_key.rs
mayeul-zama: Shortints hve been removed for the HL API so I guess this solves this issue

tmontaigu (approved):

PR #544, opened 2023-09-12
chore(ci): avoid running full suite several times on approval

soonum:
If two or more reviewers approve a Pull Request successively with no new commits in between, the full test suite would have been run for each approval. With this commit, the full test suite would be run again upon approval only if a push has occurred.


IceTDrinker (approved):
Test approve

tmontaigu (approved):

IceTDrinker (approved):
Looks good I believe

PR #547, opened 2023-09-13
chore(ci): notify slack about coverage only in case of failure

soonum:


IceTDrinker:
hmm rebase on main ? I see an unrelated commit

PR #548, opened 2023-09-13
chore(ci): improve code coverage execution

soonum:
This PR does the following:
- exclude files from coverage reports
- trigger code coverage only if sources changed
- run coverage on AWS EC2

Resolves: zama-ai/tfhe-rs-internal#234

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -6,17 +6,64 @@ env:
   RUSTFLAGS: "-C target-cpu=native"
 
 on:
-  push:
-    branches:
-      - "main"
-  pull_request:
+  # Allows you to run this workflow manually from the Actions tab as an alternative.
+  workflow_dispatch:
+    # All the inputs are provided by Slab
+    inputs:
+      instance_id:
+        description: "AWS instance ID"
+        type: string
+      instance_image_id:
+        description: "AWS instance AMI ID"
+        type: string
+      instance_type:
+        description: "AWS instance product type"
+        type: string
+      runner_name:
+        description: "Action runner name"
+        type: string
+      request_id:
+        description: 'Slab request ID'
+        type: string
+      fork_repo:
+        description: 'Name of forked repo as user/repo'
+        type: string
+      fork_git_sha:
+        description: 'Git SHA to checkout from fork'
+        type: string
 
 jobs:
   code-coverage:
-    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}_${{ github.ref }}_${{ inputs.instance_image_id }}_${{ inputs.instance_type }}
+      cancel-in-progress: true
+    runs-on: ${{ inputs.runner_name }}
     steps:
+      # Step used for log purpose.
+      - name: Instance configuration used
+        run: |
+          echo "ID: ${{ inputs.instance_id }}"
+          echo "AMI: ${{ inputs.instance_image_id }}"
+          echo "Type: ${{ inputs.instance_type }}"
+          echo "Request ID: ${{ inputs.request_id }}"
+          echo "Fork repo: ${{ inputs.fork_repo }}"
+          echo "Fork git sha: ${{ inputs.fork_git_sha }}"
+
       - name: Checkout tfhe-rs
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+        with:
+          repository: ${{ inputs.fork_repo }}
+          ref: ${{ inputs.fork_git_sha }}
+
+      - name: Set up home
+        run: |
+          echo "HOME=/home/ubuntu" >> "${GITHUB_ENV}"
+
+      - name: Install latest stable
+        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af
+        with:
+          toolchain: stable
+          default: true
 
       - name: Check for file changes
         id: changed-files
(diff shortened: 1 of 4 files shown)


ci/slab.toml
```diff
@@ -82,3 +82,8 @@ check_run_name = "WASM Client AWS Benchmarks"
 workflow = "csprng_randomness_testing.yml"
 profile = "cpu-small"
 check_run_name = "CSPRNG randomness testing"
+
+[command.code_coverage]
+workflow = "code_coverage.yml"
+profile = "cpu-small"
```
IceTDrinker, line 88: don't we want the CPU big route for more parallelism ?

Makefile
```diff
@@ -321,7 +322,8 @@ test_shortint: install_rs_build_toolchain
 test_shortint_cov: install_rs_check_toolchain install_tarpaulin
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_CHECK_TOOLCHAIN) tarpaulin --profile $(CARGO_PROFILE) \
 		--out Xml --output-dir coverage/shortint --line --engine Llvm --timeout 500 \
-		--features=$(TARGET_ARCH_FEATURE),shortint,internal-keycache,__coverage -p tfhe -- shortint::
+		--exclude-files $(COVERAGE_EXCLUDED_FILES) \
+		--features=$(TARGET_ARCH_FEATURE),shortint,internal-keycache,__coverage -p tfhe -- test_shortint_unchecked_bitand
```
IceTDrinker, line 326: test filter is not right

Makefile
```diff
@@ -16,6 +16,7 @@ PARSE_INTEGER_BENCH_CSV_FILE?=tfhe_rs_integer_benches.csv
 FAST_TESTS?=FALSE
 FAST_BENCH?=FALSE
 BENCH_OP_FLAVOR?=DEFAULT
+COVERAGE_EXCLUDED_FILES = tfhe/benches/*,apps/trivium/src/*,tfhe/examples/*,tasks/src/*
```
IceTDrinker, line 19: I think it's fine to hardcode in the recipe, those are not supposed to change right ?

.github/workflows/trigger_aws_tests_on_pr.yml
```diff
@@ -52,3 +53,4 @@ jobs:
             @slab-ci cpu_multi_bit_test
             @slab-ci cpu_wasm_test
             @slab-ci csprng_randomness_testing
+            @slab_ci code_coverage
```
IceTDrinker, line 56: does it need to run both times ?

.github/workflows/code_coverage.yml
```diff
@@ -6,23 +6,81 @@ env:
...
+            concrete_csprng:
+              - concrete-csprng/src/**
 
-      - name: Gen Keys if required
+      - name: Generate Keys
         run: |
           make GEN_KEY_CACHE_COVERAGE_ONLY=TRUE gen_key_cache
 
       - name: Run shortint coverage
+        if: steps.changed-files.outputs.tfhe_any_changed == 'true'
```
IceTDrinker, line 83: checked this worked as expected ? I don't think this would work ok for the CSPRNG e.g.

Makefile
soonum: It's because when I'll add the other layers I'll reuse this variable.

Makefile
soonum: yes my bad.

Makefile
IceTDrinker: not sure I follow 🤔

ci/slab.toml
soonum: It looks like the coverage is not parallelized, on my laptop it take only one CPU to run the tests.

.github/workflows/trigger_aws_tests_on_pr.yml
soonum: No it's not needed indeed.

.github/workflows/code_coverage.yml
soonum: The csprng will have its own step. Unless if you consider that a change in csprng could have an impact to testing in tfhe-rs

Makefile
soonum: It's to avoid copy/pasting since when I'll make `test_integer_cov`, `test_core_cov`, etc, I'll reuse this line of excluded files.

.github/workflows/code_coverage.yml
IceTDrinker: ah no you are right my bad indeed 😵‍💫

Makefile
IceTDrinker: ah yes makes sense then

ci/slab.toml
soonum: `cpu-big` profile doesn't significantly improve execution duration.
That said we still goes from 45mins on GitHub VM  down to 25mins on `cpu-small`. A good improvement though.

ci/slab.toml
```diff
@@ -1,16 +1,16 @@
 [profile.cpu-big]
 region = "eu-west-3"
-image_id = "ami-0ab73f5bd11708a85"
+image_id = "ami-051942e4055555752"
```
IceTDrinker, line 3: updated with openssl properly installed and update packages I'm guessing ?

ci/slab.toml
IceTDrinker: yes I was missing the fact the coverage was not parallelized :/

IceTDrinker (approved):
Thanks looks good to me !

ci/slab.toml
soonum: yes!

PR #549, opened 2023-09-13
Mz/just

mayeul-zama:


IceTDrinker:
not sure this is the most urgent at the moment

justfile
```diff
@@ -0,0 +1,578 @@
...
+
+
+fmt: install_rs_check_toolchain
+	{{CARGO_FMT}}
+
+check_fmt: 
+	{{CARGO_FMT}} --check
+
+
+clippy_core: install_rs_check_toolchain
```
IceTDrinker, line 122: looks like we are missing the help strings from the .PHONY targets

IceTDrinker (commented):
Thanks for looking into this !

May I suggest maybe starting with a smaller subset of the Makefile and starting by transitioning something like pcc so that we can beta test a bit ?

The reason also is that I'm not yet used to just and would like to get a good understanding from it to be able to keep editing this effectively task file

PR #551, opened 2023-09-14
chore(ci): disable codecov patch status

soonum:
This is done to avoid noisy reports in GitHub since coverage in all the library layers haven't been implemented yet.


IceTDrinker (approved):
Thanks !

tmontaigu (approved):

PR #554, opened 2023-09-14
chore(ci): add end-of-file newline checks recipe

soonum:
Resolves: zama-ai/tfhe-rs-internal#24


soonum:
As discussed with @IceTDrinker, we should assume that the linter is installed and if not print the link to the release page.
That way we don't install software in non-desired location and we don't have to worry about cross-platform installation.

.github/workflows/cargo_build.yml
```diff
@@ -23,6 +23,16 @@ jobs:
     steps:
       - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
 
+      - name: Install and run newline linter checks
+        if: matrix.os == 'ubuntu-latest'
```
IceTDrinker, line 27: good thinking

.github/workflows/cargo_build.yml
```diff
@@ -23,6 +23,16 @@ jobs:
...
+        run: |
+          curl -L "https://github.com/fernandrone/linelint/releases/download/0.0.6/linelint-linux-amd64" > linelint
```
IceTDrinker, line 29: isn't wget more suited to download stuff ?

ci/lattice_estimator.sage
```diff
@@ -36,7 +36,7 @@ def check_security(filename):
         try:
             # The lattice estimator is not able to manage such large dimension.
             # If we have the security for smaller `n` then we have security for larger ones.
-            if param.n >= 32768:
+            if param.n >= 16384:
```
IceTDrinker, line 39: `>` is probably better suited here as we are fixing stuff if we are not inside the proper value range

IceTDrinker (commented):
Very small nitpicks otherwise looks good !

Don't forget to make a small announcement to the team so that people can install linelint on their systems (and tell them that `make conformance` is awesome) :)

PR #558, opened 2023-09-15
chore(ci): add compact keys to parameters security checks

soonum:


IceTDrinker (approved):
Efficient, I like this !

ci/lattice_estimator.sage
```diff
@@ -36,7 +36,7 @@ def check_security(filename):
         try:
             # The lattice estimator is not able to manage such large dimension.
             # If we have the security for smaller `n` then we have security for larger ones.
-            if param.n == 32768:
+            if param.n >= 32768:
```
IceTDrinker, line 39: Given what we saw I think

> 16384

Is a better test, don't you think ?

PR #564, opened 2023-09-18
fix(integer): fix is_neg_possible

tmontaigu:
shortint's is_neg_possible did not check the degree on the correct value. It check the degree on the value that should be added to the next block not on the value that actually becomes the degree.

integer's is neg possible had the same problem so we also fix it and also check the next block can 'receive' the value that should be added to it.

Our tests did not catch that as they were not testing non empty carry case


tfhe/src/integer/server_key/radix/neg.rs
```diff
@@ -91,8 +91,24 @@ impl ServerKey {
...
+            let msg_mod = self.key.message_modulus.0;
+            let mut z = (ctxt.blocks[i].degree.0 + msg_mod - 1) / msg_mod;
+            z = z.wrapping_mul(msg_mod);
+
+            // z will be the new degree of ctxt.blocks[i]
+            if z > self.key.max_degree.0 {
+                return false;
+            }
+
+            let z_b = z / self.key.message_modulus.0;
```
IceTDrinker, line 105: `let z_b = z / msg_mod;` ?

IceTDrinker (commented):
Looks good, let me know if you want to update the small thing I saw, otherwise I'll approve to launch the tests

tfhe/src/integer/server_key/radix/neg.rs
tmontaigu: yes would be better will update

IceTDrinker (approved):
Thanks !

PR #568, opened 2023-09-19
chore(deps): update cargo dependencies

mayeul-zama:
Update cargo dependencies

### DIFF
--- a/concrete-csprng/Cargo.toml
+++ b/concrete-csprng/Cargo.toml
@@ -19,8 +19,8 @@ libc = "0.2.133"
 
 [dev-dependencies]
 rand = "0.8.3"
-criterion = "0.3"
-clap = "=4.2.7"
+criterion = "0.5.1"
+clap = "=4.4.4"
 
 [features]
 parallel = ["rayon"]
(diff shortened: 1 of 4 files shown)


concrete-csprng/Cargo.toml
```diff
@@ -19,8 +19,8 @@ libc = "0.2.133"
 
 [dev-dependencies]
 rand = "0.8.3"
-criterion = "0.3"
-clap = "=4.2.7"
+criterion = "0.5.1"
```
IceTDrinker, line 22: what about rayon above and libc ?

tasks/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap = "3.1"
+clap = "4.4.4"
```
IceTDrinker, line 9: please hardcode the version here, it's coming from core

concrete-csprng/Cargo.toml
mayeul-zama: The last versions should be used as they are semver compatible with the specified ones

tasks/Cargo.toml
mayeul-zama: Fixed

mayeul-zama:
@slab-ci cpu_fast_test

IceTDrinker (approved):
Thanks ! Let's wait for the tests :)

tmontaigu (approved):

PR #569, opened 2023-09-20
fix(integer): only propagate if necessary after trimming

tmontaigu:
By unconditionally propagating carries after trimming we would sometimes do work for nothing and as propagating carries not cheap at all it would degrade performances.

So only propagate when necessary


IceTDrinker (approved):
Ah damn that's the thing I missed in the review ?

tmontaigu:
> Ah damn that's the thing I missed in the review ?

Yes, but also means I missed it when writting it 😏

PR #572, opened 2023-09-20
feat(tfhe): add safe deserialiation

mayeul-zama:
Some tests will need to be added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ clippy_trivium: install_rs_check_toolchain
 .PHONY: clippy_all_targets # Run clippy lints on all targets (benches, examples, etc.)
 clippy_all_targets:
 	RUSTFLAGS="$(RUSTFLAGS)" cargo "$(CARGO_RS_CHECK_TOOLCHAIN)" clippy --all-targets \
-		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,integer,internal-keycache \
+		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,integer,internal-keycache,safe-deserialization \
 		-p tfhe -- --no-deps -D warnings
 
 .PHONY: clippy_concrete_csprng # Run clippy lints on concrete-csprng
@@ -376,6 +376,11 @@ test_integer_multi_bit_ci: install_rs_build_toolchain install_cargo_nextest
 		./scripts/integer-tests.sh --rust-toolchain $(CARGO_RS_BUILD_TOOLCHAIN) \
 		--cargo-profile "$(CARGO_PROFILE)" --multi-bit
 
+.PHONY: test_safe_deserialization # Run the tests for safe deserialization
+test_safe_deserialization: install_rs_build_toolchain install_cargo_nextest
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
+		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,integer,internal-keycache,safe-deserialization -p tfhe -- safe_deserialization::
+
 .PHONY: test_integer # Run all the tests for integer
 test_integer: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
@@ -453,7 +458,7 @@ format_doc_latex:
 .PHONY: check_compile_tests # Build tests in debug without running them
 check_compile_tests:
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --no-run \
-		--features=$(TARGET_ARCH_FEATURE),experimental,boolean,shortint,integer,internal-keycache \
+		--features=$(TARGET_ARCH_FEATURE),experimental,boolean,shortint,integer,internal-keycache,safe-deserialization \
 		-p tfhe
 
 	@if [[ "$(OS)" == "Linux" || "$(OS)" == "Darwin" ]]; then \
--- a/tfhe/Cargo.toml
+++ b/tfhe/Cargo.toml
@@ -82,14 +82,15 @@ bytemuck = "1.13.1"
 boolean = ["dep:paste"]
 shortint = ["dep:paste"]
 integer = ["shortint", "dep:paste"]
-internal-keycache = ["lazy_static", "dep:fs2", "bincode", "dep:paste"]
+internal-keycache = ["lazy_static", "dep:fs2", "dep:bincode", "dep:paste"]
+safe-deserialization = ["dep:bincode"]
 
 # Experimental section
 experimental = []
 experimental-force_fft_algo_dif4 = []
 # End experimental section
 
-__c_api = ["cbindgen", "bincode", "dep:paste"]
+__c_api = ["cbindgen", "dep:bincode", "dep:paste"]
 boolean-c-api = ["boolean", "__c_api"]
 shortint-c-api = ["shortint", "__c_api"]
 high-level-c-api = ["boolean-c-api", "shortint-c-api", "integer", "__c_api"]
@@ -101,7 +102,7 @@ __wasm_api = [
     "serde-wasm-bindgen",
     "getrandom",
     "getrandom/js",
-    "bincode",
+    "dep:bincode",
 ]
 boolean-client-js-wasm-api = ["boolean", "__wasm_api"]
 shortint-client-js-wasm-api = ["shortint", "__wasm_api"]
--- a/tfhe/src/conformance.rs
+++ b/tfhe/src/conformance.rs
@@ -4,3 +4,67 @@ pub trait ParameterSetConformant {
 
     fn is_conformant(&self, parameter_set: &Self::ParameterSet) -> bool;
 }
+
+/// A constraint on a list size
+/// The list must be composed of a number `n` of groups of size `group_size` which means list size
+/// must be a multiple of `group_size`.
+/// Moreover, `n` must be:
+/// - bigger or equal to `min_inclusive_group_count`
+/// - smaller of equal to `max_inclusive_group_count`
+#[derive(Copy, Clone)]
+pub struct ListSizeConstraint {
+    min_inclusive_group_count: usize,
+    max_inclusive_group_count: usize,
+    group_size: usize,
+}
+
+impl ListSizeConstraint {
+    pub fn exact_size(size: usize) -> ListSizeConstraint {
+        ListSizeConstraint {
+            min_inclusive_group_count: size,
+            max_inclusive_group_count: size,
+            group_size: 1,
+        }
+    }
+    pub fn try_size_in_range(
+        min_inclusive: usize,
+        max_inclusive: usize,
+    ) -> Result<ListSizeConstraint, String> {
+        if max_inclusive < min_inclusive {
+            return Err("max_inclusive < min_inclusive".to_owned());
+        }
+        Ok(ListSizeConstraint {
+            min_inclusive_group_count: min_inclusive,
+            max_inclusive_group_count: max_inclusive,
+            group_size: 1,
+        })
+    }
+    pub fn try_size_of_group_in_range(
+        group_size: usize,
+        min_inclusive_group_count: usize,
+        max_inclusive_group_count: usize,
+    ) -> Result<ListSizeConstraint, String> {
+        if max_inclusive_group_count < min_inclusive_group_count {
+            return Err("max_inclusive < min_inclusive".to_owned());
+        }
+        Ok(ListSizeConstraint {
+            min_inclusive_group_count,
+            max_inclusive_group_count,
+            group_size,
+        })
+    }
+
+    pub fn multiply_group_size(&self, group_size_multiplier: usize) -> Self {
+        ListSizeConstraint {
+            min_inclusive_group_count: self.min_inclusive_group_count,
+            max_inclusive_group_count: self.max_inclusive_group_count,
+            group_size: self.group_size * group_size_multiplier,
+        }
+    }
+
+    pub fn is_valid(&self, size: usize) -> bool {
+        size % self.group_size == 0
+            && size >= self.min_inclusive_group_count * self.group_size
+            && size <= self.max_inclusive_group_count * self.group_size
+    }
+}
--- a/tfhe/src/conformance.rs
+++ b/tfhe/src/conformance.rs
@@ -1,0 +1,6 @@
+/// A trait for objects which can be checked to be conformant with a parameter set
+pub trait ParameterSetConformant {
+    type ParameterSet;
+
+    fn is_conformant(&self, parameter_set: &Self::ParameterSet) -> bool;
+}
--- a/tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
+++ b/tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
@@ -97,13 +97,18 @@ pub fn extract_lwe_sample_from_glwe_ciphertext<Scalar, InputCont, OutputCont>(
     InputCont: Container<Element = Scalar>,
     OutputCont: ContainerMut<Element = Scalar>,
 {
-    assert!(
-        input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0
-            == output_lwe.lwe_size().to_lwe_dimension().0,
+    let in_lwe_dim = input_glwe
+        .glwe_size()
+        .to_glwe_dimension()
+        .to_equivalent_lwe_dimension(input_glwe.polynomial_size());
+
+    let out_lwe_dim = output_lwe.lwe_size().to_lwe_dimension();
+
+    assert_eq!(
+        in_lwe_dim, out_lwe_dim,
         "Mismatch between equivalent LweDimension of input ciphertext and output ciphertext. \
         Got {:?} for input and {:?} for output.",
-        LweDimension(input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0),
-        output_lwe.lwe_size().to_lwe_dimension(),
+        in_lwe_dim, out_lwe_dim,
     );
 
     assert_eq!(
@@ -354,13 +359,18 @@ pub fn par_extract_lwe_sample_from_glwe_ciphertext_with_thread_count<
     InputCont: Container<Element = Scalar> + Sync,
     OutputCont: ContainerMut<Element = Scalar>,
 {
-    assert!(
-        input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0
-            == output_lwe_list.lwe_size().to_lwe_dimension().0,
+    let in_lwe_dim = input_glwe
+        .glwe_size()
+        .to_glwe_dimension()
+        .to_equivalent_lwe_dimension(input_glwe.polynomial_size());
+
+    let out_lwe_dim = output_lwe_list.lwe_size().to_lwe_dimension();
+
+    assert_eq!(
+        in_lwe_dim, out_lwe_dim,
         "Mismatch between equivalent LweDimension of input ciphertext and output ciphertext. \
         Got {:?} for input and {:?} for output.",
-        LweDimension(input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0),
-        output_lwe_list.lwe_size().to_lwe_dimension(),
+        in_lwe_dim, out_lwe_dim,
     );
 
     assert!(
--- a/tfhe/src/core_crypto/algorithms/misc.rs
+++ b/tfhe/src/core_crypto/algorithms/misc.rs
@@ -108,6 +108,30 @@ pub fn torus_modular_diff<T: UnsignedInteger>(
     }
 }
 
+// Our representation of non native power of 2 moduli puts the information in the MSBs and leaves
+// the LSBs empty, this is what this function is checking
+pub fn check_content_respects_mod<Scalar: UnsignedInteger, Input: AsRef<[Scalar]>>(
+    input: &Input,
+    modulus: CiphertextModulus<Scalar>,
+) -> bool {
+    if modulus.is_native_modulus() {
+        true
+    } else if modulus.is_power_of_two() {
+        // If our modulus is 2^60, the scaling is 2^4 = 00...00010000, minus 1 = 00...00001111
+        // we want the bits under the mask to be 0
+        let power_2_diff_mask = modulus.get_power_of_two_scaling_to_native_torus() - Scalar::ONE;
+        input
+            .as_ref()
+            .iter()
+            .all(|&x| (x & power_2_diff_mask) == Scalar::ZERO)
+    } else {
+        // non native, not power of two
+        let scalar_modulus: Scalar = modulus.get_custom_modulus().cast_into();
+
+        input.as_ref().iter().all(|&x| x < scalar_modulus)
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
--- a/tfhe/src/core_crypto/commons/parameters.rs
+++ b/tfhe/src/core_crypto/commons/parameters.rs
@@ -108,6 +108,10 @@ impl GlweDimension {
     pub fn to_glwe_size(&self) -> GlweSize {
         GlweSize(self.0 + 1)
     }
+
+    pub fn to_equivalent_lwe_dimension(self, poly_size: PolynomialSize) -> LweDimension {
+        LweDimension(self.0 * poly_size.0)
+    }
 }
 
 /// The number of coefficients of a polynomial.
--- a/tfhe/src/core_crypto/entities/glwe_ciphertext.rs
+++ b/tfhe/src/core_crypto/entities/glwe_ciphertext.rs
@@ -116,7 +116,12 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> GlweMask<C> {
     /// let ciphertext_modulus = CiphertextModulus::new_native();
     ///
     /// let glwe_mask = GlweMask::from_container(
-    ///     vec![0u64; glwe_dimension.0 * polynomial_size.0],
+    ///     vec![
+    ///         0u64;
+    ///         glwe_dimension
+    ///             .to_equivalent_lwe_dimension(polynomial_size)
+    ///             .0
+    ///     ],
     ///     polynomial_size,
     ///     ciphertext_modulus,
     /// );
@@ -215,7 +220,9 @@ pub fn glwe_ciphertext_mask_size(
     glwe_dimension: GlweDimension,
     polynomial_size: PolynomialSize,
 ) -> usize {
-    glwe_dimension.0 * polynomial_size.0
+    glwe_dimension
+        .to_equivalent_lwe_dimension(polynomial_size)
+        .0
 }
 
 /// A [`GLWE ciphertext`](`GlweCiphertext`).
--- a/tfhe/src/core_crypto/entities/glwe_secret_key.rs
+++ b/tfhe/src/core_crypto/entities/glwe_secret_key.rs
@@ -153,7 +153,12 @@ where
         polynomial_size: PolynomialSize,
     ) -> GlweSecretKeyOwned<Scalar> {
         GlweSecretKeyOwned::from_container(
-            vec![value; glwe_dimension.0 * polynomial_size.0],
+            vec![
+                value;
+                glwe_dimension
+                    .to_equivalent_lwe_dimension(polynomial_size)
+                    .0
+            ],
             polynomial_size,
         )
     }
--- a/tfhe/src/core_crypto/entities/lwe_ciphertext.rs
+++ b/tfhe/src/core_crypto/entities/lwe_ciphertext.rs
@@ -1,7 +1,9 @@
 //! Module containing the definition of the [`LweCiphertext`].
 
+use crate::conformance::ParameterSetConformant;
 use crate::core_crypto::commons::parameters::*;
 use crate::core_crypto::commons::traits::*;
+use crate::core_crypto::prelude::misc::check_content_respects_mod;
 
 /// A convenience structure to easily manipulate the body of an [`LweCiphertext`].
 #[derive(Clone, Debug)]
@@ -687,6 +689,27 @@ pub type LweCiphertextOwned<Scalar> = LweCiphertext<Vec<Scalar>>;
 pub type LweCiphertextView<'data, Scalar> = LweCiphertext<&'data [Scalar]>;
 /// An [`LweCiphertext`] mutably borrowing memory for its own storage.
 pub type LweCiphertextMutView<'data, Scalar> = LweCiphertext<&'data mut [Scalar]>;
+
+/// Structure to store the expected properties of a ciphertext
+/// Can be used on a server to check if client inputs are well formed
+/// before running a computation on them
+pub struct LweCiphertextParameters<T: UnsignedInteger> {
+    pub lwe_dim: LweDimension,
+    pub ct_modulus: CiphertextModulus<T>,
+}
+
+impl<C: Container> ParameterSetConformant for LweCiphertext<C>
+where
+    C::Element: UnsignedInteger,
+{
+    type ParameterSet = LweCiphertextParameters<C::Element>;
+
+    fn is_conformant(&self, lwe_ct_parameters: &LweCiphertextParameters<C::Element>) -> bool {
+        check_content_respects_mod(self, lwe_ct_parameters.ct_modulus)
+            && self.lwe_size() == lwe_ct_parameters.lwe_dim.to_lwe_size()
+            && self.ciphertext_modulus() == lwe_ct_parameters.ct_modulus
+    }
+}
 
 impl<Scalar: UnsignedInteger> LweCiphertextOwned<Scalar> {
     /// Allocate memory and create a new owned [`LweCiphertext`].
--- a/tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
+++ b/tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
@@ -1,12 +1,14 @@
 //! Module containing the definition of the [`LweCompactCiphertextList`] a space efficient
 //! encryption of a list of LWE ciphertexts.
 
+use crate::conformance::{ListSizeConstraint, ParameterSetConformant};
 use crate::core_crypto::algorithms::{
     expand_lwe_compact_ciphertext_list, par_expand_lwe_compact_ciphertext_list,
 };
 use crate::core_crypto::commons::parameters::*;
 use crate::core_crypto::commons::traits::*;
 use crate::core_crypto::entities::*;
+use crate::core_crypto::prelude::misc::check_content_respects_mod;
 
 /// A [`compact list of LWE ciphertexts`](`LweCompactCiphertextList`) obtained through encryption
 /// with a [`compact LWE public key`](`super::LweCompactPublicKey`).
@@ -249,6 +251,27 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCompactCipherte
     }
 }
 
+// These accessors are used to create invalid objects and test the conformance functions
+// But these functions should not be used in other contexts, hence the `#[cfg(test)]`
+#[cfg(test)]
+impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCompactCiphertextList<C> {
+    pub fn get_mut_lwe_size(&mut self) -> &mut LweSize {
+        &mut self.lwe_size
+    }
+
+    pub fn get_mut_lwe_ciphertext_count(&mut self) -> &mut LweCiphertextCount {
+        &mut self.lwe_ciphertext_count
+    }
+
+    pub fn get_mut_ciphertext_modulus(&mut self) -> &mut CiphertextModulus<Scalar> {
+        &mut self.ciphertext_modulus
+    }
+
+    pub fn get_mut_container(&mut self) -> &mut C {
+        &mut self.data
+    }
+}
+
 impl<Scalar: UnsignedInteger, C: ContainerMut<Element = Scalar>> LweCompactCiphertextList<C> {
     /// Return a mutable view to the [`LweMaskList`] of a [`LweCompactCiphertextList`].
     pub fn get_mut_mask_list(&mut self) -> LweMaskListMutView<'_, Scalar> {
@@ -301,6 +324,33 @@ pub type LweCompactCiphertextListOwned<Scalar> = LweCompactCiphertextList<Vec<Sc
 pub type LweCompactCiphertextListView<'data, Scalar> = LweCompactCiphertextList<&'data [Scalar]>;
 pub type LweCompactCiphertextListMutView<'data, Scalar> =
     LweCompactCiphertextList<&'data mut [Scalar]>;
+
+/// Structure to store the expected properties of a ciphertext
+/// Can be used on a server to check if client inputs are well formed
+/// before running a computation on them
+pub struct LweCiphertextListParameters<T: UnsignedInteger> {
+    pub lwe_dim: LweDimension,
+    pub lwe_ciphertext_count_constraint: ListSizeConstraint,
+    pub ct_modulus: CiphertextModulus<T>,
+}
+
+impl<T: UnsignedInteger> ParameterSetConformant for LweCompactCiphertextListOwned<T> {
+    type ParameterSet = LweCiphertextListParameters<T>;
+
+    fn is_conformant(&self, param: &LweCiphertextListParameters<T>) -> bool {
+        param
+            .lwe_ciphertext_count_constraint
+            .is_valid(self.lwe_ciphertext_count.0)
+            && self.data.len()
+                == lwe_compact_ciphertext_list_size(
+                    self.lwe_size.to_lwe_dimension(),
+                    self.lwe_ciphertext_count,
+                )
+            && check_content_respects_mod(self, param.ct_modulus)
+            && self.lwe_size == param.lwe_dim.to_lwe_size()
+            && self.ciphertext_modulus == param.ct_modulus
+    }
+}
 
 impl<Scalar: UnsignedInteger> LweCompactCiphertextListOwned<Scalar> {
     /// Allocate memory and create a new owned [`LweCompactCiphertextListOwned`].
--- a/tfhe/src/high_level_api/integers/types/compact.rs
+++ b/tfhe/src/high_level_api/integers/types/compact.rs
@@ -5,6 +5,7 @@ use crate::high_level_api::integers::types::base::GenericInteger;
 use crate::high_level_api::internal_traits::TypeIdentifier;
 use crate::high_level_api::traits::FheTryEncrypt;
 use crate::integer::ciphertext::{CiphertextListConformanceParams, CompactCiphertextList};
+use crate::named::Named;
 use crate::shortint::PBSParameters;
 use crate::CompactPublicKey;
 
@@ -103,6 +104,14 @@ impl<P: IntegerParameter> ParameterSetConformant for GenericCompactInteger<P> {
         };
         self.list.is_conformant(&params)
     }
+}
+
+impl<P: IntegerParameter> Named for GenericCompactInteger<P> {
+    const NAME: &'static str = "high_level_api::GenericCompactInteger";
+}
+
+impl<P: IntegerParameter> Named for GenericCompactIntegerList<P> {
+    const NAME: &'static str = "high_level_api::GenericCompactIntegerList";
 }
 
 impl<P: IntegerParameter> ParameterSetConformant for GenericCompactIntegerList<P> {
--- a/tfhe/src/high_level_api/integers/types/compact.rs
+++ b/tfhe/src/high_level_api/integers/types/compact.rs
@@ -1,9 +1,11 @@
+use crate::conformance::{ListSizeConstraint, ParameterSetConformant};
 use crate::errors::{UninitializedPublicKey, UnwrapResultExt};
 use crate::high_level_api::integers::parameters::IntegerParameter;
 use crate::high_level_api::integers::types::base::GenericInteger;
 use crate::high_level_api::internal_traits::TypeIdentifier;
 use crate::high_level_api::traits::FheTryEncrypt;
-use crate::integer::ciphertext::CompactCiphertextList;
+use crate::integer::ciphertext::{CiphertextListConformanceParams, CompactCiphertextList};
+use crate::shortint::PBSParameters;
 use crate::CompactPublicKey;
 
 #[cfg_attr(all(doc, not(doctest)), doc(cfg(feature = "integer")))]
@@ -88,5 +90,32 @@ where
             list: ciphertext,
             id,
         })
+    }
+}
+
+impl<P: IntegerParameter> ParameterSetConformant for GenericCompactInteger<P> {
+    type ParameterSet = PBSParameters;
+    fn is_conformant(&self, pbs_parameters: &PBSParameters) -> bool {
+        let params = CiphertextListConformanceParams {
+            shortint_params: pbs_parameters.to_shortint_conformance_param(),
+            num_blocks_per_integer: P::num_blocks(),
+            num_integers_constraint: ListSizeConstraint::exact_size(1),
+        };
+        self.list.is_conformant(&params)
+    }
+}
+
+impl<P: IntegerParameter> ParameterSetConformant for GenericCompactIntegerList<P> {
+    type ParameterSet = (PBSParameters, ListSizeConstraint);
+    fn is_conformant(
+        &self,
+        (pbs_parameters, number_integers_constraint): &(PBSParameters, ListSizeConstraint),
+    ) -> bool {
+        let params = CiphertextListConformanceParams {
+            shortint_params: pbs_parameters.to_shortint_conformance_param(),
+            num_blocks_per_integer: P::num_blocks(),
+            num_integers_constraint: *number_integers_constraint,
+        };
+        self.list.is_conformant(&params)
     }
 }
--- a/tfhe/src/high_level_api/integers/types/compact.rs
+++ b/tfhe/src/high_level_api/integers/types/compact.rs
@@ -119,3 +119,240 @@ impl<P: IntegerParameter> ParameterSetConformant for GenericCompactIntegerList<P
         self.list.is_conformant(&params)
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::conformance::ParameterSetConformant;
+    use crate::core_crypto::prelude::UnsignedInteger;
+    use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+    use crate::shortint::{CiphertextModulus, PBSOrder};
+    use crate::{
+        generate_keys, set_server_key, CompactFheUint8, CompactFheUint8List, ConfigBuilder,
+    };
+    use rand::{thread_rng, Rng};
+
+    type ParameterAccessor<Ct, T> = dyn Fn(&mut Ct) -> &mut T;
+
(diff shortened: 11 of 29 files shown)


tfhe/Cargo.toml
```diff
@@ -83,6 +83,7 @@ boolean = ["dep:paste"]
 shortint = ["dep:paste"]
 integer = ["shortint", "dep:paste"]
 internal-keycache = ["lazy_static", "dep:fs2", "bincode", "dep:paste"]
+safe_serialization = ["bincode"]
```
IceTDrinker, line 86: let's try to start using the `dep:` pattern for enabling dependencies, as specifying "bincode" also specifies the "bincode" feature IIRC, let's plan an issue PR to update the pattern, let's not do it in this PR, this is not the topic here :)

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,149 @@
...
+    Ok(())
+}
+
+pub fn safe_deserialize<T: DeserializeOwned>(
+    mut reader: impl std::io::Read,
+    serialized_size_limit: u64,
+) -> Result<T, String> {
+    let options = bincode::DefaultOptions::new()
+        .with_fixint_encoding()
+        .with_limit(0);
```
IceTDrinker, line 41: set an upper limit at all times for the deserialize to avoid OOM, if we forget to call a with limit we may have some problems

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +85,49 @@ pub struct Ciphertext {
...
+
+                message_modulus = param.message_modulus;
+                ciphertext_modulus = param.ciphertext_modulus;
+                carry_modulus = param.carry_modulus
+            }
+        };
+
+        let ct_len = self.ct.as_ref().len();
+
+        ct_len == expected_dim + 1
```
IceTDrinker, line 124: I would prefer using the LweSize newtype

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +85,49 @@ pub struct Ciphertext {
...
+    type ParameterSet = super::PBSParameters;
+
+    fn conformant(&self, param: &Self::ParameterSet) -> bool {
+        let (expected_dim, message_modulus, carry_modulus, ciphertext_modulus);
+
+        match param {
+            super::PBSParameters::PBS(param) => {
+                expected_dim = match self.pbs_order {
+                    PBSOrder::KeyswitchBootstrap => {
+                        param.glwe_dimension.0 * param.polynomial_size.0
```
IceTDrinker, line 99: LweDimenstion(param.glwe_dimension.0 * param.polynomial_size.0).to_lwe_size(), similar with the other pieces of code

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: or even use a lwe_ciphertext_size function (which probably does not exist) as we have size functions for other more complex ciphertext

tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -110,6 +110,14 @@ impl GlweDimension {
...
 }
 
+impl std::ops::Mul<PolynomialSize> for GlweDimension {
+    type Output = LweDimension;
+
+    fn mul(self, poly_size: PolynomialSize) -> LweDimension {
+        LweDimension(self.0 * poly_size.0)
+    }
+}
+
```
IceTDrinker, line 120: Makes it too easy to escape the newtypes, I would prefer

LweDimension::from_glwe_parameters(glwe_dimension, polynomial_size) in that case

tfhe/src/conformance.rs
```diff
@@ -0,0 +1,5 @@
+pub trait ParameterSetConformant {
```
IceTDrinker, line 1: not sure how to structure the modules for things that are useful crate-wide

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let deserialized_version: String = options
+        .with_limit(VERSION_LENGTH_LIMIT)
+        .deserialize_from::<_, String>(&mut reader)
+        .map_err(|err| err.to_string())?;
+
+    if deserialized_version != SERIALIZATION_VERSION {
+        return Err(format!(
+            "Expected version {}, got version {}",
```
IceTDrinker, line 50: expected serialization version

maybe ?

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    let deserialized_type: String = options
+        .with_limit(TYPE_NAME_LENGTH_LIMIT)
+        .deserialize_from::<_, String>(&mut reader)
+        .map_err(|err| err.to_string())?;
+
+    if deserialized_type != T::NAME {
+        return Err(format!(
+            "Expected type {}, got type {}",
```
IceTDrinker, line 62: add somewhere that this is during serialization

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,111 @@
...
+        .with_limit(serialized_size_limit)
+        .deserialize_from(&mut reader)
+        .map_err(|err| err.to_string())
+}
+
+pub trait ParameterSetConformant {
+    type ParameterSet;
+
+    fn is_conformant(&self, param: &Self::ParameterSet) -> bool;
+}
```
IceTDrinker, line 78: redefined trait

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub fn safe_deserialize_conformant<T: DeserializeOwned + Named + ParameterSetConformant>(
```
IceTDrinker, line 80: consider adding #[track_caller] so that the panic/error shows where in the calling code this error is originating

tfhe/src/shortint/server_key/tests/shortint.rs
```diff
@@ -396,6 +401,49 @@ where
     assert_eq!(0, failures);
 }
 
+#[cfg(feature = "safe_serialization")]
+fn shortint_keyswitch_bootstrap_serialization_safe<P>(param: P)
```
IceTDrinker, line 405: not sure what we are testing here after the serialization is done

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -60,6 +61,12 @@ pub fn lwe_compact_ciphertext_list_size(
     mask_count.0 * lwe_dimension.0 + lwe_ciphertext_count.0
 }
 
+impl<Scalar: UnsignedInteger> LweCompactCiphertextList<Vec<Scalar>> {
+    pub fn get_mut_container(&mut self) -> &mut Vec<Scalar> {
+        &mut self.data
+    }
+}
+
```
IceTDrinker, line 69: use as_mut() or as_ref()

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -302,6 +321,54 @@ pub type LweCompactCiphertextListView<'data, Scalar> = LweCompactCiphertextList<
...
+    pub lwe_dim: LweDimension,
+    pub ct_modulus: CiphertextModulus<T>,
+    pub lwe_ciphertext_count_constraint: ListSizeConstraint,
+}
+
+#[derive(Copy, Clone)]
+pub enum ListSizeConstraint {
+    Exact(usize),
+    Max(usize),
+}
```
IceTDrinker, line 334: in commons::parameters::mod.rs IIRC

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -302,6 +321,54 @@ pub type LweCompactCiphertextListView<'data, Scalar> = LweCompactCiphertextList<
...
+impl std::ops::Mul<usize> for ListSizeConstraint {
+    type Output = ListSizeConstraint;
+
+    fn mul(self, scale: usize) -> ListSizeConstraint {
+        match self {
+            ListSizeConstraint::Exact(a) => ListSizeConstraint::Exact(a * scale),
+            ListSizeConstraint::Max(a) => ListSizeConstraint::Max(a * scale),
+        }
+    }
+}
```
IceTDrinker, line 354: disagree with the ops approach, prefer a from function

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +88,18 @@ pub struct Ciphertext {
...
 
+impl ParameterSetConformant for Ciphertext {
+    type ParameterSet = ShortintParams;
+
+    fn is_conformant(&self, param: &ShortintParams) -> bool {
+        self.ct.as_view().is_conformant(&param.ct)
+            && self.message_modulus == param.message_modulus
+            && self.carry_modulus == param.carry_modulus
+            && self.pbs_order == param.pbs_order
+            && self.degree.0 == self.message_modulus.0 - 1
```
IceTDrinker, line 103: prefer param on the right here to avoid mistakes

tfhe/src/shortint/parameters/mod.rs
```diff
@@ -118,6 +121,84 @@ pub enum PBSParameters {
     MultiBitPBS(MultiBitPBSParameters),
 }
 
+pub struct ShortintParams {
+    pub ct: LweCtParameters<u64>,
```
IceTDrinker, line 125: ct_params

tfhe/src/shortint/parameters/mod.rs
```diff
@@ -118,6 +121,84 @@ pub enum PBSParameters {
...
+    pub message_modulus: MessageModulus,
+    pub carry_modulus: CarryModulus,
+    pub pbs_order: PBSOrder,
+}
+
+pub struct ShortintListParams {
+    pub ct_list: LweCtListParameters<u64>,
```
IceTDrinker, line 132: ct_list_params

tfhe/src/shortint/parameters/mod.rs
```diff
@@ -118,6 +121,84 @@ pub enum PBSParameters {
...
+    pub message_modulus: MessageModulus,
+    pub carry_modulus: CarryModulus,
+    pub pbs_order: PBSOrder,
+}
+
+impl ShortintParams {
+    pub fn to_list(&self, list_constraint: ListSizeConstraint) -> ShortintListParams {
```
IceTDrinker, line 139: to_shortint_ct_list_conformance_parameters or something equivalent

tfhe/src/shortint/parameters/mod.rs
```diff
@@ -118,6 +121,84 @@ pub enum PBSParameters {
...
+            },
+            message_modulus: self.message_modulus,
+            carry_modulus: self.carry_modulus,
+            pbs_order: self.pbs_order,
+        }
+    }
+}
+
+impl PBSParameters {
+    pub fn to_shortint_param(&self) -> ShortintParams {
```
IceTDrinker, line 154: to_conformance... you get the idea

tfhe/src/integer/ciphertext/mod.rs
```diff
@@ -43,6 +46,29 @@ pub struct CompactCiphertextList {
     pub(crate) num_blocks: usize,
 }
 
+pub struct IntegerListParams {
```
IceTDrinker, line 49: Conformance Parameters etc.

IceTDrinker (commented):
thanks !

tfhe/src/core_crypto/commons/parameters.rs
IceTDrinker: Or maybe better

On GlweDimension impl

fn to_equivalent_lwe_dimension(self, polynomial_size) -> LweDimension { mul code }

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -302,6 +321,54 @@ pub type LweCompactCiphertextListView<'data, Scalar> = LweCompactCiphertextList<
...
+pub struct LweCtListParameters<T: UnsignedInteger> {
+    pub lwe_dim: LweDimension,
+    pub ct_modulus: CiphertextModulus<T>,
+    pub lwe_ciphertext_count_constraint: ListSizeConstraint,
+}
+
+#[derive(Copy, Clone)]
+pub enum ListSizeConstraint {
+    Exact(usize),
+    Max(usize),
```
mayeul-zama, line 333: Add `divisor` to Max variant

tfhe/src/safe_serialization.rs
mayeul-zama: The function is supposed to return errors, not `panic`
Isn't `track_caller` mostly useful on `panic`s?

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
mayeul-zama: We need to mutably access the `Vec` in the tests to change its length and ensure it makes it non conformant

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -169,6 +176,18 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCompactCipherte
...
+    }
+
+    pub fn get_mut_lwe_ciphertext_count(&mut self) -> &mut LweCiphertextCount {
+        &mut self.lwe_ciphertext_count
+    }
+
+    pub fn get_mut_ciphertext_modulus(&mut self) -> &mut CiphertextModulus<Scalar> {
+        &mut self.ciphertext_modulus
+    }
+
```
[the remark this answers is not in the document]
mayeul-zama: IMO `*a.get_mut_property() += 1` is easier to read than `a.set_property(a.get_property() + 1)`

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -302,6 +321,54 @@ pub type LweCompactCiphertextListView<'data, Scalar> = LweCompactCiphertextList<
 pub type LweCompactCiphertextListMutView<'data, Scalar> =
     LweCompactCiphertextList<&'data mut [Scalar]>;
 
+pub struct LweCtListParameters<T: UnsignedInteger> {
+    pub lwe_dim: LweDimension,
+    pub ct_modulus: CiphertextModulus<T>,
+    pub lwe_ciphertext_count_constraint: ListSizeConstraint,
```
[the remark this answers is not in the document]
mayeul-zama: I don't understand what you mean

tfhe/src/safe_serialization.rs
IceTDrinker: ah yes my bad, don't know why I thought we might panic in there

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
IceTDrinker: then prefer a view ?

IceTDrinker: not being able to modify a container in an entity is a conscious choice so that once you went through from_container you know the entity is valid

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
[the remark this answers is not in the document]
IceTDrinker: btw I'm thinking this is not compliant with the fact an entity is always valid, for your test you will have to prefer from_container or only enable the setters for tests

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
mayeul-zama: Put behind #[cfg(test)]

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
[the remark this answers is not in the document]
mayeul-zama: Put behind #[cfg(test)]

tfhe/src/conformance.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Moreover, `n` must be smaller of equal to `max_group_count`
+#[derive(Copy, Clone)]
+pub struct ListSizeConstraint {
+    pub max_group_count: usize,
+    pub group_size: usize,
+}
+
+impl ListSizeConstraint {
+    pub fn is_valid(&self, size: usize) -> bool {
+        0 < size && size <= self.max_group_count * self.group_size && size % self.group_size == 0
```
IceTDrinker, line 20: maybe indicate in the documentation that a list here cannot be empty as we check `0 < size`

tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
```diff
@@ -98,11 +98,17 @@ pub fn extract_lwe_sample_from_glwe_ciphertext<Scalar, InputCont, OutputCont>(
...
+            .to_glwe_dimension()
+            .to_equivalent_lwe_dimension(input_glwe.polynomial_size())
+            == output_lwe.lwe_size().to_lwe_dimension(),
         "Mismatch between equivalent LweDimension of input ciphertext and output ciphertext. \
         Got {:?} for input and {:?} for output.",
-        LweDimension(input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0),
+        input_glwe
+            .glwe_size()
+            .to_glwe_dimension()
+            .to_equivalent_lwe_dimension(input_glwe.polynomial_size()),
```
IceTDrinker, line 111: store in a temporary to avoid the duplication of that heavy expression ?

tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
```diff
@@ -355,11 +361,17 @@ pub fn par_extract_lwe_sample_from_glwe_ciphertext_with_thread_count<
     OutputCont: ContainerMut<Element = Scalar>,
 {
     assert!(
-        input_glwe.glwe_size().to_glwe_dimension().0 * input_glwe.polynomial_size().0
-            == output_lwe_list.lwe_size().to_lwe_dimension().0,
+        input_glwe
+            .glwe_size()
+            .to_glwe_dimension()
+            .to_equivalent_lwe_dimension(input_glwe.polynomial_size())
+            == output_lwe_list.lwe_size().to_lwe_dimension(),
```
IceTDrinker, line 368: same thing

tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -110,6 +110,12 @@ impl GlweDimension {
     }
 }
 
+impl GlweDimension {
```
IceTDrinker, line 113: we probably already have an impl block for this struct

tfhe/src/core_crypto/entities/glwe_ciphertext.rs
```diff
@@ -116,7 +116,7 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> GlweMask<C> {
     /// let ciphertext_modulus = CiphertextModulus::new_native();
     ///
     /// let glwe_mask = GlweMask::from_container(
-    ///     vec![0u64; glwe_dimension.0 * polynomial_size.0],
+    ///     vec![0u64; (glwe_dimension.to_equivalent_lwe_dimension(polynomial_size)).0],
```
IceTDrinker, line 119: useless parens ?

tfhe/src/core_crypto/entities/glwe_ciphertext.rs
```diff
@@ -215,7 +215,7 @@ pub fn glwe_ciphertext_mask_size(
     glwe_dimension: GlweDimension,
     polynomial_size: PolynomialSize,
 ) -> usize {
-    glwe_dimension.0 * polynomial_size.0
+    (glwe_dimension.to_equivalent_lwe_dimension(polynomial_size)).0
```
IceTDrinker, line 218: usesless parens ?

tfhe/src/core_crypto/entities/glwe_secret_key.rs
```diff
@@ -153,7 +153,7 @@ where
         polynomial_size: PolynomialSize,
     ) -> GlweSecretKeyOwned<Scalar> {
         GlweSecretKeyOwned::from_container(
-            vec![value; glwe_dimension.0 * polynomial_size.0],
+            vec![value; (glwe_dimension.to_equivalent_lwe_dimension(polynomial_size)).0],
```
IceTDrinker, line 156: useless parens ?

tfhe/src/core_crypto/entities/lwe_ciphertext.rs
```diff
@@ -688,6 +689,23 @@ pub type LweCiphertextView<'data, Scalar> = LweCiphertext<&'data [Scalar]>;
...
+impl<C: Container> ParameterSetConformant for LweCiphertext<C>
+where
+    C::Element: UnsignedInteger,
+{
+    type ParameterSet = LweCtParameters<C::Element>;
+
+    fn is_conformant(&self, lwe_ct_parameters: &LweCtParameters<C::Element>) -> bool {
+        lwe_ct_parameters.lwe_dim.to_lwe_size() == self.lwe_size()
+            && lwe_ct_parameters.ct_modulus == self.ciphertext_modulus()
+    }
```
IceTDrinker, line 706: forgot to tell you we have a `check_content_respects_mod` in tfhe/src/core_crypto/algorithms/test/mod.rs that we can move to tfhe/src/core_crypto/algorithms/misc.rs

and check that the content in the container is valid like you can see in https://github.com/zama-ai/tfhe-rs/blob/2e8189514c9e80aa4542c8b795c96f656237edd3/tfhe/src/core_crypto/algorithms/test/lwe_encryption.rs#L189

this only requires the input entity to be as_ref which all our entities are

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
```diff
@@ -249,6 +250,28 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCompactCipherte
...
+        &mut self.lwe_ciphertext_count
+    }
+
+    pub fn get_mut_ciphertext_modulus(&mut self) -> &mut CiphertextModulus<Scalar> {
+        &mut self.ciphertext_modulus
+    }
+}
+
+#[cfg(test)]
+impl<Scalar: UnsignedInteger> LweCompactCiphertextList<Vec<Scalar>> {
```
IceTDrinker, line 269: this could be in the impl block with <C: Container> I believe ?

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,205 @@
...
+            "Deserialized object of type {} not conformant with given parameter set",
+            T::NAME
+        ));
+    }
+
+    Ok(deser)
+}
+
+#[cfg(all(test, feature = "shortint"))]
+mod test {
```
IceTDrinker, line 118: test_shortint

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,205 @@
...
+        )
+        .unwrap();
+
+        let dec = ck.decrypt(&ct2);
+        assert_eq!(msg, dec);
+    }
+}
+
+#[cfg(all(test, feature = "integer"))]
+mod test2 {
```
IceTDrinker, line 157: test_integer

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -158,6 +177,18 @@ pub struct CompressedCiphertext {
...
 
+impl ParameterSetConformant for CompressedCiphertext {
+    type ParameterSet = CiphertextConformanceParams;
+
+    fn is_conformant(&self, param: &CiphertextConformanceParams) -> bool {
+        self.ct.is_conformant(&param.ct_params)
+            && param.message_modulus == self.message_modulus
+            && param.carry_modulus == self.carry_modulus
+            && param.pbs_order == self.pbs_order
+            && self.degree.0 == self.message_modulus.0 - 1
```
IceTDrinker, line 188: use param on the right, should limit mistakes

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -193,6 +224,18 @@ pub struct CompactCiphertextList {
...
 
+impl ParameterSetConformant for CompactCiphertextList {
+    type ParameterSet = CiphertextListConformanceParams;
+
+    fn is_conformant(&self, params: &CiphertextListConformanceParams) -> bool {
+        self.ct_list.is_conformant(&params.ct_list_params)
+            && params.message_modulus == self.message_modulus
+            && params.carry_modulus == self.carry_modulus
+            && params.pbs_order == self.pbs_order
+            && self.degree.0 + 1 == self.message_modulus.0
```
IceTDrinker, line 235: choose a convention for degree and message modulus please 🙂 

I would use

self.degree.0 == params.message_modulus.0 - 1

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +88,22 @@ pub struct Ciphertext {
...
+
+impl ParameterSetConformant for Ciphertext {
+    type ParameterSet = CiphertextConformanceParams;
+
+    fn is_conformant(&self, param: &CiphertextConformanceParams) -> bool {
+        self.ct.as_view().is_conformant(&param.ct_params)
+            && param.message_modulus == self.message_modulus
+            && param.carry_modulus == self.carry_modulus
+            && param.pbs_order == self.pbs_order
+            && self.degree.0 == self.message_modulus.0 - 1
```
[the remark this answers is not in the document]
IceTDrinker: param on the right please to avoid mistakes (not self) for the degree check

also maybe have all self accesses on the left ? or all params accesses, but I would like to avoid mixing

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,284 @@ where
...
+        let params = CtListConformanceParams {
+            shortint_params: pbs_parameters.to_shortint_conformance_param(),
+            num_blocks_per_integer: P::num_blocks(),
+            num_integers_constraint: *number_integers_constraint,
+        };
+        self.list.is_conformant(&params)
+    }
+}
+
+#[cfg(test)]
```
IceTDrinker, line 130: missing the enabled feature for safe deser ?

I think we can enable it by default if shortint or integer are enabled

I guess we could put

#[cfg(any(feature = shortint, feature = integer))] for the safe deserialization module ? this way it's available and bincode is enabled at the same time ?

or keep the current safe_serialization feature but shortint and integer enable that feature automatically

tfhe/src/core_crypto/entities/lwe_ciphertext.rs
```diff
@@ -688,6 +689,23 @@ pub type LweCiphertextView<'data, Scalar> = LweCiphertext<&'data [Scalar]>;
 /// An [`LweCiphertext`] mutably borrowing memory for its own storage.
 pub type LweCiphertextMutView<'data, Scalar> = LweCiphertext<&'data mut [Scalar]>;
 
+pub struct LweCtParameters<T: UnsignedInteger> {
```
IceTDrinker, line 692: Let's avoid Ct contraction and go for the full name LweCiphertextParameters ?

tfhe/src/integer/ciphertext/mod.rs
```diff
@@ -43,6 +45,34 @@ pub struct CompactCiphertextList {
     pub(crate) num_blocks: usize,
 }
 
+pub struct CtListConformanceParams {
```
IceTDrinker, line 48: Ct -> Ciphertext, same wherever applicable

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,284 @@ where
...
+mod test {
+    use super::*;
+    use crate::conformance::ParameterSetConformant;
+    use crate::core_crypto::prelude::UnsignedInteger;
+    use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
```
IceTDrinker, line 135: let's plan adding more tests later down the line using the create_parameterized_test pattern from elsewhere in the crate

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: It panicked on a test with overflow checks

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: I can do it, but I don't understand what it changes in preventing mistakes

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: 🤔 how ? message_modulus has to be 1 or higher 😵‍💫

IceTDrinker: otherwise it makes no sense right ?

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: if someone forgets to check the param.message_modulus == self.message_modulus we are comparing only with data in the ciphertext, but the proposal to create a ConformanceParam struct may allow to make this more easily

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,284 @@ where
...
+
+        type ParameterModifier<'a> = dyn Fn(&mut Ct) + 'a;
+
+        fn change_parameters<T: UnsignedInteger>(
+            func: &ParameterAccessor<T>,
+        ) -> Vec<Box<ParameterModifier<'_>>> {
+            vec![
+                Box::new(|ct| *func(ct) = T::ZERO),
+                Box::new(|ct| *func(ct) += T::ONE),
+                Box::new(|ct| *func(ct) -= T::ONE),
```
IceTDrinker, line 168: wrapping ops to avoid overflow detection ?

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn safe_ser_ct() {
+        let config = ConfigBuilder::all_disabled()
+            .enable_default_integers()
+            .build();
+
+        let (client_key, _server_key) = generate_keys(config);
+
+        let public_key = CompactPublicKey::new(&client_key);
+
+        let msg = 27u8;
```
IceTDrinker, line 178: randomize ?

tfhe/src/conformance.rs
mayeul-zama: Adding `min_group_count`

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: Sure, but we can not assume that the degree is correct here

tfhe/src/safe_serialization.rs
mayeul-zama: I prefer to make tests reproductible
Here, the decrypted value is not really what the test is about

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: I don't really see the benefit to hav this test with many parameter sets (except if they increase coverage, for exemple by using a non native modulus)

tfhe/src/shortint/ciphertext/mod.rs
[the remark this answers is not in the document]
mayeul-zama: Done

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: Anyway, the expected degree is now taken from the params

tfhe/src/lib.rs
```diff
@@ -62,3 +62,12 @@ pub use high_level_api::*;
 /// cbindgen:ignore
 #[cfg(any(test, doctest, feature = "internal-keycache"))]
 pub mod keycache;
+
+#[cfg(feature = "safe_serialization")]
+pub mod safe_serialization;
+
+pub mod conformance;
+
+pub trait Named {
```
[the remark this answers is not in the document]
mayeul-zama: I can create a module `named` but I don't see the benefit as it would only be 3 lines long

tfhe/src/core_crypto/entities/lwe_compact_ciphertext_list.rs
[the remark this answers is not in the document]
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/shortint.rs
mayeul-zama: Removed

tfhe/src/conformance.rs
```diff
@@ -0,0 +1,28 @@
...
+/// The list must be composed of a number `n` of groups of size `group_size` which means list size
+/// must be a multiple of `group_size`.
+/// Moreover, `n` must be:
+/// - bigger or equal to `min_group_count`
+/// - smaller of equal to `max_group_count`
+#[derive(Copy, Clone)]
+pub struct ListSizeConstraint {
+    pub min_group_count: usize,
+
+    pub max_group_count: usize,
```
IceTDrinker, line 18: could we rename those to min_inclusive... and max_inclusive... ? Don't forget the docstring

Makefile
```diff
@@ -376,6 +376,11 @@ test_integer_multi_bit_ci: install_rs_build_toolchain install_cargo_nextest
 		./scripts/integer-tests.sh --rust-toolchain $(CARGO_RS_BUILD_TOOLCHAIN) \
 		--cargo-profile "$(CARGO_PROFILE)" --multi-bit
 
+.PHONY: test_safe_deserialize # Run the tests for safe deserialization
+test_safe_deserialize: install_rs_build_toolchain install_cargo_nextest
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
+		--features=$(TARGET_ARCH_FEATURE),shortint,integer,internal-keycache,safe-deserialization --lib -p tfhe -- safe_deserialization::
```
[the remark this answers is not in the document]
mayeul-zama: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection
It's to avoid compiling every target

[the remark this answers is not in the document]
mayeul-zama: I'll remove it as it does not run doc tests

tfhe/src/conformance.rs
mayeul-zama: Done

tfhe/src/lib.rs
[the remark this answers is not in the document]
IceTDrinker: could it be part of the safe_deserialization ? as it's used by it ?

tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
```diff
@@ -97,14 +97,14 @@ pub fn extract_lwe_sample_from_glwe_ciphertext<Scalar, InputCont, OutputCont>(
...
-        output_lwe.lwe_size().to_lwe_dimension(),
-    );
+    let in_lwe_dim = input_glwe
+        .glwe_size()
+        .to_glwe_dimension()
+        .to_equivalent_lwe_dimension(input_glwe.polynomial_size());
+
+    let out_lwe_dim = output_lwe.lwe_size().to_lwe_dimension();
+
+    assert!(in_lwe_dim == out_lwe_dim);
```
IceTDrinker, line 107: keep the message ? 😅 with just a very small modification to the original code ?

same elsewhere

tfhe/src/core_crypto/algorithms/misc.rs
```diff
@@ -108,6 +108,25 @@ pub fn torus_modular_diff<T: UnsignedInteger>(
     }
 }
 
+// Our representation of non native power of 2 moduli puts the information in the MSBs and leaves
+// the LSBs empty, this is what this function is checking
+pub fn check_content_respects_mod<Scalar: UnsignedInteger, Input: AsRef<[Scalar]>>(
+    input: &Input,
+    modulus: CiphertextModulus<Scalar>,
+) -> bool {
+    if !modulus.is_native_modulus() {
```
IceTDrinker, line 117: I just realized this function only checks power of 2 moduli

the code needs to be slightly adapted (my bad) but this is to avoid unwanted panics/asserts

so you will need 

```rust
if modulus.is_native_modulus() {
    true
} else if modulus.is_power_of_two() {
    // the existing code
} else {
   // non native power of two 
   let scalar_modulus: Scalar = modulus.get_custom_modulus().cast_into();
   return input
            .as_ref()
            .iter()
            .all(|&x| x < scalar_modulus);
}
```

tfhe/src/high_level_api/integers/types/compact.rs
IceTDrinker: this may be still relevant if we don't auto enable the safe deserialization with shortint/integer automatically

tfhe/src/high_level_api/integers/types/compact.rs
IceTDrinker: it's just a matter of completeness :)

plus the macro instantiation of tests is really not that bad

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,273 @@ where
...
+                assert!(!ct_clone.is_conformant(&PARAM_MESSAGE_2_CARRY_2_KS_PBS.into()));
+            }
+        }
+
+        let breakers2: Vec<&ParameterModifier<'_, Ct>> = vec![
+            &|ct: &mut Ct| ct.list.ct_list.pbs_order = PBSOrder::BootstrapKeyswitch,
+            &|ct: &mut Ct| {
+                *ct.list.ct_list.ct_list.get_mut_ciphertext_modulus() =
+                    CiphertextModulus::try_new_power_of_2(1).unwrap();
+            },
```
IceTDrinker, line 199: could we have an odd modulus in addition ?

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,273 @@ where
...
+            &|ct: &mut Ct| {
+                ct.list.ct_list.ct_list.get_mut_container().pop();
+            },
+            &|ct: &mut Ct| ct.list.ct_list.ct_list.get_mut_container().push(0),
+        ];
+
+        for breaker in breakers2 {
```
IceTDrinker, line 206: we are testing each breaker separately, is it possible to use an iproduct! on those breakers maybe ? i.e. some more complicated combination ?

otherwise we can go ahead with a first implementation and then strengthen testing with the composition of several breakers, it may become complicated to break the ciphertext and be sure it's not randomly valid, so yeah maybe not right away

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,273 @@ where
...
+        set_server_key(server_key);
+
+        let ct = CompactFheUint8List::try_encrypt(&[0, 1], &public_key).unwrap();
+
+        let params = (
+            PARAM_MESSAGE_2_CARRY_2_KS_PBS.into(),
+            ListSizeConstraint {
+                min_inclusive_group_count: 1,
+                max_inclusive_group_count: 1,
+                group_size: 2,
```
IceTDrinker, line 337: shouldn't it be group_size : 1 and min/max inclusive set to 2 ?

tfhe/src/high_level_api/integers/types/compact.rs
```diff
@@ -90,3 +92,273 @@ where
...
+                .get_mut_container()
+                .fill_with(|| rng.gen::<u64>());
+
+            assert!(ct_clone.is_conformant(&params));
+
+            let mut ct_clone_expanded = ct_clone.expand();
+
+            for i in &mut ct_clone_expanded {
+                *i += i.clone();
+            }
```
IceTDrinker, line 361: Is this just a check that add operations work ok ? if so a small comment is welcome, I was a bit confused

tfhe/src/safe_deserialization.rs
```diff
@@ -0,0 +1,205 @@
...
+use serde::Serialize;
+
+// The `SERIALIZATION_VERSION` is serialized along objects serialized with `safe_serialize`.
+// This `SERIALIZATION_VERSION` should be changed on each release where any object serialization
+// details changes (this can happen when adding a field or reorderging fields of a struct).
+// When a object is deserialized using `safe_deserialize`, the deserialized version is checked
+// to be equal to SERIALIZATION_VERSION.
+// This helps prevent users from inadvertently deserializaing an object serialized in another
+// release.
+//  When this happens, it also gives a clear version mismatch error rather than a generic
```
IceTDrinker, line 14: Remove first whitespace

tfhe/src/safe_deserialization.rs
```diff
@@ -0,0 +1,205 @@
...
+mod test_integer {
+    use crate::conformance::ParameterSetConformant;
+    use crate::high_level_api::{generate_keys, ConfigBuilder};
+
+    use crate::prelude::{FheDecrypt, FheTryEncrypt};
+    use crate::safe_deserialization::{safe_deserialize_conformant, safe_serialize};
+    use crate::shortint::parameters::{
+        PARAM_MESSAGE_2_CARRY_2_KS_PBS, PARAM_MESSAGE_3_CARRY_3_KS_PBS,
+    };
+    use crate::{CompactFheUint8, CompactPublicKey};
```
IceTDrinker, line 166: can we have maybe the same for the compact list ?

tfhe/src/safe_deserialization.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        assert!(ct.is_conformant(&PARAM_MESSAGE_2_CARRY_2_KS_PBS.into()));
+
+        let mut buffer = vec![];
+
+        safe_serialize(&ct, &mut buffer, 1 << 40).unwrap();
+
+        assert!(safe_deserialize_conformant::<CompactFheUint8>(
+            buffer.as_slice(),
+            1 << 40,
```
IceTDrinker, line 190: can we chose a more reasonable size ? 1 << 40 coud exhaust a system memory and people sometimes resort to tests to find reasonable sizes

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +88,22 @@ pub struct Ciphertext {
...
 
+impl crate::Named for Ciphertext {
+    const NAME: &'static str = "shortint::Ciphertext";
+}
+
+impl ParameterSetConformant for Ciphertext {
+    type ParameterSet = CiphertextConformanceParams;
+
+    fn is_conformant(&self, param: &CiphertextConformanceParams) -> bool {
+        self.ct.as_view().is_conformant(&param.ct_params)
```
IceTDrinker, line 99: why do you need the as_view call ?

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -85,6 +88,22 @@ pub struct Ciphertext {
...
+            && self.message_modulus == param.message_modulus
+            && self.carry_modulus == param.carry_modulus
+            && self.pbs_order == param.pbs_order
+            && self.degree == param.degree
```
IceTDrinker, line 103: this is good enough for our use cas for now (BC encryption) but we should perhaps think of something for a range for the degree ? wdyt ?

IceTDrinker (commented):
Nearly there, we are down to the last details, thanks a lot for all the work up to this point !

tfhe/src/lib.rs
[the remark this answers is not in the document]
mayeul-zama: Sure, but it forces the `impl Named` to be behing a `#[cfg(safe-deserialization)]` which is why I put it out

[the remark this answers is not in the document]
IceTDrinker: hmm how bad does it complicate things ?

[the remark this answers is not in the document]
mayeul-zama: It's not too bad

tfhe/src/core_crypto/algorithms/glwe_sample_extraction.rs
mayeul-zama: Done

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: `conformance` is not part of `safe-deserialization` so it's not feature gated

IceTDrinker: ah right, thanks

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: Removed

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: I think we can add more complex constraints later when needed

tfhe/src/safe_deserialization.rs
mayeul-zama: Changed to 1 << 20

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: Added comment

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: Sure, it's equivalent but makes more sense the way you propose
When there was no `min_inclusive_group_count` this was the only way

mayeul-zama: Changed

tfhe/src/safe_deserialization.rs
mayeul-zama: Done

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: I agree on the principle

But it seems complicated and I don't really see the benefit it would provide

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: follow up issue then please :)

tfhe/src/high_level_api/integers/types/compact.rs
IceTDrinker: Re-opening as I'm worried for corner cases here, where a spec may validate something else but we may not have the right number of blocks ?

Could you check ?

tfhe/src/high_level_api/integers/types/compact.rs
IceTDrinker: I would say a follow up issue would be the way to go there

tfhe/src/high_level_api/integers/types/compact.rs
IceTDrinker: I was worried about the check_content_respects_mod potentially panicking (as it could have before the code change) hence why just having an odd modulus in there allows to be safer

tfhe/src/lib.rs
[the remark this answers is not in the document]
tmontaigu: Would https://doc.rust-lang.org/std/any/fn.type_name.html be an alternative to having a custom `Named` trait ?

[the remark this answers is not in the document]
IceTDrinker: from the page you linked 

`This is intended for diagnostic use. The exact contents and format of the string returned are not specified,`

it was the first solution but did not seem robust

tfhe/src/safe_deserialization.rs
```diff
@@ -0,0 +1,205 @@
...
+        .with_fixint_encoding()
+        .with_limit(0);
+
+    options
+        .with_limit(VERSION_LENGTH_LIMIT)
+        .serialize_into::<_, String>(&mut writer, &SERIALIZATION_VERSION.to_owned())?;
+
+    options
+        .with_limit(TYPE_NAME_LENGTH_LIMIT)
+        .serialize_into::<_, String>(&mut writer, &T::NAME.to_owned())?;
```
tmontaigu, line 45: Do we really need these .to_owned() ? passign the &'static str should work

mayeul-zama: For deserialization, I think we have to use the `String` type
I'm not sure a serialized `&str` is guaranted to be deserializable as a `String` which is why I did it this way

tfhe/src/lib.rs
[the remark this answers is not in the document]
tmontaigu: Yeah, and also just moving the struct from one module to another would change the string thus breaking serialization

tfhe/src/safe_deserialization.rs
mayeul-zama: Done

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/270

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/271

tfhe/src/high_level_api/integers/types/compact.rs
mayeul-zama: Not sure what you have in mind

`ListSizeConstraint::is_valid` looks correct to me

IceTDrinker: the constraint validates for group min = 1 and group max = 1 and group size = 2

but also for

group min = 2 and group max = 2 and group size = 1

right ?

which are not the same thing at all from a data layout perspective ?

mayeul-zama: Yes
A list of 2 cts could be used as 1 group of 2 cts or as 2 groups of 1 ct, I don´t see where is the problem

Lists of ciphertexts from a client don't have a specific data layout (only a list length), do they?

IceTDrinker: so the other size checks probably mean we are good but I stand by what I said, let's take an LweCiphertextList from core

it could contain a single LWE ciphertext of size 2*n or two ciphertexts of size n and the layout consequences for algorithms mean that some asserts will not hold

I believe the group size may not be required for the ListSize constraint ? and we just need to get the number of entities/count which are allowed within a min/max range ?

Can we plan to add a try_new for the ListContstrainSize ? To avoid people mistakenly giving a max < min

otherwise plan the use of Rust range expressions which could solve both problems at once, wdyt ?

IceTDrinker: the try new stuff/rust range thing can come in a follow up issue or PR once we work more on this, but I belive the ListConstraint is a bit brittle at the moment

tfhe/src/conformance.rs
```diff
@@ -0,0 +1,70 @@
...
+/// - smaller of equal to `max_inclusive_group_count`
+#[derive(Copy, Clone)]
+pub struct ListSizeConstraint {
+    min_inclusive_group_count: usize,
+    max_inclusive_group_count: usize,
+    group_size: usize,
+}
+
+impl ListSizeConstraint {
+    pub fn exact_size(size: usize) -> ListSizeConstraint {
```
IceTDrinker, line 22: let's plan an issue to add docstrings to those functions, but otherwise I think it's good for me

IceTDrinker (approved):
All right let's start with this, thanks a lot for the work !

tfhe/src/conformance.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/272

PR #573, opened 2023-09-20
chore(bench): add missing unsigned integer operations

soonum:
Could you tell me if there are some benchmarks to comment out ?

Resolves: zama-ai/tfhe-rs-internal#224

### DIFF
--- a/.github/workflows/boolean_benchmark.yml
+++ b/.github/workflows/boolean_benchmark.yml
@@ -19,6 +19,14 @@ on:
       request_id:
         description: "Slab request ID"
         type: string
+      # This input is not used in this workflow but still mandatory since a calling workflow could
+      # use it. If a triggering command include a user_inputs field, then the triggered workflow
+      # must include this very input, otherwise the workflow won't be called.
+      # See start_full_benchmarks.yml as example.
+      user_inputs:
+        description: "Type of benchmarks to run"
+        type: string
+        default: "weekly_benchmarks"
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/integer_full_benchmark.yml
+++ b/.github/workflows/integer_full_benchmark.yml
@@ -19,6 +19,10 @@ on:
       request_id:
         description: "Slab request ID"
         type: string
+      user_inputs:
+        description: "Type of benchmarks to run"
+        type: string
+        default: "weekly_benchmarks"
 
 env:
   CARGO_TERM_COLOR: always
@@ -26,8 +30,33 @@ env:
   ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
 jobs:
+  prepare-matrix:
+    name: Prepare operations matrix
+    runs-on: ubuntu-latest
+    outputs:
+      op_flavor: ${{ steps.set_op_flavor.outputs.op_flavor }}
+    steps:
+      - name: Weekly benchmarks
+        if: ${{ github.event.inputs.user_inputs == 'weekly_benchmarks' }}
+        run: |
+          echo "OP_FLAVOR=[\"default\", \"default_comp\", \"default_scalar\", \"default_scalar_comp\"]" >> ${GITHUB_ENV}
+
+      - name: Quarterly benchmarks
+        if: ${{ github.event.inputs.user_inputs == 'quarterly_benchmarks' }}
+        run: |
+          echo "OP_FLAVOR=[\"default\", \"default_comp\", \"default_scalar\", \"default_scalar_comp\", \
+          \"smart\", \"smart_comp\", \"smart_scalar\", \"smart_parallelized\", \"smart_parallelized_comp\", \"smart_scalar_parallelized\", \"smart_scalar_parallelized_comp\", \
+          \"unchecked\", \"unchecked_comp\", \"unchecked_scalar\", \"unchecked_scalar_comp\", \
+          \"misc\"]" >> ${GITHUB_ENV}
+
+      -  name: Set operation flavor output
+         id: set_op_flavor
+         run: |
+          echo "op_flavor=${{ toJSON(env.OP_FLAVOR) }}" >> ${GITHUB_OUTPUT}
+
   integer-benchmarks:
     name: Execute integer benchmarks for all operations flavor
+    needs: prepare-matrix
     runs-on: ${{ github.event.inputs.runner_name }}
     if: ${{ !cancelled() }}
     continue-on-error: true
@@ -35,7 +64,7 @@ jobs:
       max-parallel: 1
       matrix:
         command: [ integer, integer_multi_bit]
-        op_flavor: [ default, default_comp, default_scalar, default_scalar_comp, smart, smart_comp, smart_scalar, smart_parallelized, smart_parallelized_comp, smart_scalar_parallelized, unchecked, unchecked_comp, unchecked_scalar, unchecked_scalar_comp, misc ]
+        op_flavor: ${{ fromJson(needs.prepare-matrix.outputs.op_flavor) }}
     steps:
       - name: Instance configuration used
         run: |
--- a/.github/workflows/start_full_benchmarks.yml
+++ b/.github/workflows/start_full_benchmarks.yml
@@ -3,9 +3,21 @@ name: Start full suite benchmarks
 
 on:
   schedule:
-    # Job will be triggered each Saturday at 1a.m.
+    # Weekly benchmarks will be triggered each Saturday at 1a.m.
     - cron: '0 1 * * 6'
+    # Quarterly benchmarks will be triggered right before end of quarter, the 25th of the current month at 4a.m.
+    # These benchmarks are far longer to execute hence the reason to run them only four time a year.
+    - cron: '0 4 25 MAR,JUN,SEP,DEC *'
   workflow_dispatch:
+    inputs:
+      benchmark_type:
+        description: 'Benchmark type'
+        required: true
+        default: 'weekly'
+        type: choice
+        options:
+          - weekly
+          - quarterly
 
 jobs:
   start-benchmarks:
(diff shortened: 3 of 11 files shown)


.github/workflows/integer_full_benchmark.yml
```diff
@@ -35,7 +35,10 @@ jobs:
       max-parallel: 1
       matrix:
         command: [ integer, integer_multi_bit]
-        op_flavor: [ default, default_comp, default_scalar, default_scalar_comp, smart, smart_comp, smart_scalar, smart_parallelized, smart_parallelized_comp, smart_scalar_parallelized, unchecked, unchecked_comp, unchecked_scalar, unchecked_scalar_comp, misc ]
+        op_flavor: [ default, default_comp, default_scalar, default_scalar_comp,
+                     smart, smart_comp, smart_scalar, smart_parallelized, smart_parallelized_comp, smart_scalar_parallelized, smart_scalar_parallelized_comp,
+                     unchecked, unchecked_comp, unchecked_scalar, unchecked_scalar_comp,
```
IceTDrinker, line 40: question between default (which are all parallelized at the moment) and unchecked, some of which parallelized what should be done for the naming here ?

soonum: I see three solutions to this issue:

1. create a new group containing only `unchecked_scalar_[add|sub|mul]` operation and rename the other by adding the `parallelized` in the name.
2. write a parallel implementation for the operations mentioned above
3. remove from the benchmarks these 3 sequential operations.

soonum:
@slab-ci integer_full_bench

.github/workflows/integer_full_benchmark.yml
tmontaigu: The way negation and subtraction are written is not parallelizable i think

we could do subtraction in a different way to have it parallelizable, however I don't really thing there any benefit to it (much like when testing, there does not seem to have any real benefit of using unchecked_add or unchecked_add_parallelized in the default_add_parallelized)

tfhe/benches/integer/bench.rs
```diff
@@ -684,6 +696,38 @@ define_server_key_bench_default_fn!(
...
+    method_name: unchecked_scalar_bitand_parallelized,
+    display_name: bitand,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_fn!(
+    method_name: unchecked_scalar_bitor_parallelized,
+    display_name: bitor,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_fn!(
```
tmontaigu, line 726: They should use bench_scalar_default_fn

tfhe/benches/integer/bench.rs
```diff
@@ -716,6 +760,46 @@ define_server_key_bench_scalar_fn!(
...
+    method_name: smart_scalar_bitxor_parallelized,
+    display_name: bitxor,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_default_fn!(
+    method_name: smart_scalar_rotate_left_parallelized,
+    display_name: rotate_left,
+    rng_func: shift_scalar
+);
+define_server_key_bench_scalar_default_fn!(
```
tmontaigu, line 798: Theses should not be using bench_default_fn

tfhe/benches/integer/bench.rs
```diff
@@ -803,6 +908,47 @@ define_server_key_bench_scalar_default_fn!(
...
+    display_name: greater_or_equal,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_default_fn!(
+    method_name: smart_scalar_gt_parallelized,
+    display_name: greater_than,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_default_fn!(
+    method_name: smart_scalar_max_parallelized,
```
tmontaigu, line 942: To be consisten default fn should use non default-fn tests

tfhe/benches/integer/bench.rs
```diff
@@ -857,9 +1045,35 @@ define_server_key_bench_default_fn!(
...
+    method_name: unchecked_scalar_gt_parallelized,
+    display_name: greater_than,
+    rng_func: default_scalar
+);
+define_server_key_bench_scalar_fn!(
+    method_name: unchecked_scalar_ge_parallelized,
+    display_name: greater_or_equal,
+    rng_func: default_scalar
+);
+
```
tmontaigu, line 1072: should be using default bench fn

soonum:
Now we can differentiate weekly and quarterly benchamarks.

tfhe/benches/integer/bench.rs
```diff
@@ -952,11 +1178,14 @@ criterion_group!(
...
 criterion_group!(
     default_parallelized_ops,
+    neg_parallelized,
+    abs_parallelized,
     add_parallelized,
     sub_parallelized,
     mul_parallelized,
     div_parallelized,
     rem_parallelized,
+    div_rem_parallelized,
```
IceTDrinker, line 1188: thanks for adding the benchmarks, here I believe we may want only the div_rem version

tfhe/benches/integer/bench.rs
```diff
@@ -952,11 +1178,14 @@ criterion_group!(
 
 criterion_group!(
     default_parallelized_ops,
+    neg_parallelized,
```
IceTDrinker, line 1181: I see it line 1189

tfhe/benches/integer/bench.rs
```diff
@@ -993,6 +1222,26 @@ criterion_group!(
     smart_scalar_add_parallelized,
     smart_scalar_sub_parallelized,
     smart_scalar_mul_parallelized,
+    smart_scalar_div_parallelized,
+    smart_scalar_rem_parallelized,
+    smart_scalar_div_rem_parallelized,
```
IceTDrinker, line 1227: probably only div rem

tfhe/benches/integer/bench.rs
```diff
@@ -1002,10 +1251,14 @@ criterion_group!(
     scalar_mul_parallelized,
     scalar_div_parallelized,
     scalar_rem_parallelized,
+    scalar_div_rem_parallelized,
```
IceTDrinker, line 1254: probably only div_rem

tfhe/benches/integer/bench.rs
```diff
@@ -1030,11 +1283,40 @@ criterion_group!(
...
 );
 
+criterion_group!(
+    unchecked_parallelized_ops,
+    unchecked_abs_parallelized,
+    unchecked_add_parallelized,
+    unchecked_mul_parallelized,
+    unchecked_div_parallelized,
+    unchecked_rem_parallelized,
+    unchecked_div_rem_parallelized,
```
IceTDrinker, line 1293: only div_rem potentially

tfhe/benches/integer/bench.rs
```diff
@@ -1030,11 +1283,40 @@ criterion_group!(
     unchecked_bitxor,
 );
 
+criterion_group!(
+    unchecked_parallelized_ops,
+    unchecked_abs_parallelized,
+    unchecked_add_parallelized,
+    unchecked_mul_parallelized,
+    unchecked_div_parallelized,
+    unchecked_rem_parallelized,
```
IceTDrinker, line 1292: btw @tmontaigu do we have a super fast rem/div for a scalar a power of 2 ?

div is a shift, rem is an and with a well chosen mask

tfhe/benches/integer/bench.rs
```diff
@@ -1046,20 +1328,28 @@ criterion_group!(
     unchecked_scalar_add,
     unchecked_scalar_sub,
     unchecked_scalar_mul_parallelized,
-    unchecked_bitand_parallelized,
-    unchecked_bitor_parallelized,
-    unchecked_bitxor_parallelized,
+    unchecked_scalar_div_parallelized,
+    unchecked_scalar_div_rem_parallelized,
+    unchecked_scalar_rem_parallelized,
```
IceTDrinker, line 1333: div_rem

IceTDrinker (commented):
Not sure as to what we want to keep, I would guess div rem is the only one we need for div/rem

maybe the bit ops only need one as well, I let Thomas complete with his knowledge of the algorithms

.github/workflows/boolean_benchmark.yml
```diff
@@ -19,6 +19,10 @@ on:
       request_id:
         description: "Slab request ID"
         type: string
+      user_inputs:
+        description: "Type of benchmarks to run"
+        type: string
+        default: "weekly_benchmarks"
```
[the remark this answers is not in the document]
soonum: Since I'm passing `user_inputs` to all the workflow I need to declare this input otherwise the API call fail and the workflow is not triggered.

.github/workflows/start_full_benchmarks.yml
```diff
@@ -27,10 +30,20 @@ jobs:
...
+      - name: Set benchmarks type as weekly
+        if: github.event_name == 'workflow_dispatch' || github.event.schedule == '0 1 * * 6'
+        run: |
+          echo "BENCH_TYPE=weekly_benchmarks" >> "${GITHUB_ENV}"
+
+      - name: Set benchmarks type as quarterly
+        if: github.event.schedule == '0 4 25 MAR,JUN,SEP,DEC *'
+        run: |
+          echo "BENCH_TYPE=quarterly_benchmarks" >> "${GITHUB_ENV}"
+
```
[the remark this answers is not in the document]
soonum: I can add a choice along with `workflow_dispatch` event.

.github/workflows/boolean_benchmark.yml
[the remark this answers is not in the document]
IceTDrinker: ah I see, but some don't use that input ? then a comment would be welcome

tfhe/benches/integer/bench.rs
tmontaigu: Yes there is

tfhe/benches/integer/bench.rs
tmontaigu: In the case of scalar division, div is done on its own, rem needs division

In ciphertext, ciphertext  division, they are both computed by div_rem, so whether you need only the div or remainder, the cost is the same

soonum: So what should we do for scalar ? div and div_rem ?

IceTDrinker: probably the 3 ops for scalar, div is its own thing, rem uses div, and div_rem does both

tmontaigu: For scalars `(div_rem` == rem) != div`
For ciphertexts `div == rem == div_rem`

soonum: I think we are almost good now with this last push.

tmontaigu (approved):

soonum:
@slab-ci cpu_wasm_test

PR #579, opened 2023-09-21
fix(integer): better estimate which algorithm to choose

tmontaigu:
Use a more precise way to choose between using parallel or non parallel carry propagation.

Fixes an issue where for 256bits scalar div, it would select the non parallel carry propagation slowing things down

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -2,6 +2,8 @@ use crate::integer::ciphertext::IntegerRadixCiphertext;
 use crate::integer::{RadixCiphertext, ServerKey};
 use crate::shortint::Ciphertext;
 
+use crate::core_crypto::commons::numeric::UnsignedInteger;
+use crate::core_crypto::prelude::misc::divide_ceil;
 use rayon::prelude::*;
 
 #[repr(u64)]
@@ -35,6 +37,40 @@ fn prefix_sum_carry_propagation(msb: u64, lsb: u64) -> u64 {
     } else {
         msb
     }
+}
+
+fn should_hillis_steele_propagation_be_faster(num_blocks: usize, num_threads: usize) -> bool {
+    // Measures have shown that using a parallelized algorithm degrades
+    // the latency of a PBS, so we take that into account.
+    // (This factor is a bit pessimistic).
+    const PARALLEL_LATENCY_PENALTY: usize = 2;
+    // However that penalty only kicks in when certain level of
+    // parallelism is used
+    let penalty_threshold = num_threads / 2;
+
+    // The unit of latency is a PBS
+    let compute_latency_of_one_layer = |num_blocks: usize, num_threads: usize| -> usize {
+        let latency = divide_ceil(num_blocks, num_threads);
+        if num_blocks >= penalty_threshold {
+            latency * PARALLEL_LATENCY_PENALTY
+        } else {
+            latency
+        }
+    };
+
+    // Estimate the latency of the parallelized algorithm
+    let mut parallel_expected_latency = 2 * compute_latency_of_one_layer(num_blocks, num_threads);
+    let max_depth = num_blocks.ceil_ilog2();
+    let mut space = 1;
+    for _ in 0..max_depth {
+        let num_block_at_iter = num_blocks - space;
+        let iter_latency = compute_latency_of_one_layer(num_block_at_iter, num_threads);
+        parallel_expected_latency += iter_latency;
+        space *= 2;
+    }
+
+    // the other algorithm has num_blocks latency
+    parallel_expected_latency < num_blocks
 }
 
 impl ServerKey {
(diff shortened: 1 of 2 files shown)


tmontaigu:
@slab-ci cpu_integer_test

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -37,6 +39,40 @@ fn prefix_sum_carry_propagation(msb: u64, lsb: u64) -> u64 {
     }
 }
 
+fn should_hillis_steele_propagation_be_faster(num_blocks: usize, num_threads: usize) -> bool {
+    // Measures have shown that using a parallelized algorithm degrades
+    // the latency of a PBS, so we take that into account.
+    // (This factor is a bit pessimistic).
+    const PARALLEL_LATENCY_PENALTY: usize = 2;
```
IceTDrinker, line 46: should we consider the worst case for parallelization in that case ? As we don't necesseraliy know if the penalty is the same ? As you point out it's a bit pessimisitc, realistically the only way to know for sure would be to benchmark but it's not something we'll do right away (there are some thoughts on that for a functionality on auto tuning down the line)

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -709,3 +731,58 @@ impl ServerKey {
         reduce_impl(self, ct_seq, op)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::should_hillis_steele_propagation_be_faster;
```
IceTDrinker, line 737: are those run properly with the filters we have ?

tmontaigu: they should on the non fast test

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: What is the worst cae you think about?

IceTDrinker: the worst case you are talking of here, it's an uncertain worst case is what I'm trying to say, even if we see empirically it's mostly true

tmontaigu: I think keeping that worst case penalty is ok as thats something we always observed for now

I'll rerun the tests without taking into account that penalty and see if it changes the chosen algorithms.

If it does change them then we will have to keep the penalty thing for sure

IceTDrinker: Ok

tmontaigu: I reran the tests, witout the penalty factor the only change is that for 12 threads, 8 blocks it selects the parallel version (with the penalty factor it selects the non parallel one)

From benchmarks they are close (the parallel version being 5-7% slower)

tmontaigu: Actually ran the thing without penalty, it chooses the paralle version for all block size using 12 threads which does not match  benchmarks results

IceTDrinker: so better with penalty ? my only concern is potentially having a bad surprise with large thread count, I don't know what the graph of those functions look like

IceTDrinker: ah I think I know why, it thinks the whole layer is parallelized, while for small core counts it will get serialized by chunks of 12 e.g.

IceTDrinker: so the true latency per layer is num_pbs_per_layer / thread_count (with some rounding, ceil or floor in there)

tmontaigu: > so better with penalty ? my only concern is potentially having a bad surprise with large thread count, I don't know what the graph of those functions look like

With the penalty we get a choice that matches performances measured by benchmarks

> so the true latency per layer is num_pbs_per_layer / thread_count (with some rounding, ceil or floor in there)
We do one PBS per block, so divide_ceil(num_blocks, num_threads) is a somewhat good estimation i would say

IceTDrinker: ah yeah indeed, it's already in there, then let's go with the empirical formula for now and revisit if we see that these things change

IceTDrinker (approved):
Looks good, let's revisit at a later stage if we see some stuff evolving a lot on that front

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test

IceTDrinker:
Are we waiting for after 0.4.0 to release ?

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
```diff
@@ -885,6 +885,7 @@ fn integer_signed_unchecked_div_rem(param: impl Into<PBSParameters>) {
 
     // Div is the slowest operation
     for _ in 0..NB_TEST_SMALLER {
+        println!("lol");
```
IceTDrinker, line 888: lol

IceTDrinker (approved):

PR #580, opened 2023-09-22
chore(integer): add tests for default signed rotations/shifts

tmontaigu:

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
@@ -1103,6 +1103,50 @@ create_parametrized_test!(integer_signed_default_bitand);
 create_parametrized_test!(integer_signed_default_bitor);
 create_parametrized_test!(integer_signed_default_bitxor);
 create_parametrized_test!(integer_signed_default_absolute_value);
+create_parametrized_test!(integer_signed_default_left_shift {
+    // Requires 3 bits, so 1_1 parameters are not supported
+    // until they get their own version of the algorithm
+    PARAM_MESSAGE_2_CARRY_2_KS_PBS,
+    PARAM_MESSAGE_3_CARRY_3_KS_PBS,
+    PARAM_MESSAGE_4_CARRY_4_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_2_CARRY_2_GROUP_2_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_3_CARRY_3_GROUP_2_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_2_CARRY_2_GROUP_3_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_3_CARRY_3_GROUP_3_KS_PBS
+});
+create_parametrized_test!(integer_signed_default_right_shift {
+    // Requires 3 bits, so 1_1 parameters are not supported
+    // until they get their own version of the algorithm
+    PARAM_MESSAGE_2_CARRY_2_KS_PBS,
+    PARAM_MESSAGE_3_CARRY_3_KS_PBS,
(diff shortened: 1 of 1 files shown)


tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
```diff
@@ -1474,6 +1518,342 @@ fn integer_signed_default_absolute_value(param: impl Into<PBSParameters>) {
...
+
+    sks.set_deterministic_pbs_execution(true);
+
+    let mut rng = rand::thread_rng();
+
+    let modulus = (cks.parameters().message_modulus().0.pow(NB_CTXT as u32) / 2) as i64;
+
+    assert!(modulus > 0);
+    assert!((modulus as u64).is_power_of_two());
+    let nb_bits = modulus.ilog2() + 1; // We are using signed numbers
```
IceTDrinker, line 1536: how does that change the nb_bits and increases it ?

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
```diff
@@ -1474,6 +1518,342 @@ fn integer_signed_default_absolute_value(param: impl Into<PBSParameters>) {
...
+            let mut shift = cks.encrypt(nb_bits);
+            sks.unchecked_add_assign(&mut shift, &tmp);
+            clear_shift += nb_bits;
+            assert!(!shift.block_carries_are_empty());
+
+            let ct_res = sks.left_shift_parallelized(&ct, &shift);
+            let dec_res: i64 = cks.decrypt_signed(&ct_res);
+            // When nb_bits is not a power of two
+            // then the behaviour is not the same
+            let mut nb_bits = modulus.ilog2();
```
IceTDrinker, line 1585: can you elaborate a bit ?

IceTDrinker (commented):
Same comments apply for the various tests

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
tmontaigu: Above, we computed the modulus as `let modulus = (cks.parameters().message_modulus().0.pow(NB_CTXT as u32) / 2) as i64;` so the unsigned modulus divided by two. So two have the number of bits we have to add that bit

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
tmontaigu: It comes from how barrel shifter is/are made (so thats also the case for unsigned radix)

If you want to shift an integer type with N bits, the barrel shift only consider the log2(N) bits from the value that gives the shift amount.

This works well when N is a power of two as it is the case for 8, 16, 32, 64, 128.
e.g for a u8 -> 8 bits -> log2(8) = 4, barrel shifter only consider bits in range 0..4 (so 3 bits), with 3 bits we can only give input shift in range 0..=7 -> great u8 has 8 bits
u16 -> 16 bits -> bits in range 0..4 considered -> possible shift values 0..15.
Basically the barrel shift naturally does `shift_amount % nb_bits` if N is a power of two.

for u12 (4 blocks of 3_3)  -> log2(12) = 3.58 -> we have to ceil() -> 4 -> bits in range 0..4 -> shift values 0..15 which is > 12.  Which means that the behaviour is not the same if the input shift value is not already in the good range

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
IceTDrinker: indeed

IceTDrinker (approved):

tmontaigu:
Rebased on main to have fix for division

IceTDrinker (approved):

IceTDrinker:
needs make fmt

IceTDrinker (approved):

PR #587, opened 2023-09-26
chore(hlapi): remove unused keychain_member from macro

tmontaigu:


tfhe/src/high_level_api/integers/keys.rs
```diff
@@ -119,6 +119,11 @@ impl IntegerServerKey {
         let Some(cks) = &client_key.key else {
             return Self::default();
         };
+        assert_eq!(
+            cks.parameters().message_modulus().0,
+            1 << 2,
+            "This API only supports integers with 2 bits per block (MessageModulus(16))",
```
IceTDrinker, line 125: MessageModulus(4) ? in that case you could use the debug format to avoid having to write it yourself :)

tfhe/src/high_level_api/integers/keys.rs
```diff
@@ -119,6 +119,11 @@ impl IntegerServerKey {
         let Some(cks) = &client_key.key else {
             return Self::default();
         };
+        assert_eq!(
+            cks.parameters().message_modulus().0,
+            1 << 2,
```
IceTDrinker, line 124: any reason for that ? I thought we were ok for 1_1 ?

tmontaigu: No were not (and still are not) ok with 1_1, in the HLAPI number of blocks are statically fixed for each types assuming 2_2 parameters.

And we don't support all operations on 1_1 parameters

So first we should support all ops on 1_1 at the integer level then we can add suport for 1_1 params in HLAPI

IceTDrinker: > in the HLAPI number of blocks are statically fixed for each types assuming 2_2 parameters

that's surprising, does it mean 3_3 or 4_4 params would not work as expected ?

IceTDrinker: ah ok I just saw it's an assert_eq

IceTDrinker (approved):

PR #588, opened 2023-09-27
fix(integer): StaticSignedBigInt right shift

tmontaigu:
### Check-list:


IceTDrinker (approved):
Can you quickly indicate what wrong value we were getting before ?

tmontaigu:
For -1 >> 1 we were getting [0, 0, 0, u64::MAX], so 64 `1` in the MSB then zeros
when we expect 2 `1` in MSB and the rest to be zero

IceTDrinker:
ok arithmetic shift

tmontaigu:
Yes but we were putting the value of the sign bit all over the most significant block and not only in places where sign bits would actually need to be used as padding

PR #590, opened 2023-09-27
add asserts on smart ops

mayeul-zama:


mayeul-zama:
@slab-ci cpu_fast_test

mayeul-zama:
@slab-ci cpu_fast_test

tfhe/src/shortint/engine/server_side/mul.rs
```diff
@@ -179,6 +181,7 @@ impl ShortintEngine {
                 self.message_extract_assign(server_key, ct_left)?;
                 self.message_extract_assign(server_key, ct_right)?;
             }
+            assert!(server_key.is_mul_small_carry_possible(ct_left, ct_right));
```
mayeul-zama, line 184: Remove duplicate

IceTDrinker:
needs a rebase to get rid of the wasm issue

IceTDrinker (commented):
general question, what are the conditions that would make the smart operations fail ? i.e. a bad parameter set e.g. ?

Are the smart operations the only operations needing those checks ?

mayeul-zama:
> general question, what are the conditions that would make the smart operations fail ? i.e. a bad parameter set e.g. ?

It should not fail if operations are used properly (except if we have bugs).
It helped catch bugs fixed in 76aaa5669198aaf7d27de294fda263bb45543dfb, 096e320b97a1cda873ff2599ad8dd1c6b332a490 and f08ea8cf85a7ca6dee85fd1bae98be0b796bd262

But for example, if you call `smart_small_scalar_mul_parallelized` with a big scalar (as before fix commit 76aaa5669198aaf7d27de294fda263bb45543dfb), it used to do a full propagate and then do the `unchecked_small_scalar_mul_parallelized` which was incorrect. Now it should panic.

> Are the smart operations the only operations needing those checks ?

IMO it could make sense for some default operations but I'm not sure.
For unchecked and checked operations, I don't think it would make sense

IceTDrinker (approved):

PR #591, opened 2023-09-27
chore(integer): impl RecomposableSignedInteger for StaticSignedBigInt

tmontaigu:


tfhe/src/integer/block_decomposition.rs
```diff
@@ -48,9 +49,21 @@ macro_rules! impl_recomposable_decomposable {
     };
 }
 
-impl_recomposable_decomposable!(
-    u8, u16, u32, u64, u128, U256, U512, i8, i16, i32, i64, i128, I256, I512
-);
+impl_recomposable_decomposable!(u8, u16, u32, u64, u128, i8, i16, i32, i64, i128,);
+
+impl<const N: usize> Decomposable for StaticSignedBigInt<N> {}
```
IceTDrinker, line 54: no methods on these traits ? just curious

IceTDrinker (approved):

tfhe/src/integer/block_decomposition.rs
tmontaigu: No they are super traits that require some std::ops (like <<, |, &) to avoid repeating long trait bounds

PR #597, opened 2023-09-29
chore(csprng): the stabilized aarch64 intrisics were in Rust 1.72

IceTDrinker:


mayeul-zama:
What doe this PR do ?
Just replaces a `unresolved imports std::?` by a `upgrade to 1.72`  for people using an old toolchain?

IceTDrinker:
> What doe this PR do ? Just replaces a `unresolved imports std::?` by a `upgrade to 1.72` for people using an old toolchain?

Something like that potentially yes

require people to use the 1.72 toolchain so that CSPRNG builds on stable toolchains no matter the cpu arch x86 or aarch64, it's also the reason the minimal rust version was raised on TFHE

IceTDrinker:
@slab-ci cpu_fast_test

PR #599, opened 2023-10-02
feat(hlapi): tie scalar ops with corresponding clear type

tmontaigu:
Operations that used a scalar as right operand where generically implemented meaning a user could, for example, add a u32 to a FheUint8.

Rust only allows operations between matching types, so we do the same thing.

BREAKING CHANGE: This is a braking change on the Rust API, but for the better I believe. On the C API it is not a breaking change as we already made that assocation as it was simpler to implement

### Check-list:

### DIFF
--- a/tfhe/src/high_level_api/integers/types/base.rs
+++ b/tfhe/src/high_level_api/integers/types/base.rs
@@ -22,7 +22,6 @@ use crate::high_level_api::traits::{
 use crate::high_level_api::{ClientKey, PublicKey};
 use crate::integer::block_decomposition::DecomposableInto;
 use crate::integer::ciphertext::RadixCiphertext;
-use crate::integer::server_key::{Reciprocable, ScalarMultiplier};
 use crate::integer::U256;
 use crate::CompactPublicKey;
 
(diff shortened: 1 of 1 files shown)


tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -919,31 +952,343 @@ generic_integer_impl_operation_assign!(RotateRightAssign(rotate_right_assign) =>
...
+        (super::FheUint64, u8, u16, u32, u64, u128),
+        (super::FheUint128, u8, u16, u32, u64, u128),
+        (super::FheUint256, u8, u16, u32, u64, u128, U256),
+);
+generic_integer_impl_scalar_operation_assign!(
+    rust_trait: ShrAssign(shr_assign),
+    key_method: scalar_right_shift_assign_parallelized,
+    fhe_and_scalar_type:
+        (super::FheUint8, u8, u16, u32, u64, u128),
+        (super::FheUint10,u8, u16, u32, u64, u128),
```
IceTDrinker, line 1227: formatting for all FheUint10 lines is off, missing a space after the first comma

IceTDrinker (commented):
nitpick

tfhe/src/high_level_api/integers/types/base.rs
tmontaigu: fixed

IceTDrinker (approved):

PR #603, opened 2023-10-03
refactor(core): use avx512 intrinsics when available for data conversions

IceTDrinker:


IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test

IceTDrinker:
actually there appears to be no rounding, but it should be fine in our case for the native case as we rescale things to [-2^63, 2^63[ which is a 2^64 wide interval, we end up with only proper integer values in the floating point domain, the 2^63 edge case seems to be handled well

sarah-ek:
what we need is VCVTTPD2QQ instead of VCVTPD2QQ, since i think we want rounding with truncation

IceTDrinker:
indeed to match the existing behavior

IceTDrinker:
with 0.4.0 released and the AVX512 enabled integer test + updated conversion tests I feel confident moving ahead with this @sarah-ek :)

sarah-ek (approved):

PR #605, opened 2023-10-03
feat(hlapi): add FheInt{8,16,32,64,128,256}

tmontaigu:


tfhe/src/high_level_api/errors.rs
```diff
@@ -53,6 +53,18 @@ pub enum Type {
     FheUint128,
     #[cfg(feature = "integer")]
     FheUint256,
+    #[cfg(feature = "integer")]
+    FheInt8,
```
IceTDrinker, line 57: skipping the "shortints", do we have an issue to remove those from the HL ?

tfhe/src/high_level_api/integers/tests_unsigned.rs
```diff
@@ -779,12 +780,43 @@ fn test_integer_casting() {
...
-        let clear = 43_129u16;
         let a = FheUint16::encrypt(clear, &client_key);
         let a = FheUint16::cast_from(a);
         let da: u16 = a.decrypt(&client_key);
         assert_eq!(da, clear);
     }
+
+    // Downcasting to smaller signed integer then Upcasting back to unsigned
+    {
+        let clear = rng.gen_range((i16::MAX) as u16 + 1..u16::MAX);
```
IceTDrinker, line 791: not sure I understand why the range is written this way

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -111,38 +111,76 @@ where
...
+                    let num_blocks_to_remove = current_num_blocks - target_num_blocks;
+                    integer_key.trim_radix_blocks_msb_assign(
+                        &mut ct_as_unsigned_radix,
+                        num_blocks_to_remove,
+                    );
+                }
+                ct_as_unsigned_radix.blocks
+            };
+
+            assert_eq!(blocks.len(), P2::num_blocks());
```
IceTDrinker, line 169: error message could be welcome

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -111,38 +111,76 @@ where
...
-            if target_num_blocks > current_num_blocks {
-                let num_blocks_to_add = target_num_blocks - current_num_blocks;
-                integer_key.extend_radix_with_trivial_zero_blocks_msb_assign(
-                    &mut self.ciphertext,
-                    num_blocks_to_add,
-                );
+            let blocks = if P::InnerCiphertext::IS_SIGNED {
+                if target_num_blocks > current_num_blocks {
+                    let mut ct_as_signed_radix =
+                        SignedRadixCiphertext::from_blocks(self.ciphertext.into_blocks());
```
IceTDrinker, line 133: not sure I understand the from blocks and the check on InnerCiphertext IS_SIGNED and the need for the from_blocks, is it a matter of trait not knowing the concrete type ?

tfhe/src/high_level_api/integers/tests_signed.rs
```diff
@@ -0,0 +1,650 @@
+use crate::integer::I256;
```
IceTDrinker, line 1: I believe HL API tests are not run 🤔 could we add a Makefile target ?

(could not find them in the integer tests run)

IceTDrinker: it's done in the fast tests it looks like my bad

tfhe/src/high_level_api/errors.rs
tmontaigu: I don't think so

tfhe/src/high_level_api/integers/types/base.rs
tmontaigu: I did not put one because if that assert is hit, that means we (and not the user) did something wrong
but i can add a small one

tfhe/src/high_level_api/integers/tests_unsigned.rs
tmontaigu: Im downcasting a u16 to an i8, and wanted the u16 value to be out of the range of the i8, I could have use (i8::MAX as u16)+ 1 instead.

not sure doing that range is that useful

tfhe/src/high_level_api/integers/types/base.rs
tmontaigu: Yes we don't know the concrete type, and extend_radix_with_sign_msb only works for SignedRadixCiphertext

tfhe/src/high_level_api/integers/types/base.rs
IceTDrinker: ah right, it never hurts to have a bit of an explicit message (even just for us, this way we can search/grep for the error message)

tfhe/src/high_level_api/integers/types/base.rs
IceTDrinker: ok

tfhe/src/high_level_api/integers/tests_unsigned.rs
IceTDrinker: ah I know what confused me it's the + 1..u16::MAX

IceTDrinker (approved):

PR #607, opened 2023-10-04
chore(bench): fix keygen location to avoid slow matching on benchmarks

IceTDrinker:


tmontaigu (approved):

IceTDrinker:
this is making benchmarks much slower so I guess we have to chose our poison, slow matching or slow bench, I vote for slow matching, we know how to change this if required, closing

PR #609, opened 2023-10-04
feat(integer): add unsigned_overflowing_sub

tmontaigu:


tfhe/src/integer/server_key/radix/sub.rs
```diff
@@ -323,4 +324,119 @@ impl ServerKey {
...
+        for (lhs_b, rhs_b) in lhs.blocks.iter().zip(rhs.blocks.iter()) {
+            let mut result_block = self.key.unchecked_sub(lhs_b, rhs_b);
+            // Here unchecked_sub_assign does not give correct result, we don't want
+            // the correcting term to be used
+            // -> This is ok as the value returned by unchecked_sub is in range 1..(message_mod * 2)
+            crate::core_crypto::algorithms::lwe_ciphertext_sub_assign(
+                &mut result_block.ct,
+                &borrow.ct,
+            );
+            let (msg, new_borrow) = rayon::join(
```
IceTDrinker, line 426: taking a simple case where lhs is a single block and rhs as well where lhs = 0 and rhs = 2 I seem to not find the right result

the unchecked sub produces the modular opposite so -2 mod msg_mod which is greater than msg_mod

the borrow is zero, so the next sub still has that -2 mod msg_mod

the apply lookup table will return a message encrypting a zero right ?

am I missing something ?

tfhe/src/integer/server_key/radix/sub.rs
```diff
@@ -323,4 +324,119 @@ impl ServerKey {
...
+                self.full_propagate_parallelized(&mut tmp_lhs);
+                (&tmp_lhs, ctxt_right)
+            }
+            (false, false) => {
+                tmp_lhs = ctxt_left.clone();
+                tmp_rhs = ctxt_right.clone();
+                rayon::join(
+                    || self.full_propagate_parallelized(&mut tmp_lhs),
+                    || self.full_propagate_parallelized(&mut tmp_rhs),
+                );
```
IceTDrinker, line 387: is this ok in the radix module ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -257,4 +260,211 @@ impl ServerKey {
...
+                        &mut block.ct,
+                        &input_borrow.ct,
+                    );
+                    self.key.message_extract_assign(block);
+                });
+            assert!(ct.block_carries_are_empty());
+            // we know here that the result is a boolean value
+            // however the lut used has a degree of 2.
+            output_borrow.degree.0 = 1;
+            //assert!(output_borrow.degree.0 == 1);
```
IceTDrinker, line 377: still required ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -257,4 +260,211 @@ impl ServerKey {
...
+            (ct, output_borrow)
+        } else {
+            self.unsigned_overflowing_sub(lhs, rhs)
+        }
+    }
+
+    pub(super) fn generate_init_borrow_array(&self, sum_ct: &RadixCiphertext) -> Vec<Ciphertext> {
+        let modulus = self.key.message_modulus.0 as u64;
+
+        // This used for the first pair of blocks
```
IceTDrinker, line 387: This is used

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -257,4 +260,211 @@ impl ServerKey {
...
+        });
+
+        let mut generates_or_propagates = Vec::with_capacity(sum_ct.blocks.len());
+        sum_ct
+            .blocks
+            .par_iter()
+            .enumerate()
+            .map(|(i, block)| {
+                if i == 0 {
+                    // The first block can only output a borrow
```
IceTDrinker, line 418: is borrow "just" a carry in the end ? if so maybe changing the name is better for consistency with function calls ? unless it's the standardized term for the subtraction

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -257,4 +260,211 @@ impl ServerKey {
...
+    pub(crate) fn compute_borrow_propagation_parallelized_low_latency(
+        &self,
+        generates_or_propagates: Vec<Ciphertext>,
+    ) -> (Vec<Ciphertext>, Ciphertext) {
+        let lut_borrow_propagation_sum = self
+            .key
+            .generate_lookup_table_bivariate(prefix_sum_borrow_propagation);
+
+        fn prefix_sum_borrow_propagation(msb: u64, lsb: u64) -> u64 {
+            if msb == 2 {
```
IceTDrinker, line 440: small comment for the signification of the 2 or have an enum repr(u64) or whichever works

IceTDrinker (commented):
One question about the correctness which I could not confirm alone (I might be misunderstanding the algorithm) and some other more minor remarks

tfhe/src/integer/server_key/radix/sub.rs
tmontaigu: I don't understand 🤔

IceTDrinker: parallelized ops in the non parallel module ?

tfhe/src/integer/server_key/radix/sub.rs
tmontaigu: Lhs: [0]
Rhs: [2]

Lhs - Rhs: [0 - 2 + msg_ms] = [0 - 2 + 4] = [2]
2 < msg_mod -> output_borrow 1.

So at the end we have (0 - 2) % 4 = 2 with an output borrow of one, meaning overflow occured

IceTDrinker: thanks !

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -338,18 +339,47 @@ impl ServerKey {
     ///
     /// Requires the blocks to have at least 4 bits
     pub(crate) fn compute_carry_propagation_parallelized_low_latency(
+        &self,
+        generates_or_propagates: Vec<Ciphertext>,
+    ) -> (Vec<Ciphertext>, Ciphertext) {
```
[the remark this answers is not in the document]
tmontaigu: At the integer level, having a Ciphertext seems ok to me as this is likely to be used in cmux

If it were an integer, would it be 1 block sized, or same num blocks a input like I did for comparisons ?

Talking of comparisons, they return an integer, but in practice a lot of time they are used you do cmp_result.blocks[0], because you know others are trivials.


A boolean like integer will likely be needed at some point, probably when we can migrate the hlapi to being integers only

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -338,18 +339,47 @@ impl ServerKey {
...
+
+        let num_blocks = generates_or_propagates.len();
+        let mut carries_out =
+            self.compute_prefix_sum_hillis_steele(generates_or_propagates, sum_function);
+        let mut last_block_out_carry = self.key.create_trivial(0);
+        std::mem::swap(&mut carries_out[num_blocks - 1], &mut last_block_out_carry);
+        // The output carry of block i-1 becomes the input
+        // carry of block i
+        carries_out.rotate_right(1);
+        self.key.create_trivial_assign(&mut carries_out[0], 0);
```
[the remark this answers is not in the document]
tmontaigu: Yes technically not needed, it allows to simply `zip(carry, mesage).map(|c, m| message_etrac(m + c))` without having an if inside.

tfhe/src/integer/server_key/radix_parallel/add.rs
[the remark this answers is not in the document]
IceTDrinker: I believe the boolean integer would be a single shortint Ciphertext underneath, just that we have two ways of representing booleans with this code change/comparisons

tfhe/src/integer/server_key/radix_parallel/add.rs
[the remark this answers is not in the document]
IceTDrinker: no I mean you swapped a trivial zero before ?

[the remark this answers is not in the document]
IceTDrinker: I could be wrong

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Depends on how we see things, the max value of the lut used is 2, but we know that the output carry block of the propagation steps will always be either 0 or 1.

So if I don't force the degree, we may have the problem of the block not looking like it encrypts a boolean value, when we know it in fact does

IceTDrinker: I mean the commented assert ;)

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Well, a borrow is the opposite of a carry, you give a carry to the next block, you borrw from the next block

tmontaigu: And its a standard name borrow == retenue

IceTDrinker: ok thanks !

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -1717,6 +1732,95 @@ where
...
+    let cks = RadixClientKey::from((cks, NB_CTXT));
+
+    sks.set_deterministic_pbs_execution(true);
+
+    let mut rng = rand::thread_rng();
+
+    // message_modulus^vec_length
+    let modulus = cks.parameters().message_modulus().0.pow(NB_CTXT as u32) as u64;
+
+    executor.setup(&cks, sks.clone());
```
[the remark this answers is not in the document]
tmontaigu: Yeah, the executor receives ownerhsip of the sks, however in tests like this, the test case still need the server key to be able to modify degrees.

If we are worried about this 2 copes being there, maybe the key could be put in a Rc/Arc

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: 😅 no its not needed, I think i put it there when debugging the division that uses that overflowing_sub

tfhe/src/integer/server_key/radix/sub.rs
tmontaigu: Ah right, yes I guess it should use non parallelized propagate

tfhe/src/integer/server_key/radix/sub.rs
```diff
@@ -323,4 +324,119 @@ impl ServerKey {
...
+        &self,
+        lhs: &RadixCiphertext,
+        rhs: &RadixCiphertext,
+    ) -> (RadixCiphertext, Ciphertext) {
+        assert!(
+            lhs.blocks.len() >= rhs.blocks.len(),
+            "Left hand side must must have a number of blocks greater or equal \
+            to the number of blocks of the right hand side: lhs {} blocks, rhs {} blocks",
+            lhs.blocks.len(),
+            rhs.blocks.len(),
```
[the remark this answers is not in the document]
tmontaigu: We could be more strict and require same length

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Added an enum just like it's done for addition's carry propagation

tfhe/src/integer/server_key/radix/sub.rs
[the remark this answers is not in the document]
tmontaigu: replace with assert_eq

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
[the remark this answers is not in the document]
IceTDrinker: ok to be seen down the line this is purely internal

tfhe/src/integer/server_key/radix/sub.rs
```diff
@@ -323,4 +324,120 @@ impl ServerKey {
...
+            }
+            (false, true) => {
+                tmp_lhs = ctxt_left.clone();
+                self.full_propagate(&mut tmp_lhs);
+                (&tmp_lhs, ctxt_right)
+            }
+            (false, false) => {
+                tmp_lhs = ctxt_left.clone();
+                tmp_rhs = ctxt_right.clone();
+                rayon::join(
```
IceTDrinker, line 384: last nitpick, is rayon ok in the sequential module ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -257,4 +272,203 @@ impl ServerKey {
...
+        self.unchecked_unsigned_overflowing_sub_parallelized(lhs, rhs)
+    }
+
+    pub fn unchecked_unsigned_overflowing_sub_parallelized(
+        &self,
+        lhs: &RadixCiphertext,
+        rhs: &RadixCiphertext,
+    ) -> (RadixCiphertext, Ciphertext) {
+        assert!(
+            lhs.blocks.len() >= rhs.blocks.len(),
```
IceTDrinker, line 349: should this be an assert_eq as well ?

tfhe/src/integer/server_key/radix/sub.rs
tmontaigu: Humpf, we can remove them from this function

However you can notice we also use rayon in the unchecked version, and these should not be removed because its the version used when the ciphertext does not support the parallel prefix sum (either due to parameters like 1_1, or because the ciphertext does not have enough blocks for it to be worth)

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: I guess so to be consistent

IceTDrinker (approved):

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

Issue #615, opened 2023-10-06
`if_then_else_parallelized`: "Condition ciphertext does not encrypt a boolean (0 or 1) value"


matthiasgeihs:
**Describe the bug**
`if_then_else_parallelized` panics with "Condition ciphertext does not encrypt a boolean (0 or 1) value", but the condition ciphertext decrypts to 0.

**To Reproduce**
Minimal example:
```rust
use tfhe::integer::{gen_keys_radix, RadixCiphertext, ServerKey};
use tfhe::shortint::prelude::PARAM_MESSAGE_2_CARRY_2_KS_PBS;

fn main() {
    let params = PARAM_MESSAGE_2_CARRY_2_KS_PBS;
    let precision = 16;
    let num_blocks = precision / params.message_modulus.0.ilog2() as usize;
    let (ck, sk) = gen_keys_radix(params, num_blocks);

    let a = 0u32;
    let b = 0u32;
    let a_enc = ck.encrypt(a);
    let b_enc = ck.encrypt(b);

    let c_enc = binary_or(&sk, &a_enc, &b_enc);
    let c_dec = ck.decrypt::<u8>(&c_enc);

    println!("condition = {c_dec}"); // Prints "0".
    let _ = sk.if_then_else_parallelized(&c_enc, &a_enc, &b_enc); // Panics.
}

// Returns `a or b`, assuming `a` and `b` are encryptions of binary values.
pub fn binary_or(k: &ServerKey, a: &RadixCiphertext, b: &RadixCiphertext) -> RadixCiphertext {
    // a + b - a * b
    let a_add_b = k.add_parallelized(a, b);
    let a_mul_b = k.mul_parallelized(a, b);
    k.sub_parallelized(&a_add_b, &a_mul_b)
}

```

**Additional context**
```
[dependencies]
tfhe = { git = "https://github.com/zama-ai/tfhe-rs.git", rev = "251ee9aa0e011b2f377640c6ecfd727b6108ebc9", features = [ "boolean", "shortint", "integer", "aarch64-unix" ] }
```

tmontaigu:
Yes, thats because we don't actually know the value since its encrypted, we have some checks in place that were meant as safe guards, but that means the condition value must be the result of a comparison `lt, le, eq`, etc to work. 

But these checks are too limitating, and are probably going to remove them

MakisChristou:
> ... but that means the condition value must be the result of a comparison `lt, le, eq`, etc to work.


[...]

PR #617, opened 2023-10-09
fix(integer): remove remove if_then_else assert

tmontaigu:
unchecked_if_then_else had an assert that required that the condition value looked like it encrypts a boolean usind the degree

However, the only cases where a value looks like it encrypts a boolean value is when they are the result of a comparison (lt, le, eq, etc).

But there are other cases were the value holds a boolean value but due to how degree works, it's not possible to know thus limiting the use of if_then_else.

So we remove that assert, and rely on the developpe knowing its condition is 0 or 1.

Fixes: https://github.com/zama-ai/tfhe-rs/issues/615
Fixes: https://github.com/zama-ai/tfhe-rs-internal/issues/246

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/cmux.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/cmux.rs
@@ -12,11 +12,6 @@ impl ServerKey {
     where
         T: IntegerRadixCiphertext,
     {
-        assert!(
-            condition.holds_boolean_value(),
-            "The condition ciphertext does not encrypt a boolean (0 or 1) value"
-        );
-
         let condition_block = &condition.blocks()[0];
         self.unchecked_programmable_if_then_else_parallelized(
             condition_block,


IceTDrinker (approved):
Thanks, we'll check to add the "boolean" type down the line for integer

IceTDrinker:
should we maybe have a warning in the style of "undefined" behavior if the ciphertext is not a boolean ?

tmontaigu:
> _then_else.
> 
> So we remove that assert, and rely on the developpe knowing its condition is 0 or 1

You mean a runtime print ?

The doc of the function alreay explains the condition value must encrypt a zero or one

tmontaigu:
> > So we remove that assert, and rely on the developpe knowing its condition is 0 or 1
> 
 You mean a runtime print ?
 
The doc of the function alreay explains the condition value must encrypt a zero or one

PR #618, opened 2023-10-09
refactor(core): refactor conversion code from f64 to i64

IceTDrinker:
- observed that the subnormal case is already handled by the shift logic so the special handling was not required


IceTDrinker:
I think this is worth merging first as it's a less radical change

PR #619, opened 2023-10-09
feat(integer): speed-up division by using overflowing_sub

tmontaigu:
using overflowing sub allows to remove the comparison used in the algorithm, giving significant performance boost.

```
before
                8        16       32      40       64       128        256
hpc7a:    `981 ms` `2.53 s` `6.41 s` `9.04 s` `16.1 s` `39.3 s` `1.55 min`
m6i:        `1.10 s` `2.97 s` `7.17 s` `10.5 s` `19.7 s` `50.2 s` `2.11 min`

afer:
                8        16       32      40       64       128        256
hpc7a:   `604 ms` `1.6 s`  `3.8 s`  `5.14 s` `9.4 s`  `22.4 s`  `54.613 s`
m6i:      `659 ms` `1.77 s` `4.4 s`  `5.9 s`  `11.5 s` `29.8 s`  `79.95 s`
```

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/div_mod.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/div_mod.rs
@@ -2,8 +2,7 @@ use crate::integer::ciphertext::{IntegerRadixCiphertext, RadixCiphertext, Signed
 use crate::integer::server_key::comparator::ZeroComparisonType;
 use crate::integer::{IntegerCiphertext, ServerKey};
 
-use super::bit_extractor::BitExtractor;
-
+use crate::shortint::MessageModulus;
 use rayon::prelude::*;
 
 impl ServerKey {
@@ -73,6 +72,7 @@ impl ServerKey {
                     &quotient_minus_one,
                     &quotient,
                     |x| x == 2,
+                    true,
                 )
             },
             || {
@@ -81,6 +81,7 @@ impl ServerKey {
                     &remainder_plus_divisor,
                     &remainder,
                     |x| x == 2,
+                    true,
                 )
             },
         );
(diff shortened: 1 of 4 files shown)


tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
         //     Q(i) := 1
         //   end
         // end
+        assert_eq!(
```
IceTDrinker, line 117: why all the asserts compared to before ?

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
...
-        let merge_two_cmp_lut = self
-            .key
-            .generate_lookup_table_bivariate(|x, y| u64::from(x == 1 && y == 1));
-
-        let bit_extractor = BitExtractor::new(self, num_bits_in_block as usize);
+        let bit_extractor = BitExtractor::new(self, num_bits_in_message as usize);
         let numerator_bits = bit_extractor.extract_all_bits(&numerator.blocks);
 
+        // The overflow flag is computed by combining 2 separate values,
+        // this vec will contain the lut that merges this two flag.
```
IceTDrinker, line 160: this -> these

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
...
-            self.key
-                .unchecked_add_assign(&mut interesting_remainder.blocks[0], &numerator_bits[i]);
+            // We split the divisor at a block position, when in reality the split should be at a
+            // bit position meaning that potentially (depending on msb_bit_set) the
+            // split versions share some bits they should not. So we do one PBS on the
+            // last block of the interesting_divisor, and first block of divisor_ms_blocks
+            // to trim out bits which should not be there
+
+            let mut trim_last_interesting_divisor_bits = || {
+                if ((msb_bit_set + 1) % num_bits_in_message as usize) == 0 {
```
IceTDrinker, line 208: msb_bit_set + 1 seems to appear in several places, should it be stored in a variable with an explicit name ?

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
...
+                if divisor_ms_blocks.blocks.is_empty()
+                    || ((msb_bit_set + 1) % num_bits_in_message as usize) == 0
+                {
+                    return;
+                }
+                // As above, we need to zero out some bits, but here its in the
+                // first block of most significant blocks of the divisor.
+                // The block has the same value as the last block of interesting_divisor.
+                // Here we will zero out the bits that the spawn above
+                // wanted to keep.
```
IceTDrinker, line 245: spawn ? not sure I understand the sentence 🤔

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
...
+                let pos_in_block = msb_bit_set as u64 % num_bits_in_message;
+
+                // e.g 2 bits in message:
+                // if pos_in_block is 0, then we want to discard the first bit (left shift mask
+                // by 1) if pos_in_block is 1, then we want to discard the
+                // two bits (left shift mask by 2) let shift_amount =
+                // num_bits_in_message - pos_in_block as u64;
+                let shift_amount = pos_in_block + 1;
+                let full_message_mask = self.key.message_modulus.0 as u64 - 1;
+                let shifted_mask = full_message_mask << shift_amount;
```
IceTDrinker, line 257: this requires the carry bits to be empty, is it always the case ? I think there is an alternative where you can start with 1, and conditionally AND it with 0 while shifting it by one to only the select the bits you want

IceTDrinker (commented):
Some nitpicks, if you have some time to quickly discuss the changes you did, they seem pretty extensive and I'm not sure I got all of them

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
tmontaigu: To be more explicit on the requirements of the imputs, before some of them were implicit or assert later (like in the comparisons)

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
tmontaigu: Ah yes, it references the `trim_last_interesting_divisor_bits` function which I used to  call in a rayon spaw:
```rust
rayon::scope(|s| {
    s.spawn(||  ...   );
});
```

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
tmontaigu: It does not require empty carry bits, we use a non-bivariate PBS, and the mask computed masks the bits we want, which are some bits of the message part.

Its true tho, that instead of calling applu_lut_assign, I could call bitand_assign

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
IceTDrinker: I meant it may have some better human readable name if it does not, it's fine like this

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,134 +114,241 @@ impl ServerKey {
...
+
+            remainder.blocks[..first_trivial_block]
+                .iter_mut()
+                .zip(interesting_remainder.blocks.iter())
+                .for_each(|(remainder_block, new_value)| {
+                    remainder_block.clone_from(new_value);
+                });
         }
 
         // Even though, the quotient does not have any actual carries,
```
IceTDrinker, line 354: this does not look true anymore for the remainder, remainder is coming from interestring_remainder itself coming from the unchecked if_then_else without message cleaning it seems to me

tmontaigu: Yep, depends on if we can keep the fact of not cleaning in the if_then_else

tfhe/src/integer/server_key/radix_parallel/div_mod.rs
```diff
@@ -113,146 +114,381 @@ impl ServerKey {
...
+
+            let overflow_sum = self.key.unchecked_add(
+                &subtraction_overflowed,
+                &at_least_one_upper_block_is_non_zero,
+            );
+            // Give name to closures to improve readability
+            let overflow_happened = |overflow_sum: u64| overflow_sum != 0;
+            let overflow_did_not_happen = |overflow_sum: u64| !overflow_happened(overflow_sum);
+
+            // Here, we will do what zero_out_if does, but to stay within noise constrains,
```
IceTDrinker, line 382: constrains -> constraints

IceTDrinker (commented):
Looks good noise wise (hopefully I did not miss anything) small nitpicks in comments, let me know if you want to proceed without them

IceTDrinker (approved):
Thanks a lot !

tmontaigu:
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

tmontaigu:
@slab-ci cpu_integer_test

tmontaigu:
@slab-ci cpu_fast_test
@slab-ci cpu_test

IceTDrinker:
if we are getting sig kills we may want to change the number of cargo nextest processes we launch in parallel

IceTDrinker:
or limit the tests given the biggest server key size so that it fits in RAM, as the tests are multi threaded anyways we should not be losing too much hopefully

btw the test executor taking the key by value will make things worse right ? Could we maybe move the key in the executor and have an accessor on the executor to get a ref to the server key so that we don't have an extra copy in memory ?

tmontaigu:
> btw the test executor taking the key by value will make things worse right ? Could we maybe move the key in the executor and have an accessor on the executor to get a ref to the server key so that we don't have an extra copy in memory ?

Yes that should work

PR #621, opened 2023-10-10
Full propa test

tmontaigu:
### Check-list:

5

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
@@ -1,3 +1,4 @@
+use crate::integer::block_decomposition::BlockDecomposer;
 use crate::integer::keycache::KEY_CACHE;
 use crate::integer::{IntegerRadixCiphertext, RadixCiphertext, RadixClientKey, ServerKey};
 use crate::shortint::parameters::*;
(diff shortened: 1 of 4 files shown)


tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -3096,3 +3096,155 @@ where
...
+            "Invalid degree after propagation"
+        );
+    }
+
+    if block_carry_mod == block_msg_mod {
+        // This test is easier to write with this assumption
+        // which, conveniently is true for our radix type
+        //
+        // In this test, we are creating a ciphertext which is at full capacity
+        // which just enough room that allows sequential (non-parallel)
```
tmontaigu, line 3157: with just enough

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -3096,3 +3096,155 @@ where
...
+        );
+
+        // Take the initial value, but remove any bits below absober block
+        // as the bits below will have changed, but bits above will not.
+        let mut expected_built_by_hand =
+            initial_value & (u64::MAX << ((absorber_block_index + 1) as u32 * num_bits_in_msg));
+        // The first block generated a carry,
+        // but also results in a non zero block.
+        //
+        // The carry gets propagated until by other blocks
```
tmontaigu, line 3242: `get propagated by other blocks`

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -3096,3 +3096,159 @@ where
...
+        );
+        assert!(
+            ct.blocks
+                .iter()
+                .all(|b| b.degree.0 as u64 == block_msg_mod - 1),
+            "Invalid degree after propagation"
+        );
+    }
+
+    if block_carry_mod == block_msg_mod {
```
IceTDrinker, line 3152: block_carry_mod >= block_msg_mod ? I don't know if we have such parameters though

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -3096,3 +3096,159 @@ where
...
+        let msg = cks.encrypt(clear_max_value);
+        let mut ct = cks.encrypt(clear_max_value);
+        while sks.is_add_possible(&ct, &msg) {
+            sks.unchecked_add_assign(&mut ct, &msg);
+            expected_result = expected_result.wrapping_add(clear_max_value) % modulus;
+        }
+        assert!(ct
+            .blocks
+            .iter()
+            .all(|b| b.degree.0 as u64 == (block_total_mod - 1) - (block_msg_mod - 1)),);
```
IceTDrinker, line 3171: `- (block_carry_mod -1)` ?

IceTDrinker (commented):
Will check the expected built by hand tomorrow some small comments

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -3096,3 +3096,159 @@ where
...
+        // but also results in a non zero block.
+        //
+        // The carry gets propagated by other blocks
+        // until it hits the absorber block, which takes the value of the carry
+        // (1) as its new value. Blocks that propagated the carry will have as new value
+        // 0 as for these block we did: ((block_msg_mod - 1  + 1) % block_msg_modulus) == 0
+        // and carry = ((block_msg_mod - 1  + 1) / block_msg_modulus) == 1
+        //
+        // Set the value of first block
+        expected_built_by_hand |= (2 * (block_msg_mod - 1)) % block_msg_mod;
```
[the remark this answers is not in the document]
IceTDrinker: nevermind, my bad

IceTDrinker:
looks good, just the minor comments for which I may be wrong

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: Yes I think `(block_carry_mod -1)` is the correct thing if we want to take into account  `block_carry_mod >= block_msg_mod`, but then In that case our full_propagate won't be a real full_propagate and so this test won't be correct

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: We have for shortints, but we don't use them in integers.

In integers params needs as many carry bit as message bits to have all operations
But if their are more carry bits than message bits, then we would need 2 full_propagation
to actually fully propagate.

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: Not sure about that, need to think about it 🤔

tmontaigu: Ok so in the case where params are such that `block_carry_mod > block_msg_mod`, the sequential propagation is fine as long as is_op_possible do proper checks.

The parallel carry propagation would require to be adapted to support that case

IceTDrinker: feel free to open follow up issues

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
IceTDrinker: ok so this is correct with our current implementation of full propagate for integers right ?

tmontaigu: Ok Finally plugged my brain,

So using params where `block_carry_mod >= block_msg_mod`, the full_propagations (both sequential and parallel)
will correctly work, so as long as the block has enough space to receive the carry of the next block.
Which is what is_thing_possible where reworked to ensure (there is probably is_scalar_thing_possible left to update).

What confused me, is that the parallel carry propagation also works  by default in the even in the case where the whole block is completely full.

For example, in 2_2, the sequential propagation requires that there is enough room to receive worst case carry of preceding block, so max degree is 15 - 3 = 12. Using the parallel algorithm we can propagate carries where blocks
have a degree of 15 (ie degree == max_degree), however that's only true if `carry_modulus <= message_modulus`.

Parallel propagation could be improved to support the case where `carry_modulus  > message_modulus` and we sum until degree = (message_modulus * message_modulus)- 1.

tmontaigu: I'm adding PARAM_MESAGE_2_CARRY_3 in the instanciated tests

IceTDrinker: ok !

IceTDrinker (approved):
Thanks ! Looks good !

tmontaigu:
@slab-ci cpu_test

tmontaigu:
@slab-ci cpu_integer_test

tmontaigu:
Realized that the test was not meaningful as it did not check manually each block to verify its value (e.g no carry after propagation). Checking the result of the decryption is not a complete check, because decrypting propagtes carries, so an implemenation that did nothing would pass the tests

IceTDrinker (approved):
Looks good, and good catch on decryption doing the propagate automatically

tmontaigu:
@slab-ci cpu_integer_test

IceTDrinker:
@slab-ci cpu_integer_test

PR #623, opened 2023-10-11
fix(crt): fix mul for non symmetrical parameters

IceTDrinker:
- add non reg test for 32 bits mul with 5_1 parameters


IceTDrinker:
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test

PR #628, opened 2023-10-13
feat(core): specialize keyswitch implementation for small scalar values

sarah-ek:

### PR content/description

this improves the keyswitch performance by 35% with avx2, and 15% with avx512

### Check-list:

### DIFF
--- a/tfhe/benches/core_crypto/ks_bench.rs
+++ b/tfhe/benches/core_crypto/ks_bench.rs
@@ -1,0 +1,91 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use tfhe::core_crypto::prelude::*;
+use tfhe::keycache::NamedParam;
+use tfhe::shortint::prelude::*;
+
+fn criterion_bench(criterion: &mut Criterion) {
+    type Scalar = u64;
+
+    let mut bench_group = criterion.benchmark_group("KS");
+    bench_group
+        .sample_size(15)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    for params in [
+        PARAM_MESSAGE_1_CARRY_1_KS_PBS,
+        PARAM_MESSAGE_2_CARRY_2_KS_PBS,
+        PARAM_MESSAGE_3_CARRY_3_KS_PBS,
+        PARAM_MESSAGE_4_CARRY_4_KS_PBS,
+    ]
+    .into_iter()
+    {
+        let lwe_dimension = params.lwe_dimension;
+        let lwe_modular_std_dev = params.lwe_modular_std_dev;
+        let ciphertext_modulus = params.ciphertext_modulus;
+        let encoding_with_padding = if ciphertext_modulus.is_native_modulus() {
+            Scalar::ONE << (Scalar::BITS - 1)
+        } else {
+            Scalar::cast_from(ciphertext_modulus.get_custom_modulus() / 2)
+        };
+        let glwe_dimension = params.glwe_dimension;
+        let polynomial_size = params.polynomial_size;
+        let ks_decomp_base_log = params.ks_base_log;
+        let ks_decomp_level_count = params.ks_level;
+        let msg_modulus: Scalar = params.message_modulus.0.cast_into();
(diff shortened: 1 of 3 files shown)


tfhe/src/core_crypto/algorithms/slice_algorithms.rs
```diff
@@ -251,9 +251,73 @@ pub fn slice_wrapping_sub_scalar_mul_assign<Scalar>(
...
+            spec_constant!(!6);
+            spec_constant!(!7);
+            spec_constant!(!8);
+            spec_constant!(!9);
+            spec_constant!(!10);
+            spec_constant!(!11);
+            spec_constant!(!12);
+            spec_constant!(!13);
+            spec_constant!(!14);
+            spec_constant!(!15);
```
IceTDrinker, line 295: I think a small comment here explaining that !0 == -1 ... !15 == -16 could be welcome :)

do you think there would be a way to write a match statement ? I know the scalar type makes it awkward to work with a bit, just curious

tfhe/benches/core_crypto/ks_bench.rs
```diff
@@ -0,0 +1,81 @@
...
+        let lwe_dimension = params.lwe_dimension;
+        let lwe_modular_std_dev = params.lwe_modular_std_dev;
+        let ciphertext_modulus = params.ciphertext_modulus;
+        let encoding_with_padding = Scalar::ONE << (Scalar::BITS - 1);
+        let glwe_dimension = params.glwe_dimension;
+        let polynomial_size = params.polynomial_size;
+        let ks_decomp_base_log = params.ks_base_log;
+        let ks_decomp_level_count = params.ks_level;
+
+        let msg_modulus = Scalar::ONE << 4;
```
IceTDrinker, line 25: should probably use the message modulus from the parameter set

tfhe/src/core_crypto/algorithms/slice_algorithms.rs
sarah-ek: i don't think it'll work with a match statement. it'll end up being something like
```rust
match scalar {
  n if n == Scalar::cast_from(0) => { ... }
  n if n == Scalar::cast_from(1) => { ... }
  // ...
```

IceTDrinker:
updated with some comments I made, will ask for reviews so that we can merge this significant improvement for keyswtich :)

tfhe/src/core_crypto/algorithms/slice_algorithms.rs
```diff
@@ -245,15 +250,93 @@ pub fn slice_wrapping_sub_scalar_mul_assign<Scalar>(
...
+    impl<Scalar: UnsignedInteger> pulp::NullaryFnOnce for Impl<'_, Scalar> {
+        type Output = ();
+
+        #[inline(always)]
+        fn call(self) -> Self::Output {
+            let Self { lhs, rhs, scalar } = self;
+
+            macro_rules! spec_constant {
+                ($constant: expr) => {
+                    if scalar == Scalar::cast_from($constant) {
```
mayeul-zama, line 268: I think it would be cleaner to have
```
let signed_constant: i128 = $constant;
Scalar::cast_from(signed_constant as u128)
```
so that you can have simple signed constants without type annotations in the macro call

IceTDrinker: just cast the constant directly in the Scalar::cast_from

mayeul-zama: Defining `signed_constant` does not force us to add type annotations in the macro call

IceTDrinker: I'm not a fan of the multiple let statements this creates 🤔 plus in rust most of the time signed constants are i32 so yeah I think I like the fact it is explicit here ?

IceTDrinker: plus I put a comment above the spec constant chain

IceTDrinker:
@slab-ci cpu_test

PR #631, opened 2023-10-13
chore(doc): encourage users to use dedicated keys to Radix or CRT

IceTDrinker:


IceTDrinker:
> There is one thing that slightly bothers me, is that RadixClientKey takes a number of blocks, so if you create a radix client key giving it 4 blocks, you will only be able to encrypt ciphertexts with 4 blocks. Maybe that would be too restrictive for some cases

What do you suggest then to make this more user friendly and usable ?

IceTDrinker:
we can add a method with an "encrypt with custom block number" or something of the sort

tmontaigu (approved):

PR #644, opened 2023-10-19
refactor(c_api): do not require a seed in the multi bit PBS keygen

IceTDrinker:
BREAKING CHANGE:
function signature has changed, this is a breaking change


IceTDrinker:
apparently it's not clear we want to give up the control of the seed

IceTDrinker:
closing for now as there was a miscommunication

PR #645, opened 2023-10-19
chore(ci): enable integer multi bit tests on M1

IceTDrinker:


IceTDrinker:
@slab-ci cpu_fast_test

soonum (approved):
Let's gooo

PR #651, opened 2023-10-23
add noise level

mayeul-zama:


tfhe/src/shortint/server_key/mod.rs
```diff
@@ -276,6 +277,7 @@ impl ServerKey {
             carry_modulus: self.carry_modulus,
             degree: Degree(self.message_modulus.0 - 1),
             pbs_order: self.pbs_order,
+            noise_level: NoiseLevel::NOMINAL,
```
IceTDrinker, line 280: Let's open an issue for the conformance to manage NoiseLevel Ranges as valid ciphertexts may have noise levels which are not nominal (like being loaded from disk after some unchecked operations or equivalent), those ciphertexts can very well be valid still

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -777,6 +780,8 @@ impl ShortintEngine {
         trivially_encrypt_lwe_ciphertext(&mut ct.ct, encoded);
 
         ct.degree = Degree(modular_value);
+        ct.noise_level = NoiseLevel::NOMINAL;
+
```
IceTDrinker, line 784: trivial would have a noise of 0 ?

IceTDrinker (changes requested):
Missing tests

tfhe/src/shortint/parameters/multi_bit.rs
```diff
@@ -66,6 +66,8 @@ impl MultiBitPBSParameters {
 
         let degree = Degree(message_modulus.0 - 1);
 
+        let noise_level = NoiseLevel::NOMINAL;
```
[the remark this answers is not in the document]
mayeul-zama: To build the `struct` a few lines later

tfhe/src/shortint/engine/server_side/mod.rs
mayeul-zama: Yes, fixed

tfhe/src/shortint/engine/server_side/mod.rs
```diff
@@ -756,6 +758,7 @@ impl ShortintEngine {
             message_modulus: server_key.message_modulus,
             carry_modulus: server_key.carry_modulus,
             pbs_order: server_key.pbs_order,
+            noise_level: NoiseLevel::NOMINAL,
```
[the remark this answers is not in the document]
mayeul-zama: Yes, fixed

tfhe/src/shortint/server_key/mod.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/300

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -83,6 +123,7 @@ impl Degree {
 pub struct Ciphertext {
     pub ct: LweCiphertextOwned<u64>,
     pub degree: Degree,
+    pub noise_level: NoiseLevel,
```
tmontaigu, line 126: So as we know adding this field breaks deserialization

Do we plan to do something to mitigate that ?

IceTDrinker: with the 0.5.0 we will do the semver trick, we will most likely provide a way to load old ciphertext and set the noise to the max for that ciphertext or do a bootstrapping right away I would say

for 0.5.0 I would try to break as many things as possible

tfhe/src/shortint/engine/server_side/add.rs
```diff
@@ -21,6 +21,7 @@ impl ShortintEngine {
     ) -> EngineResult<()> {
         lwe_ciphertext_add_assign(&mut ct_left.ct, &ct_right.ct);
         ct_left.degree = Degree(ct_left.degree.0 + ct_right.degree.0);
+        ct_left.noise_level += ct_right.noise_level;
```
IceTDrinker, line 24: this needs a test

tfhe/src/shortint/engine/server_side/sub.rs
```diff
@@ -49,6 +49,7 @@ impl ShortintEngine {
 
         lwe_ciphertext_add_assign(&mut ct_left.ct, &neg_right.ct);
 
+        ct_left.noise_level += ct_right.noise_level;
```
IceTDrinker, line 52: this as well

tfhe/src/shortint/engine/server_side/scalar_mul.rs
```diff
@@ -24,6 +24,7 @@ impl ShortintEngine {
         match scalar {
             0 => {
                 trivially_encrypt_lwe_ciphertext(&mut ct.ct, Plaintext(0));
+                ct.noise_level = NoiseLevel::ZERO;
```
IceTDrinker, line 27: and for this one as well unless I missed them

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,31 @@
+use crate::shortint::ciphertext::NoiseLevel;
+use crate::shortint::keycache::KEY_CACHE;
+use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+
+#[test]
+fn test_noise_level_propagation() {
```
IceTDrinker, line 6: one function per operator being tested please, with corner cases (i.e. 0 and other values for scalars)

IceTDrinker: and when I say per operator, it includes, default, checked, smart and unchecked with assign versions to be sure we cover all of the paths

IceTDrinker: except smart versions because of the mutable references potentially but it would automate most of the rest to check all paths

tfhe/src/shortint/parameters/multi_bit.rs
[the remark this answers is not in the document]
IceTDrinker: ok the other parts of the code were not following that style I believe in your PR

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: @mayeul-zama didn't we say we want the noise level to be private to be able to crash with a given feature if the noise level exceeds a valid noise level ?

mayeul-zama: I don't remember that
I don't see how it's needed to have this crash feature

IceTDrinker: how do you guarantee we never go over the limit without a setter ?

mayeul-zama: I was thinking of adding the assert in the few places where it's updated
But indeed your proposition is safer
I'll implement it

IceTDrinker: The setter will likely need an input to indicate the max noise level, which will require an input from the Server Key most likely, the extra input would be ignored if we are not checking (i.e. if the noise assert feature is disabled) but used otherwise

once we have that I believe most pieces of the puzzle will be ready

IceTDrinker: there remains a question for the maximum level of noise that should be stored in parameters

IceTDrinker: or each ciphertext brings it's own max noise level which may be a bit wasteful

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+fn test_1_ct_noise_level_propagation(sk: &ServerKey, ct: &Ciphertext) {
+    let test_fn = |f: &dyn Fn(&ServerKey, &Ciphertext) -> Ciphertext,
+                   g: &dyn Fn(NoiseLevel) -> NoiseLevel| {
+        assert!(f(sk, ct).noise_level == g(ct.noise_level));
+    };
+
+    test_fn(&ServerKey::unchecked_neg, &|a| a);
+    test_fn(
+        &|sk, ct| ServerKey::unchecked_neg_with_correcting_term(sk, ct).0,
+        &|a| a,
```
IceTDrinker, line 15: maybe rename a to "noise_is_unchanged" or "input_ct_noise" so that reading the test is made easier, as here we don't have a Noise expression (same thing elsewhere)

It's a detail but should help readability if ever we are down there debugging noise stuff

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+    );
+}
+
+fn test_2_ct_noise_level_propagation(sk: &ServerKey, ct1: &Ciphertext, ct2: &Ciphertext) {
+    let test_fn = |f: &dyn Fn(&ServerKey, &Ciphertext, &Ciphertext) -> Ciphertext,
+                   g: &dyn Fn(NoiseLevel, NoiseLevel) -> NoiseLevel| {
+        assert!(f(sk, ct1, ct2).noise_level == g(ct1.noise_level, ct2.noise_level));
+    };
+
+    test_fn(&ServerKey::unchecked_add, &|a, b| a + b);
```
IceTDrinker, line 56: rename a and b to lhs_noise and rhs_noise this way we know what's happening, same thing everywhere

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
+use crate::shortint::ciphertext::NoiseLevel;
+use crate::shortint::keycache::KEY_CACHE;
+use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+use crate::shortint::{Ciphertext, ServerKey};
+
+fn test_1_ct_noise_level_propagation(sk: &ServerKey, ct: &Ciphertext) {
```
IceTDrinker, line 6: test_unary_op_... ? it's a proposal but I think it's the notation we use in the C api IIRC, the end of the name is good 👍 , nitpick

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    let acc = sk.generate_lookup_table(|_| 0);
+
+    test_fn(
+        &|sk, ct| ServerKey::apply_lookup_table(sk, ct, &acc),
+        &|_| NoiseLevel::NOMINAL,
+    );
+}
+
+fn test_1_ct_assign_noise_level_propagation(sk: &ServerKey, ct: &Ciphertext) {
```
IceTDrinker, line 26: test_unary_op_assign_... ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    let acc = sk.generate_lookup_table(|_| 0);
+
+    test_fn(
+        &|sk, ct| ServerKey::apply_lookup_table_assign(sk, ct, &acc),
+        &|_| NoiseLevel::NOMINAL,
+    );
+}
+
+fn test_2_ct_noise_level_propagation(sk: &ServerKey, ct1: &Ciphertext, ct2: &Ciphertext) {
```
IceTDrinker, line 50: test_binary_op_... ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+    let test_fn = |f: &dyn Fn(&ServerKey, &Ciphertext, &Ciphertext) -> Ciphertext,
+                   g: &dyn Fn(NoiseLevel, NoiseLevel) -> NoiseLevel| {
+        assert!(f(sk, ct1, ct2).noise_level == g(ct1.noise_level, ct2.noise_level));
+    };
```
IceTDrinker, line 54: can this be made a function with #[track_caller] so that we have the test triggering the error if a test fails ? unless closures can be track_caller bu I don't think it's the case, to combine with the naming improvement for f and g proposed earlier in the file and applied to all similar test_fn

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    test_fn(&ServerKey::unchecked_add, &|a, b| a + b);
+    test_fn(&ServerKey::unchecked_sub, &|a, b| a + b);
+    test_fn(
+        &|sk, ct1, ct2| ServerKey::unchecked_sub_with_correcting_term(sk, ct1, ct2).0,
+        &|a, b| a + b,
+    );
+
+    let any_trivially_encrypted = ct1.degree.0 == 0 || ct2.degree.0 == 0;
```
IceTDrinker, line 63: isn't this better detected with the Noise level being 0 ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+    });
+    test_fn(&ServerKey::unchecked_not_equal, &|_, _| NoiseLevel::NOMINAL);
+    test_fn(
+        &|sk, ct1, ct2| ServerKey::unchecked_evaluate_bivariate_function(sk, ct1, ct2, |_, _| 0),
+        &|_, _| NoiseLevel::NOMINAL,
+    );
+}
+
+fn test_2_ct_assign_noise_level_propagation(sk: &ServerKey, ct1: &Ciphertext, ct2: &Ciphertext) {
+    let test_fn = |f: &dyn Fn(&ServerKey, &mut Ciphertext, &Ciphertext),
```
IceTDrinker, line 103: comments of test_fn apply here

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+                   g: &dyn Fn(NoiseLevel, NoiseLevel) -> NoiseLevel| {
+        let mut clone = ct1.clone();
+        f(sk, &mut clone, ct2);
+        assert!(clone.noise_level == g(ct1.noise_level, ct2.noise_level));
+    };
+
+    test_fn(&ServerKey::unchecked_add_assign, &|a, b| a + b);
+    test_fn(&ServerKey::unchecked_sub_assign, &|a, b| a + b);
```
IceTDrinker, line 111: lhs_noise, rhs_noise and similar later in the function

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+    test_fn(
+        &|sk, ct1, ct2| {
+            ServerKey::unchecked_sub_with_correcting_term_assign(sk, ct1, ct2);
+        },
+        &|a, b| a + b,
+    );
+
+    let any_trivially_encrypted = ct1.degree.0 == 0 || ct2.degree.0 == 0;
```
IceTDrinker, line 119: remark on the noise == 0 for trivial

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+        NoiseLevel::NOMINAL
+    });
+    test_fn(&ServerKey::unchecked_not_equal, &|_, _| NoiseLevel::NOMINAL);
+    test_fn(
+        &|sk, ct1, ct2| ServerKey::unchecked_evaluate_bivariate_function(sk, ct1, ct2, |_, _| 0),
+        &|_, _| NoiseLevel::NOMINAL,
+    );
+}
+
+fn test_2_ct_assign_noise_level_propagation(sk: &ServerKey, ct1: &Ciphertext, ct2: &Ciphertext) {
```
IceTDrinker, line 102: binary_op_assign

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+        NoiseLevel::NOMINAL
+    });
+}
+
+#[test]
+fn test_noise_level_propagation() {
+    let keys = KEY_CACHE.get_from_param(PARAM_MESSAGE_2_CARRY_2_KS_PBS);
+    let (ck, sk) = (keys.client_key(), keys.server_key());
+
+    let trivial = sk.create_trivial(0);
```
IceTDrinker, line 240: let's have some randomized values in addition to a hard coded 0 for the trivial encryption as it's better detected with noise level == 0

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    let ct2 = ck.encrypt(0);
```
IceTDrinker, line 242: let's have randomized values for ct2

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
...
+    test_fn(&ServerKey::unchecked_scalar_right_shift_assign, &|_, _| {
+        NoiseLevel::NOMINAL
+    });
+}
+
+#[test]
+fn test_noise_level_propagation() {
+    let keys = KEY_CACHE.get_from_param(PARAM_MESSAGE_2_CARRY_2_KS_PBS);
+    let (ck, sk) = (keys.client_key(), keys.server_key());
+
```
IceTDrinker, line 239: let's have several test runs with the randomized values in addition to the 0s

IceTDrinker (commented):
Some minor stuff on testing (naming) and some randomized tests

IceTDrinker:
thanks for the base of the testing infrastructure, some improvements can be made I believe but we are nearly there

tfhe/src/shortint/ciphertext/mod.rs
tmontaigu: Maybe putting an assert on the noise before bootstrap might be enough to catch errors since bootstraps are very frequent and the whole norm2 is computed around that ?

IceTDrinker: that's a very good point

IceTDrinker: it might make identifying the faulty operation harder though if you have a chain of adds, so... not sure I like it much

mayeul-zama: Yes, and if you too many linear operations without a PBS after, the assert won't be triggered

tmontaigu: Yeah its less precise and uses the fact that in practive you are very limited in the number of leveled operation you can do

mayeul-zama: But making it private forces the use of a constructor instead of a normal struct construction which is less ergonomic IMO

IceTDrinker: yes but that's a very small price to pay, even though it crossed my mind

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,263 @@
+use crate::shortint::ciphertext::NoiseLevel;
+use crate::shortint::keycache::KEY_CACHE;
+use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+use crate::shortint::{Ciphertext, ServerKey};
+
+fn test_1_ct_noise_level_propagation(sk: &ServerKey, ct: &Ciphertext) {
+    let test_fn = |f: &dyn Fn(&ServerKey, &Ciphertext) -> Ciphertext,
+                   g: &dyn Fn(NoiseLevel) -> NoiseLevel| {
```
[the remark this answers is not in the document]
mayeul-zama: Renamed `op` and `predicate`

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Renamed `ct_noise`

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Renamed `ct1_noise` and `ct2_noise`

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Does not seem to work on closures

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: IMO it's better to use the same condition as in the op themselves
Otherwise the assert may not pass

tfhe/src/shortint/server_key/tests/noise_level.rs
IceTDrinker: can't the test_fn be defined like

```rust
#[track_caller]
fn test_fn(...) {
}
```
 ?

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/tests/noise_level.rs
IceTDrinker: ok but it means we may need to progressively improve the operations handling of trivial having a fn is_trivial on ciphertexts checking their noise, can you open a follow up issue please ?

IceTDrinker: with a comment on those checks with a TODO

mayeul-zama: The ops multiplication optimization needs one of the encrypted value to be `0` which is independent (a priori) from its noise level to be 0.
Other optimizations of trivial encryptions could be interesting but not in this case I think

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: To keep the rest of the code simple (not passing sk, ct1, ct2 explicitely everywhere), `test_fn` (as you just defined it) would have to be wrapped in a closure

IceTDrinker: Ok should be fine with backtrace anyways

tfhe/src/shortint/server_key/tests/noise_level.rs
IceTDrinker: Did we open a follow up issue to have a is_trivial function on shortint ?

IceTDrinker: if we don't have an issue we need one, and yes to keep the check consistent with ops for now

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+    test_fn(&ServerKey::unchecked_scalar_bitand_assign, &|_, _| {
+        NoiseLevel::NOMINAL
+    });
+    test_fn(&ServerKey::unchecked_scalar_bitor_assign, &|_, _| {
+        NoiseLevel::NOMINAL
+    });
+    test_fn(&ServerKey::unchecked_scalar_bitxor_assign, &|_, _| {
+        NoiseLevel::NOMINAL
+    });
+    if scalar < 8 {
```
IceTDrinker, line 245: why this special case ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+        NoiseLevel::NOMINAL
+    });
+}
+
+#[test]
+fn test_noise_level_propagation() {
+    let keys = KEY_CACHE.get_from_param(PARAM_MESSAGE_2_CARRY_2_KS_PBS);
+    let (ck, sk) = (keys.client_key(), keys.server_key());
+
+    let modulus: u64 = 4;
```
IceTDrinker, line 261: get from params

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    for _ in 0..10 {
+        let trivial = sk.create_trivial(rand::random::<u64>() % modulus);
```
IceTDrinker, line 264: let's have a trivial 0 as well as a trivial random please, Maybe I was not clear in my previous comment :)

IceTDrinker (commented):
Minor details, next one should be good to merge

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        for ct1 in [&trivial, &ct1, &ct2] {
+            for ct2 in [&trivial, &ct1, &ct2] {
+                test_ct_binary_op_noise_level_propagation(sk, ct1, ct2);
+                test_ct_binary_op_assign_noise_level_propagation(sk, ct1, ct2);
+            }
+        }
+
+        for ct in [&trivial, &ct1, &ct2] {
+            for scalar in 0..12 {
```
IceTDrinker, line 282: could we have a scalar based on the parameter set ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+        NoiseLevel::NOMINAL
+    });
+    test_fn(&ServerKey::unchecked_not_equal, &|_, _| NoiseLevel::NOMINAL);
+    test_fn(
+        &|sk, ct1, ct2| ServerKey::unchecked_evaluate_bivariate_function(sk, ct1, ct2, |_, _| 0),
+        &|_, _| NoiseLevel::NOMINAL,
+    );
+}
+
+fn test_ct_binary_op_assign_noise_level_propagation(
```
IceTDrinker, line 106: we don't have an equivalent for smarts ? smart would need two things, clean ciphertexts and then ciphertexts with filled degree which can be simulated by setting the ct degree to Degree(msg_mod * carry_mod - 1)

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+
+fn test_ct_binary_op_noise_level_propagation(sk: &ServerKey, ct1: &Ciphertext, ct2: &Ciphertext) {
+    let test_fn = |op: &dyn Fn(&ServerKey, &Ciphertext, &Ciphertext) -> Ciphertext,
+                   predicate: &dyn Fn(NoiseLevel, NoiseLevel) -> NoiseLevel| {
+        assert!(op(sk, ct1, ct2).noise_level() == predicate(ct1.noise_level(), ct2.noise_level()));
+    };
+
+    test_fn(&ServerKey::unchecked_add, &|ct1_noise, ct2_noise| {
+        ct1_noise + ct2_noise
+    });
```
IceTDrinker, line 58: I think you can easily add the default ops ?

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -9,10 +9,50 @@ use std::fmt::Debug;
 
 use super::parameters::{CiphertextConformanceParams, CiphertextListConformanceParams};
 
-/// This tracks the number of operations that has been done.
+/// This tracks the max degree that can be in a Ciphertext.
```
tmontaigu, line 12: Was this change meant ? I'm worried the using `max degree` might create confusions with the real max degree

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    let modulus: u64 = 4;
+
+    for _ in 0..10 {
+        let trivial = sk.create_trivial(rand::random::<u64>() % modulus);
+
+        let ct1 = ck.encrypt(rand::random::<u64>() % modulus);
+        let ct2 = sk.add(&ct1, &ct1);
+
+        for ct in [&trivial, &ct1, &ct2] {
```
tmontaigu, line 269: I think here we have that trivial has its noise at 0, ct1 has its noise at nominal, ct2 also has its noise as nominal since its the result from a default ops, which is made using PBS

It would be better if ct2 had non nominal noise (which Is what you meant I think), so it should be `let ct2 = sk.unchecked_add(&ct1, &ct1)`;

tfhe/src/shortint/ciphertext/mod.rs
IceTDrinker: I think the docstring is not correct, we want the degree in the ct

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Indeed, that's what I had in mind
Thanks!

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: Indeed max degree is a ambiguous name.
IMO the name `degree` is ambiguous itself because it's not well defined (setting the degree to `msg_modulus-1` after a comparison instead of `1` is correct even if not optimal)

What do you suggest replacing this comment by?
Maybe we could add a paragraph explaining what is the degree and the subtleties

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: I plan to handle the noise level for operations (other than unchecked) and associated tests in subsequent PRs
In this PR, the noise level is updated but never used

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: See https://github.com/zama-ai/tfhe-rs/pull/651#discussion_r1377656150

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: To avoid an overflow in `let scalar = 1_u8 << shift;`

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -156,13 +202,39 @@ impl Clone for Ciphertext {
...
         self.degree.0 < self.message_modulus.0
     }
+
+    pub fn noise_level(&self) -> NoiseLevel {
+        self.noise_level
+    }
+
+    pub fn set_noise_level(&mut self, noise_level: NoiseLevel) {
+        self.noise_level = noise_level
+    }
```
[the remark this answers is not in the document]
mayeul-zama: I plan to add it in a subsequent PR

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Added

tfhe/src/shortint/server_key/tests/noise_level.rs
tmontaigu: He probably meant taking a scalar in range 0..message_modulus or 0..carry_modulus * message_modulus

tfhe/src/shortint/ciphertext/mod.rs
tmontaigu: One point is that having the noise separated from the degree, is that after comparisons degree can be set to 1.

I don't think its necessary to change the docstring of degree now, it would make sense once the noise integration is done. Then it should be simpler to explain as degree would just be the max possible value

tmontaigu:
Changes looks good to me

tfhe/src/shortint/server_key/tests/noise_level.rs
mayeul-zama: Done

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: Reverted

tmontaigu (approved):

tfhe/src/shortint/server_key/tests/noise_level.rs
IceTDrinker: ok thanks

tfhe/src/shortint/server_key/tests/noise_level.rs
IceTDrinker: let's have a follow up issue if we don't have one :)

tfhe/src/shortint/ciphertext/mod.rs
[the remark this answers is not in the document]
IceTDrinker: ok ! follow up issue please :)

IceTDrinker (approved):
Looks good thanks !

tfhe/src/shortint/ciphertext/mod.rs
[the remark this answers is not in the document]
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/313

PR #654, opened 2023-10-25
fix(integer): fix worst case noise growth in encrypted shifts

tmontaigu:
In encrypted shifts we pack 3 bits from 3 different blocks into the same blocks by doing `b0 * 4 + 2 * b1 + b2`, and then do a PBS to simulate a hardware mux gate.

If the inputs of shift (ie, in lhs << rhs, lhs != rhs, ie we don't do lhs << lhs) this is fine regarding the norm2 noise.

However if we do things like `a << a` or `a >> a`, which is probably a very rare thing but not impossible, the norm2 noise would go above the limit that guarantees our error probability.

To fix that, we extract the bits that tells shift amount, so that they are already properly aligned to their mux input position. The packing becomes `b0 + 2 * b1 + b2` and so,
the noise growth is ok even in the worst case of doind `a << a`.

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/bit_extractor.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/bit_extractor.rs
@@ -12,9 +12,29 @@ pub(crate) struct BitExtractor<'a> {
 
 impl<'a> BitExtractor<'a> {
     pub(crate) fn new(server_key: &'a ServerKey, bits_per_block: usize) -> Self {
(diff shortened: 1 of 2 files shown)


IceTDrinker (approved):
Looks good thanks !

tmontaigu:
@slab-ci cpu_integer_test

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_wasm_test

PR #655, opened 2023-10-25
Boolean value wrapper

tmontaigu:
### PR content/description

The BooleanValue wrapper type is meant to convey the fact that
the ciphertext encrypts a 0 or 1.

Since its meant to be a simple wrapper, the goal for is to be flexible
and not add more burden than usefulness.

Hopefully this implementation somehow achieves that

Breaking Changes:
 - This changes the return type of comparisons from a T to
   a BooleanBlock. Requiring existing code to explicitely convert
   using `.into_radix`.
 - This makes the cmux/if_then_else functions take a BooleanValue
   as the input type  Requiring existing code to wrap their condition
   ciphertext in a new BooleanValue

### Check-list:

### DIFF
--- a/tfhe/src/integer/ciphertext/boolean_value.rs
+++ b/tfhe/src/integer/ciphertext/boolean_value.rs
@@ -1,0 +1,133 @@
+use super::{IntegerCiphertext, IntegerRadixCiphertext};
+use crate::integer::{RadixCiphertext, ServerKey};
+use crate::shortint::Ciphertext;
+use serde::{Deserialize, Serialize};
+
+/// Wrapper type used to signal that the inner value encrypts 0 or 1
+///
+/// Since values ares encrypted, it is not possible to know whether a
+/// ciphertext encrypts a boolean value (0 or 1). However some algorithms
+/// require that the ciphertext does indeed encrypt a boolean value.
+///
+/// This wrapper serves as making it explicit that it is known that the value
+/// encrypted is 0 or 1. And that if a function taking a BooleanBlock as input
+/// returns incorrect value, it may be due to the value not really being 0 or 1.
+///
+/// Also some function such as comparisons are known to return an encrypted value
+/// that is either 0 or 1, and thus return a Ciphertext wrapped in a [BooleanBlock].
+///
+/// # Examples
+/// ## Converting a [BooleanBlock] to a [RadixCiphertext]
+///
+/// ```
+/// use tfhe::integer::{gen_keys_radix, BooleanBlock};
+/// use tfhe::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+///
+/// // We have 4 * 2 = 8 bits of message
+/// let size = 4;
+/// let (cks, sks) = gen_keys_radix(PARAM_MESSAGE_2_CARRY_2_KS_PBS, size);
+///
+/// let a = 128u8;
+/// let b = 55u8;
+///
+/// let ct_a = cks.encrypt(a);
+/// let ct_b = cks.encrypt(b);
+///
+/// let ct_res = sks.ge_parallelized(&ct_a, &ct_b);
+/// // Convert the boolean value to a RadixCiphertext of size blocks
+/// // so we can use it in operations
+/// let ct_res = ct_res.into_radix(size, &sks);
+/// // Decrypt:
+/// let dec: u8 = cks.decrypt(&ct_res);
+/// assert_eq!(u8::from(a >= b), dec);
+/// ```
+#[derive(Clone, Deserialize, Serialize)]
+pub struct BooleanBlock(pub(crate) Ciphertext);
+
+impl BooleanBlock {
+    /// Creates a new BooleanBlock without checks.
+    ///
+    /// You have to be sure the ciphertext encrypts 0 or 1 otherwise
+    /// functions expecting a BooleanBlock could result in wrong computation
+    pub fn new_unchecked(block: Ciphertext) -> Self {
+        Self(block)
+    }
+
+    /// Creates a new BooleanBlock, but does some checks to see
+    /// if it seems plausible that it encrypts a boolean.
+    ///
+    /// Sometimes these checks may prove to be too strict,
+    /// and you might need to use [BooleanBlock::new_unchecked] if you know it actually does encrypt
+    /// a boolean, or [BooleanBlock::convert] if there is some uncertainty.
+    pub fn try_new<T>(ct: &T) -> Option<Self>
+    where
+        T: IntegerRadixCiphertext,
+    {
+        if ct.holds_boolean_value() {
+            Some(Self(ct.blocks()[0].clone()))
+        } else {
+            None
+        }
+    }
+
+    /// Creates a new BooleanBlock by doing a PBS to ensure the result in either O or 1
+    ///
+    /// It is equivalent to doing `let b: bool = value != 0`
+    ///
+    /// # Examples
+    /// ## Converting a [BooleanBlock] to a [RadixCiphertext]
+    ///
+    /// ```
+    /// use tfhe::integer::{gen_keys_radix, BooleanBlock};
+    /// use tfhe::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+    ///
+    /// // We have 4 * 2 = 8 bits of message
+    /// let size = 4;
+    /// let (cks, sks) = gen_keys_radix(PARAM_MESSAGE_2_CARRY_2_KS_PBS, size);
+    ///
+    /// let a = 128u8;
+    /// let ct_a = cks.encrypt(a);
+    ///
+    /// let ct_b = BooleanBlock::convert(&ct_a, &sks);
+    ///
+    /// // Decrypt:
+    /// let dec = cks.decrypt_bool(&ct_b);
+    /// assert_eq!(a != 0, dec);
+    /// ```
+    pub fn convert<T>(ct: &T, sks: &ServerKey) -> Self
+    where
+        T: IntegerRadixCiphertext,
+    {
+        if !ct.holds_boolean_value() {
+            sks.scalar_ne_parallelized(ct, 0)
+        } else {
+            let block = ct.blocks()[0].clone();
+            Self(block)
+        }
+    }
+
+    /// Consumes the BooleanBlock to return its inner ciphertext
+    pub fn into_inner(self) -> Ciphertext {
+        self.0
+    }
+
+    /// Consumes and converts the BooleanBlock to a [RadixCiphertext] / [SignedRadixCiphertext]
+    /// with the given number of blocks.
+    ///
+    /// [SignedRadixCiphertext]: crate::integer::SignedRadixCiphertext
+    pub fn into_radix<T>(self, num_blocks: usize, sks: &ServerKey) -> T
+    where
+        T: IntegerRadixCiphertext,
+    {
+        let mut radix_ct = RadixCiphertext::from_blocks(vec![self.0]);
+        let missing_blocks = num_blocks.saturating_sub(1);
+        sks.extend_radix_with_trivial_zero_blocks_msb_assign(&mut radix_ct, missing_blocks);
+        T::from_blocks(radix_ct.blocks)
+    }
+}
+
+impl AsRef<Ciphertext> for BooleanBlock {
+    fn as_ref(&self) -> &Ciphertext {
+        &self.0
+    }
+}
--- a/tfhe/src/integer/ciphertext/mod.rs
+++ b/tfhe/src/integer/ciphertext/mod.rs
@@ -1,4 +1,6 @@
 //! This module implements the ciphertext structures.
+pub mod boolean_value;
+
 use super::parameters::{
     RadixCiphertextConformanceParams, RadixCompactCiphertextListConformanceParams,
 };
@@ -191,15 +193,17 @@ impl From<CompressedSignedRadixCiphertext> for SignedRadixCiphertext {
 }
 
 pub trait IntegerCiphertext: Clone {
-    fn from_blocks(blocks: Vec<Ciphertext>) -> Self;
     fn blocks(&self) -> &[Ciphertext];
-    fn blocks_mut(&mut self) -> &mut [Ciphertext];
     fn moduli(&self) -> Vec<u64> {
         self.blocks()
             .iter()
             .map(|x| x.message_modulus.0 as u64)
             .collect()
     }
+
+    fn from_blocks(blocks: Vec<Ciphertext>) -> Self;
+
+    fn blocks_mut(&mut self) -> &mut [Ciphertext];
 }
 
 pub trait IntegerRadixCiphertext: IntegerCiphertext + Sync + Send + From<Vec<Ciphertext>> {
@@ -222,13 +226,14 @@ pub trait IntegerRadixCiphertext: IntegerCiphertext + Sync + Send + From<Vec<Cip
 }
 
 impl IntegerCiphertext for RadixCiphertext {
+    fn blocks(&self) -> &[Ciphertext] {
+        &self.blocks
+    }
+
     fn from_blocks(blocks: Vec<Ciphertext>) -> Self {
         Self::from(blocks)
     }
 
-    fn blocks(&self) -> &[Ciphertext] {
-        &self.blocks
-    }
     fn blocks_mut(&mut self) -> &mut [Ciphertext] {
         &mut self.blocks
     }
@@ -243,12 +248,14 @@ impl IntegerRadixCiphertext for RadixCiphertext {
 }
 
 impl IntegerCiphertext for SignedRadixCiphertext {
-    fn from_blocks(blocks: Vec<Ciphertext>) -> Self {
-        Self::from(blocks)
-    }
     fn blocks(&self) -> &[Ciphertext] {
         &self.blocks
     }
+
+    fn from_blocks(blocks: Vec<Ciphertext>) -> Self {
+        Self::from(blocks)
+    }
+
     fn blocks_mut(&mut self) -> &mut [Ciphertext] {
         &mut self.blocks
     }
@@ -263,13 +270,15 @@ impl IntegerRadixCiphertext for SignedRadixCiphertext {
 }
 
 impl IntegerCiphertext for CrtCiphertext {
+    fn blocks(&self) -> &[Ciphertext] {
+        &self.blocks
+    }
+
     fn from_blocks(blocks: Vec<Ciphertext>) -> Self {
         let moduli = blocks.iter().map(|x| x.message_modulus.0 as u64).collect();
         Self { blocks, moduli }
     }
-    fn blocks(&self) -> &[Ciphertext] {
-        &self.blocks
-    }
+
     fn blocks_mut(&mut self) -> &mut [Ciphertext] {
         &mut self.blocks
     }
--- a/tfhe/src/integer/client_key/mod.rs
+++ b/tfhe/src/integer/client_key/mod.rs
@@ -20,6 +20,7 @@ use serde::{Deserialize, Serialize};
 pub use utils::radix_decomposition;
 
 use crate::integer::bigint::static_signed::StaticSignedBigInt;
+use crate::integer::ciphertext::boolean_value::BooleanBlock;
 pub use crt::CrtClientKey;
 pub use radix::RadixClientKey;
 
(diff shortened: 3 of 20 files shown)


tfhe/src/integer/server_key/radix_parallel/comparison.rs
```diff
@@ -77,33 +74,36 @@ impl ServerKey {
...
         }
 
-        block_comparisons.resize_with(lhs.blocks().len(), || self.key.create_trivial(0));
-
-        T::from_blocks(block_comparisons)
+        BooleanBlock::new(
+            block_comparisons
+                .into_iter()
+                .next()
+                .unwrap_or_else(|| self.key.create_trivial(0)),
```
IceTDrinker, line 81: no way to use block_comparisons.blocks().get(0).unwrap_or_else() ?

I think the iterating part is something we sometimes use but it looks really odd

tfhe/src/integer/server_key/radix/mod.rs
```diff
@@ -100,7 +100,7 @@ impl ServerKey {
     /// # Example
     ///
     ///```rust
-    /// use tfhe::integer::{gen_keys_radix, IntegerCiphertext};
+    /// use tfhe::integer::{gen_keys_radix, IntegerCiphertext, IntegerCiphertextRef};
```
IceTDrinker, line 103: I think you had a plan for an integer prelude ? could it be the time to do it ? maybe not in this PR but I think you mentioned it recently

tfhe/src/integer/client_key/mod.rs
```diff
@@ -425,13 +427,24 @@ impl ClientKey {
         self.key.encrypt(message)
     }
 
+    pub fn encrypt_bool(&self, msg: bool) -> BooleanBlock {
```
IceTDrinker, line 430: docstring maybe ?

tfhe/src/integer/client_key/mod.rs
```diff
@@ -425,13 +427,24 @@ impl ClientKey {
...
+    }
+
     /// Decrypts one block.
     ///
     /// This takes a shortint ciphertext as input.
     pub fn decrypt_one_block(&self, ct: &Ciphertext) -> u64 {
         self.key.decrypt(ct)
     }
 
+    pub fn decrypt_bool<Inner>(&self, ct: &BooleanValue<Inner>) -> bool
```
IceTDrinker, line 441: Here a docstring to indicate it uses the C-like meaning for integers and bools would be nice (maybe same thing in the radix client key stuff)

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,191 @@
...
+///
+/// Since values ares encrypted, it is not possible to know whether a
+/// ciphertext encrypts a boolean value (0 or 1). However some algorithms
+/// require that the ciphertext does indeed encrypt a boolean value.
+///
+/// This wrapper serves as making it explicit that it is known that the value
+/// encrypted is 0 or 1. And that if a function taking a BooleanValue as input
+/// returns incorrect value, it may be due to the value not really being 0 or 1.
+///
+/// Also some function such as comparisons are know to return an encrypted value
```
IceTDrinker, line 16: are know -> are known

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,191 @@
...
+/// let size = 4;
+/// let (cks, sks) = gen_keys_radix(PARAM_MESSAGE_2_CARRY_2_KS_PBS, size);
+///
+/// let clear_condition = 1u8;
+/// let a = 128u8;
+/// let b = 55u8;
+///
+/// let ct_a = cks.encrypt(a);
+/// let ct_b = cks.encrypt(b);
+/// let ct_condition = cks.encrypt(clear_condition);
```
IceTDrinker, line 62: isn't it better to advertise encrypt_boolean here ?

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,191 @@
...
+impl<Inner> BooleanValue<Inner>
+where
+    Inner: IntegerRadixCiphertext,
+{
+    pub fn into_radix(self, num_blocks: usize, sks: &ServerKey) -> Inner {
+        let mut radix_ct = RadixCiphertext::from_blocks(self.0.into_blocks());
+        let missing_blocks = num_blocks.saturating_sub(radix_ct.blocks().len());
+        sks.extend_radix_with_trivial_zero_blocks_msb_assign(&mut radix_ct, missing_blocks);
+        Inner::from_blocks(radix_ct.blocks)
+    }
```
IceTDrinker, line 107: Does this handle the case where the boolean value comes from a u32 and then is casted to a u16 ? i.e. the blocks may need trimming ?

what about always storing a RadixCiphertext that always has a single block ? or always a shortint Ciphertext ?

I don't remember what we said on that when we discussed it recently (a single shortint block vs something else).

tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -30,10 +31,10 @@ impl ServerKey {
     /// let ct_res = sks.unchecked_eq(&ct1, &ct2);
     ///
     /// // Decrypt:
-    /// let dec_result: u64 = cks.decrypt(&ct_res);
+    /// let dec_result: u64 = cks.decrypt_one_block(&ct_res.0);
```
IceTDrinker, line 34: why can't decrypt be used here ?

IceTDrinker (commented):
Thanks a few questions here and there and maybe a slight issue for into_radix on the new Boolean type

tfhe/src/integer/server_key/radix/mod.rs
tmontaigu: Yes i planned on adding the prelude after this PR as we don't want users (and potentially us) to have to remember similar but differently named traits

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: It should be `decrypt_bool`, to use decrypt I would need to convert the ct_res to a radix using .into_radix as ct_res is a BooleanBlock

tfhe/src/integer/ciphertext/boolean_value.rs
tmontaigu: You're right it does not handle the case where the target num blocks is smaller than the current number.

If we were to force the use of one representation I would be in favor of single block (and not a Vec with single block)

However the idea was for this wrapper to be as frictionless as possible, and so allow it to wrap a RadixCiphertext or a &RadixCiphertext in case that is was the user have its 'hands'

tfhe/src/integer/ciphertext/boolean_value.rs
tmontaigu: Yes it makes sense

tfhe/src/integer/ciphertext/boolean_value.rs
IceTDrinker: ah right the ref case is a bit annoying, but I guess we could have BooleanBlock<&Ciphertext> and you could do something like from_ct_ref(&radix_ct) { return BooleanVec(&radix_ct.blocks()[0]) } ?

As always frictionless is good, just need to find a design that avoids as many mistakes as possible

tfhe/src/integer/server_key/radix/comparison.rs
IceTDrinker: can't decrypt accept a BooleanBlock ? I would expect to get back 0 or 1 in that case

IceTDrinker: I mean via some trait or mechanic to make it seamless

tfhe/src/integer/server_key/radix/mod.rs
IceTDrinker: fine by me

tfhe/src/integer/server_key/radix_parallel/comparison.rs
IceTDrinker: remove ?

IceTDrinker: I think you used it, but maybe it's more costly than it needs to be

IceTDrinker: or drain, but yes I see your point

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: I can do `pub trait IntegerUnsignedRadixCiphertext: IntegerRadixCiphertext {}` 
impl it on RadixCiphertext and BooleanValue, and use it as a bound to decrypt_radix.

It breaks usage of `cks.decrypt::<u64>(&value)`  they have to be changed to `cks.decrypt::<u64, _>(&value)`;

Is that ok ?


Sadly I can't do `pub trait IntegerUnsignedRadixCiphertext: IntegerRadixCiphertext<IS_SIGNED=false> {}` because  Is_SIGNED is a const generic not a type.


we could try to do 
```rust
pub trait SignednessMarker : sealed {
   const IS_SIGNED: bool;
}

pub struct UnsignedMarker;
pub struct SignedMarker;

impl SignednessMarker for SignedMarker {
  IS_SIGNED = true;
}

// same for unsigned

pub trait IntegerRadixCiphertext {
    type Signedness: SignednessMarker;
}


pub trait IntegerUnsignedRadixCiphertext: IntegerRadixCiphertext<Signedness=UnsignedMarker> {}
```

but its out of the scope of this pr so maybe later if it seems worth and doable

IceTDrinker: IntegerDecryptableCiphertext maybe ?

IceTDrinker: it's not critical right now but could be nice to handle that down the line

tmontaigu: `IntegerDecryptableCiphertext` is not explicit enough, decrypt_radix is meant to decrypt unsigned ciphertext

We have decrypt_radix for unsigned radix and decrypt_signed for signed radix, the seperation is there to avoid user decrypting a  mismatching clear

tmontaigu: We could have only one decrypt/encrypt function but that means users would have less safeguards , if they mismatch a type, it would be as if an implicit cast had been done

IceTDrinker: ok yeah and implementing the IntegerRadixCipehrtext for Boolean defeats the whole purpose of the type

IceTDrinker: I'm not sure yet, I'm just wondering if we have a way to make it a bit more uniform, I guess it may be possible, I just don't quite see it yet

tfhe/src/integer/ciphertext/boolean_value.rs
tmontaigu: Actually no because the example is about showing we can wrap a regular radix ciphertext

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: I already implemented IntegerRadixCipehrtext for BooleanValue.

Yes it means that it can be used in additions, making it non boolean anymore, (but with the way are signatures are, you can only add a boolean to a boolean)

But it also makes boolean usable in bitand/or/xor

The idea is that since we cannot verify if a ciphertext actually is boolean, it does not really matter if we allow addition/sub on booleans because in the end the user can wrap any ciphertext to be marked as boolean.


So to me the question is more: "does this bring enough value"

tmontaigu: If not, we could just go with Returning single shortint block / taking single shortint block when manipulating/returning boolean valued ciphertext and use documentation to tell when the returned value/input value has to be boolean

IceTDrinker: I think the fact any ciphertext can be wrapped defeats the purpose unless we do a computation like value != 0 to store the boolean

tmontaigu: In that case I don't think introducing this is a good idea, because I could see cases where the users know the value in the ciphertext is either 0 or 1 and so making it pay the cost of a pbs for nothing does not sounds like a good idea

IceTDrinker: so let's not do that change for now ? and revert the decrypt change ?

tmontaigu: :thinking:  so what should we keep ?

IceTDrinker: you original one that did not change the decryption I think ?

tmontaigu: But that still has that any ciphertext can be wwrapper as boolean, and BooleanValue implements IntegerRadixCiphertext (if the wrapped thing is a radix)

IceTDrinker: ah damn

IceTDrinker: ok sorry maybe I did not follow enough what you wanted to do.

Why do we implement the trait for the Boolean thingy and why do we allow anything to be wrapped ?

tmontaigu:
Updated so its closer to how we agreed on what it should be

tmontaigu:
@slab-ci cpu_fast_test

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Sometimes these checks may prove to be too strict,
+    /// and you might need to use [BooleanBlock::new_unchecked] if you know it actually does encrypt
+    /// a boolean, or [BooleanBlock::convert] if there is some uncertainty.
+    pub fn try_new(block: Ciphertext) -> Option<Self> {
+        if block.degree.0 > 1 {
+            None
+        } else {
+            Some(Self(block))
+        }
+    }
```
IceTDrinker, line 68: should we take a reference to an &IntegerCiphertext or IntegerRadixCiphertext and perform the checks for all blocks non only the LSB one ?

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Creates a new BooleanBlock by doing a PBS to ensure the result in either O or 1
+    ///
+    /// It is equivalent to doing `let b: bool = value != 0`
+    pub fn convert(mut block: Ciphertext, sks: &ServerKey) -> Self {
+        if block.degree.0 > 1 {
+            Self(sks.key.scalar_not_equal(&mut block, 1))
+        } else {
+            Self(block)
+        }
```
IceTDrinker, line 78: same thing a ref to a radix ciphertext ?

tfhe/src/integer/ciphertext/boolean_value.rs
tmontaigu: made `trye_new`, `convert` take `&T: IntegerRadixCiphertext,`

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
```diff
@@ -474,7 +470,7 @@ impl ServerKey {
             match self.is_scalar_out_of_bounds(lhs, rhs) {
                 Some(std::cmp::Ordering::Greater) | Some(std::cmp::Ordering::Less) => {
                     // Scalar is not within bounds so its not equal
-                    return self.create_trivial_radix(1, lhs.blocks().len());
+                    return BooleanBlock::new_unchecked(self.key.create_trivial(1));
```
IceTDrinker, line 473: new trivial boolean rather (I think you added that function) ? I think there are quite a bit of new_unchecked on trivial 1s and 0s

tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -82,12 +83,14 @@ impl ServerKey {
                 .collect::<Vec<_>>();
         }
 
-        block_comparisons.resize_with(lhs.blocks().len(), || self.key.create_trivial(0));
-
-        T::from_blocks(block_comparisons)
+        let result = block_comparisons
+            .into_iter()
+            .next()
+            .unwrap_or_else(|| self.key.create_trivial(1));
```
IceTDrinker, line 89: why is the default value true here ?

the into iter next fails if block_comparisons are empty i.e. if the LHS is empty 🤔 I've seen the pattern elsewhere, maybe we should not set a default value with the unwrap_or_else but rather expect with an internal error message ?

(same where this pattern is used elsewhere, there are a few places) unless I'm missing something

tfhe/src/integer/public_key/tests.rs
```diff
@@ -97,7 +97,7 @@ fn radix_encrypt_decrypt_compact_128_bits_list(params: ClassicPBSParameters) {
 
         let compact_encrypted_list = pk.encrypt_slice_radix_compact(&clear_vec, num_block);
 
-        let ciphertext_vec = compact_encrypted_list.expand();
+        let ciphertext_vec: Vec<RadixCiphertext> = compact_encrypted_list.expand();
```
IceTDrinker, line 100: why is the annotation needed now ?

tfhe/src/integer/ciphertext/boolean_value.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Creates a new BooleanBlock by doing a PBS to ensure the result in either O or 1
+    ///
+    /// It is equivalent to doing `let b: bool = value != 0`
+    pub fn convert<T>(ct: &T, sks: &ServerKey) -> Self
+    where
+        T: IntegerRadixCiphertext,
+    {
+        if !ct.holds_boolean_value() {
+            let mut block = ct.blocks()[0].clone();
+            Self(sks.key.scalar_not_equal(&mut block, 1))
```
IceTDrinker, line 82: not_equal 1 or 0 ? does not seem to match the docstring

tfhe/src/integer/ciphertext/mod.rs
```diff
@@ -263,20 +270,18 @@ impl IntegerRadixCiphertext for SignedRadixCiphertext {
...
-        &self.blocks
-    }
+
     fn blocks_mut(&mut self) -> &mut [Ciphertext] {
         &mut self.blocks
     }
-
-    fn moduli(&self) -> Vec<u64> {
-        self.moduli.clone()
-    }
```
IceTDrinker, line 279: never used ?

tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -30,10 +31,10 @@ impl ServerKey {
     /// let ct_res = sks.unchecked_eq(&ct1, &ct2);
     ///
     /// // Decrypt:
-    /// let dec_result: u64 = cks.decrypt(&ct_res);
+    /// let dec_result = cks.decrypt_one_block(ct_res.as_ref());
```
IceTDrinker, line 34: use the decrypt boolean instead ? unsure

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
```diff
@@ -372,7 +373,7 @@ impl ServerKey {
             match self.is_scalar_out_of_bounds(lhs, rhs) {
                 Some(std::cmp::Ordering::Greater) | Some(std::cmp::Ordering::Less) => {
                     // Scalar is not within bounds so it cannot be equal
-                    return self.create_trivial_radix(0, lhs.blocks().len());
+                    return BooleanBlock::new_unchecked(self.key.create_trivial(0));
```
IceTDrinker, line 376: should we have a new_trivial ? so that we can do BooleanBlock::new_trivial(true)

tfhe/src/integer/ciphertext/boolean_value.rs
tmontaigu: indeed

tfhe/src/integer/ciphertext/mod.rs
tmontaigu: It was removed by mistake, I will re-add it, removing it makes it use the default impl of the trait which is still correct, but a bit less optimal

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
tmontaigu: If think I already added it, its a matter of using it there

tfhe/src/integer/client_key/mod.rs
```diff
@@ -425,13 +426,67 @@ impl ClientKey {
...
+    ///
+    /// let a = cks.encrypt_one_block(1u64);
+    /// let wrapped = BooleanBlock::new_unchecked(a);
+    /// let dec = cks.decrypt_bool(&wrapped);
+    /// assert_eq!(dec, true);
+    ///
+    /// // Here the value is not a boolean
+    /// // This is ok for decrypt function as it treats non zero as true
+    /// // However, generally when a function requires an input to be wrapped in
+    /// // BooleanBlock, it means the ciphertext must be 0 or 1
```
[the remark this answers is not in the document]
tmontaigu: Yes I was unsure about that, but if we don't show the thing, do we keep the `Treats 0 as false and the rest as true` or should it be a hidden detail since BooleanBlock is assumed to be 0 or 1 ?


 or maybe after decryption assert that the value is 0 or 1 before casting it to a bool ?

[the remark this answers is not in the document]
tmontaigu: Yeah I prefer having new_unchecked in case there's some cases where its the only choice (you don't want to pay for the price of convertion pbs).

Manual degree adjustment could be a potential solution, I mean, user can already do that to bypass the try_new, Its just 
that until noise and degree and not split it could be risky.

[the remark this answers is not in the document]
IceTDrinker: ok so let's have a follow up issue so that degree gets fixed manually later ?

tfhe/src/integer/server_key/radix/comparison.rs
tmontaigu: It was to handle the case when inputs are empty, if both lhs and rhs are emtpy then they are equal.

but actually it will also return the trivial true if lhs is empty but rhs is not, which is not correct.
Maybe its better to assert lhs.len() == rhs.len(), and at the begining

IceTDrinker: yep

tfhe/src/integer/client_key/mod.rs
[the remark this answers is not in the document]
tmontaigu: > ok so let's have a follow up issue so that degree gets fixed manually later ?

You mean, where we remove new_unchecked from pub, and tel user to fix degree before calling try_new ?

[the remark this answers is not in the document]
IceTDrinker: the first solution I would say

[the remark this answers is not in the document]
IceTDrinker: alright status quo then thanks for the discussion

IceTDrinker (approved):

IceTDrinker:
wut ?

IceTDrinker:
looks like I approved 😵‍💫 what is going on with the GH interface

tfhe/src/integer/server_key/radix/comparison.rs
```diff
@@ -125,9 +132,13 @@ impl ServerKey {
...
         }
 
-        block_comparisons.resize_with(lhs.blocks().len(), || self.key.create_trivial(0));
-
-        T::from_blocks(block_comparisons)
+        let result = block_comparisons
+            .into_iter()
+            .next()
+            // if block_comparisons is empty then both lhs and rhs were empty
+            // so they are equal (i.e different)
```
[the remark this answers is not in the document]
tmontaigu: yes true, at first I wrote `not not equal` but it felt too strange I removed one too many not

IceTDrinker:
https://github.com/zama-ai/tfhe-rs/actions/runs/6785671350/job/18444505568

tmontaigu:
@slab-ci cpu_fast_test

tmontaigu:
@IceTDrinker I fixed a doctest that was wrong and made test fail, I would need a new approval so tests run again

IceTDrinker (approved):
For the doctest fixes

tmontaigu:
Still did a mistake, lets wait for the fast test at least 😓

tmontaigu:
@slab-ci cpu_test
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test

PR #657, opened 2023-10-27
chore(integer): use Arc<ServerKey> for executor

tmontaigu:
The goal is to avoid holding the key twice in memory when both the executor and the test case needs the key


IceTDrinker:
Just a quick question why do we clone the key in the executor again ? I don't quite remember, it seems we are not using the sks outside the executor ? 🤔

tmontaigu:
the executor does not reclone the key

The test case sometimes clones the key because it needs a sks to be able to create ciphertexts with non clean carries

IceTDrinker (approved):
Yep ok, I misread a case where we create two executors

PR #658, opened 2023-10-27
feat(integer): add signed_overflowing_sub

tmontaigu:
### Check-list:


tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,305 @@ impl ServerKey {
...
+    ) -> Ciphertext {
+        let bits_of_message = self.key.message_modulus.0.ilog2();
+        let message_bit_mask = (1 << bits_of_message) - 1;
+
+        // This lut will generate a block that contains the information
+        // of how carry propagation happens in the last block, until the last bit.
+        let last_block_inner_propagation_lut =
+            self.key
+                .generate_lookup_table_bivariate(|lhs_block, rhs_block| {
+                    // half_modulus allows to c
```
tmontaigu, line 434: remove that

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,305 @@ impl ServerKey {
...
+
+                    // subtraction is done by doing addition of negation
+                    // negation(x) = bit_flip(x) + 1
+                    // We only add the flipped value, the + 1 well be resolved by
```
tmontaigu, line 438: will be

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,303 @@ impl ServerKey {
...
+
+            let output_carry = block_add_assign_returning_carry(
+                self,
+                &mut result.blocks[num_blocks - 1],
+                &input_carry,
+            );
+
+            // Encode the rule
+            // "Overflow occurred if the two inputs add the same sign and output has a different
+            // sign" using 2 PBS
```
IceTDrinker, line 697: I'm not a 2s complement expert like you are, what you are saying is that when seeing a - b as a + (-b), if a is > 0 e.g. a = 4 and -b is also positive so let's say b = 5

a + (-b) = 4 - 5 = -1 then this is an overflow ? I'm positive I'm wrong but this is how I understand the sentence, if you have a simple formula to explain that sentence I would put it in the comment :)

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -486,4 +784,108 @@ impl ServerKey {
...
+                    || self.full_propagate_parallelized(&mut tmp_rhs),
+                );
+                (&tmp_lhs, &tmp_rhs)
+            }
+        };
+
+        self.unchecked_signed_overflowing_sub_parallelized(lhs, rhs)
+    }
+
+    pub fn signed_overflowing_sub(
```
IceTDrinker, line 855: should this be in the parallel module ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,303 @@ impl ServerKey {
...
+        // So the carry propagation is done in 2 steps, first we compute the carry propagation
+        // in the last block to be able at the second step, to know the actual carry that
+        // the last bit receives.
+        //
+        // These are done in parallel to other stuff, and so no additional 'latency cost'
+        // should occur.
+
+        let mut result = lhs.clone();
+
+        let neg = self.unchecked_neg(rhs);
```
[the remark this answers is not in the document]
tmontaigu: It does not exists, I also don't think it can be made parallelized, and from measures of `unchecked_add_parallelized` it  would not bring measurable improbements

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,303 @@ impl ServerKey {
...
+    // does not have enough block.
+    pub(crate) fn unchecked_signed_overflowing_sub_parallelized_impl(
+        &self,
+        lhs: &SignedRadixCiphertext,
+        rhs: &SignedRadixCiphertext,
+    ) -> (SignedRadixCiphertext, Ciphertext) {
+        // This assert is here because this overflow computation requires these preconditions
+        // which is_eligible_for_parallel_single_carry_propagation, but it could change in the
+        // future
+        assert!(self.key.message_modulus.0 >= 4 && self.key.carry_modulus.0 >= 4);
```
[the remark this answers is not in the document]
tmontaigu: No because is_eligible_for_parallel_single_carry_propagation  checks for other stuff as well (like number of blocks to select between fully parallelized propagation or sequential)

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
IceTDrinker: ok !

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
IceTDrinker: should this sub check be a function that is called both here and in the is_elligible for parallel etc ?

Not saying we need it but in case some pre conditions change ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Well, the comment is not up to date with the code, the code 'encodes' a different rule for overflow, its the one I first used

- The comment says `Overflow occurred if the two inputs had the same sign and output has a different`

`4 + (-5)` -> 4 is positive, -5 is negative, the two input signs are different, this can never overflow
`4 + 5` -> both are positive (same sign), assuming 4 bits twos complement number, the result is `1001` = -7 -> different sign so overflow occured
`-3 + -3` -> both are negative , assuming 4 bits twos complement number, the result is  -6 -> same sign no overflow occured.

The code now uses a different rule that is, 'if the carry onto the last pair of bit is different that the output carry of said pair of bits, then overflow occurent`

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: I put it there because it uses `full_propagate_parallelized` and calls `unchecked_signed_overflowing_sub` which also uses rayon.

`unchecked_signed_overflowing_sub` implements the sequential propagation algorithm
that is used when the parallel version would be slower or when the parameters do not support it.

There's a bit of a naming problem because we have two 'parallel' propagation:
- One that is truly parallelized using hillis-steele prefix sum
- One that is not truly parallelized, its the sequential algorithm, but uses rayon::join
  to parallelized some message and carry extract.

And we need to have both of those.

For `signed_overflowing_sub` I can simply remove it and only keep `signed_overflowing_sub_parallelized`

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
tmontaigu: `assert!(self.key.message_modulus.0 >= 4 && self.key.carry_modulus.0 >= 4);` is a precondition for both parallel propagation and `unchecked_signed_overflowing_sub_parallelized_impl`

If the precondition changes for parallel_propagation (and is updated in is_eligible_for_parallel_single_carry_propagation), that does not mean the precondition for `unchecked_signed_overflowing_sub_parallelized_impl` also changes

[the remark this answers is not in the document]
IceTDrinker: ok then, I was guessing something like this may happen, all good

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
tmontaigu: hum does not seems a good thing, I'm worried ppl will say `hey why is unchecked_neg_parallelized not faster than unchecked_neg ?`

tfhe/src/integer/server_key/radix_parallel/sub.rs
IceTDrinker: ok then let's update the wording in that case, I don't think I understand the new rule which seems more complex 😅 ransom of performance I guess !

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,303 @@ impl ServerKey {
...
+            // Encode the rule
+            // "Overflow occurred if the two inputs add the same sign and output has a different
+            // sign" using 2 PBS
+            let overflowed = self.key.not_equal(&output_carry, &input_carry);
+            return (result, overflowed);
+        }
+
+        panic!(
+            "Invalid combo of message modulus ({}) and carry modulus ({})",
+            self.key.message_modulus.0, self.key.carry_modulus.0
```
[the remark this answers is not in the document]
tmontaigu: hum the combo is X_Y where X >= 1 and Y >= X, so 1_1, 2_2, 3_3, 4_4, 1_2, 2_3
but not 2_1, 1_0

tfhe/src/integer/server_key/radix_parallel/sub.rs
IceTDrinker: in any case I guess rayon join is never guaranteed to be parallelized, but here this signed overflowing sub does not really have the parallelized algorithm right ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
IceTDrinker: the equation is good it indicates clearly what is required for the moduli 👍 as long as it's something like

"let P be the message modulus and C the carry modulus, this algorithm requires, {your equation}, got msg_mod and carry mod == etc."

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -407,6 +408,303 @@ impl ServerKey {
...
+        self.key.unchecked_apply_lookup_table_bivariate(
+            last_lhs_block,
+            last_rhs_block,
+            &last_block_inner_propagation_lut,
+        )
+    }
+
+    // - last_block_inner_propagation must be the result of generate_last_block_inner_propagation
+    // - last_block_input_carry: carry that the last pair of blocks (lhs, rhs) receives as input
+    // - last_block_output_carry: carry that the last pair of blocks (lhs, rhs) output
```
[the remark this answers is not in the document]
tmontaigu: We could, but to me it would seem a bit zealous especially given that they would be used in 3-4 fn at most, and not by users

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Yeah the first rules needs that we have done carry_propgation + carry-add + message_extract.

Whereas the rule on input/output carry of last bit allows to compute overflow flag while we do the carry-add+message_extract

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
IceTDrinker: just a thought, Rust does that a lot for their std lib even if some of those types are interacted with by users

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: If you do 
```
 4: 0100
+5: 0101
----------
(-7) 1001
```

The last pair of bit, (`0`, `0`), receives as input the carry of `1 + 1`, which is `1`, so, `1 + 0 +0` emits 0 as output carry.

-> Input 1, output 0 => overflow

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: If you do not count rayon::join as parallelism then appart from the calls to `full_propagate_parallelized` then yes `signed_overflowing_sub` does not have parallelism.

It also call `unchecked_signed_overflowing_sub` which uses `rayon::join`

tmontaigu: What should we do, keep `signed_overflowing_sub`, move it to non-parallel folder, and still allow use of rayon::join.

or remove `signed_overflowing_sub` (only keep `signed_overflowing_sub_parallelized`), and rename `unchecked_signed_overflowing_sub` to `unchecked_sequential_signed_overflowing_sub_parallelized` as a private fn called when parameters are not compatible with fully parallelized variant ?

IceTDrinker: Could be moved to the sequential module without removing the rayon stuff, I think some functions already do a similar thing anyways

tmontaigu: I checked, and indeed there is the unsigned non parallelized variant of overflowing_sub which uses rayon, I did that for the same reasons

tfhe/src/integer/server_key/radix_parallel/sub.rs
IceTDrinker: Yep ok, still boggles the mind

IceTDrinker (approved):
Thanks

tmontaigu:
@slab-ci cpu_wasm_test

IceTDrinker (approved):

PR #662, opened 2023-10-30
refactor(hlapi): Remove shortint support from HLAPI

tmontaigu:
### PR content/description

This removes the wrapping of shortints from the HLAPI, the reasons are:

Contrary to integers for which we have different bit size by combining different number of blocks from the _same_ key. shortints had different bit size, but also different keys which lead to:

- Not being able to cast between 2 different shortint type and between 1 shortint and 1 integer. Technically these casts are possible, but requires a keyswitch (and likely a PBS). But the keyswitch requires parameters, which may not always exists.

- Due to each shortint having different keys, the internal code to manage that made heavy use of macros to avoid having thousands of repeated lines. However, this made the code harder to follow / modify especially for people that were not familiar with that.

- In practive to really benefit from shortints, proper management of carry space is needed, however the HLAPI completely hides that, resulting in less optimal performances. In short, shortints are better used as a low level construct.

- Building a FheUint4 with two block of message_2_carry_2 is likely to be faster the one message_4_carry_4 for most use cases.

So removing the wrapping of shortints will simplify the code, and allow for more simplification later.
Also, it will allow us to expose Fhe{Ui/I}nt{2, 4, 6} types which are compatible (cast_from/into) with Fhe{Ui/I}nt{8, 16, 32, etc}.

BREAKING CHANGE:
    - FheUint{2,3,4} removed from HLAPI
    - All HLAPI functions thied to shortints are removed

### Check-list:


tfhe/src/high_level_api/details.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(any(feature = "boolean", feature = "shortint", feature = "integer"))]
```
mayeul-zama, line 1: Why remove integer here?

tfhe/src/high_level_api/global_state.rs
```diff
@@ -101,7 +101,7 @@ where
 /// - The identifier (or identifier chain) that points to the member in the `ServerKey` that holds
 ///   the key for which the trait is implemented.
 /// - Type Variant used to identify the type at runtime (see `error.rs`)
-#[cfg(any(feature = "integer", feature = "shortint", feature = "boolean"))]
```
mayeul-zama, line 104: Why remove integer here?

tfhe/src/high_level_api/keys/client.rs
```diff
@@ -112,7 +104,7 @@ pub trait RefKeyFromKeyChain: Sized {
 /// - The identifier (or identifier chain) that points to the member in the `ClientKey` that holds
 ///   the key for which the trait is implemented.
 /// - Type Variant used to identify the type at runtime (see `error.rs`)
-#[cfg(any(feature = "integer", feature = "shortint", feature = "boolean"))]
```
mayeul-zama, line 115: Why remove integer here?

tfhe/src/high_level_api/keys/public.rs
```diff
@@ -66,7 +60,7 @@ pub trait RefKeyFromPublicKeyChain: Sized {
     }
 }
 
-#[cfg(any(feature = "integer", feature = "shortint", feature = "boolean"))]
```
mayeul-zama, line 69: Why remove integer here?

tfhe/src/high_level_api/mod.rs
```diff
@@ -45,11 +45,6 @@ expand_pub_use_fhe_type!(
         FheUint256, FheInt8, FheInt16, FheInt32, FheInt64, FheInt128, FheInt256
     };
 );
-#[cfg(feature = "shortint")]
-pub use crate::high_level_api::shortints::{
-    CompressedFheUint2, CompressedFheUint3, CompressedFheUint4, FheUint2, FheUint2Parameters,
```
mayeul-zama, line 50: Do you plan to reintroduce `FheUint2` and `FheUint4` as integers in another PR?

tfhe/src/high_level_api/tests/test_fhe_number_ops.rs
```diff
@@ -1,207 +0,0 @@
-#![cfg(any(feature = "integer", feature = "shortint"))]
```
mayeul-zama, line 1: Do these tests not make sense anymore without shortints?

tfhe/src/high_level_api/keys/public.rs
tmontaigu: because this macro is not used in integer part of the API

tfhe/src/high_level_api/tests/test_fhe_number_ops.rs
tmontaigu: Only shortints were really tested in this, there was just a tiny integer test, but not particularly worth

tfhe/src/high_level_api/keys/client.rs
tmontaigu: This macro is not used in the integer part of the hlapi

tfhe/src/high_level_api/details.rs
tmontaigu: This macro is (and was) not used in the integer part of the hlapi, so it should not have been there

tfhe/src/high_level_api/global_state.rs
tmontaigu: This macro is not used in the integer part of the hlapi

mayeul-zama (approved):

IceTDrinker:
as an FYI this also breaks serialization for some structs

PR #663, opened 2023-11-02
refactor(hlapi): remove wrapping of booleans

tmontaigu:
This commit removes the wrapping of the `tfhe::boolean` that was done in the HLAPI, effectively making the HLAPI only wrapping `tfhe::integer`.

FheBool is now reused to be a single shortint block compatible with other type FheUint8,16,etc (previously they were not).

In the future, `tfhe::boolean` could be re-wrapped in hlapi, but this time, to be used as a base for all integers and not just FheBool.

BREAKING CHANGE:
- hlapi no longer wraps tfhe::boolean API.
- tfhe::ConfigBuilder::enable_bool/disable_bool/all_disabled/all_enabled removed. Now default configuration should be done using `tfhe::ConfigBuilder::default()`.
- Uninitialied{ClientKey, PublicKey, CompressedPublicKey} error types removed as these erros are no longer possible


tfhe/src/c_api/high_level_api/keys.rs
```diff
@@ -52,13 +52,20 @@ pub unsafe extern "C" fn set_server_key(server_key: *const ServerKey) -> c_int {
 }
 
 /// result can be null
+///
+/// If result may be set to null if no previous server key was set
```
mayeul-zama, line 56: Leading "If" should not be here I guess

tfhe/src/high_level_api/global_state.rs
```diff
@@ -11,7 +11,7 @@ use crate::high_level_api::keys::ServerKey;
 /// (eg a web server that processes multiple requests in multiple threads).
 /// The user however, has to initialize the internal keys each time it starts a thread.
 thread_local! {
-    static INTERNAL_KEYS: RefCell<ServerKey> = RefCell::new(ServerKey::default());
+    static INTERNAL_KEYS: RefCell<Option<ServerKey>> = RefCell::new(None);
```
mayeul-zama, line 14: What's the impact of this change?

tfhe/src/high_level_api/integers/keys.rs
```diff
@@ -25,38 +25,22 @@ impl IntegerConfig {
...
-            ),
+            block_parameters: crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_PBS_KS.into(),
             wopbs_block_parameters: None,
         }
     }
 
     pub fn enable_wopbs(&mut self) {
-        let block_parameter = self
-            .block_parameters
-            .get_or_insert(crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS.into());
```
mayeul-zama, line 57: Why was it done this way?
Why change (even if IMO it's better in this PR)?

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -299,11 +281,10 @@ where
 impl<P, ClearType> FheDecrypt<ClearType> for GenericInteger<P>
 where
     P: IntegerParameter,
-    P::Id: RefKeyFromKeyChain<Key = crate::integer::ClientKey>,
     crate::integer::ClientKey: DecryptionKey<P::InnerCiphertext, ClearType>,
 {
     fn decrypt(&self, key: &ClientKey) -> ClearType {
-        let key = self.id.unwrapped_ref_key(key);
```
mayeul-zama, line 306: What was the use of id here?

tfhe/src/high_level_api/global_state.rs
tmontaigu: There is no impact (either performance wise or api wise).

That option used to be a bit deeper inside the `ServerKey` struct, it was in `ServerKey::integer_key::key`
Each level of API wrapped had an `Option<ServerKey>`.

Now that the API only uses integers, the `ServerKey` always has an inner integer server key.

But we still need an option as the thread_local storage cannot be instanciated with a key until the user calls `set_server_key`

tfhe/src/high_level_api/integers/keys.rs
tmontaigu: Previously users had to 'enable' the types they wanted to use (FheBool, FheUint2, FheUin3, etc)

So potentially a user could have called the `ConfigBuilder::enable_function_evaluation_integer` without having called `enable_default_integers` before, and in that case, instead of panicking to require calling `enable_default_integers` we chose to set the block parameters ourselves.

But now that there is no choice of which type to enable, there is always a block_parameters

tfhe/src/high_level_api/integers/types/base.rs
tmontaigu: Previously each data type had its own set of keys, FheBoolClentKey, FheUint2ClientKey

Ids was the mechanism that allowd to write generic function were the id was responsible of retrieving the 
correct key for the given type.

Now that there is only one key shared by all types, it's necessary to use the id to get the inner key.

mayeul-zama: I guess you meant "not necessary" in your last sentence.
Then, is it still necessary to have `id`s now?

tmontaigu: > I guess you meant "not necessary" in your last sentence.
Indeed

> Then, is it still necessary to have `id`s now?

They are still somewhat useful as they are the zero sized typed stored in a GenericInteger<P> to have a difference 
between FheUin8,16,etc.

mayeul-zama: Ok

tmontaigu: Like maybe instead of making `GenericInteger` generic over some parameters `P` (which has an associated Id),
make it generic over the `Id` directly, has now all types without exception shares parameters

mayeul-zama (approved):

tmontaigu:
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test
@slab-ci cpu_test

tmontaigu:
@slab-ci cpu_test

tmontaigu:
@slab-ci cpu_test

tmontaigu:
@slab-ci cpu_test

PR #664, opened 2023-11-03
chore(ci): add core_crypto layer to code coverage

soonum:
One more layer about to be implemented :tada:

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -38,6 +38,7 @@ jobs:
       group: ${{ github.workflow }}_${{ github.ref }}_${{ inputs.instance_image_id }}_${{ inputs.instance_type }}
       cancel-in-progress: true
     runs-on: ${{ inputs.runner_name }}
+    timeout-minutes: 1080
     steps:
       # Step used for log purpose.
       - name: Instance configuration used
@@ -79,6 +80,12 @@ jobs:
         if: steps.changed-files.outputs.tfhe_any_changed == 'true'
         run: |
           make GEN_KEY_CACHE_COVERAGE_ONLY=TRUE gen_key_cache
+          make gen_key_cache_core_crypto
+
+      - name: Run coverage for core_crypto
+        if: steps.changed-files.outputs.tfhe_any_changed == 'true'
+        run: |
+          make test_core_crypto_cov AVX512_SUPPORT=ON
 
       - name: Run coverage for boolean
         if: steps.changed-files.outputs.tfhe_any_changed == 'true'
@@ -97,7 +104,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           directory: ./coverage/
           fail_ci_if_error: true
-          files: shortint/cobertura.xml,boolean/cobertura.xml
+          files: shortint/cobertura.xml,boolean/cobertura.xml,core_crypto/cobertura.xml,core_crypto_avx512/cobertura.xml
 
       - name: Slack Notification
         if: ${{ failure() }}
--- a/Makefile
+++ b/Makefile
@@ -225,13 +225,6 @@ clippy_js_wasm_api clippy_tasks clippy_core clippy_concrete_csprng clippy_triviu
 clippy_fast: clippy clippy_all_targets clippy_c_api clippy_js_wasm_api clippy_tasks clippy_core \
 clippy_concrete_csprng
 
-.PHONY: gen_key_cache # Run the script to generate keys and cache them for shortint tests
-gen_key_cache: install_rs_build_toolchain
-	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) run --profile $(CARGO_PROFILE) \
-		--example generates_test_keys \
-		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,internal-keycache -- \
-		$(MULTI_BIT_ONLY) $(COVERAGE_ONLY)
-
 .PHONY: build_core # Build core_crypto without experimental features
 build_core: install_rs_build_toolchain install_rs_check_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) build --profile $(CARGO_PROFILE) \
@@ -317,6 +310,21 @@ test_core_crypto: install_rs_build_toolchain install_rs_check_toolchain
 	@if [[ "$(AVX512_SUPPORT)" == "ON" ]]; then \
 		RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_CHECK_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
 			--features=$(TARGET_ARCH_FEATURE),experimental,$(AVX512_FEATURE) -p $(TFHE_SPEC) -- core_crypto::; \
+	fi
+
+.PHONY: test_core_crypto_cov # Run the tests of the core_crypto module with code coverage
+test_core_crypto_cov: install_rs_build_toolchain install_rs_check_toolchain install_tarpaulin
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) tarpaulin --profile $(CARGO_PROFILE) \
+		--out xml --output-dir coverage/core_crypto --line --engine llvm --timeout 500 \
+		--implicit-test-threads $(COVERAGE_EXCLUDED_FILES) \
+		--features=$(TARGET_ARCH_FEATURE),experimental,internal-keycache,__coverage \
+		-p tfhe -- core_crypto::
+	@if [[ "$(AVX512_SUPPORT)" == "ON" ]]; then \
+		RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_CHECK_TOOLCHAIN) tarpaulin --profile $(CARGO_PROFILE) \
+			--out xml --output-dir coverage/core_crypto_avx512 --line --engine llvm --timeout 500 \
+			--implicit-test-threads $(COVERAGE_EXCLUDED_FILES) \
+			--features=$(TARGET_ARCH_FEATURE),experimental,internal-keycache,__coverage,$(AVX512_FEATURE) \
+			-p tfhe -- core_crypto::; \
 	fi
 
 .PHONY: test_boolean # Run the tests of the boolean module
@@ -627,6 +635,18 @@ ci_bench_web_js_api_parallel: build_web_js_api_parallel
 #
 # Utility tools
 #
+.PHONY: gen_key_cache # Run the script to generate keys and cache them for shortint tests
+gen_key_cache: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) run --profile $(CARGO_PROFILE) \
+		--example generates_test_keys \
+		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,internal-keycache -- \
+		$(MULTI_BIT_ONLY) $(COVERAGE_ONLY)
+
+.PHONY: gen_key_cache_core_crypto # Run function to generate keys and cache them for core_crypto tests
+gen_key_cache_core_crypto: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --tests --profile $(CARGO_PROFILE) \
+		--features=$(TARGET_ARCH_FEATURE),experimental,internal-keycache -p tfhe -- --nocapture \
+		core_crypto::keycache::generate_keys
 
 .PHONY: measure_hlapi_compact_pk_ct_sizes # Measure sizes of public keys and ciphertext for high-level API
 measure_hlapi_compact_pk_ct_sizes: install_rs_check_toolchain
--- a/tfhe/src/core_crypto/algorithms/mod.rs
+++ b/tfhe/src/core_crypto/algorithms/mod.rs
@@ -44,7 +44,7 @@ pub mod seeded_lwe_public_key_decompression;
 pub mod slice_algorithms;
 
 #[cfg(test)]
-mod test;
+pub(crate) mod test;
 
 // No pub use for slice and polynomial algorithms which would not interest higher level users
 // They can still be used via `use crate::core_crypto::algorithms::slice_algorithms::*;`
--- a/tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs
+++ b/tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs
@@ -1,36 +1,94 @@
 use super::*;
+use crate::core_crypto::keycache::KeyCacheAccess;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
 
-pub struct MultiBitParams<Scalar: UnsignedInteger> {
-    pub input_lwe_dimension: LweDimension,
-    pub lwe_modular_std_dev: StandardDev,
-    pub decomp_base_log: DecompositionBaseLog,
-    pub decomp_level_count: DecompositionLevelCount,
-    pub glwe_dimension: GlweDimension,
-    pub polynomial_size: PolynomialSize,
-    pub glwe_modular_std_dev: StandardDev,
-    pub message_modulus_log: CiphertextModulusLog,
-    pub ciphertext_modulus: CiphertextModulus<Scalar>,
-    pub grouping_factor: LweBskGroupingFactor,
-    pub thread_count: ThreadCount,
+#[cfg(not(feature = "__coverage"))]
+const NB_TESTS: usize = 10;
+#[cfg(not(feature = "__coverage"))]
+// Divided by two compared to other tests, we are running the algorithm twice for determinism
+const NB_TESTS_LIGHT: usize = 5;
+#[cfg(feature = "__coverage")]
+const NB_TESTS: usize = 1;
+#[cfg(feature = "__coverage")]
+const NB_TESTS_LIGHT: usize = 1;
+
+pub fn generate_keys<
+    Scalar: UnsignedTorus + Sync + Send + CastFrom<usize> + CastInto<usize> + Serialize + DeserializeOwned,
+>(
+    params: MultiBitTestParams<Scalar>,
+    rsc: &mut TestResources,
+) -> MultiBitBootstrapKeys<Scalar> {
+    // Keygen is a bit slow on this one so we keep it out of the testing loop
+
+    // Create the LweSecretKey
+    let input_lwe_secret_key = allocate_and_generate_new_binary_lwe_secret_key(
+        params.input_lwe_dimension,
+        &mut rsc.secret_random_generator,
+    );
+    let output_glwe_secret_key = allocate_and_generate_new_binary_glwe_secret_key(
+        params.glwe_dimension,
+        params.polynomial_size,
+        &mut rsc.secret_random_generator,
+    );
+    let output_lwe_secret_key = output_glwe_secret_key.clone().into_lwe_secret_key();
+
+    let mut bsk = LweMultiBitBootstrapKey::new(
+        Scalar::ZERO,
+        params.glwe_dimension.to_glwe_size(),
+        params.polynomial_size,
+        params.decomp_base_log,
+        params.decomp_level_count,
+        params.input_lwe_dimension,
+        params.grouping_factor,
+        params.ciphertext_modulus,
+    );
+
+    par_generate_lwe_multi_bit_bootstrap_key(
+        &input_lwe_secret_key,
+        &output_glwe_secret_key,
+        &mut bsk,
+        params.glwe_modular_std_dev,
+        &mut rsc.encryption_random_generator,
+    );
+
+    let mut fbsk = FourierLweMultiBitBootstrapKey::new(
+        params.input_lwe_dimension,
+        params.glwe_dimension.to_glwe_size(),
+        params.polynomial_size,
+        params.decomp_base_log,
+        params.decomp_level_count,
+        params.grouping_factor,
+    );
+
+    par_convert_standard_lwe_multi_bit_bootstrap_key_to_fourier(&bsk, &mut fbsk);
+
+    MultiBitBootstrapKeys {
+        small_lwe_sk: input_lwe_secret_key,
+        big_lwe_sk: output_lwe_secret_key,
+        bsk,
+        fbsk,
+    }
 }
 
-fn lwe_encrypt_multi_bit_pbs_decrypt_custom_mod<
-    Scalar: UnsignedTorus + Sync + Send + CastFrom<usize> + CastInto<usize>,
->(
-    params: MultiBitParams<Scalar>,
-) {
-    let input_lwe_dimension = params.input_lwe_dimension;
+fn lwe_encrypt_multi_bit_pbs_decrypt_custom_mod<Scalar>(params: MultiBitTestParams<Scalar>)
+where
+    Scalar: UnsignedTorus
+        + Sync
+        + Send
+        + CastFrom<usize>
+        + CastInto<usize>
+        + Serialize
+        + DeserializeOwned,
+    MultiBitTestParams<Scalar>: KeyCacheAccess<Keys = MultiBitBootstrapKeys<Scalar>>,
+{
     let lwe_modular_std_dev = params.lwe_modular_std_dev;
-    let glwe_modular_std_dev = params.glwe_modular_std_dev;
     let ciphertext_modulus = params.ciphertext_modulus;
     let message_modulus_log = params.message_modulus_log;
     let msg_modulus = Scalar::ONE.shl(message_modulus_log.0);
     let encoding_with_padding = get_encoding_with_padding(ciphertext_modulus);
     let glwe_dimension = params.glwe_dimension;
     let polynomial_size = params.polynomial_size;
-    let decomp_base_log = params.decomp_base_log;
-    let decomp_level_count = params.decomp_level_count;
-    let grouping_factor = params.grouping_factor;
     let thread_count = params.thread_count;
 
     let mut rsc = TestResources::new();
@@ -43,7 +101,6 @@ fn lwe_encrypt_multi_bit_pbs_decrypt_custom_mod<
 
     let delta: Scalar = encoding_with_padding / msg_modulus;
     let mut msg = msg_modulus;
-    const NB_TESTS: usize = 10;
 
     let accumulator = generate_accumulator(
         polynomial_size,
@@ -59,55 +116,16 @@ fn lwe_encrypt_multi_bit_pbs_decrypt_custom_mod<
         ciphertext_modulus
     ));
 
-    // Keygen is a bit slow on this one so we keep it out of the testing loop
-    // Create the LweSecretKey
-    let input_lwe_secret_key = allocate_and_generate_new_binary_lwe_secret_key(
-        input_lwe_dimension,
-        &mut rsc.secret_random_generator,
-    );
-    let output_glwe_secret_key = allocate_and_generate_new_binary_glwe_secret_key(
-        glwe_dimension,
-        polynomial_size,
-        &mut rsc.secret_random_generator,
-    );
-    let output_lwe_secret_key = output_glwe_secret_key.clone().into_lwe_secret_key();
+    let mut keys_gen = |params| generate_keys(params, &mut rsc);
 
-    let mut bsk = LweMultiBitBootstrapKey::new(
-        Scalar::ZERO,
-        glwe_dimension.to_glwe_size(),
-        polynomial_size,
-        decomp_base_log,
-        decomp_level_count,
-        input_lwe_dimension,
-        grouping_factor,
-        ciphertext_modulus,
-    );
-
-    par_generate_lwe_multi_bit_bootstrap_key(
-        &input_lwe_secret_key,
-        &output_glwe_secret_key,
-        &mut bsk,
-        glwe_modular_std_dev,
-        &mut rsc.encryption_random_generator,
-    );
+    let keys = gen_keys_or_get_from_cache_if_enabled(params, &mut keys_gen);
+    let (input_lwe_secret_key, output_lwe_secret_key, bsk, fbsk) =
+        (keys.small_lwe_sk, keys.big_lwe_sk, keys.bsk, keys.fbsk);
 
     assert!(check_encrypted_content_respects_mod(
         &*bsk,
         ciphertext_modulus
     ));
-
-    let mut fbsk = FourierLweMultiBitBootstrapKey::new(
-        input_lwe_dimension,
-        glwe_dimension.to_glwe_size(),
-        polynomial_size,
-        decomp_base_log,
-        decomp_level_count,
-        grouping_factor,
-    );
-
-    par_convert_standard_lwe_multi_bit_bootstrap_key_to_fourier(&bsk, &mut fbsk);
-
-    drop(bsk);
 
     while msg != Scalar::ZERO {
         msg = msg.wrapping_sub(Scalar::ONE);
@@ -152,26 +170,33 @@ fn lwe_encrypt_multi_bit_pbs_decrypt_custom_mod<
 
             assert_eq!(decoded, f(msg));
         }
+
+        // In coverage, we break after one while loop iteration, changing message values does not
+        // yield higher coverage
+        #[cfg(feature = "__coverage")]
+        break;
     }
 }
 
-fn lwe_encrypt_multi_bit_deterministic_pbs_decrypt_custom_mod<
-    Scalar: UnsignedTorus + Sync + Send + CastFrom<usize> + CastInto<usize>,
->(
-    params: MultiBitParams<Scalar>,
-) {
-    let input_lwe_dimension = params.input_lwe_dimension;
+fn lwe_encrypt_multi_bit_deterministic_pbs_decrypt_custom_mod<Scalar>(
+    params: MultiBitTestParams<Scalar>,
+) where
+    Scalar: UnsignedTorus
+        + Sync
+        + Send
+        + CastFrom<usize>
+        + CastInto<usize>
+        + Serialize
+        + DeserializeOwned,
+    MultiBitTestParams<Scalar>: KeyCacheAccess<Keys = MultiBitBootstrapKeys<Scalar>>,
+{
     let lwe_modular_std_dev = params.lwe_modular_std_dev;
-    let glwe_modular_std_dev = params.glwe_modular_std_dev;
     let ciphertext_modulus = params.ciphertext_modulus;
     let message_modulus_log = params.message_modulus_log;
     let msg_modulus = Scalar::ONE.shl(message_modulus_log.0);
     let encoding_with_padding = get_encoding_with_padding(ciphertext_modulus);
     let glwe_dimension = params.glwe_dimension;
     let polynomial_size = params.polynomial_size;
-    let decomp_base_log = params.decomp_base_log;
-    let decomp_level_count = params.decomp_level_count;
-    let grouping_factor = params.grouping_factor;
     let thread_count = params.thread_count;
 
     let mut rsc = TestResources::new();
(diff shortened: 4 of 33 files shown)


soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

.github/workflows/code_coverage.yml
```diff
@@ -76,28 +77,34 @@ jobs:
               - concrete-csprng/src/**
 
       - name: Generate Keys
-        if: steps.changed-files.outputs.tfhe_any_changed == 'true'
+#        if: steps.changed-files.outputs.tfhe_any_changed == 'true'
```
IceTDrinker, line 80: maybe remove the comments now ?

Makefile
```diff
@@ -585,6 +593,18 @@ ci_bench_web_js_api_parallel: build_web_js_api_parallel
...
+.PHONY: gen_key_cache # Run the script to generate keys and cache them for shortint tests
+gen_key_cache: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) run --profile $(CARGO_PROFILE) \
+		--example generates_test_keys \
+		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,internal-keycache -p tfhe -- \
+		$(MULTI_BIT_ONLY) $(COVERAGE_ONLY)
+
+.PHONY: gen_key_cache_core_crypto # Run function to generate keys and cache them for core_crypto tests
+gen_key_cache_core_crypto: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --tests --profile devo \
```
IceTDrinker, line 605: devo -> $(CARGO_PROFILE)

tfhe/src/core_crypto/algorithms/mod.rs
```diff
@@ -44,7 +44,7 @@ pub mod seeded_lwe_public_key_decompression;
 pub mod slice_algorithms;
 
 #[cfg(test)]
-mod test;
+pub mod test;
```
IceTDrinker, line 47: why is that needed ?

tfhe/src/shortint/server_key/tests/noise_level.rs
```diff
@@ -254,6 +254,7 @@ fn test_ct_scalar_op_assign_noise_level_propagation(sk: &ServerKey, ct: &Ciphert
 }
 
 #[test]
+#[cfg(not(tarpaulin))] // This test is ignored in coverage, it takes around 4 hours to run otherwise.
```
IceTDrinker, line 257: question should we use cfg(not(tarpaulin)) or cfg(not(__coverage)) ?

tfhe/src/core_crypto/mod.rs
```diff
@@ -17,3 +17,6 @@ pub mod prelude;
 pub mod seeders;
 
 pub mod fft_impl;
+
+#[cfg(any(test, doctest, feature = "internal-keycache"))]
+pub mod keycache;
```
IceTDrinker, line 22: no, only for internal-keycache enabled, otherwise we have the keycache on for all tests

IceTDrinker (commented):
Haven't read everything yet, first batch of comment already has some stuff in it

tfhe/src/keycache/mod.rs
```diff
@@ -214,4 +214,73 @@ pub mod utils {
...
+    where
+        P: Copy + PartialEq + NamedParam,
+        S: PersistentStorage<P, K>,
+    {
+        pub fn get_with_closure<C: FnMut(P) -> K>(&self, param: P, c: &mut C) -> SharedKey<K> {
+            self.with_closure(param, c)
+        }
+
+        pub fn with_closure<C>(&self, param: P, c: &mut C) -> SharedKey<K>
+        // pub fn with_closure<C, R: Clone>(&self, param: P, c: C) -> R
```
tmontaigu, line 228: remove

tfhe/src/keycache/mod.rs
```diff
@@ -214,4 +214,73 @@ pub mod utils {
...
+
+                        Ok(shared_cell)
+                    },
+                    Ok,
+                )
+            };
+
+            try_load_from_memory_and_init().ok().unwrap()
+        }
+    }
```
tmontaigu, line 285: Now that we have this, shouldn't we able to implement `Keycache::get` as 
```rust
self.get_with_closure(params, K::from)
```
Or something similar

IceTDrinker: sounds very likely indeed

tfhe/src/core_crypto/algorithms/test/lwe_packing_keyswitch.rs
```diff
@@ -197,6 +198,11 @@ fn lwe_list_encrypt_pks_to_glwe_decrypt_custom_mod<Scalar: UnsignedTorus + Send
 
             assert_eq!(decrypted_plaintext_list, input_plaintext_list);
         }
+
+        // In coverage, we break after one while loop iteration. This test case is quite long,
+        // around 400s per loop.
+        #[cfg(feature = "__coverage")]
```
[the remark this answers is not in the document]
IceTDrinker: and that for all functions where we loop on message values

tfhe/src/core_crypto/algorithms/mod.rs
soonum: To get access to the keys generation function from outside the algorithms/ directory. I need the functions in the keycache thus outside of this very module.

tfhe/src/core_crypto/algorithms/test/mod.rs
```diff
@@ -40,28 +41,14 @@ impl TestResources {
...
-    pub pfks_base_log: DecompositionBaseLog,
-    pub pfks_modular_std_dev: StandardDev,
-    pub cbs_level: DecompositionLevelCount,
-    pub cbs_base_log: DecompositionBaseLog,
-    pub message_modulus_log: CiphertextModulusLog,
-    pub ciphertext_modulus: CiphertextModulus<Scalar>,
+impl Default for TestResources {
+    fn default() -> Self {
+        Self::new()
+    }
```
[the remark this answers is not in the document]
soonum: Clippy is forcing me to do this

tfhe/src/core_crypto/algorithms/mod.rs
IceTDrinker: hmm, ah for the fft impl stuff ? pub(crate) should suffice

tfhe/src/core_crypto/algorithms/test/mod.rs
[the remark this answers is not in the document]
IceTDrinker: some of those lints are a bit ridiculous 🙄

tfhe/src/core_crypto/mod.rs
soonum: I need to mention at least `test`otherwise `pcc` command fails.

IceTDrinker: still not good enough, there must be some missing internal-keycache stuff then in the tests, we must be able to run without keycache

soonum: Ok I'll look at it.

tfhe/src/core_crypto/algorithms/mod.rs
soonum: No it for pre-generation, so it's in core_crypto/keycache.rs but in anycase pub(crate) seems to work.

IceTDrinker: can't the test be in core_crypto::algorithms::tests to make it easier ?

tfhe/src/core_crypto/algorithms/test/lwe_packing_keyswitch.rs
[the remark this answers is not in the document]
soonum: Nice, it will dramatically speed up coverage duration with this trick! :tada:

tfhe/src/shortint/server_key/tests/noise_level.rs
soonum: We should select the second option indeed, just for the sake of being consistent with the rest of the `cfg` directives related to coverage.

soonum:
@slab-ci code_coverage

tfhe/src/shortint/keycache.rs
```diff
@@ -427,7 +418,10 @@ impl KeycacheWopbsV0 {
     pub fn get_from_param<T: Into<WopbsParamPair>>(&self, params: T) -> SharedWopbsKey {
         let params = params.into();
         let key = KEY_CACHE.get_from_param(params.0);
-        let wk = self.inner.get(params);
+        // let wk = self.inner.get(params);
```
IceTDrinker, line 421: comment to remove

tfhe/src/shortint/keycache.rs
```diff
@@ -496,7 +479,14 @@ impl KeycacheKeySwitchingKey {
         let params = params.into();
         let key_1 = KEY_CACHE.get_from_param(params.0);
         let key_2 = KEY_CACHE.get_from_param(params.1);
-        let ksk = self.inner.get(params);
+        // let ksk = self.inner.get(params);
```
IceTDrinker, line 482: comment to remove

tfhe/src/keycache/mod.rs
```diff
@@ -155,28 +155,38 @@ pub mod utils {
...
+    where
+        P: Copy + PartialEq + NamedParam,
+        S: PersistentStorage<P, K>,
+    {
+        pub fn get_with_closure<C: FnMut(P) -> K>(&self, param: P, c: &mut C) -> SharedKey<K> {
+            self.with_closure(param, c)
         }
 
-        pub fn with_key<F, R>(&self, param: P, f: F) -> R
+        pub fn with_closure<C>(&self, param: P, c: &mut C) -> SharedKey<K>
```
IceTDrinker, line 171: could rename c to be a bit more readable key_gen_closure or something

tfhe/src/keycache/mod.rs
```diff
@@ -186,32 +196,34 @@ pub mod utils {
...
-                    shared_cell.inner.get_or_init(load_from_persistent_storage);
-                    Ok(shared_cell)
-                } else {
-                    Err(())
-                }
+                maybe_shared_cell.map_or_else(
+                    || {
+                        {
+                            // we only hold a write lock for a short duration to push the lazily
+                            // evaluated key without actually evaluating the key
```
IceTDrinker, line 203: that comment does not look accurate, the code below does the generation while holding the lock.

I think we should check if a key is in there, if it's in there then done, otherwise generate it and then re-acquire the lock to write in there

In practice it makes little difference for us for now as we pre-generate keys, but I tjink I would prefer have this done once and for all

tfhe/src/core_crypto/mod.rs
```diff
@@ -17,3 +17,6 @@ pub mod prelude;
 pub mod seeders;
 
 pub mod fft_impl;
+
+#[cfg(all(test, feature = "internal-keycache"))]
```
IceTDrinker, line 21: does that work in the end ? 🤔

IceTDrinker (commented):
Haven't finished the review but I'm not sure it's the latest version of the code ? I could be wrong

tfhe/src/core_crypto/algorithms/test/mod.rs
```diff
@@ -225,4 +363,26 @@ macro_rules! create_parametrized_test{
...
+            fn [<test_ $name _ $param:lower>]() {
+                $name($param)
+            }
+            )*
+        }
+    };
+     ($name:ident)=> {
+        create_parametrized_test!($name
+        {
+            TEST_PARAMS_4_BITS_NATIVE_U64
```
IceTDrinker, line 381: this is lowering coverage vs. the other create parameterized tests

soonum: Adding the other parameters set will probably dramatically increase coverage duration.

soonum: I'm running the early break currently to see how long it last.

IceTDrinker: potentially the avx512 does not need to re run both parameter sets

IceTDrinker: but it starts to make a very convoluted test set up

tfhe/src/core_crypto/keycache.rs
```diff
@@ -0,0 +1,319 @@
...
+        } else if scalar_type_id == TypeId::of::<u64>() {
+            unsafe {
+                &*(std::ptr::addr_of!(keycache.u64_multi_bit_cache)
+                    as *const KeyCacheCoreImpl<Self, Self::Keys>)
+            }
+        } else {
+            panic!("No keycache for given Scalar type")
+        }
+    }
+}
```
tmontaigu, line 118: I think you could avoid use of unsafe by doing something like 
```rust
impl<Scalar: UnsignedInteger + Serialize + DeserializeOwned> KeyCacheAccess
    for MultiBitTestParams<u32>
{
    type Keys = MultiBitBootstrapKeys<u32>;

    fn access(keycache: &KeyCache) -> &KeyCacheCoreImpl<Self, Self::Keys> {
        use std::any::TypeId;

    &keycache.u32_multi_bit_cache
  }
}

impl<Scalar: UnsignedInteger + Serialize + DeserializeOwned> KeyCacheAccess
    for MultiBitTestParams<u64>
{
    type Keys = MultiBitBootstrapKeys<u64>;

    fn access(keycache: &KeyCache) -> &KeyCacheCoreImpl<Self, Self::Keys> {
        use std::any::TypeId;

    &keycache.u32_multi_bit_cache
  }
}
```

Its more code tho

IceTDrinker: that would not work from the gen key cache thing though if the tests are generic over Scalar ? (is that the more code you mention ?)

tmontaigu: It will work
```rust

#[derive(Default)]
struct KeyCacheImpl<T>(T);

#[derive(Default)]
struct KeyCache{
    u32: KeyCacheImpl<u32>,
    u64: KeyCacheImpl<u64>,
}


trait KeyCacheAccess: Sized {
    fn access(kc: &KeyCache) -> &KeyCacheImpl<Self>;
}


impl KeyCache {
    fn get<Scalar>(&self) -> Scalar 
    where Scalar: KeyCacheAccess + Copy {
        Scalar::access(self).0
    }
}

impl KeyCacheAccess for u64 {
    fn access(kc: &KeyCache) -> &KeyCacheImpl<Self> {
        &kc.u64
    }
}

impl KeyCacheAccess for u32 {
    fn access(kc: &KeyCache) -> &KeyCacheImpl<Self> {
        &kc.u32
    }
}



fn test<Scalar: KeyCacheAccess + Copy + std::fmt::Display>(kc: &KeyCache) {
    println!("{}", kc.get::<Scalar>());
} 


fn main() {
    let keycache = KeyCache::default();
    test::<u32>(&keycache);
    test::<u64>(&keycache);
}
```

IceTDrinker: So we would need to parameterize with the parameters, as we have several parameter sets for tests which are themselves generic over the scalar

IceTDrinker: But yeah this has a chance to work well

tfhe/src/core_crypto/algorithms/test/mod.rs
soonum: Weirdly, the early break gives us little gains, around 30 min.
I see some tests that can be improved, especially the ones related to key generation.

Does using the second parameters set will improve coverage ? I.e is the code taking other branches in this case ?

IceTDrinker: yes

soonum: I'll add the other parameter set and run the test to check how long it takes.

IceTDrinker: don't forget the keygen

but I think we need to identify the slowest tests and limit the runs of those

tfhe/src/core_crypto/keycache.rs
IceTDrinker: but I know rust would not let us do that at some point because the keycache was not implemented for the Scalar trait with a blanket implementation but rather for specific types implementing the trait

tmontaigu: That sounds extremely strange as  that not how traits should works

IceTDrinker: we were not following the exact same pattern as you so to be checked

soonum:
@slab-ci code_coverage

tfhe/src/keycache/mod.rs
```diff
@@ -186,32 +200,42 @@ pub mod utils {
...
+                        if !has_found_key {
+                            let shared_key = SharedKey {
+                                inner: Arc::new(OnceLock::new()),
+                            };
+                            shared_key.inner.get_or_init(load_from_persistent_storage);
+                            {
+                                // we only hold a write lock for a short duration to push the lazily
+                                // evaluated key without actually evaluating the key
+                                let mut memory_storage = self.memory_storage.write().unwrap();
+                                memory_storage.push((param, shared_key));
```
IceTDrinker, line 222: there needs to be another check if the key already exist here (after getting the lock), to avoid pushing another copy of a key

tfhe/src/core_crypto/keycache.rs
IceTDrinker: it is as I remember, when instantiating the test with a set of parameter the key cache access implementation is not generic enough as it's not implemented for all scalar implementing types

```
error[E0277]: the trait bound `params::MultiBitTestParams<Scalar>: KeyCacheAccess` is not satisfied
   --> tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs:107:18
    |
107 |     let shared = gen_keys_or_get_from_cache_if_enabled(params, &mut keys_gen);
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `KeyCacheAccess` is not implemented for `params::MultiBitTestParams<Scalar>`
```
if you have an implementation with 

```rust
impl KeyCacheAccess for MultiBitTestParams<u32> {
    type Keys = MultiBitBootstrapKeys<u32>;

    fn access(keycache: &KeyCache) -> &KeyCacheCoreImpl<Self, Self::Keys> {
        &keycache.u32_multi_bit_cache
    }
}

impl KeyCacheAccess for MultiBitTestParams<u64> {
    type Keys = MultiBitBootstrapKeys<u64>;

    fn access(keycache: &KeyCache) -> &KeyCacheCoreImpl<Self, Self::Keys> {
        &keycache.u64_multi_bit_cache
    }
}
```

and test instantiation with concrete parameters

IceTDrinker: example of test function being instantiated

```rust
fn std_lwe_encrypt_multi_bit_deterministic_pbs_decrypt_custom_mod<
    Scalar: UnsignedTorus + Sync + Send + CastFrom<usize> + CastInto<usize> + Serialize + DeserializeOwned,
>(
    params: MultiBitTestParams<Scalar>,
) {
```

IceTDrinker: one possibility might be to return a Result to an &dyn Keycache, we return an error if such a keycache is not available (for u128 for example) and return a trait object otherwise maybe

soonum: Don't worry, @tmontaigu walked me through his idea. I've implemented it and it works like a charm.

IceTDrinker: I think I saw, and indeed this is very smart

soonum: Problem is we need to have the keycache always enabled with this solution.
```rust
error[E0432]: unresolved import `crate::core_crypto::keycache`
  --> tfhe/src/core_crypto/fft_impl/common.rs:94:29
   |
94 |     use crate::core_crypto::keycache::KeyCacheAccess;
   |                             ^^^^^^^^ could not find `keycache` in `core_crypto`

error[E0405]: cannot find trait `KeyCacheAccess` in this scope
  --> tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs:78:33
   |
78 |     MultiBitTestParams<Scalar>: KeyCacheAccess<Keys = MultiBitBootstrapKeys<Scalar>>,
   |                                 ^^^^^^^^^^^^^^ not found in this scope


```

tfhe/src/keycache/mod.rs
```diff
@@ -186,32 +200,44 @@ pub mod utils {
...
+                        if !has_found_key {
+                            let shared_key = SharedKey {
+                                inner: Arc::new(OnceLock::new()),
+                            };
+                            shared_key.inner.get_or_init(load_from_persistent_storage);
+                            {
+                                // we only hold a write lock for a short duration to push the lazily
+                                // evaluated key without actually evaluating the key
+                                let mut memory_storage = self.memory_storage.write().unwrap();
+                                if !memory_storage.iter().any(|(p, _)| *p == param) {
```
IceTDrinker, line 222: I think I prefer

```rust
if memory_storage.iter().all(|(p, _)| *p != param)
```

i.e. if all available params are not my params then it is missing the key

tfhe/src/core_crypto/keycache.rs
IceTDrinker: so it's an any enable but as long as the internal-keycache feature is not given it's good right ?

tfhe/src/core_crypto/algorithms/test/mod.rs
```diff
@@ -137,6 +130,133 @@ pub const DUMMY_31_U32: TestParams<u32> = TestParams {
...
+    glwe_dimension: GlweDimension(1),
+    polynomial_size: PolynomialSize(2048),
+    glwe_modular_std_dev: StandardDev(0.0000000000000003152931493498455),
+    message_modulus_log: CiphertextModulusLog(4),
+    ciphertext_modulus: CiphertextModulus::new_native(),
+    grouping_factor: LweBskGroupingFactor(3),
+    thread_count: ThreadCount(12),
+};
+
+#[cfg(feature = "internal-keycache")]
```
IceTDrinker: not required I would think ?

same for the ones below I believe

tfhe/src/core_crypto/algorithms/test/lwe_programmable_bootstrapping.rs
```diff
@@ -251,8 +247,6 @@ fn lwe_encrypt_pbs_f128_decrypt_custom_mod<
 
         convert_standard_lwe_bootstrap_key_to_fourier_128(&bsk, &mut fbsk);
 
-        drop(bsk);
```
IceTDrinker, line 254: why remove this one ? it ensures the key is dropped here, is it a lint from clippy ?

tfhe/src/core_crypto/algorithms/test/params.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub glwe_dimension: GlweDimension,
+    pub polynomial_size: PolynomialSize,
+    pub glwe_modular_std_dev: StandardDev,
+    pub message_modulus_log: CiphertextModulusLog,
+    pub ciphertext_modulus: CiphertextModulus<Scalar>,
+    pub grouping_factor: LweBskGroupingFactor,
+    pub thread_count: ThreadCount,
+}
+
+// Toy parameters to test FFT implementation
```
IceTDrinker, line 79: just parameters, what makes them "toys" are their actual values if they are very small, same for all param struct definitions

tfhe/src/core_crypto/algorithms/test/lwe_programmable_bootstrapping.rs
```diff
@@ -209,37 +232,10 @@ fn lwe_encrypt_pbs_f128_decrypt_custom_mod<
...
-        par_generate_lwe_bootstrap_key(
-            &input_lwe_secret_key,
-            &output_glwe_secret_key,
-            &mut bsk,
-            glwe_modular_std_dev,
-            &mut rsc.encryption_random_generator,
-        );
-
-        assert!(check_content_respects_mod(&*bsk, ciphertext_modulus));
+        let mut keys_gen = |_| generate_keys(params, &mut rsc);
```
IceTDrinker, line 243: shouldn't the closure take params as input ? same everywhere I guess

tfhe/src/core_crypto/algorithms/test/params.rs
```diff
@@ -0,0 +1,289 @@
...
+                "PARAM_PKS_glwe_{}_poly_{}_decomp_base_log_{}_decomp_level_{}_lwe_dim_{}_ct_modulus_{}_lwe_std_dev_{}_glwe_std_dev_{}",
+                self.glwe_dimension.0, self.polynomial_size.0, self.pbs_base_log.0,
+                self.pbs_level.0,
+                self.lwe_dimension.0, self.ciphertext_modulus, self.lwe_modular_std_dev.0.to_string().replace('.', "-"),
+                self.glwe_modular_std_dev.0.to_string().replace('.', "-"),
+            )
+    }
+}
+
+// impl KeyCacheAccess for MultiBitTestParams<u32> {
```
IceTDrinker, line 203: comment to remove

tfhe/src/core_crypto/algorithms/test/mod.rs
soonum: The cost of adding the second parameters set is negligible. It also improve coverage a bit. :tada:

tfhe/src/keycache/mod.rs
soonum: It seems not right, in that case it would means that all keys stored in the directory must have the same parameters set in order for this conditional test to pass no ?

IceTDrinker: we check that all parameters in the vector are not equal to our parameters, which means our parameters are not in the vector, if that's the case we add our parameters + key we just generated

IceTDrinker: it's a matter of taste, both conditions should be equivalent, I just find my formulation easier to read, but I guess it varies from person to person :)

soonum: Oh yes, I miss the *not* equal part in my head. So your proposal makes more sense and is definitely clearer.

soonum:
@slab-ci code_coverage

tfhe/src/core_crypto/keycache.rs
```diff
@@ -0,0 +1,279 @@
...
+
+lazy_static! {
+    pub static ref KEY_CACHE: KeyCache = Default::default();
+}
+
+// This function populates keycache for core_crypto. This is very helpful when doing code coverage
+// since generating keys in that case is very time consuming.
+//#[cfg(not(feature = "__coverage"))]
```
IceTDrinker, line 188: this comment needs to be removed, I think it is the thing slowing down coverage as we are re generating keys as we run tests.

Also I would gate it behind internal-keycache to avoid generating keys for "fun" when running tests without keycache

tfhe/src/keycache/mod.rs
```diff
@@ -186,32 +200,44 @@ pub mod utils {
...
+                            let shared_key = SharedKey {
+                                inner: Arc::new(OnceLock::new()),
+                            };
+                            shared_key.inner.get_or_init(load_from_persistent_storage);
+                            {
+                                // we only hold a write lock for a short duration to push the lazily
+                                // evaluated key without actually evaluating the key
+                                let mut memory_storage = self.memory_storage.write().unwrap();
+                                if memory_storage.iter().all(|(p, _)| *p != param) {
+                                    memory_storage.push((param, shared_key));
```
IceTDrinker, line 223: here we can clone the shared key (it is an Arc so small in memory) instead of reloading it below

IceTDrinker (commented):
Some small things still

tfhe/src/core_crypto/fft_impl/common.rs
```diff
@@ -143,13 +128,56 @@ pub mod tests {
...
+        let params = FftTestParams {
+            lwe_dimension: LweDimension(742),
+            glwe_dimension: GlweDimension(1),
+            polynomial_size: PolynomialSize(2048),
+            lwe_modular_std_dev,
+            glwe_modular_std_dev,
+            pbs_base_log: DecompositionBaseLog(23),
+            pbs_level: DecompositionLevelCount(1),
+            ciphertext_modulus: CiphertextModulus::new_native(),
+        };
```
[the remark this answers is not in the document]
soonum: It looks like the noise used in the tests are the same for the five bootstrap test case.
Shouldn't we create a fixed test parameters set in this case ?

tfhe/src/core_crypto/algorithms/test/params.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub pfks_base_log: DecompositionBaseLog,
+    pub pfks_modular_std_dev: StandardDev,
+    pub cbs_level: DecompositionLevelCount,
+    pub cbs_base_log: DecompositionBaseLog,
+    pub message_modulus_log: CiphertextModulusLog,
+    pub ciphertext_modulus: CiphertextModulus<Scalar>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
+pub struct MultiBitTestParams<Scalar: UnsignedInteger> {
```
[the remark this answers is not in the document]
soonum: If we go down that path we should also avoid comparing against `message_modulus_log` and `ciphertext_modulus` no to match your comment above ?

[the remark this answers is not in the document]
IceTDrinker: ciphertext modulus is critical, message_modulus_log is not

tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -599,179 +532,98 @@ fn std_lwe_encrypt_multi_bit_deterministic_pbs_decrypt_custom_mod<
...
 }
 
 #[test]
 pub fn test_lwe_encrypt_multi_bit_pbs_decrypt_factor_3_thread_12_custom_mod() {
-    lwe_encrypt_multi_bit_pbs_decrypt_custom_mod::<u64>(MultiBitParams {
-        thread_count: ThreadCount(12),
-        message_modulus_log: CiphertextModulusLog(3),
-        ciphertext_modulus: CiphertextModulus::try_new_power_of_2(63).unwrap(),
-        ..MULTI_BIT_2_2_3_PARAMS
-    });
```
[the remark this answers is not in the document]
soonum: Yes I've factorized the parameters as they are using the same values.
If we keep them explicit like that it would painful to pre generate keycache.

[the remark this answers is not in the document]
IceTDrinker: no of course, I just wanted to be sure we tested the same values for the thread counts etc.

soonum:
@slab-ci code_coverage

soonum:
@slab-ci code_coverage

tfhe/src/core_crypto/algorithms/test/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -534,7 +476,7 @@ fn std_lwe_encrypt_multi_bit_deterministic_pbs_decrypt_custom_mod<
 
     while msg != Scalar::ZERO {
         msg = msg.wrapping_sub(Scalar::ONE);
-        for _ in 0..NB_TESTS {
+        for _ in 0..NB_TESTS_LIGHT {
```
IceTDrinker, line 479: no need to change it here I believe

tfhe/src/core_crypto/algorithms/test/mod.rs
```diff
@@ -153,7 +144,162 @@ pub const TEST_PARAMS_3_BITS_SOLINAS_U64: TestParams<u64> = TestParams {
...
+    glwe_dimension: GlweDimension(1),
+    polynomial_size: PolynomialSize(2048),
+    lwe_modular_std_dev: StandardDev(0.12345),
+    glwe_modular_std_dev: StandardDev(0.00000000000000000000000000000008645717832544903),
+    pbs_base_log: DecompositionBaseLog(23),
+    pbs_level: DecompositionLevelCount(1),
+    ciphertext_modulus: CiphertextModulus::<u128>::new_native(),
+};
+
+pub const FFT_WOPBS_PARAMS: FftWopPbsTestParams<u64> = FftWopPbsTestParams {
```
IceTDrinker, line 253: As indicated elsewhere, not sure why the nois parameter is not stored here ? 🤔

IceTDrinker (commented):
Ok final stretch

tfhe/src/keycache/mod.rs
```diff
@@ -155,63 +155,75 @@ pub mod utils {
...
-                let maybe_shared_cell = {
-                    let memory_storage = self.memory_storage.read().unwrap();
-                    memory_storage
-                        .iter()
-                        .find(|(p, _)| *p == param)
-                        .map(|param_key| param_key.1.clone())
+            let try_load_from_memory_and_init = || -> Result<_, ()> {
+                let maybe_shared_key = {
+                    let mut res = None;
+                    let memory_iter = self.memory_storage.read().unwrap().clone().into_iter();
```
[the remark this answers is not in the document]
soonum: clippy forced me to do it.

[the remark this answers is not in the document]
IceTDrinker: just iter() vs clone().into_iter() does not compile ?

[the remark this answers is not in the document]
soonum: Yes because of possible deadlock due the for loop.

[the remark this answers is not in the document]
IceTDrinker: I don't understand how it can deadlock 😵‍💫 

the previous code is 

```rust
                let maybe_shared_cell = {
                    let memory_storage = self.memory_storage.read().unwrap();
                    memory_storage
                        .iter()
                        .find(|(p, _)| *p == param)
                        .map(|param_key| param_key.1.clone())
                };
```

i.e. we drop the lock right at the end of the block, the code above is rejected by clippy with what error message ?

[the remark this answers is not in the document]
soonum: Here is the working implementation of what you've suggested previousely:
```rust
let maybe_shared_key = {
                    let mut res = None;
                    for (p, key) in self.memory_storage.read().unwrap().iter() {
                        if *p == param {
                            res = Some(key.clone());
                            break;
                        }
                    }
                    res
                };
```
And here's the associated error:
```rust
rror: temporary with significant `Drop` in `for` loop condition will live until the end of the `for` expression
   --> tfhe/src/keycache/mod.rs:197:37
    |
197 |                     for (p, key) in self.memory_storage.read().unwrap().iter() {
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
202 |                     }
    |                      - temporary lives until here
    |
    = note: this might lead to deadlocks or other unexpected behavior
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee
    = note: `-D clippy::significant-drop-in-scrutinee` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::significant_drop_in_scrutinee)]`

```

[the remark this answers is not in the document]
IceTDrinker: yes I meant the previous code ? why was clippy complaining ?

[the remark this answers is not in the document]
IceTDrinker: this one https://github.com/zama-ai/tfhe-rs/pull/664#discussion_r1400690128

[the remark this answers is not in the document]
soonum: Clippy wasn't complaining with the previous code, it's the refactor suggestion you made that triggered clippy.

IceTDrinker (approved):
All right let's go 🚀

soonum:
Resolves: zama-ai/tfhe-rs-internal#230

PR #665, opened 2023-11-03
enable pedantic clippy lints

mayeul-zama:


IceTDrinker:
change does not look too big from afar which is nice, needs a rebase though :)

tfhe/src/c_api/shortint/parameters.rs
```diff
@@ -47,6 +41,7 @@ pub struct ShortintPBSParameters {
     pub encryption_key_choice: ShortintEncryptionKeyChoice,
 }
 
+#[allow(clippy::fallible_impl_from)]
 impl From<ShortintPBSParameters> for crate::shortint::ClassicPBSParameters {
     fn from(c_params: ShortintPBSParameters) -> Self {
```
IceTDrinker, line 46: maybe we should have a try from here then ?

IceTDrinker (commented):
Could we fix stuff instead of adding allows ?

mayeul-zama:
> Could we fix stuff instead of adding allows ?

Sure for most of them I think it makes sense.
For some others (especially nursery ones) it makes less sens IMO.

I was thinking that we could fix them one by one or in small groups to have small PRs

tfhe/src/lib.rs
```diff
@@ -2,6 +2,57 @@
 //!
 //! TFHE-rs is a fully homomorphic encryption (FHE) library that implements Zama's variant of TFHE.
 
+#![warn(clippy::pedantic)]
+#![allow(clippy::must_use_candidate)]
```
IceTDrinker, line 6: erm, no ?

IceTDrinker:
I would probably do it the other way around, add some more restrictive lints and fix as we go, e.g. for the Self stuff turn this on with the fix and commit the small corresponding PR, here we allow a shit ton of stuff which may not even be relevant for now

tfhe/src/lib.rs
IceTDrinker: After checking the pedantic group looks like we can't do much about this :| so I guess would be the way to go

tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs
```diff
@@ -358,6 +358,7 @@ where
             fft: Fft128View<'_>,
             stack: PodStack<'_>,
         ) {
+            #[allow(clippy::transmute_undefined_repr)]
```
IceTDrinker, line 361: is there something we can do for the undefined repr ?

tfhe/src/js_on_wasm_api/js_high_level_api/mod.rs
```diff
@@ -2,6 +2,7 @@ use wasm_bindgen::prelude::*;
 
 pub(crate) mod config;
 pub(crate) mod integers;
+#[allow(clippy::use_self)]
```
IceTDrinker, line 5: this can probably be fixed without too much of a hassle ?

tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -97,11 +97,11 @@ impl Degree {
             }
         }
 
-        Degree(result)
+        Self(result)
     }
 
     #[allow(dead_code)]
-    pub(crate) fn after_pbs<F>(&self, f: F) -> Degree
+    pub(crate) fn after_pbs<F>(&self, f: F) -> Self
```
IceTDrinker, line 104: this can probably be removed as we have the degree in the accumulator directly

IceTDrinker: in a follow up PR potentially

tfhe/src/shortint/server_key/scalar_bitwise_op.rs
```diff
@@ -57,7 +57,6 @@ impl ServerKey {
     }
 
     // by convention smart operations take mut refs to their inputs, even if they do not modify them
-    #[allow(clippy::needless_pass_by_ref_mut)]
```
IceTDrinker, line 60: why are these ones gone but the comment remains ?

tfhe/src/lib.rs
```diff
@@ -2,6 +2,61 @@
 //!
 //! TFHE-rs is a fully homomorphic encryption (FHE) library that implements Zama's variant of TFHE.
 
+// Enable pedantic lints
+#![warn(clippy::pedantic)]
+// Fixing the following lints may be useful
+#![allow(clippy::must_use_candidate)]
```
IceTDrinker, line 8: add a comment to indicate that by default those are set to allow and that for now we are introducing lints progressively

tfhe/src/js_on_wasm_api/js_high_level_api/mod.rs
mayeul-zama: Added a comment to explain why not (using Self does not work well with #[wasm_bindgen] macro)

tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs
mayeul-zama: Would probably need a refactor to treat u128 differently

IceTDrinker: yeah no worries, we'll go from the allow list in lib.rs and remove the allows progressively for what makes sense

tfhe/src/shortint/server_key/scalar_bitwise_op.rs
mayeul-zama: I missed it
Will remove them

tfhe/src/c_api/shortint/parameters.rs
mayeul-zama: Fixed

tfhe/src/shortint/ciphertext/mod.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/316

tfhe/src/core_crypto/commons/math/random/mod.rs
```diff
@@ -67,9 +67,9 @@ where
         distribution: D,
         slice: &mut [Self],
     ) {
-        slice.iter_mut().for_each(|s| {
+        for s in slice.iter_mut() {
             *s = Self::generate_one(generator, distribution);
-        });
+        }
```
IceTDrinker, line 72: what motivated that change ?

tfhe/src/js_on_wasm_api/js_high_level_api/mod.rs
tmontaigu: I don't think its possible for all in the wasm part.

e.g:

```
#[wasm_bindgen]
impl TfheClientKey {
    #[wasm_bindgen]
    pub fn generate(config: &TfheConfig) -> Result<TfheClientKey, JsError> {
        catch_panic(|| Self(hlapi::ClientKey::generate(config.0.clone())))
    }
}
```

We would want to return `Result<Self, JsError>`, but `#[wasm_bindgen]` does not allow that so we have to keep `Result<TfheClientKey, JsError>`

tfhe/src/core_crypto/commons/math/random/mod.rs
tmontaigu: using for_each is an 'anti-pattern', you should always (unless its not possible) prefer for loop instead of using for_each

> This is equivalent to using a [for](https://doc.rust-lang.org/book/ch03-05-control-flow.html#looping-through-a-collection-with-for) loop on the iterator, although break and continue are not possible from a closure. *It’s generally more idiomatic to use a for loop*

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

IceTDrinker: huh, I thought the for_each approach was preferred for some iterator specific optimization 🤔

mayeul-zama: https://rust-lang.github.io/rust-clippy/master/index.html#/needless_for_each

tfhe/src/lib.rs
mayeul-zama: Changed the comment
Let me know if you have suggestions to improve it

tfhe/src/core_crypto/commons/math/random/mod.rs
IceTDrinker: ok yeah then my bad, I started to adopt the "for each is generally better", which seems to be the case unless you have only a single variable or need some break/continue functionality

tfhe/src/lib.rs
IceTDrinker: just curious what are the comments after each lint ?

mayeul-zama: It's the current number of warnings emitted if we remove the line
This may change with time so it's just to have an idea of the amount of work needed to fix the lint

IceTDrinker: thx

tfhe/src/shortint/public_key/compact.rs
```diff
@@ -86,7 +88,7 @@ impl CompactPublicKey {
     }
 
     pub fn encrypt(&self, message: u64) -> Ciphertext {
-        let plain = to_plaintext_iterator([message].iter().copied(), &self.parameters)
+        let plain = to_plaintext_iterator(once(message), &self.parameters)
```
IceTDrinker, line 91: neat, did not even know this existed

tfhe/src/integer/block_decomposition.rs
```diff
@@ -65,7 +65,7 @@ impl<const N: usize> RecomposableFrom<u8> for StaticUnsignedBigInt<N> {}
 impl<const N: usize> DecomposableInto<u64> for StaticUnsignedBigInt<N> {}
 impl<const N: usize> DecomposableInto<u8> for StaticUnsignedBigInt<N> {}
 
-#[derive(Copy, Clone)]
+#[derive(Clone)]
```
IceTDrinker, line 68: why is Copy removed here we could want the Copy derive if T is copy right ?

mayeul-zama: https://rust-lang.github.io/rust-clippy/master/index.html#/copy_iterator

IceTDrinker (approved):
Good for me, all PRs will need a rebase though

PR #668, opened 2023-11-06
Generic int over Id

tmontaigu:
Have GenericInteger be generic over the `Id` directly and not some `P` (Parameters) from which we get the `Id`

### DIFF
--- a/tfhe/src/high_level_api/integers/parameters.rs
+++ b/tfhe/src/high_level_api/integers/parameters.rs
@@ -1,5 +1,3 @@
-use crate::high_level_api::internal_traits::ParameterType;
-
 /// Meant to be implemented on the inner server key
 /// eg the crate::integer::ServerKey
 pub trait EvaluationIntegerKey<ClientKey> {
@@ -12,8 +10,8 @@ pub trait EvaluationIntegerKey<ClientKey> {
     ) -> crate::integer::wopbs::WopbsKey;
 }
 
-/// Trait to mark parameters type for integers
-pub trait IntegerParameter: ParameterType {
+/// Trait to mark Id type for integers
+pub trait IntegerId: Copy + Default {
     type InnerCiphertext: crate::integer::ciphertext::IntegerRadixCiphertext;
     type InnerCompressedCiphertext;
 
--- a/tfhe/src/high_level_api/integers/types/base.rs
+++ b/tfhe/src/high_level_api/integers/types/base.rs
@@ -6,9 +6,9 @@ use std::ops::{
 
 use crate::conformance::ParameterSetConformant;
 use crate::high_level_api::global_state::WithGlobalKey;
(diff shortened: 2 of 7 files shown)


tmontaigu:
@slab-ci cpu_fast_test

tfhe/src/high_level_api/integers/parameters.rs
```diff
@@ -13,7 +11,7 @@ pub trait EvaluationIntegerKey<ClientKey> {
 }
 
 /// Trait to mark parameters type for integers
-pub trait IntegerParameter: ParameterType {
+pub trait IntegerId: Copy + Default {
```
IceTDrinker, line 14: why do you prefer the Id wording to the Parameter one in that case ? I know it's something I never quite understood from the HL API design given I never really edited code in there

tfhe/src/high_level_api/integers/types/base.rs
```diff
@@ -78,55 +78,53 @@ impl std::fmt::Display for GenericIntegerBlockError {
...
-        ciphertext: P::InnerCiphertext,
-        id: P::Id,
+        ciphertext: Id::InnerCiphertext,
+        id: Id,
     ) -> Self {
         Self { ciphertext, id }
     }
 
-    pub fn cast_from<P2>(other: GenericInteger<P2>) -> Self
+    pub fn cast_from<Id2>(other: GenericInteger<Id2>) -> Self
```
IceTDrinker, line 111: FromId ?

tmontaigu: What do you mean ?

tfhe/src/high_level_api/integers/parameters.rs
tmontaigu: Originally, each type (FheUint8, 16, 32, etc) was meant to have its own parameters, so there was `FheUint{8,16,32}Parameters`. Each parameter had an associated `Id` `FheUint{8,16,32}d` that is a zero sized typed stored in the GenericInteger, and this Id has some behabiour attached to it (like how to retrieve the correct client/server key)

However things changed now all types share the same key, the same base block and same parameters, so it makes more sense to just keep the `Id` thing.

tfhe/src/high_level_api/integers/types/base.rs
IceTDrinker: I would call Id2 "FromId" ?

tfhe/src/high_level_api/integers/parameters.rs
IceTDrinker: Ok thanks, I think there are some docstrings needing update just above that struct in that case, not sure about other locations

tfhe/src/high_level_api/integers/types/base.rs
tmontaigu: A yes good suggestion

PR #673, opened 2023-11-08
Am/chore/next tfhe wip

IceTDrinker:
### PR content/description

Various improvements to the build system to be able to use the crate as a dependency in a semver-trick setting, main change is a way to by pass the build.rs script which was erroring due to weird cargo conflicts on package names (and maybe the fact our build.rs is not that smart to begin with with path allocations and such, so being able to skip it entirely just makes things easier for us for the semver trick)

The added TFHE_SPEC in the Makefile is to make sure Cargo can disambiguate the fact we are talking about our package and not a dependency, see https://github.com/rust-lang/cargo/issues/12891

### Check-list:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ TARGET_ARCH_FEATURE:=$(shell ./scripts/get_arch_feature.sh)
 RS_BUILD_TOOLCHAIN:=stable
 CARGO_RS_BUILD_TOOLCHAIN:=+$(RS_BUILD_TOOLCHAIN)
 CARGO_PROFILE?=release
-MIN_RUST_VERSION:=$(shell grep rust-version tfhe/Cargo.toml | cut -d '=' -f 2 | xargs)
+MIN_RUST_VERSION:=$(shell grep '^rust-version[[:space:]]*=' tfhe/Cargo.toml | cut -d '=' -f 2 | xargs)
 AVX512_SUPPORT?=OFF
 WASM_RUSTFLAGS:=
 BIG_TESTS_INSTANCE?=FALSE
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,6 +17,10 @@ FAST_TESTS?=FALSE
 FAST_BENCH?=FALSE
 BENCH_OP_FLAVOR?=DEFAULT
 NODE_VERSION=20
+TFHE_CURRENT_VERSION:=$(shell grep '^version[[:space:]]*=' tfhe/Cargo.toml | cut -d '=' -f 2 | xargs)
```
mayeul-zama, line 20: This could give the wrong result if a crate dependency is defined before the package version and its version is in new line 

That's unlikely

tfhe/cbindgen.toml
```diff
@@ -114,14 +114,15 @@ bitflags = false
 ############## Options for How Your Rust library Should Be Parsed ##############
 
 [parse]
-parse_deps = true
-include = ["tfhe"]
+parse_deps = false
+include = []
```
mayeul-zama, line 118: What does this change? Why is this necessary?

tfhe/build.rs
```diff
@@ -63,8 +64,8 @@ fn gen_c_api() {
     };
 
     cbindgen::Builder::new()
-        .with_crate(crate_dir.clone())
-        .with_config(cbindgen::Config::from_root_or_default(crate_dir))
+        .with_crate(crate_dir.as_path())
+        .with_config(cbindgen::Config::from_file(crate_dir.join("cbindgen.toml")).unwrap())
```
mayeul-zama, line 68: Is this to be more strict by precaution?

tfhe/cbindgen.toml
IceTDrinker: it was a mistake from the time the C API required macro expansion, I think I made the mistake at the time, we don't need to parse deps, only our own crate, the include here is not required as it refers to dependencies as far as I understand the cbindgen doc, the parse expand is handled in the build.rs to programatically feed features that are enabled

tfhe/build.rs
IceTDrinker: yes

Makefile
mayeul-zama: Not sure if it exist, but inserting a utility which asserts that a stream is only one line and forwards it would be more secure (even if not perfect)

IceTDrinker: though it looks like the Cargo.toml format does not allow new lines for { version = "..." } stuff

mayeul-zama: I think you can have 
```
[dependencies.dep_name]
version = "1.0"
```

IceTDrinker:
gotta love that the unambiguous spec for run (and only run) does not work

Makefile
IceTDrinker: yeah :|

IceTDrinker: I guess I could trim the file once we find the word "dependencies"

IceTDrinker: something like this maybe https://stackoverflow.com/questions/7103531/how-to-get-the-part-of-a-file-after-the-first-line-that-matches-a-regular-expres

IceTDrinker: or maybe a light toml parser

mayeul-zama: That seems overkill to me

IceTDrinker: got something a bit more robust

IceTDrinker:
> looks ok to me

Approve if you think it can go for merging pretty please 😄

tmontaigu (approved):

PR #674, opened 2023-11-08
feat(integer): add unsigned_oveflowing_add

tmontaigu:


tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -232,12 +232,94 @@ impl ServerKey {
...
+    /// let dec_overflowed = cks.decrypt_one_block(&overflowed);
+    /// let (expected_result, expected_overflow) = msg1.overflowing_add(msg2);
+    /// assert_eq!(dec_result, expected_result);
+    /// assert_eq!(dec_overflowed, u64::from(expected_overflow));
+    /// ```
+    pub fn unsigned_overflowing_add_parallelized(
+        &self,
+        ct_left: &RadixCiphertext,
+        ct_right: &RadixCiphertext,
+    ) -> (RadixCiphertext, Ciphertext) {
```
IceTDrinker, line 272: should the overflow flag be a boolean block or a plain ciphertext ?

I know there is likely the question of a potential PBS in there

I guess the question is: does the carry block contain a 1 or a 0 ? if so then we can likely wrap it in a boolean block

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -73,6 +73,11 @@ fn overflowing_sub_under_modulus(lhs: u64, rhs: u64, modulus: u64) -> (u64, bool
     (r, q != 0)
 }
 
+fn overflowing_add_under_modulus(lhs: u64, rhs: u64, modulus: u64) -> (u64, bool) {
+    let result = lhs.wrapping_add(rhs);
+    (result % modulus, result >= modulus)
+}
```
IceTDrinker, line 79: let's say your modulus is 2^63 + 1

I take 2^63 + 2^63 result is 2^64 or 0 in a u64, the computation overflowed but we would not see it here, am I correct ?

Maybe the sub under modulus is wrong as well 🤔

IceTDrinker (commented):
Big doubt on the add under modulus which makes me wonder if sub under modulus is correct 🤔

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
IceTDrinker:
```rust
fn overflowing_sub_under_modulus(lhs: u64, rhs: u64, modulus: u64) -> (u64, bool) {
    let overflowed = rhs > lhs;
    let result = if overflowed {
        todo!() // maybe use wrapping_sub_custom mod from the unsigned integer trait
    } else {
        lhs - rhs
    };
    (result, overflowed)
}
```

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: I planned on making it a BooleanBlock after the merge (this branch is not up to date with the merge of booleanblock)
and also the 2 overflowing_sub do not return boolean block yet

IceTDrinker: ok !

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: Yeah, but these functions are written only for the tests where we don't go beyond a modulus of 2^16

IceTDrinker: understood, could you add the assert in the two functions to avoid issues on the value ranges ? or update them so that we don't mistakenly use them in contexts where they are not meant to be used ?

IceTDrinker: and sorry for being annoying with this but I just battled against moduli 😅 I'd rather we don't have weird surprises

IceTDrinker: indeed should be good

IceTDrinker: could you take a look for the sub as well ?

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -65,12 +65,13 @@ fn rotate_right_helper(value: u64, n: u32, actual_bit_size: u32) -> u64 {
...
-    let (q, r) = (result / modulus, result % modulus);
+    let (result, overflowed) = lhs.overflowing_sub(rhs);
+    (result % modulus, overflowed)
+}
 
-    (r, q != 0)
+fn overflowing_add_under_modulus(lhs: u64, rhs: u64, modulus: u64) -> (u64, bool) {
+    let (result, overflowed) = lhs.overflowing_add(rhs);
+    (result % modulus, overflowed || result >= modulus)
 }
```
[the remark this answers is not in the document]
tmontaigu: It works when modulus is not a power of two, the case where is will not work is if the modulus is not a power of two **and** the sum of two values overflows u64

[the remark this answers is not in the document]
tmontaigu: Actually I tried and it works even in this case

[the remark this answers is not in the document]
IceTDrinker: hmm yes I can see it, sub will not work though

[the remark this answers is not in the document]
tmontaigu: yep sub won't work I added the assert for it

IceTDrinker (approved):
Thanks !

PR #676, opened 2023-11-09
chore(integer): restore assert after using 3_3 params for CRT doctests

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/256

### PR content/description

- fix max degree for CRT keys which don't need to propagate carries

also prevent users to have access to gen_keys which was wrong for the CRT max degree, the old Integer ClientKey is still accessible through as_ref calls, but some convenience functions have been added on the key itself

Some API breaks because of the change

### Check-list:

### DIFF
--- a/tfhe/src/integer/mod.rs
+++ b/tfhe/src/integer/mod.rs
@@ -20,14 +20,14 @@
 //! //4 blocks for the radix decomposition
 //! let number_of_blocks = 4;
 //! // Modulus = (2^2)*4 = 2^8 (from the parameters chosen and the number of blocks
-//! let modulus = 1 << 8;
+//! let modulus = 1u64 << 8;
 //!
 //! // Generation of the client/server keys, using the default parameters:
 //! let (mut client_key, mut server_key) =
 //!     gen_keys_radix(PARAM_MESSAGE_2_CARRY_2_KS_PBS, number_of_blocks);
 //!
-//! let msg1 = 153;
-//! let msg2 = 125;
+//! let msg1 = 153u64;
+//! let msg2 = 125u64;
 //!
 //! // Encryption of two messages using the client key:
 //! let ct_1 = client_key.encrypt(msg1);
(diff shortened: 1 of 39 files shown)


IceTDrinker:
@slab-ci cpu_fast_test

tfhe/src/integer/mod.rs
```diff
@@ -76,6 +76,14 @@ pub use client_key::{ClientKey, CrtClientKey, RadixClientKey};
 pub use public_key::{CompressedCompactPublicKey, CompressedPublicKey, PublicKey};
 pub use server_key::{CheckError, CompressedServerKey, ServerKey};
 
+/// Enum to indicate which kind of computations the [`ServerKey`] will be performing, this changes
+/// the parameterization of the key to manage carries in the Radix case.
+#[derive(Clone, Copy, Debug)]
+pub enum IntegerKeyKind {
+    Radix,
+    CRT,
+}
```
tmontaigu, line 85: Would it make sense to later have to disctinct ServerKey struct instead ? 
As with this enum a use could potentially using function that do not match the kind

IceTDrinker: I was wondering about that too, at the moment the design is this way, and that's why I asked if we imagined having a key able to do both

tmontaigu: We'll see how it should evolve

tmontaigu (approved):

tfhe/src/integer/mod.rs
IceTDrinker: Ok, will open a follow up issue

IceTDrinker: https://github.com/zama-ai/tfhe-rs-internal/issues/321

IceTDrinker:

IceTDrinker:
the failure is linked to keys being generated for radix and then loaded for CRT from the cache, several functions are only for the radix case on server keys, I will take the time to update the various functions

or it is time to introduce a separate radix and crt server keys, will wait next week

IceTDrinker:
doctests run ok on my m6i, still would prefer a proof read for the last things I changed

tmontaigu (approved):

PR #678, opened 2023-11-09
chore(tfhe): fix builds on main

tmontaigu:


IceTDrinker (approved):
Thanks, ideas as to why it was not caught during the PR process ? Clippy tests, etc. ?

IceTDrinker (approved):

PR #680, opened 2023-11-10
fix(integer): fix signed_overflowing_sub using trivial 0

tmontaigu:


tfhe/src/shortint/engine/server_side/neg.rs
```diff
@@ -41,7 +42,9 @@ impl ShortintEngine {
     ) -> EngineResult<u64> {
         // z = ceil( degree / 2^p ) * 2^p
         let msg_mod = ct.message_modulus.0;
-        let mut z = ((ct.degree.0 + msg_mod - 1) / msg_mod) as u64;
+        // Ensure z is always >= 1 (which would not be the case if degree == 0)
+        // some algorithms (e.g. overflowing_sub) require this even for trivial zeros
+        let mut z = divide_ceil(ct.degree.0, msg_mod).max(1) as u64;
```
IceTDrinker, line 47: care to elaborate the fix ? otherwise looks good

tmontaigu: negation is (0 - value), but that would overflow the padding bit, so we add a correcting term
the correcting term depends on the degree. eg, degree = 3, param 2_2 -> correcting term = 4

If the degree is 0, the correcting term is 0, the negation is still ok (0 - 0 + 0) = 0 = -0,
but, for properoverflow detection we need the correcting term to always be added regardless.
so it will now do (0 - 0 + 4) % 4 = 4 % 4 = 0.

IceTDrinker: ok z is always at least one and so multiplies the msg_mod below, got it thanks !

IceTDrinker (approved):
Thank you !

tmontaigu:
@slab-ci cpu_test

PR #681, opened 2023-11-10
Fhebool boolean block

tmontaigu:
- This makes FheBool use integer::BooleanBlock internally.
- It makes comparisons (eq, ne, le, etc) return a FheBool instead of
  FheUint/FheInt.
- It also moves the if_then_else and cmux methods to FheBool.
- Adds casting from FheBool to FheUint/FheInt (but not from
  FheUint/FheInt to FheBool as we expect users to do `a.ne(0)`
  as its matches Rust)

BREAKING CHANGE:
    - Comparisons now return FheBool
    - if_then_else/cmux are now methods of FheBool.

### DIFF
--- a/tfhe/src/high_level_api/booleans/mod.rs
+++ b/tfhe/src/high_level_api/booleans/mod.rs
@@ -3,11 +3,13 @@ use std::ops::{BitAnd, BitOr, BitXor};
 
 use crate::errors::Type;
 use crate::high_level_api::global_state::WithGlobalKey;
+use crate::high_level_api::integers::{GenericInteger, IntegerId};
 use crate::high_level_api::internal_traits::TypeIdentifier;
 use crate::high_level_api::keys::{ClientKey, PublicKey};
 use crate::high_level_api::traits::{
     FheDecrypt, FheEq, FheTrivialEncrypt, FheTryEncrypt, FheTryTrivialEncrypt,
 };
+use crate::integer::BooleanBlock;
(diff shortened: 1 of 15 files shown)


tfhe/src/high_level_api/booleans/mod.rs
```diff
@@ -159,7 +195,7 @@ impl FheTryEncrypt<bool, CompressedPublicKey> for crate::FheBool {
     fn try_encrypt(value: bool, key: &CompressedPublicKey) -> Result<Self, Self::Error> {
         let key = &key.key;
         let ciphertext = key.key.encrypt(u64::from(value));
-        Ok(Self::new(ciphertext))
+        Ok(Self::new(BooleanBlock::new_unchecked(ciphertext)))
```
IceTDrinker, line 198: internally those ops still return a shortint ciphertext ?

tmontaigu: Hum, Yeah I haven't added encryt_bool for the different PublicKey types of integer. Should I add them in this PR ?

IceTDrinker: is it a large change ? otherwise it's fine to put it in a separate PR I would say

tmontaigu: I added the 2 missing functions

IceTDrinker (approved):
Thanks a lot !

PR #682, opened 2023-11-10
chore(tfhe): fix required features for the generate_test_keys util

IceTDrinker:


soonum (approved):
Thanks for the quick fix.

PR #684, opened 2023-11-13
clean shortintengine

mayeul-zama:


tfhe/src/js_on_wasm_api/shortint.rs
```diff
@@ -340,17 +340,17 @@ impl Shortint {
...
         let seed: u128 = (seed_high_bytes << 64) | seed_low_bytes;
 
         let mut seeder = DeterministicSeeder::<ActivatedRandomGenerator>::new(Seed(seed));
-        crate::shortint::engine::ShortintEngine::new_from_seeder(&mut seeder)
-            .new_client_key(parameters.0.try_into().unwrap())
-            .map_err(|e| wasm_bindgen::JsError::new(format!("{e:?}").as_str()))
-            .map(ShortintClientKey)
+        ShortintClientKey(
+            crate::shortint::engine::ShortintEngine::new_from_seeder(&mut seeder)
+                .new_client_key(parameters.0.try_into().unwrap()),
```
IceTDrinker, line 352: makes me think we may want to forward the try into potential error to JS, can be done in a separate PR/follow up issue, it's not critical as it was already like that

IceTDrinker (approved):
Looks good thanks !

tfhe/src/js_on_wasm_api/shortint.rs
IceTDrinker: thanks

mayeul-zama:
@slab-ci cpu_test

PR #687, opened 2023-11-15
feat(integer): add accessors to inner shortint sks

tmontaigu:
Users can access blocks from an integer but they don't have the ability to use the inner shortint server key to process individual blocks.

This adds an AsRef impl on integer ServerKey to allow that.

This also adds shortcuts to the integer ServerKey to get the MessageModulus/CarryModulus (these are shorticuts because users could do `integer_key.as_ref().message_modulus`.

Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/182


IceTDrinker (approved):
Thanks a lot !

PR #688, opened 2023-11-15
chore(ci): spread tests between two CI machines/workflow for faster runtime

IceTDrinker:


soonum (approved):

soonum (approved):

.github/workflows/aws_tfhe_multi_bit_tests.yml
```diff
@@ -77,6 +77,12 @@ jobs:
         run: |
           AVX512_SUPPORT=ON make test_integer_multi_bit_ci
 
+      - name: Run signed integer tests
```
IceTDrinker, line 80: I will add the key gen for the integers as well in the a final commit/edit once the current tests run (to check distribution is helping and that I did not break the AVX512 in the Makefile)

IceTDrinker:
@tmontaigu opinions on that, I see there are several files with test cases and it's not clear all signed/unsigned cases are in separate sub modules

IceTDrinker:
@slab-ci cpu_integer_test
@slab-ci cpu_multi_bit_test

IceTDrinker:
@slab-ci cpu_fast_test

tmontaigu:
> @tmontaigu opinions on that, I see there are several files with test cases and it's not clear all signed/unsigned cases are in separate sub modules

Tests are in 4 files
- tests_cases_unsigned.rs contains test cases for unsigned radix, to be used (but it does not contain any #[tests]
- tests_unsigned.rs contains the actual unsigne #[test] made using tests cases
- tests_signed.rs contains actual signed #[test]
- tests_cases_comparisons.rs contains #[tests] for both signed/unsigned comparisons

In any case, signed tests are prefixed with `signed` :`integer_default_signed_etc`

IceTDrinker:
> In any case, signed tests are prefixed with `signed` :`integer_default_signed_etc`

Yes I was able to filter on _signed was just wondering if you had a preference for test organization, but this should be ok

tmontaigu:
That looks ok for me

PR #689, opened 2023-11-16
chore(integer): make oveflowing_add/sub return BooleanBlock

tmontaigu:


tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -1817,15 +1816,14 @@ where
         let (ct_res, result_overflowed) = executor.execute((&ctxt_0, &ctxt_1));
         let (tmp_ct, tmp_o) = executor.execute((&ctxt_0, &ctxt_1));
         assert!(ct_res.block_carries_are_empty());
-        assert!(result_overflowed.carry_is_empty());
```
IceTDrinker, line 1820: in what case can the carry be non empty ?

tmontaigu: Since its a boolean block, carries are always empty

IceTDrinker: that's very nice, cool thanks !

IceTDrinker (approved):
Thanks !

PR #690, opened 2023-11-16
feat(tfhe): add LWE encryption with non power of 2 moduli

IceTDrinker:
### PR content/description

Add support for non power of 2 q for LWE encryption and linalg, LWE encryption tests and LWE linear algebra tests have been updated to make use of the Solinas set of parameter

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
@@ -3,6 +3,7 @@
 
 use crate::core_crypto::algorithms::slice_algorithms::*;
 use crate::core_crypto::algorithms::*;
+use crate::core_crypto::commons::ciphertext_modulus::CiphertextModulusKind;
 use crate::core_crypto::commons::dispersion::DispersionParameter;
 use crate::core_crypto::commons::generators::{EncryptionRandomGenerator, SecretRandomGenerator};
 use crate::core_crypto::commons::math::random::{ActivatedRandomGenerator, RandomGenerator};
@@ -36,25 +37,129 @@ pub fn fill_lwe_mask_and_body_for_encryption<Scalar, KeyCont, OutputCont, Gen>(
 
     let ciphertext_modulus = output_mask.ciphertext_modulus();
 
+    if ciphertext_modulus.is_compatible_with_native_modulus() {
+        fill_lwe_mask_and_body_for_encryption_native_mod_compatible(
+            lwe_secret_key,
+            output_mask,
+            output_body,
+            encoded,
+            noise_parameters,
+            generator,
+        );
+    } else {
+        fill_lwe_mask_and_body_for_encryption_other_mod(
+            lwe_secret_key,
+            output_mask,
+            output_body,
+            encoded,
+            noise_parameters,
+            generator,
+        );
+    }
+}
+
+pub fn fill_lwe_mask_and_body_for_encryption_native_mod_compatible<
+    Scalar,
+    KeyCont,
+    OutputCont,
+    Gen,
+>(
+    lwe_secret_key: &LweSecretKey<KeyCont>,
+    output_mask: &mut LweMask<OutputCont>,
+    output_body: &mut LweBodyRefMut<Scalar>,
+    encoded: Plaintext<Scalar>,
+    noise_parameters: impl DispersionParameter,
+    generator: &mut EncryptionRandomGenerator<Gen>,
+) where
+    Scalar: UnsignedTorus,
+    KeyCont: Container<Element = Scalar>,
+    OutputCont: ContainerMut<Element = Scalar>,
+    Gen: ByteRandomGenerator,
+{
+    assert_eq!(
+        output_mask.ciphertext_modulus(),
+        output_body.ciphertext_modulus(),
+        "Mismatched moduli between mask ({:?}) and body ({:?})",
+        output_mask.ciphertext_modulus(),
+        output_body.ciphertext_modulus()
+    );
+
+    let ciphertext_modulus = output_mask.ciphertext_modulus();
+
     assert!(ciphertext_modulus.is_compatible_with_native_modulus());
 
+    // generate a randomly uniform mask
     generator.fill_slice_with_random_mask_custom_mod(output_mask.as_mut(), ciphertext_modulus);
 
     // generate an error from the normal distribution described by std_dev
-    *output_body.data = generator.random_noise_custom_mod(noise_parameters, ciphertext_modulus);
-    *output_body.data = (*output_body.data).wrapping_add(encoded.0);
-
-    if !ciphertext_modulus.is_native_modulus() {
-        let torus_scaling = ciphertext_modulus.get_power_of_two_scaling_to_native_torus();
-        slice_wrapping_scalar_mul_assign(output_mask.as_mut(), torus_scaling);
-        *output_body.data = (*output_body.data).wrapping_mul(torus_scaling);
-    }
-
+    let noise = generator.random_noise_custom_mod(noise_parameters, ciphertext_modulus);
     // compute the multisum between the secret key and the mask
-    *output_body.data = (*output_body.data).wrapping_add(slice_wrapping_dot_product(
+    let mask_key_dot_product = (*output_body.data).wrapping_add(slice_wrapping_dot_product(
         output_mask.as_ref(),
         lwe_secret_key.as_ref(),
     ));
+
+    // Store sum(ai * si) + delta * m + e in the body
+    *output_body.data = mask_key_dot_product
+        .wrapping_add(encoded.0)
+        .wrapping_add(noise);
+
+    match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native => (),
+        CiphertextModulusKind::NonNativePowerOfTwo => {
+            // Manage power of 2 encoding to map to the native case
+            let torus_scaling = ciphertext_modulus.get_power_of_two_scaling_to_native_torus();
+            slice_wrapping_scalar_mul_assign(output_mask.as_mut(), torus_scaling);
+            *output_body.data = (*output_body.data).wrapping_mul(torus_scaling);
+        }
+        CiphertextModulusKind::Other => unreachable!(),
+    };
+}
+
+pub fn fill_lwe_mask_and_body_for_encryption_other_mod<Scalar, KeyCont, OutputCont, Gen>(
+    lwe_secret_key: &LweSecretKey<KeyCont>,
+    output_mask: &mut LweMask<OutputCont>,
+    output_body: &mut LweBodyRefMut<Scalar>,
+    encoded: Plaintext<Scalar>,
+    noise_parameters: impl DispersionParameter,
+    generator: &mut EncryptionRandomGenerator<Gen>,
+) where
+    Scalar: UnsignedTorus,
+    KeyCont: Container<Element = Scalar>,
+    OutputCont: ContainerMut<Element = Scalar>,
+    Gen: ByteRandomGenerator,
+{
+    assert_eq!(
+        output_mask.ciphertext_modulus(),
+        output_body.ciphertext_modulus(),
+        "Mismatched moduli between mask ({:?}) and body ({:?})",
+        output_mask.ciphertext_modulus(),
+        output_body.ciphertext_modulus()
+    );
+
+    let ciphertext_modulus = output_mask.ciphertext_modulus();
+
+    assert!(!ciphertext_modulus.is_compatible_with_native_modulus());
+
+    // generate a randomly uniform mask
+    generator.fill_slice_with_random_mask_custom_mod(output_mask.as_mut(), ciphertext_modulus);
+
+    // generate an error from the normal distribution described by std_dev
+    let noise = generator.random_noise_custom_mod(noise_parameters, ciphertext_modulus);
+
+    let ciphertext_modulus_as_scalar: Scalar = ciphertext_modulus.get_custom_modulus().cast_into();
+
+    // compute the multisum between the secret key and the mask
+    let mask_key_dot_product = slice_wrapping_dot_product_custom_mod(
+        output_mask.as_ref(),
+        lwe_secret_key.as_ref(),
+        ciphertext_modulus_as_scalar,
+    );
+
+    // Store sum(ai * si) + delta * m + e in the body
+    *output_body.data = mask_key_dot_product
+        .wrapping_add_custom_mod(encoded.0, ciphertext_modulus_as_scalar)
+        .wrapping_add_custom_mod(noise, ciphertext_modulus_as_scalar);
 }
 
 /// Encrypt an input plaintext in an output [`LWE ciphertext`](`LweCiphertext`).
@@ -294,15 +399,15 @@ pub fn trivially_encrypt_lwe_ciphertext<Scalar, OutputCont>(
     output.get_mut_mask().as_mut().fill(Scalar::ZERO);
 
     let output_body = output.get_mut_body();
-
-    *output_body.data = encoded.0;
-
     let ciphertext_modulus = output_body.ciphertext_modulus();
-    assert!(ciphertext_modulus.is_compatible_with_native_modulus());
-    if !ciphertext_modulus.is_native_modulus() {
-        *output_body.data = (*output_body.data)
-            .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus());
-    }
+
+    *output_body.data = match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native | CiphertextModulusKind::Other => encoded.0,
+        CiphertextModulusKind::NonNativePowerOfTwo => {
+            // Manage non native power of 2 encoding
+            encoded.0 * ciphertext_modulus.get_power_of_two_scaling_to_native_torus()
+        }
+    };
 }
 
 /// A trivial encryption uses a zero mask and no noise.
@@ -373,15 +478,15 @@ where
     *new_ct.get_mut_body().data = encoded.0;
 
     let output_body = new_ct.get_mut_body();
-
-    *output_body.data = encoded.0;
-
     let ciphertext_modulus = output_body.ciphertext_modulus();
-    assert!(ciphertext_modulus.is_compatible_with_native_modulus());
-    if !ciphertext_modulus.is_native_modulus() {
-        *output_body.data = (*output_body.data)
-            .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus());
-    }
+
+    *output_body.data = match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native | CiphertextModulusKind::Other => encoded.0,
+        CiphertextModulusKind::NonNativePowerOfTwo => {
+            // Manage non native power of 2 encoding
+            encoded.0 * ciphertext_modulus.get_power_of_two_scaling_to_native_torus()
+        }
+    };
 
     new_ct
 }
@@ -395,6 +500,24 @@ where
 /// See the [`LWE ciphertext formal definition`](`LweCiphertext#lwe-decryption`) for the definition
 /// of the encryption algorithm.
 pub fn decrypt_lwe_ciphertext<Scalar, KeyCont, InputCont>(
+    lwe_secret_key: &LweSecretKey<KeyCont>,
+    lwe_ciphertext: &LweCiphertext<InputCont>,
+) -> Plaintext<Scalar>
+where
+    Scalar: UnsignedInteger,
+    KeyCont: Container<Element = Scalar>,
+    InputCont: Container<Element = Scalar>,
+{
+    let ciphertext_modulus = lwe_ciphertext.ciphertext_modulus();
+
+    if ciphertext_modulus.is_compatible_with_native_modulus() {
+        decrypt_lwe_ciphertext_native_mod_compatible(lwe_secret_key, lwe_ciphertext)
+    } else {
+        decrypt_lwe_ciphertext_other_mod(lwe_secret_key, lwe_ciphertext)
+    }
+}
+
+pub fn decrypt_lwe_ciphertext_native_mod_compatible<Scalar, KeyCont, InputCont>(
     lwe_secret_key: &LweSecretKey<KeyCont>,
     lwe_ciphertext: &LweCiphertext<InputCont>,
 ) -> Plaintext<Scalar>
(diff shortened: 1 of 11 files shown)


tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -434,6 +547,39 @@ where
...
+
+    let (mask, body) = lwe_ciphertext.get_mask_and_body();
+
+    Plaintext((*body.data).wrapping_sub_custom_mod(
+        slice_wrapping_dot_product_custom_mod(
+            mask.as_ref(),
+            lwe_secret_key.as_ref(),
+            ciphertext_modulus.get_custom_modulus().cast_into(),
+        ),
+        ciphertext_modulus.get_custom_modulus().cast_into(),
```
mayeul-zama, line 579: Store in variable

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -298,10 +393,10 @@ pub fn trivially_encrypt_lwe_ciphertext<Scalar, OutputCont>(
...
 
     let ciphertext_modulus = output_body.ciphertext_modulus();
-    assert!(ciphertext_modulus.is_compatible_with_native_modulus());
-    if !ciphertext_modulus.is_native_modulus() {
+    // Manage non native power of 2 encoding
+    if ciphertext_modulus.kind() == CiphertextModulusKind::NonNativePowerOfTwo {
         *output_body.data = (*output_body.data)
-            .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus());
+            .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus())
     }
```
mayeul-zama, line 400: Would be cleaner to have 
```
 *output_body.data = match kind{
    nnp2 => encoded * scaling,
    _ => encoded
}
```

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -377,8 +472,8 @@ where
     *output_body.data = encoded.0;
 
     let ciphertext_modulus = output_body.ciphertext_modulus();
-    assert!(ciphertext_modulus.is_compatible_with_native_modulus());
-    if !ciphertext_modulus.is_native_modulus() {
+    // Manage the non native power of 2 encoding
+    if ciphertext_modulus.kind() == CiphertextModulusKind::NonNativePowerOfTwo {
         *output_body.data = (*output_body.data)
             .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus());
```
mayeul-zama, line 478: Would be cleaner to have 
```
 *output_body.data = match kind{
    nnp2 => encoded * scaling,
    _ => encoded
}
```

Maybe better to call `trivially_encrypt_lwe_ciphertext` instead, even if it fills `0`s twice

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -434,6 +547,39 @@ where
...
+        lwe_secret_key.lwe_dimension()
+    );
+
+    let ciphertext_modulus = lwe_ciphertext.ciphertext_modulus();
+
+    assert!(!ciphertext_modulus.is_compatible_with_native_modulus());
+
+    let (mask, body) = lwe_ciphertext.get_mask_and_body();
+
+    Plaintext((*body.data).wrapping_sub_custom_mod(
```
mayeul-zama, line 573: Shouldn't we do a division by the scale in the power of 2 case?
Isn't this tested?

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
```diff
@@ -235,20 +279,51 @@ pub fn lwe_ciphertext_plaintext_add_assign<Scalar, InCont>(
 ) where
     Scalar: UnsignedInteger,
     InCont: ContainerMut<Element = Scalar>,
+{
+    let ciphertext_modulus = lhs.ciphertext_modulus();
+    if ciphertext_modulus.is_compatible_with_native_modulus() {
+        lwe_ciphertext_plaintext_add_assign_native_mod_compatible(lhs, rhs)
+    } else {
+        lwe_ciphertext_plaintext_add_assign_non_native_mod(lhs, rhs)
+    }
```
mayeul-zama, line 288: As the three cases do different things, I would have 3 functions

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -933,7 +1063,7 @@ pub fn encrypt_lwe_ciphertext_with_seeded_public_key<Scalar, KeyCont, OutputCont
         let (mut mask, body) = tmp_zero_encryption.get_mut_mask_and_body();
         random_generator
             .fill_slice_with_random_uniform_custom_mod(mask.as_mut(), ciphertext_modulus);
-        if !ciphertext_modulus.is_native_modulus() {
+        if ciphertext_modulus.is_non_native_power_of_two() {
             slice_wrapping_scalar_mul_assign(
                 mask.as_mut(),
                 ciphertext_modulus.get_power_of_two_scaling_to_native_torus(),
```
mayeul-zama, line 1069: I think it would be cleaner if  `fill_slice_with_random_uniform_custom_mod` returned a scaled mask

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
```diff
@@ -71,6 +72,22 @@ pub fn lwe_ciphertext_add_assign<Scalar, LhsCont, RhsCont>(
     Scalar: UnsignedInteger,
     LhsCont: ContainerMut<Element = Scalar>,
     RhsCont: Container<Element = Scalar>,
+{
+    let ciphertext_modulus = rhs.ciphertext_modulus();
+    if ciphertext_modulus.is_compatible_with_native_modulus() {
+        lwe_ciphertext_add_assign_native_mod_compatible(lhs, rhs)
+    } else {
+        lwe_ciphertext_add_assign_non_native_mod(lhs, rhs)
```
mayeul-zama, line 80: As both non native cases do different things, I would put them in different functions

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -31,6 +31,13 @@ pub struct CiphertextModulus<Scalar: UnsignedInteger> {
     _scalar: PhantomData<Scalar>,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum CiphertextModulusKind {
+    Native,
+    NonNativePowerOfTwo,
+    NonNative,
+}
```
mayeul-zama, line 39: `CiphertextModulusInner` could have these 3 variants

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
mayeul-zama: If it's already the case, the multiplication should be removed here

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -225,6 +232,13 @@ impl<Scalar: UnsignedInteger> CiphertextModulus<Scalar> {
         self.is_native_modulus() || self.is_power_of_two()
     }
 
+    pub const fn is_non_native_power_of_two(&self) -> bool {
+        match self.inner {
+            CiphertextModulusInner::Native => false,
+            CiphertextModulusInner::Custom(modulus) => modulus.is_power_of_two(),
```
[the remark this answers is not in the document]
IceTDrinker: maybe, but don't you find this formulation more easily readable ?

plus it looks like what you propose is wrong for the power of 2 case ?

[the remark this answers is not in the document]
mayeul-zama: Indeed, I misread the name of the function

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -57,6 +107,51 @@ pub fn fill_lwe_mask_and_body_for_encryption<Scalar, KeyCont, OutputCont, Gen>(
...
+    );
+
+    // compute the multisum between the secret key and the mask
+    *output_body.data = (*output_body.data).wrapping_add_custom_mod(
+        slice_wrapping_dot_product_custom_mod(
+            output_mask.as_ref(),
+            lwe_secret_key.as_ref(),
+            ciphertext_modulus.get_custom_modulus().cast_into(),
+        ),
+        ciphertext_modulus.get_custom_modulus().cast_into(),
```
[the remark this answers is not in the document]
IceTDrinker: the compiler should optimize that away but yes I think I did that a bit quickly, will make it better

and here we can't have it as a parameter, it's coming from the masks and body of an LWE ciphertext

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: works for me

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: this is tested in the test/lwe_encryption file normally

the way the API works is that from the user perspective everything behaves 'as if' the modulus is the custom modulus, we do tricks in the background to map power of 2 cases to the native case

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
IceTDrinker: well we have 3 functions, this one dispatching, the compatible mod managing the native and power of 2 case, and the non native case

I don't know if having a function for each is better

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: no the random should behave properly with respect to the modulus as we expose those functions IIRC, we have a special handling of the power of 2 at encryption and decryption time and that's on us to manage those cases, the rest should behave properly without our special encoding

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
IceTDrinker: no, as then the ciphertext modulus is not a u128 anymore

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: this part of the code is not for power of 2 so no need to divide, though I think the naming NonNative/NonNativePowerOfTwo is maybe misleading, open to better naming

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: used a slightly more explicit version

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
IceTDrinker: opinion on that @tmontaigu ? as well as the naming for the CiphertextModulusKind

Native => 2^n_bits for a n_bits integer (I think this one is fine)
NonNativePowerOfTwo => any 2^b for b != n_bits
NonNative => everything else

the last two are not too great

tmontaigu: `Native`,
`PowerOfTwo`,
`Custom`
?

tmontaigu: One advantage of having 3 separate functions is that it might be better for tight loop performances
```rust
match modulus_type {
Native => for bidule in machin { do_native(bidule) }
PowerOfTwo=> for bidule in machin { do_power_of_two(bidule) }
Custom=> for bidule in machin { do_custom(bidule) }
}
```

But since I dont `core` well, I don't know it that would actually be useful / measurable

IceTDrinker: That's a fair point, though I think most loops would be in the underlying function anyway. It's just that I think we were fearing the growth in terms of number of lines of code. I'll take a second look, but generally speaking from a code point of view the power of 2 case is mapped to the native case via encryption, the few places it shows up are for plaintext encoding and removing noisy bits after the fft domain shenanigans. If the number of lines of code is not a cause for concern then I'll add specific variants in places where the code for native and power of 2 differ, if they are the same then I'm not a fan of duplicating the code as asserts would need to be adapted but the code would be essentially the same

tmontaigu: I don't have strong opinions on whether we should have 3 variants or not, unless we can measure a perf difference for some of them

IceTDrinker:
last push is just the rebase

tmontaigu (approved):

IceTDrinker:
I haven't yet had the time to check the 3 vs 2 function thing 😅 last push was just a rebase given the latest big pushes

IceTDrinker:
@slab-ci cpu_wasm_test

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
IceTDrinker: I think the NonNative will become Custom or Other, NonNativePowerOfTwo will stay as it describes well what it is

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -294,15 +389,15 @@ pub fn trivially_encrypt_lwe_ciphertext<Scalar, OutputCont>(
...
-            .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus());
-    }
+
+    *output_body.data = match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native | CiphertextModulusKind::NonNative => encoded.0,
+        CiphertextModulusKind::NonNativePowerOfTwo => {
+            // Manage non native power of 2 encoding
+            encoded.0 * ciphertext_modulus.get_power_of_two_scaling_to_native_torus()
+        }
+    };
```
tmontaigu, line 400: Does not look equivalent to what was done before, should it not be ?
```rust 
       CiphertextModulusKind::Native  => encoded.0,
        CiphertextModulusKind::NonNativePowerOfTwo | CiphertextModulusKind::NonNative => {
            // Manage non native power of 2 encoding
            encoded.0 * ciphertext_modulus.get_power_of_two_scaling_to_native_torus()
        }
```

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -373,15 +468,15 @@ where
...
```
tmontaigu, line 479: Same as above

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: That is because before this pr we only had

native
Non native power of 2

So !native modulus == power 2 before, now we have to handle an additional case and only the power of 2 case has a special handling

tmontaigu (approved):

IceTDrinker:
changes done after reviewing this -> the NonNative modulus kind is now called Other

there are two version of the crypto primitives when it's better to split them to avoid having very long functions native_mod_compatible and other_mod

when a behavior depends on the kind of the modulus all three cases are managed, with an unreachable for the cases that don't make sense, I checked on godbolt, the asserts means the unreachable parts are eliminated as dead code with optimized builds

IceTDrinker:
you can check https://github.com/zama-ai/tfhe-rs/compare/c0b8812d373e7acb66b51b151b3e162954281ecd..af197a81451e7d9cdf5996279a9dc9e70731ed23 for the last push diff it's small and readable

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -417,21 +535,58 @@ where
 
     let (mask, body) = lwe_ciphertext.get_mask_and_body();
 
-    if ciphertext_modulus.is_native_modulus() {
-        Plaintext((*body.data).wrapping_sub(slice_wrapping_dot_product(
+    match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native => Plaintext((*body.data).wrapping_sub(
```
mayeul-zama, line 539: Could factor ` Plaintext((*body.data).wrapping_sub(` out of the `match`

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
```diff
@@ -235,20 +279,51 @@ pub fn lwe_ciphertext_plaintext_add_assign<Scalar, InCont>(
...
     assert!(ciphertext_modulus.is_compatible_with_native_modulus());
-    if ciphertext_modulus.is_native_modulus() {
-        *body.data = (*body.data).wrapping_add(rhs.0);
-    } else {
-        *body.data = (*body.data).wrapping_add(
-            rhs.0
-                .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus()),
-        );
+    match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native => *body.data = (*body.data).wrapping_add(rhs.0),
```
mayeul-zama, line 302: Could factor `*body.data = (*body.data).wrapping_add(` out of the `match`

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
```diff
@@ -311,20 +386,51 @@ pub fn lwe_ciphertext_plaintext_sub_assign<Scalar, InCont>(
...
-    if ciphertext_modulus.is_native_modulus() {
-        *body.data = (*body.data).wrapping_sub(rhs.0);
-    } else {
-        *body.data = (*body.data).wrapping_sub(
-            rhs.0
-                .wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus()),
-        );
+    match ciphertext_modulus.kind() {
+        CiphertextModulusKind::Native => *body.data = (*body.data).wrapping_sub(rhs.0),
+        CiphertextModulusKind::NonNativePowerOfTwo => {
```
mayeul-zama, line 410: Could factor `*body.data = (*body.data).wrapping_sub(` out of the `match`

tfhe/src/core_crypto/algorithms/seeded_lwe_ciphertext_decompression.rs
```diff
@@ -26,16 +27,20 @@ pub fn decompress_seeded_lwe_ciphertext_with_existing_generator<Scalar, OutputCo
...
+        // Manage the specific encoding for non native power of 2
+        CiphertextModulusKind::NonNativePowerOfTwo => {
+            slice_wrapping_scalar_mul_assign(
+                output_mask.as_mut(),
+                ciphertext_modulus.get_power_of_two_scaling_to_native_torus(),
+            );
+        }
+        // Nothing to do
+        CiphertextModulusKind::Native | CiphertextModulusKind::Other => (),
     }
```
mayeul-zama, line 44: I think it would be better to have `output_mask` written to only once
Could be done cleanly (IMO) by adding a `scale` argument to `fill_slice_with_random_mask_custom_mod`

mayeul-zama: Applies to `decompress_seeded_lwe_ciphertext_list_with_existing_generator`  and `par_decompress_seeded_lwe_ciphertext_list_with_existing_generator` too

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -57,6 +112,51 @@ pub fn fill_lwe_mask_and_body_for_encryption<Scalar, KeyCont, OutputCont, Gen>(
...
+
+    // compute the multisum between the secret key and the mask
+    *output_body.data = (*output_body.data).wrapping_add_custom_mod(
+        slice_wrapping_dot_product_custom_mod(
+            output_mask.as_ref(),
+            lwe_secret_key.as_ref(),
+            ciphertext_modulus_as_scalar,
+        ),
+        ciphertext_modulus_as_scalar,
+    );
```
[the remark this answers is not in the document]
IceTDrinker: what do you mean ?

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: I don't think it helps readability having a Plaintext around the match statement

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
mayeul-zama:
```
let random_noise = ...
let dot_product = ...
*output_body.data = dot_product.wrapping_add(random_noise).wrapping_add(encoded)
```

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
mayeul-zama: Me neither, but it could be stored in a variable before being wrapped in a `Plaintext`

IceTDrinker: then rust will complain of an unnecessary binding or something, so disagree on that one

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
IceTDrinker: same as above

tfhe/src/core_crypto/algorithms/seeded_lwe_ciphertext_decompression.rs
IceTDrinker: I don't think it's possible with the current CSPRNG design and randomness design, plus this way you have a tight loop around the CSPRNG and a scalar mul which can easily be vectorized by the compiler, so don't think I'll change it (the CSPRNG fills a full slice and it's a design I don't want to update for now this is some of the oldest legacy code and I don't thing it brings value to refactor at this point in time)

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
mayeul-zama: IMO
```
let scaled_rhs =  if ciphertext_modulus.is_native_modulus() {
    rhs.0
} else {
    rhs.0.wrapping_mul(ciphertext_modulus.get_power_of_two_scaling_to_native_torus())
};

*body.data = (*body.data).wrapping_add(scaled_rhs);
```
is cleaner

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
mayeul-zama: Something like
```
let dot_product = match {};
Plaintext((*body.data).wrapping_sub(dot_product));
```
I don´t see why Rust would complain

tfhe/src/core_crypto/algorithms/seeded_lwe_ciphertext_decompression.rs
mayeul-zama: Ok

tfhe/src/core_crypto/algorithms/lwe_linear_algebra.rs
mayeul-zama: Sure, a match seems better
But I still think it's better to have the assignment out of the conditional

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -36,6 +37,55 @@ pub fn fill_lwe_mask_and_body_for_encryption<Scalar, KeyCont, OutputCont, Gen>(
 
     let ciphertext_modulus = output_mask.ciphertext_modulus();
 
+    if ciphertext_modulus.is_compatible_with_native_modulus() {
```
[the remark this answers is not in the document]
IceTDrinker: Also I think at the time when we were discussing we wanted to potentially have special arithmetic for some specific moduli (like the goldilocks/solinas prime can have some super optimized modular operations) so we would use the other_mod function as fallback, so keeping it like this

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
IceTDrinker: re reading the code I think my idea was : have CSPRNG actions close together (having the CSPRNG code already "hot" from the mask generation)
never store the noise in a temporary to avoid leaking the value if possible, though this may not matter much

I'll see if I can achieve this and have something less complicated to read

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
mayeul-zama: Ok

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
IceTDrinker: not sure having the dot product in a temporary is a good idea either...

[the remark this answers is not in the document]
IceTDrinker: does this look ok ? keeping noise generation close to mask generation
```rust
    // Store delta * m + error, keeping noise generation close to mask generation as that code path
    // will already be hot
    *output_body.data = encoded
        .0
        // add the noise: an error generated from the normal distribution described by std_dev
        .wrapping_add(generator.random_noise_custom_mod(noise_parameters, ciphertext_modulus));

    // compute the multisum between the secret key and the mask
    let mask_key_dot_product = (*output_body.data).wrapping_add(slice_wrapping_dot_product(
        output_mask.as_ref(),
        lwe_secret_key.as_ref(),
    ));

    // Store sum(ai * si) + delta * m + e in the body
    *output_body.data = (*output_body.data).wrapping_add(mask_key_dot_product);
    // (power of 2 encoding omitted but managed as the last step)
```

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
IceTDrinker: I think this is ok and factorizes what can be

```rust
    let mask_key_dot_product = slice_wrapping_dot_product(mask.as_ref(), lwe_secret_key.as_ref());
    let plaintext = (*body.data).wrapping_sub(mask_key_dot_product);

    match ciphertext_modulus.kind() {
        CiphertextModulusKind::Native => Plaintext(plaintext),
        CiphertextModulusKind::NonNativePowerOfTwo => {
            // Manage power of 2 encoding
            Plaintext(
                plaintext
                    .wrapping_div(ciphertext_modulus.get_power_of_two_scaling_to_native_torus()),
            )
        }
        CiphertextModulusKind::Other => unreachable!(),
    }
```

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
mayeul-zama: I don´t see how not explicitly storing temporaries in local variables would prevent anything
In LLVM IR, these temporaries will be represented explicitly

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
mayeul-zama: LGTM

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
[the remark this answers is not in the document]
IceTDrinker: good point

mayeul-zama (approved):
Thanks !

IceTDrinker:
@slab-ci cpu_integer_test

IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_integer_test

PR #691, opened 2023-11-16
chore(ci): refactor integer script and skip div and rem preferring div_rem

IceTDrinker:
Rejoice, as div_rem takes over div and rem!


IceTDrinker:
I recommend a split diff

tmontaigu (approved):

PR #695, opened 2023-11-17
feat(integer): avoid un-necessary computations in mul params 1_X

tmontaigu:
When the parameters have 1 bit of message (message modulus == 2) then the multiplication of 2 blocks does not create a result that can go into the carry space.

We use that fact to avoid doig un-necessary computations when multiplying integers encrypted under parameters with 1 bit.


IceTDrinker (approved):
Looks good, thanks !

PR #696, opened 2023-11-17
chore(shortint): fix typo for NoiseLevel variant UNKNOWN

IceTDrinker:


tfhe/src/shortint/ciphertext/mod.rs
```diff
@@ -24,7 +24,7 @@ impl NoiseLevel {
     // for formats which did not have noise levels saved
     pub const MAX: Self = Self(usize::MAX);
     // As a safety measure the unknwon noise level is set to the max value
-    pub const UNKOWN: Self = Self::MAX;
+    pub const UNKNOWN: Self = Self::MAX;
```
tmontaigu, line 27: there is also a type in the comment above

IceTDrinker: arf thanks

IceTDrinker: fixed thanks

tmontaigu (approved):

PR #697, opened 2023-11-20
chore(ci): change benchmark aws ec2 machine type

soonum:
This instance type hpc7a.96xlarge yields better performances for nearly the same hourly cost.


soonum:
@slab-ci boolean_bench

PR #699, opened 2023-11-20
chore(tfhe): force compilation fail on non-optimized build

tmontaigu:
We've had a lot of complaints about performances that turned out to be users not turning on optimization using `release` or other means.

As FHE is already slow on its own in optimized build, building non-optimized version makes little to no sense.

This commit adds a way to check via the build.rs the optimization level and if it is to low, force fail the compilation, with a long message explaining why the compilation failed, and how to fix it.

This might be a bit intrusive, but but ints sort of intended as non-optimized make non sense.

As everything happens in the build.rs, it wont affect user how build tfhe-rs using another build system than cargo.

Rendered error

### DIFF
--- a/tfhe/build.rs
+++ b/tfhe/build.rs
@@ -73,7 +73,71 @@ fn gen_c_api() {
         .write_to_file(output_file);
 }
 
+fn panic_if_non_optimized_build() {
+    // This has a lot of \n and \ to control formatting precisely
+    const ERROR_MESSAGE: &str = "\
+    It seems that tfhe-rs is being compiled with insufficient optimization level.\n\
+    FHE is already slow on its own, and so compiler optimizations have to be turned on.\n\
+    \n\
+    By default rust/cargo does not enable optimizations, make sure you build/run with the \n\
+    release flag:\n    \
+    - `cargo run --release`\n    \
+    - `cargo build --release`\n\
+    (Note that doing `cargo build --release && cargo run`, will build in release mode\n\
+     but then rebuild in debug mode and run in debug mode)\n
+    \n\
+    * If you wish to be able to use a debugger, consider adding the following lines \n\
+    to the projects's Cargo.toml and compile in release\n\
+    \t```\n\
+    \t[profile.release]\n\
+    \tdebug = true\n\
+    \t```\n\
+    \t\n\
+    * It is also possible to override the opt-level only for tfhe-rs in the Cargo.toml\n\
+    \t```\n\
+    \t[profile.dev.package.tfhe]\n\
+    \topt-level=3\n\
+    \t```\n\
+    \tMore generally:\n\
+    \t```\n\
+    \t[profile.profile_name.package.tfhe]\n\
+    \topt-level=3\n\
+    \t```\n\
+    \t\n\
+    * If you really need a non optimized / less optimized build set the environment variable:\n\
+    \t```\n\
+    \texport TFHE_RS_ALLOW_NON_OPTIMIZED_BUILD=true\n\
+    \t```\n\
+    Or in .cargo/config.toml add\n\
+    \t```\n\
+    \t[env]\n\
+    \tTFHE_RS_ALLOW_NON_OPTIMIZED_BUILD=\"true\"\n\
+    ```\n\
+";
+    if option_env!("CARGO_PRIMARY_PACKAGE").is_none() {
+        // If CARGO_PRIMARY_PACKAGE is set, then we are build tfhe-rs is not being built
+        // as a dependency.
+        match std::env::var("TFHE_RS_ALLOW_NON_OPTIMIZED_BUILD") {
+            Ok(value) if value == "true" => {}
+            // Either the variable is defined but with not the correct value
+            // Or undefined, or other some kind of errors we consider those cases equal anyway
+            _ => {
+                let opt_level_string = std::env::var("OPT_LEVEL").expect(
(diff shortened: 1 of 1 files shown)


tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
...
+    \t```\n\
+    Or in .cargo/config.toml add\n\
+    \t```\n\
+    \t[env]\n\
+    \tTFHE_RS_ALLOW_NON_OPTIMIZED_BUILD=\"true\"\n\
+    ```\n\
+";
+    if option_env!("CARGO_PRIMARY_PACKAGE").is_none() {
+        // If CARGO_PRIMARY_PACKAGE is set, then we are build tfhe-rs is not being built
+        // as a dependency.
```
tmontaigu, line 119:
```
       // If CARGO_PRIMARY_PACKAGE is set, then tfhe-rs is not being built
        // as a dependency.
```

IceTDrinker: maybe I should use that for the C API thing from last time

IceTDrinker: we need to set the escape hatch in our Makefile though right ?

tmontaigu: CARGO_PRIMARY_PACKAGE  is used so that we don't have to use optimized build / set the env var to allow them as we know what we are doing 😬, so we are exempt of the opt-level check

IceTDrinker: ha nice

IceTDrinker: could you maybe update the code for the force skip cbindgen to use the primary package thing ? in a separate commit but as it's the same kind of idea it could be good to do at the same time ?

IceTDrinker: though I don't know if that would work for Google who use rustc directly 🤔

IceTDrinker: Also I hope Cargo will not get confused with the two tfhe crates... maybe better in a separate PR then

tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
...
+        match std::env::var("TFHE_RS_ALLOW_NON_OPTIMIZED_BUILD") {
+            Ok(value) if value == "true" => {}
```
IceTDrinker, line 121: maybe use to_lower to avoid some minor annoyance ?

tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
...
+    const ERROR_MESSAGE: &str = "\
+    It seems that tfhe-rs is being compiled with insufficient optimization level.\n\
+    FHE is already slow on its own, and so compiler optimizations have to be turned on.\n\
+    \n\
+    By default rust/cargo does not enable optimizations, make sure you build/run with the \n\
+    release flag:\n    \
+    - `cargo run --release`\n    \
+    - `cargo build --release`\n\
+    (Note that doing `cargo build --release && cargo run`, will build in release mode\n\
+     but then rebuild in debug mode and run in debug mode)\n
```
IceTDrinker, line 87: indicate they should then use cargo run --release

tfhe/build.rs
IceTDrinker: I'll experiment with the primary package thing to remove the cbindgen related feature

tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
...
+    \n\
+    * If you wish to be able to use a debugger, consider adding the following lines \n\
+    to the projects's Cargo.toml and compile in release\n\
+    \t```\n\
+    \t[profile.release]\n\
+    \tdebug = true\n\
```
mayeul-zama, line 93: You could also propose to set opt-level to 3 in the debug profile

tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
...
+    \t```\n\
+    \t\n\
+    * It is also possible to override the opt-level only for tfhe-rs in the Cargo.toml\n\
```
mayeul-zama, line 96: This could be the first suggestion as it is the less intrusive for a user (the long compilation time is limited to tfhe dependency)

In the same time, for some users unfamiliar to Rust, suggesting release profile is useful.

Maybe we could put this suggestion ("opt-level only for tfhe-rs") first and explain in the release profile suggestion that it's better to have all the code optimized to do performance tests

tfhe/build.rs
```diff
@@ -73,7 +73,71 @@ fn gen_c_api() {
         .write_to_file(output_file);
 }
 
+fn panic_if_non_optimized_build() {
+    // This has a lot of \n and \ to control formatting precisely
+    const ERROR_MESSAGE: &str = "\
+    It seems that tfhe-rs is being compiled with insufficient optimization level.\n\
```
mayeul-zama, line 79: We could clarify that the only accepted opt-level is 3

mayeul-zama (commented):
I guess you tested all the suggestions
Do you want me to also test them to be sure ?

tfhe/build.rs
tmontaigu: its proposed a few lines below

tfhe/build.rs
tmontaigu: I also think opt-level should be first because its needed for cargo check/clippy to work without having to
set --release all the time

tfhe/build.rs
mayeul-zama: I meant the global dev profile, not the tfhe specific one

PR #706, opened 2023-11-27
test(integer): use keycache for comparisons and wopbs

soonum:


tfhe/src/integer/wopbs/test.rs
```diff
@@ -58,7 +58,7 @@ pub fn wopbs_native_crt() {
 
     let params = crate::shortint::parameters::parameters_wopbs::PARAM_4_BITS_5_BLOCKS;
```
IceTDrinker, line 59: probably need to add this to the keygen, but I think we want to get rid of that parameter set

soonum: Ok let's add it anyway, since removing it will trigger a compilation error in the future.

soonum: After second look, I revert the changes on line 61. Otherwise I have to convert this very parameters set to a `PBSParameters` one, this is a bit to much especially if we intend to remove this parameters set.

IceTDrinker: hmm ok

IceTDrinker:
cargo builds are red and fast tests are red

IceTDrinker (approved):
Looks good thanks !

PR #712, opened 2023-11-29
chore(shortint): remove doctest as tests as it is confirmed they fail

IceTDrinker:
- doctest were also failing as tests and so it is not linked to doctests
- still unclear what is causing the issue, the results are sometimes way off
- the concentration of failed tests can indicate a miscompile as those tests never fail on the M1 CI, some alignment is causing issues from time to time

this will lower the amount of flakies


mayeul-zama (approved):

IceTDrinker:
what now

IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_signed_integer_test
@slab-ci cpu_wasm_test

IceTDrinker:
will merge once cargo builds are green, this is a simple test code removal

PR #726, opened 2023-12-01
feat(integer): expose default and smart ciphertext sum

tmontaigu:
This expose the ciphertext sum default and smart variants

This also removes the par_seq_op functions as they are less optimal

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -755,172 +755,14 @@ impl ServerKey {
         generates_or_propagates
     }
 
-    /// op must be associative and commutative
-    pub fn smart_binary_op_seq_parallelized<'this, 'item, T>(
-        &'this self,
-        ct_seq: impl IntoIterator<Item = &'item mut T>,
-        op: impl for<'a> Fn(&'a Self, &'a mut T, &'a mut T) -> T + Sync,
-    ) -> Option<T>
+    /// See [Self::unchecked_sum_ciphertexts_vec_parallelized]
+    pub fn unchecked_sum_ciphertexts_parallelized<'a, T, C>(&self, ciphertexts: C) -> Option<T>
     where
-        T: IntegerRadixCiphertext + 'item + From<Vec<crate::shortint::Ciphertext>>,
+        C: IntoIterator<Item = &'a T>,
+        T: IntegerRadixCiphertext + 'a,
     {
-        enum CiphertextCow<'a, C: IntegerRadixCiphertext> {
-            Borrowed(&'a mut C),
-            Owned(C),
-        }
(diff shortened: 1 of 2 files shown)


tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -1094,6 +934,42 @@ impl ServerKey {
 
         Some(result)
     }
+
+    /// Computes the sum of the ciphertexts in parallel.
+    ///
+    /// - Returns None if ciphertexts is empty
+    ///
+    /// See [Self::unchecked_sum_ciphertexts_vec_parallelized] for constraints
+    pub fn sum_ciphertexts_vec_parallelized<T>(&self, mut ciphertexts: Vec<T>) -> Option<T>
```
IceTDrinker, line 943: is this not a smart operation as well ?

also do we often have the case where we move a vec in there ? Not a blocker but was wondering if the just the slice was not enough

Those are nitpicks, not blockers

IceTDrinker: if it's good for API consistency I'm good for it, I just don't know how large the vectors our users would want to sum are, so if we end up cloning large amounts of data I guess a very explicit indication should be made in the docstring along the lines of "this function allocates temporary buffers to compute the sum of the input ciphertexts, if the input vector is large, this may result in a slow down linked to buffer allocation"

worst case they still have the possibility to do it in place ? I guess that may not make much sense as it would be trashing the content of the vec

Let me know if you prefer modifiying that API for now, I see the slice variants doing a to_vec anyways

tmontaigu: Updated to use AsRef<[T]> bounds, as it felt simpler that `IntoIterator<Item = &a T>>` and should be general enough for most (if not all) actual use cases.

There is a 'sad' path in the default_sum, if one of the ciphertext in the container has non empty carries we are going to clone the thing 2 times

IceTDrinker: maybe a pub(crate) fn to avoid that additional clone ? unsure, let me know otherwise good for approval

tmontaigu: The way to avoir it would be to have unchecked_sum take `mut Vec<T>` (or have 2 variants for unchecked, one which has the same API as default and smart, the other that take `mut Vec<T>` (and could possibly be hidden non pub fn)

IceTDrinker: up to you, can be a follow up issue if you want, I just want to avoid running tests multiple times if we are just adding possibilities in the API

tmontaigu: changed bounds a bit, this time I would say its okay for now, plugging this into HLAPI and C API will be the battle-test

PR #733, opened 2023-12-06
feat(integer): add unsigned_overflowing_scalar_add

tmontaigu:


tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
```diff
@@ -1,8 +1,50 @@
...
+        self.unchecked_scalar_add_assign(lhs, scalar);
+        let overflowed = self.unsigned_overflowing_propagate_addition_carry(lhs);
+
+        let num_scalar_block =
+            BlockDecomposer::with_early_stop_at_zero(scalar, self.key.message_modulus.0.ilog2())
+                .count();
+
+        if num_scalar_block > lhs.blocks.len() {
+            // Scalar has more blocks so addition counts as overflowing
+            BooleanBlock::new_unchecked(self.key.create_trivial(1))
```
IceTDrinker, line 28: care to elaborate why this is required, given we already have an overflow block, I'm guessing the block can be wrong ?

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
```diff
@@ -112,4 +116,57 @@ impl ServerKey {
             self.full_propagate_parallelized(ct);
         }
     }
+
+    pub fn unsigned_overflowing_scalar_sub_parallelized<T>(
```
IceTDrinker, line 120: any reason you don't do the sale as for add creating the sub_assign function and calling it in the non assign version ?

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
```diff
@@ -112,4 +116,57 @@ impl ServerKey {
...
+                    (1_u64 << 63) / (self.key.message_modulus.0 * self.key.carry_modulus.0) as u64;
+
+                let encoded_scalar = Plaintext(rhs_scalar as u64 * delta);
+                lwe_ciphertext_plaintext_sub_assign(&mut result.ct, encoded_scalar);
+
+                let encoded_msg_mod = Plaintext(self.key.message_modulus.0 as u64 * delta);
+                lwe_ciphertext_plaintext_add_assign(&mut result.ct, encoded_msg_mod);
+
+                result.degree += Degree::new(self.key.message_modulus.0);
+                result
```
IceTDrinker, line 157: is that a scalar with correcting term ? I'm surprised we don't have that in shortint 🤔

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -365,19 +365,41 @@ impl ServerKey {
...
+    /// It propagates the borrows in-place, making the ciphertext clean and returns
+    /// the boolean indicating overflow
+    pub(in crate::integer) fn unsigned_overflowing_propagate_subtraction_borrow(
+        &self,
+        ct: &mut RadixCiphertext,
+    ) -> BooleanBlock {
+        if self.is_eligible_for_parallel_single_carry_propagation(ct) {
             // Here we have to use manual unchecked_sub on shortint blocks
             // rather than calling integer's unchecked_sub as we need each subtraction
             // to be independent from other blocks.
```
IceTDrinker, line 400: comment can likely be removed

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
```diff
@@ -2649,6 +2648,140 @@ where
...
+
+    executor.setup(&cks, sks.clone());
+
+    for _ in 0..NB_TESTS_SMALLER {
+        let clear_0 = rng.gen::<u64>() % modulus;
+        let clear_1 = rng.gen::<u64>() % modulus;
+
+        let ctxt_0 = cks.encrypt(clear_0);
+
+        let (ct_res, result_overflowed) = executor.execute((&ctxt_0, clear_1));
```
IceTDrinker, line 2675: should there be a degree/noise check on the result_overflowed ?

IceTDrinker (commented):
a few comments/questions nothing major

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
tmontaigu: Our scalar addition signature allows any clear integer type, so I can be adding a u16 to a radixciphertext with 8 bits
The way it works it that it truncates the u16 to 8bits, ignoring the rest, which is correct, but in the overflowing version
we have to make sure the ignored blocks are non zero, if any of the ignored block is non zero it means we tried to add a value that is bigger than the ciphertext max to the ciphertext, so we count that as overflow

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
tmontaigu: Nope it does not exists

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: We can add some

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
IceTDrinker: ah ok thanks and the LSB blocks may not be overflowing

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
IceTDrinker: do you think it could be interesting to add ?

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
tmontaigu: added some noise/degree checks

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
tmontaigu: I moved that part into its  own unchecked_scalar_sub_with_correcting_term

tmontaigu:
(local tests failed)

tmontaigu:
Fixed, lets wait for fast tests

IceTDrinker (approved):
Thanks, looks good !

PR #753, opened 2024-01-02
docs(capi): fix C API example

tmontaigu:


tmontaigu:
(This one if  for target 0.5 release)

IceTDrinker (approved):
Thanks, added a non regression test for the C API and an output so that users can see the execution went well !

Now enjoy your vacation please :) 🌴

PR #754, opened 2024-01-03
fix(hlapi): fix safe-deserialization feature gating

mayeul-zama:
Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/373#event-11372892071

### DIFF
--- a/tfhe/Cargo.toml
+++ b/tfhe/Cargo.toml
@@ -33,7 +33,6 @@ serde_json = "1.0.94"
 # changes often between minor releases, which breaks our CI
 clap = { version = "=4.4.4", features = ["derive"] }
 # Used in user documentation
-bincode = "1.3.3"
 fs2 = { version = "0.4.3" }
 itertools = "0.11.0"
 # For erf and normality test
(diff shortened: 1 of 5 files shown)


IceTDrinker (commented):
Should we add some Makefile target for when the safe deserialize is not active ? or enable the safe serialization at all time ?

mayeul-zama:
> Should we add some Makefile target for when the safe deserialize is not active ? or enable the safe serialization at all time ?

Removing safe serialization feature (enabling it all the time) would be simpler.
But `bincode` can't be an optional dependency anymore then. It's not really a problem IMO.

IceTDrinker:
> > Should we add some Makefile target for when the safe deserialize is not active ? or enable the safe serialization at all time ?
> 
> Removing safe serialization feature (enabling it all the time) would be simpler. But `bincode` can't be an optional dependency anymore then. It's not really a problem IMO.

Yep was thinking about that and bincode

tfhe/Cargo.toml
```diff
@@ -56,12 +54,12 @@ concrete-csprng = { version = "0.4.0", path = "../concrete-csprng", features = [
 lazy_static = { version = "1.4.0", optional = true }
 serde = { version = "1.0", features = ["derive"] }
 rayon = { version = "1.5.0" }
-bincode = { version = "1.3.3", optional = true }
+bincode = "1.3.3"
```
IceTDrinker, line 57: looks like there is a space after the version specifier ?

On VS Code I like the even better toml extension for formatting the toml automatically via alt + shift + F

mayeul-zama: Done

IceTDrinker (approved):
Thanks !

PR #755, opened 2024-01-03
chore(tfhe): update 2023 to 2024 in license files and other places

IceTDrinker:


mayeul-zama (commented):
There's also a 2022 in README.md to cite tfhe-rs
Not sure if we should also update it

Otherwise, LGTM

IceTDrinker:
Hmm I think citation is supposed to be stable in time, as the 0.1.0 release was in 2022 I think it's ok, will check that tomorrow

mayeul-zama (approved):

Issue #761, opened 2024-01-05
Documentation should state `signed` instead of `unsigned` for `FheIntX`


matthiasgeihs:
Was just going through the `docs.rs` of `tfhe-rs` and noticed that types `FheIntX` are described as `unsigned integer type`.
https://docs.rs/tfhe/0.4.1/tfhe/ (e.g., for `FheInt8` it states "An unsigned integer type with8bits")

This should be `signed integer type`?

Here is the relevant code location:
https://github.com/zama-ai/tfhe-rs/blob/fef6d1860518576a6d073e7c2f583aa11bb5061a/tfhe/src/high_level_api/integers/types/static_.rs#L86

IceTDrinker:
Hello @matthiasgeihs likely a typo/error with copy pasting some docstring code, we'll take a look, thanks for the report !

PR #762, opened 2024-01-05
chore(doc): fix docstring for FheIntX types

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs/issues/761

### PR content/description

~~Probably a copy paste mistake, fixed~~

Macro was shared between types and the doc! macro arguments were not managing signed vs. unsigned.

Also fixes a spacing issue


IceTDrinker:
ah no wait 🤔 it's shared between uint/int types it looks like

mayeul-zama (approved):

PR #766, opened 2024-01-12
feat(integer): add cast_to_signed/unsigned

tmontaigu:
These functions does the logic of uX/iX as iX and uX/iX as uX


IceTDrinker (commented):
Could we have some randomized tests ? Otherwise looks good, if you are pressed by time we can do a follow up issue

tmontaigu:
follow up issue https://github.com/zama-ai/tfhe-rs-internal/issues/392

IceTDrinker (approved):
Ok for the follow-up issue

PR #770, opened 2024-01-15
chore(ci): gather common benchmarks groups

soonum:
Instead of having multiple BENCH_OP_FLAVOR to run for full benchmarks, criterion groups are now gathered by operation kind (default, smart, unchecked).


IceTDrinker (approved):
Thx

PR #771, opened 2024-01-15
feat(integer): add naive signed overflowing_mul

tmontaigu:


tfhe/src/integer/server_key/radix/mod.rs
```diff
@@ -417,6 +417,9 @@ impl ServerKey {
         ct: &mut SignedRadixCiphertext,
         num_blocks: usize,
     ) {
+        if !ct.block_carries_are_empty() {
+            self.full_propagate_parallelized(ct)
+        }
```
IceTDrinker, line 422: do we have a test to capture the previous (I assume faulty) behavior ?

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
```diff
@@ -1468,6 +1474,17 @@ create_parametrized_test!(integer_signed_default_neg);
...
 create_parametrized_test!(integer_signed_default_overflowing_sub);
 create_parametrized_test!(integer_signed_default_mul);
+create_parametrized_test!(integer_signed_default_overflowing_mul {
+    // Uses comparisnons internally, so no 1_1
+    PARAM_MESSAGE_2_CARRY_2_KS_PBS,
+    PARAM_MESSAGE_3_CARRY_3_KS_PBS,
+    PARAM_MESSAGE_4_CARRY_4_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_2_CARRY_2_GROUP_2_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_3_CARRY_3_GROUP_2_KS_PBS,
+    PARAM_MULTI_BIT_MESSAGE_1_CARRY_1_GROUP_3_KS_PBS,
```
IceTDrinker, line 1484: I see a 1_1 parameter set

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
```diff
@@ -1941,6 +1958,136 @@ where
...
+fn integer_signed_default_overflowing_mul(param: impl Into<PBSParameters>) {
+    let (cks, mut sks) = KEY_CACHE.get_from_params(param, IntegerKeyKind::Radix);
+    let cks = RadixClientKey::from((cks, NB_CTXT));
+
+    sks.set_deterministic_pbs_execution(true);
+
+    let mut rng = rand::thread_rng();
+
+    // message_modulus^vec_length
+    let modulus = (cks.parameters().message_modulus().0.pow(NB_CTXT as u32) / 2) as i64;
```
IceTDrinker, line 1970: why the / 2 here for the modulus ?

tfhe/src/integer/server_key/radix_parallel/mul.rs
```diff
@@ -755,4 +756,106 @@ impl ServerKey {
...
+                || self.unchecked_scalar_lt_parallelized(&full_result, min_scalar),
+                || self.unchecked_scalar_gt_parallelized(&full_result, max_scalar),
+            )
+        };
+
+        assert_eq!(is_gt_max.0.degree.get(), 1);
+        assert_eq!(is_gt_max.0.degree.get(), 1);
+
+        let mut overflowed = self.boolean_bitor(&is_lt_min, &is_gt_max);
+        // after_bitxor does not give the correct degree
```
IceTDrinker, line 855: after_bitor

tfhe/src/integer/server_key/radix_parallel/mul.rs
```diff
@@ -755,4 +756,106 @@ impl ServerKey {
...
+        let bigger_lhs = self.extend_radix_with_sign_msb(lhs, lhs.blocks.len());
+        let bigger_rhs = self.extend_radix_with_sign_msb(rhs, rhs.blocks.len());
+        let mut full_result = self.mul_parallelized(&bigger_lhs, &bigger_rhs);
+
+        let num_bits_of_message = self.message_modulus().0.ilog2();
+        let total_num_bits = num_bits_of_message * lhs.blocks.len() as u32;
+
+        let (is_lt_min, is_gt_max) = if total_num_bits > I256::BITS {
+            // The max of a two's complement number is a 0 in msb and then only 1
+            // The min is the opposite, 1 in msb, then zero
```
IceTDrinker, line 807: do we have a test for the 256 bit case ?

follow up question : is there a way to have a dynamic bigint that would be helpful here ? Not necessary for the PR and not sure it's something we really need at the moment

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
tmontaigu: because its signed numbers

tfhe/src/integer/server_key/radix/mod.rs
tmontaigu: No

tfhe/src/integer/server_key/radix_parallel/mul.rs
tmontaigu: We don't have dynamic bigint but we have all the backbone stuff already implemented for it

I tested this path localy, maybe I can add a small case in the test to do 1 or 2 mul > 256 bits

tmontaigu: Or slightly better maybe, put the code that builds the trivial min/max into separate  pub functions, that we test

tfhe/src/integer/server_key/radix_parallel/tests_signed.rs
IceTDrinker: thanks

tfhe/src/integer/server_key/radix_parallel/mul.rs
IceTDrinker: if you feel that's better and it does not take you too much time why not yes

tfhe/src/integer/server_key/radix/mod.rs
tmontaigu: Yes the signed mul does plaintext addition to raise degree, which made some tests fail because the sign extend was not propagating carries

But there is no specially crafted test case that triggers the issue fixed

IceTDrinker: yep all good then

tmontaigu:
Added test functions that create trivial min/max
Added fix for decyption of negative type with more than 32 bits

IceTDrinker (approved):

PR #772, opened 2024-01-15
refactor(c_api): use DynamicBuffer as the buffer type between C and Rust

IceTDrinker:
This is a required step for the 0.5 to be ready for forward compatibility with the 0.4 C API

### PR content/description

- allows to share data safely between C and Rust and make sure it gets destroyed with the right destructor
- apply formatting to C tests
- add a script to symlink the dependency lib to a fixed name
- the rust build system adds a fingerprint to the lib name which is not practical when we are linking our C test executable
- link the most recent artifacts corresponding to the dependency we have


tfhe/Cargo.toml
```diff
@@ -51,6 +51,9 @@ concrete-csprng = { version = "0.4.0", path = "../concrete-csprng", features = [
     "generator_fallback",
     "parallel",
 ] }
+tfhe-c-api-dynamic-buffer = { version = "0.1.0", optional = true, features = [
+    "c_api",
```
tmontaigu, line 55: Strange to have a c-api feature when its in the name of the crate that the libs is for c-api :thinking:

tmontaigu (approved):

tfhe/Cargo.toml
IceTDrinker: I know, but it's generating files in the target directory and cargo was being annoying when publishing the crate, so I went for this workaround...

IceTDrinker: otherwise it's near impossible to get the actual artifact as the out directory is somewhere in a hashed/fingerprinted directory very hard to find again

tmontaigu: hum ok

IceTDrinker: basically we write the header file in the target directory and cargo complained that the project generated the file and so considered the test build as dirty

PR #775, opened 2024-01-16
refactor(shortint): make scalar_eq/scalar_not_eq take & not &mut

tmontaigu:
BREAKING_CHANGE: change mutability of scalar_equal/scalar_not_equal arguement


IceTDrinker (approved):
thanks!

PR #788, opened 2024-01-18
docs(tfhe): add fhe_strings example

JoseSK999:
This PR adds the FHE str API example.

### DIFF
--- a/tfhe/examples/fhe_strings/README.md
+++ b/tfhe/examples/fhe_strings/README.md
@@ -1,0 +1,53 @@
+# FHE Strings
+
+This example contains the implementation of a str API in FHE, featuring 30 methods. This API allows the user to:
+* Encrypt the `str` with or without padding nulls (i.e. encrypted `0u8`s at the end of the string), which serve to obfuscate the length but are ignored by algorithms
+* Encrypt any kind of pattern (`pat`, `from`, `to`, `rhs`) with or without padding nulls
+* Encrypt the number of repetitions `n`, allowing to provide a clear `max` to restrict the range of the encrypted `n`
+* Provide a cleartext pattern when algorithms can run faster. Otherwise, it's possible to trivially encrypt the pattern with `FheString::trivial`
+
+Encrypted strings contain a flag indicating whether they have padding nulls or not. Algorithms are optimized to differentiate between the two kind of strings. For instance, in some cases we can skip entirely the FHE computations if we know the true lengths of the string or pattern.
+
+Just like the clear str API, any encrypted string returned by a function can be used as input to other functions. For instance when `trim_start` is executed, or a `Split` iterator instance is advanced with `next`, the result will only have nulls at the end. The decryption function `decrypt_ascii` will panic if it encounters with malformed encrypted strings, including padding inconsistencies.
+
+### Example
+
+```rust
+let (ck, sk) = gen_keys();
+let s = "Zama ";
+let padding = Some(2);
+
+let enc_s = FheString::new(&ck, &s, padding);
+let clear_count = UIntArg::Clear(3);
+
+// All the nulls are shifted to the right end
+let result_repeat = sk.repeat(&enc_s, &clear_count);
+let result_trim_end = sk.trim_end(&result_repeat);
+let result_uppercase = sk.to_uppercase(&result_trim_end);
+
+let clear = ck.decrypt_ascii(&result_uppercase);
+
+assert_eq!(clear, "ZAMA ZAMA ZAMA");
+```
+
+## Technical Details
+
+We have implemented conversions between encrypted strings (`FheString`) and UInts (`RadixCiphertext`). This is useful for:
+
+- Speeding up comparisons and pattern matching: We perform a _single comparison_ between two numbers. This is more efficient than many u8 comparisons.
+- Shifting by an encrypted number of characters: By treating the string as a `RadixCiphertext` we can use the tfhe-rs shifting operations, and then convert back to `FheString`.
+
+Similarly, when a pattern is provided in the clear (`ClearString`) we convert it to `StaticUnsignedBigInt<N>`. This type requires a constant `N` for the u64 length of the clear UInt, and we have set it to 4, allowing for up to 32 characters in `ClearString`.
+
+`N` can be increased in `main.rs` to enable longer clear patterns, or reduced to improve performance (if we know that we will work with smaller clear patterns).
+
+## Test Cases
+
+We have handled corner cases like empty strings and empty patterns (with and without padding), the number of repetitions `n` (clear and encrypted) being zero, etc. A complete list of tests can be found at `assert_functions/test_vectors.rs`.
+
+## Usage
+To run all the functions and see the comparison with the clear Rust API you can specify the following arguments:
+
+```--str <"your str"> --pat <"your pattern"> --to <"argument used in replace"> --rhs <"used in comparisons and concat"> --n <number of repetitions> --max <clear max n>```
+
+To optionally specify a number of padding nulls for any argument you can also use: ``--str_pad``, ``--pat_pad``, ``--to_pad`` and ``--rhs_pad``.
--- a/tfhe/examples/fhe_strings/assert_functions/test_vectors.rs
+++ b/tfhe/examples/fhe_strings/assert_functions/test_vectors.rs
@@ -1,0 +1,462 @@
+use crate::Keys;
+
+const TEST_CASES_MATCH: [(&str, u32); 15] = [
+    ("", 0),
+    ("", 1),
+    ("", 2),
+    ("", 3),
+    ("a", 0),
+    ("e", 1),
+    ("o", 2),
+    (" ", 2),
+    ("?", 3),
+    ("<3", 3),
+    ("foo", 0),
+    ("foofoo", 0),
+    ("foofoo", 1),
+    (" don't", 12),
+    ("What is <3? Baby don't hurt me", 0),
+];
+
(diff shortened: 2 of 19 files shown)


tfhe/examples/fhe_strings/README.md
```diff
@@ -0,0 +1,20 @@
+# fhe_strings
+
+This repo contains the implementation of a str API in FHE, featuring 30 methods. This API allows the user to:
+* Encrypt the `str` with or without padding nulls (i.e. encrypted `0u8`s at the end of the string), which serve to obfuscate the length but are ignored by algorithms
```
IceTDrinker, line 4: nitpick: Perhaps to update, as it's not a repo anymore, rather an example

tfhe/examples/fhe_strings/README.md
```diff
@@ -0,0 +1,20 @@
...
+* Encrypt the number of repetitions `n`, allowing to provide a clear `max` to restrict the range of the encrypted `n`
+* Provide a cleartext pattern when algorithms can run faster. Otherwise, it's possible to trivially encrypt the pattern with `FheString::trivial`
+
+Encrypted strings contain a flag indicating whether they have padding nulls or not. Algorithms are optimized to differentiate between the two kind of strings. For instance, in some cases we can skip entirely the FHE computations if we know the true lengths of the string or pattern.
+
+Just like the clear str API, any encrypted string returned by a function can be used as input to other functions. For instance when `trim_start` is executed, or a `Split` iterator instance is advanced with `next`, the result will only have nulls at the end. The decryption function `decrypt_ascii` will panic if it encounters with malformed encrypted strings, including padding inconsistencies.
+
+We have handled corner cases like empty strings and empty patterns (with and without padding), the number of repetitions `n` (clear and encrypted) being zero, etc. A complete list of tests can be found at `src/assert_functions/test_vectors.rs`.
+
```
IceTDrinker, line 14: maybe a small example demonstrating using some basic usage of the lib with maybe two chained methods could be nice and compare the result to the rust clear result

tfhe/examples/fhe_strings/assert_functions/test_vectors.rs
```diff
@@ -0,0 +1,462 @@
...
+    ("a", 1),
+    ("o", 2),
+    (" ", 2),
+    ("I F*", 3),
+    ("<3", 3),
+    ("foo", 0),
+    ("foofoo", 0),
+    ("foofoo", 1),
+    (" zelda", 12),
+    ("I F*** <3 zelda", 0),
```
IceTDrinker, line 18: hmm, maybe something a bit less edgy here please ;)

tfhe/examples/fhe_strings/assert_functions/test_vectors.rs
```diff
@@ -0,0 +1,462 @@
...
+const TEST_CASES_MATCH: [(&str, u32); 15] = [
+    ("", 0),
+    ("", 1),
+    ("", 2),
+    ("", 3),
+    ("a", 0),
+    ("a", 1),
+    ("o", 2),
+    (" ", 2),
+    ("I F*", 3),
```
IceTDrinker, line 12: same here something a bit less edgy please :)

tfhe/examples/fhe_strings/README.md
```diff
@@ -0,0 +1,20 @@
+# fhe_strings
```
IceTDrinker, line 1: title:

FHE Strings

tfhe/examples/fhe_strings/README.md
```diff
@@ -0,0 +1,20 @@
...
+
```
IceTDrinker, line 2: I think you use an interesting method by "converting" a string to a uint in some algorithms (if I understand correctly) it could be worth explaining the approach and the intuition behind that transformation in this readme, a small paragraph could be nice

IceTDrinker (commented):
Hello @JoseSK999 the PR is fairly large, so we will be reviewing in chunks, I did some of the high level nitpicks, mainly on the readme, I will try to take a look at the docstrings and some of my colleagues will check the algorithms and test cases in more depth

Generally the code is clean and is one of the reasons you are a winner, still we'll likely have some comments regarding some stuff, we have been pretty busy, so sorry for the delay !

IceTDrinker (commented):
Some additional minor stuff for docstrings, but looks good overall for documentation !

tfhe/examples/fhe_strings/server_key/comp.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///
+    /// Returns `false` if they are not equal.
+    ///
+    /// The pattern for comparison (`rhs`) can be specified as either `GenericPattern::Clear` for a
+    /// clear string or `GenericPattern::Enc` for an encrypted string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let (ck, sk) = gen_keys();
```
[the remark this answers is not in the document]
JoseSK999: I think docstrings are not compiled as tests because this is actually a binary, and doctests seem to only work in libraries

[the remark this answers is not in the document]
IceTDrinker: yep, I would still add the imports as this example will likely get in the lib down the line in form or the other :)

tfhe/examples/fhe_strings/server_key/pattern/contains.rs
```diff
@@ -0,0 +1,247 @@
...
+
+                    self.asciis_eq(a.into_iter(), b.into_iter())
+                }
+            })
+            .collect();
+
+        for match_case in matched {
+            // One of the possible values of pat must match the str
+            self.key.boolean_bitor_assign(&mut result, &match_case);
+        }
```
tmontaigu, line 40: Actually here, you should be able to transform all these matched booleanblocks into a RadixCiphertext, and use scalar_ne_parallelized(&radix, 0) on it, it should be faster

tfhe/examples/fhe_strings/server_key/pattern/contains.rs
```diff
@@ -0,0 +1,247 @@
...
+                let a: Vec<&FheAsciiChar> = str_chars.collect();
+
+                self.clear_asciis_eq(a.into_iter(), pat)
+            })
+            .collect();
+
+        for match_case in matched {
+            // One of the possible values of pat must match the str
+            self.key.boolean_bitor_assign(&mut result, &match_case);
+        }
```
tmontaigu, line 65: Same as in `compare_shifted`

tfhe/examples/fhe_strings/server_key/pattern/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+// Useful for handling cases in which we know if there is or there isn't a match just by looking at
+// the lengths
+enum IsMatch {
+    Clear(bool),
+    Cipher(BooleanBlock),
+    None,
+}
+
+impl ServerKey {
+    fn length_checks(&self, str: &FheString, pat: &FheString) -> IsMatch {
```
tmontaigu, line 21: This could use a small descriptions

tfhe/examples/fhe_strings/ciphertext.rs
```diff
@@ -0,0 +1,234 @@
...
+            .collect();
+
+        let mut uint = RadixCiphertext::from_blocks(blocks);
+
+        if uint.blocks().is_empty() {
+            sk.key()
+                .extend_radix_with_trivial_zero_blocks_lsb_assign(&mut uint, 4);
+        }
+
+        uint
```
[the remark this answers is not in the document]
JoseSK999: Yess that makes sense

tfhe/examples/fhe_strings/server_key/pattern/contains.rs
JoseSK999: Thanks! Very useful

JoseSK999:
Everything has been addressed. The docstring should now work in case they were compiled, some comments, docstrings and README updated, and the improvement in `len`, `compare_shifted`, `compare_shifted_clear` and `to_uint`.

Thanks!

IceTDrinker:
@slab-ci cpu_fast_test
@slab-ci gpu_test

IceTDrinker:
@slab-ci cpu_test
@slab-ci cpu_signed_integer_test
@slab-ci cpu_wasm_test

IceTDrinker (approved):

IceTDrinker (changes requested):
Hmm actually @JoseSK999 it looks like there are test failures with the latest version of the code

e.g.

test assert_functions::test_vectors::test_split_ascii_whitespace ... FAILED
test assert_functions::test_vectors::test_trim ... FAILED
test assert_functions::test_vectors::test_trim_start ... FAILED

could you check on your end ?

JoseSK999:
So the problem is that in the new version of `len` we assume a padded string to have at least one char (a null), which is correct.
```rust
let len = self
    .key
    .sum_ciphertexts_parallelized(non_zero_chars.iter())
    .expect("There's at least one padding character");
```

However in `trim_start` we set the trimmed result to padded to check the new length (in FHE to discard potential nulls). The diff between new and old lengths is the amount to shift the trimmed string left, to make it null-terminated.

So it panics because in `trim_start` we set an empty string as padded. Actually the `trim` functions and `split_ascii_whitespace` should return early in the empty str case. It's better for efficiency and also fixes this.

JoseSK999:
Fixed

IceTDrinker (approved):
Thanks for the fix

JoseSK999:
Clippy is happy now ;)

PR #789, opened 2024-01-18
feat(hlapi): add gpu support

tmontaigu:


tfhe/src/high_level_api/keys/server.rs
```diff
@@ -104,6 +104,67 @@ impl CompressedServerKey {
             key: Arc::new(self.integer_key.decompress()),
         }
     }
+
+    #[cfg(feature = "gpu")]
+    pub fn decompress_to_gpu(&self) -> CudaServerKey {
+        // TODO this should a its own shortint function and standard key, with its integer wrapper
```
mayeul-zama, line 110: `a` instead of `have`

mayeul-zama (approved):

PR #797, opened 2024-01-22
fix(integer): is_scalar_out_of_bounds handles bigger ct

tmontaigu:
Fix a bug where in is_scalar_out_of_bounds, if the scalar was negative and the ciphertext a signed one with more blocks than the decomposed scalar, we would do an out of bound access (i.e a panic).

This fixes that, this will fix doing signed_overflowing_mul on 256 bits where the bug first appeared

### Check-list:


IceTDrinker (approved):
Thanks !

PR #798, opened 2024-01-22
fix(integer): fix cast in scalar_shift/rotate

tmontaigu:
In scalar_shift/rotate, we get the number of bits to shift/rotate as a generic type, the can be casted to u64.

We compute the total number of bits the ciphertext has, cast that number to the same type as the scalar, and do "shift % num_bits".

However, if the number of bits computed exceeds the max value the scalar type can hold, we could end up doing a remainder with 0.

e.g 256bits ciphertext and scalar type u8, 256u64 casted to u8 results in 0.

Fix that by casting the scalar value to u64.

### DIFF
--- a/tfhe/src/high_level_api/integers/unsigned/tests.rs
+++ b/tfhe/src/high_level_api/integers/unsigned/tests.rs
@@ -809,3 +809,31 @@ fn test_if_then_else() {
         if clear_a <= clear_b { clear_b } else { clear_a }
     );
 }
+
+#[test]
+fn test_scalar_shift_when_clear_type_is_small() {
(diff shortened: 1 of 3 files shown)


IceTDrinker (commented):
Could we have a small regression test for this ? Otherwise looks good

tmontaigu:
Is it ok if its a HLAPI test ?

IceTDrinker:
> Is it ok if its a HLAPI test ?

Sure as long as we have a comment on the test indicating it's a regression test and should be kept under the same form

IceTDrinker (approved):
Thanks !

PR #807, opened 2024-01-22
fix(core): ignore value in the body when doing LWE encryption

IceTDrinker:
### PR content/description

- update tests to have non zero values in structures to avoid depending on the initialization of the structs

### Check-list:

~~* [ ] Docs have been added / updated (for bug fixes / features)~~
~~* [ ] Relevant issues are marked as resolved/closed, related issues are linked in the description~~
~~* [ ] Check for breaking changes (including serialization changes) and add them to commit message following the conventional commit [specification][conventional-breaking]~~

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
@@ -94,10 +94,8 @@ pub fn fill_lwe_mask_and_body_for_encryption_native_mod_compatible<
     // generate an error from the normal distribution described by std_dev
     let noise = generator.random_noise_custom_mod(noise_parameters, ciphertext_modulus);
     // compute the multisum between the secret key and the mask
-    let mask_key_dot_product = (*output_body.data).wrapping_add(slice_wrapping_dot_product(
-        output_mask.as_ref(),
-        lwe_secret_key.as_ref(),
-    ));
+    let mask_key_dot_product =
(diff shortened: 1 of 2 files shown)


tfhe/src/core_crypto/algorithms/test/ggsw_encryption.rs
```diff
@@ -72,7 +72,7 @@ fn test_parallel_and_seeded_ggsw_encryption_equivalence<Scalar>(
         );
 
         let mut par_ggsw = GgswCiphertext::new(
-            Scalar::ZERO,
+            Scalar::MAX,
```
mayeul-zama, line 75: Why change these values?
Is there a specific reason to take MAX in particular?

IceTDrinker: hmm actually it might not be enough to trigger the bug the hw encountered :/

IceTDrinker: I just wanted some broken value and did not want to run the rng as it would slow the coverage down potentially

IceTDrinker:
I think I'll revert the test changes and instead do the sale encryption two times in the same LWE, if the bug is still present the decryption will be wrong

IceTDrinker:
Changed as indicated, the test for lwe encryption specifically has the non regression test, doing the encryption twice

mayeul-zama (approved):

PR #814, opened 2024-01-23
chore(ci): add gpu tests from user documentation

soonum:


agnesLeroy (commented):
Looks good to me

agnesLeroy:
Hmm no, looks like the CPU tests try to execute the how_to_run_on_gpu doc test...

agnesLeroy (approved):

soonum:

tfhe/src/test_user_docs.rs
```diff
@@ -1,102 +1,116 @@
...
+fn test_gpu_doc() {
+    doctest!("../docs/how_to/run_on_gpu.md", how_to_run_on_gpu);
+}
+
+fn main() {
+    #[cfg(not(feature = "gpu"))]
+    test_cpu_doc();
+    #[cfg(feature = "gpu")]
+    test_gpu_doc()
+}
```
IceTDrinker, line 116: main ?

IceTDrinker (commented):
I don't think we should merge with a main in the test_user_docs.rs file ? 🤔

tfhe/src/test_user_docs.rs
IceTDrinker: I think we are not running any doctests here ? as main is a private function called nowhere ?

IceTDrinker (changes requested):
Requesting change as I'm really unsure on that main/doc change

tfhe/src/test_user_docs.rs
```diff
@@ -1,102 +1,109 @@
...
-// Tutorials
-doctest!(
-    "../docs/tutorials/ascii_fhe_string.md",
-    ascii_fhe_string_tutorial
-);
-doctest!("../docs/tutorials/parity_bit.md", parity_bit_tutorial);
+#[cfg(feature = "gpu")]
+fn test_gpu_doc() {
+    doctest!("../docs/how_to/run_on_gpu.md", how_to_run_on_gpu);
+}
```
IceTDrinker, line 109: I think it should be a module

like

```rust
#[cfg(featyre = "gpu")]
mod test_gpu_user_doc {
    doctest!(...);
}
```

the same for cpu but with the

#[cfg(not(feature = "gpu"))]

as the functions are not called, the doctest! macro creates the entry points

soonum: You're right

IceTDrinker (approved):

PR #822, opened 2024-01-25
chore(gpu): check for all cuda errors and abort in device.cu/.h

agnesLeroy:
solves a part of: https://github.com/zama-ai/tfhe-rs-internal/issues/276

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/device.h
+++ b/backends/tfhe-cuda-backend/cuda/include/device.h
@@ -11,6 +11,22 @@
 
 extern "C" {
 
+#define check_cuda_error(ans)                                                  \
+  { cuda_error((ans), __FILE__, __LINE__); }
+inline void cuda_error(cudaError_t code, const char *file, int line) {
+  if (code != cudaSuccess) {
+    std::fprintf(stderr, "Cuda error: %s %s %d\n", cudaGetErrorString(code), 
+              file, line);
+    std::abort();
+  }
+}
+#define PANIC(format, ...)                                                     \
+  {                                                                            \
+    std::fprintf(stderr, "%s::%d::%s: panic.\n" format "\n", __FILE__,         \
+            __LINE__, __func__, ##__VA_ARGS__);                                \
+    std::abort();                                                              \
+  }
+
 struct cuda_stream_t {
   cudaStream_t stream;
   uint32_t gpu_index;
@@ -18,68 +34,59 @@ struct cuda_stream_t {
   cuda_stream_t(uint32_t gpu_index) {
     this->gpu_index = gpu_index;
 
-    cudaStreamCreate(&stream);
+    check_cuda_error(cudaStreamCreate(&stream));
   }
 
   void release() {
-    cudaSetDevice(gpu_index);
-    cudaStreamDestroy(stream);
+    check_cuda_error(cudaSetDevice(gpu_index));
+    check_cuda_error(cudaStreamDestroy(stream));
   }
 
-  void synchronize() { cudaStreamSynchronize(stream); }
+  void synchronize() { 
+    check_cuda_error(cudaStreamSynchronize(stream)); 
+  }
 };
 
 cuda_stream_t *cuda_create_stream(uint32_t gpu_index);
 
-int cuda_destroy_stream(cuda_stream_t *stream);
+void cuda_destroy_stream(cuda_stream_t *stream);
 
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
 
 void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream);
 
-int cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
+void cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
 
-int cuda_check_support_cooperative_groups();
+bool cuda_check_support_cooperative_groups();
 
-int cuda_memcpy_to_cpu(void *dest, const void *src, uint64_t size);
+void cuda_memcpy_to_cpu(void *dest, const void *src, uint64_t size);
 
-int cuda_memcpy_async_to_gpu(void *dest, void *src, uint64_t size,
+void cuda_memcpy_async_to_gpu(void *dest, void *src, uint64_t size,
                              cuda_stream_t *stream);
 
-int cuda_memcpy_async_gpu_to_gpu(void *dest, void *src, uint64_t size,
+void cuda_memcpy_async_gpu_to_gpu(void *dest, void *src, uint64_t size,
                                  cuda_stream_t *stream);
 
-int cuda_memcpy_to_gpu(void *dest, void *src, uint64_t size);
+void cuda_memcpy_to_gpu(void *dest, void *src, uint64_t size);
 
-int cuda_memcpy_async_to_cpu(void *dest, const void *src, uint64_t size,
+void cuda_memcpy_async_to_cpu(void *dest, const void *src, uint64_t size,
                              cuda_stream_t *stream);
 
-int cuda_memset_async(void *dest, uint64_t val, uint64_t size,
+void cuda_memset_async(void *dest, uint64_t val, uint64_t size,
                       cuda_stream_t *stream);
 
 int cuda_get_number_of_gpus();
 
-int cuda_synchronize_device(uint32_t gpu_index);
+void cuda_synchronize_device(uint32_t gpu_index);
 
-int cuda_drop(void *ptr, uint32_t gpu_index);
+void cuda_drop(void *ptr, uint32_t gpu_index);
 
-int cuda_drop_async(void *ptr, cuda_stream_t *stream);
+void cuda_drop_async(void *ptr, cuda_stream_t *stream);
 
 int cuda_get_max_shared_memory(uint32_t gpu_index);
 
-int cuda_synchronize_stream(cuda_stream_t *stream);
+void cuda_synchronize_stream(cuda_stream_t *stream);
 
-#define check_cuda_error(ans)                                                  \
-  { cuda_error((ans), __FILE__, __LINE__); }
-inline void cuda_error(cudaError_t code, const char *file, int line,
-                       bool abort = true) {
-  if (code != cudaSuccess) {
-    fprintf(stderr, "Cuda error: %s %s %d\n", cudaGetErrorString(code), file,
-            line);
-    if (abort)
-      exit(code);
-  }
-}
 }
 
 template <typename Torus>
--- a/backends/tfhe-cuda-backend/cuda/src/device.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/device.cu
@@ -4,25 +4,23 @@
 
 /// Unsafe function to create a CUDA stream, must check first that GPU exists
 cuda_stream_t *cuda_create_stream(uint32_t gpu_index) {
-  cudaSetDevice(gpu_index);
+  check_cuda_error(cudaSetDevice(gpu_index));
   cuda_stream_t *stream = new cuda_stream_t(gpu_index);
   return stream;
 }
 
 /// Unsafe function to destroy CUDA stream, must check first the GPU exists
-int cuda_destroy_stream(cuda_stream_t *stream) {
+void cuda_destroy_stream(cuda_stream_t *stream) {
   stream->release();
-  return 0;
 }
 
 /// Unsafe function that will try to allocate even if gpu_index is invalid
 /// or if there's not enough memory. A safe wrapper around it must call
 /// cuda_check_valid_malloc() first
 void *cuda_malloc(uint64_t size, uint32_t gpu_index) {
-  cudaSetDevice(gpu_index);
+  check_cuda_error(cudaSetDevice(gpu_index));
   void *ptr;
-  cudaMalloc((void **)&ptr, size);
-  check_cuda_error(cudaGetLastError());
+  check_cuda_error(cudaMalloc((void **)&ptr, size));
 
   return ptr;
 }
@@ -30,7 +28,7 @@ void *cuda_malloc(uint64_t size, uint32_t gpu_index) {
 /// Allocates a size-byte array at the device memory. Tries to do it
 /// asynchronously.
 void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
-  cudaSetDevice(stream->gpu_index);
+  check_cuda_error(cudaSetDevice(stream->gpu_index));
   void *ptr;
 
 #ifndef CUDART_VERSION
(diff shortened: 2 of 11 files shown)


backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -11,75 +11,78 @@
 
 extern "C" {
 
+#define check_cuda_error(ans)                                                  \
+  { cuda_error((ans), __FILE__, __LINE__); }
+inline void cuda_error(cudaError_t code, const char *file, int line,
+                       bool abort = true) {
```
IceTDrinker, line 17: any reason to have abort as a flag here ? looks like we are not forwarding anything to it in the macro ?

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -4,33 +4,31 @@
...
+  check_cuda_error(cudaMalloc((void **)&ptr, size));
 
   return ptr;
 }
 
 /// Allocates a size-byte array at the device memory. Tries to do it
 /// asynchronously.
 void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
-  cudaSetDevice(stream->gpu_index);
+  check_cuda_error(cudaSetDevice(stream->gpu_index));
```
IceTDrinker, line 31: I see the check cuda error was already used below, IIRC for async function it has no penalty as the error is only fetched once a synchronize happen ? My CUDA is a bit rusty to be honest :)

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
+void cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index) {
+  check_cuda_error(cudaSetDevice(gpu_index));
   size_t total_mem, free_mem;
-  cudaMemGetInfo(&free_mem, &total_mem);
+  check_cuda_error(cudaMemGetInfo(&free_mem, &total_mem));
   if (size > free_mem) {
-    // error code: not enough memory
-    return -1;
+    printf("Cuda error: not enough memory on device. "
+            "Available: %d vs Requested: %d", free_mem, total_mem);
```
IceTDrinker, line 60: %d is not the right specififer for size_t it seems https://stackoverflow.com/a/2125854/18088947

what sort of warning/lint level is enabled for the compiler on the cuda code ? Not saying we need to turn it on right now, but could be interesting to add more lints/warning levels as time goes on :)

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
+    abort();
   }
-  return 0;
 }
 
 /// Returns
 ///  -> 0 if Cooperative Groups is not supported.
 ///  -> 1 otherwise
 int cuda_check_support_cooperative_groups() {
```
IceTDrinker, line 68: any reason to return an int rather than a bool here ? 🤔

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-int cuda_memcpy_async_to_gpu(void *dest, void *src, uint64_t size,
+/// Copy memory to the GPU asynchronously
+void cuda_memcpy_async_to_gpu(void *dest, void *src, uint64_t size,
                              cuda_stream_t *stream) {
   if (size == 0) {
-    // error code: zero copy size
-    return -3;
+    printf("Cuda error: zero copy size in async copy to GPU.");
+    abort();
   }
```
IceTDrinker, line 82: Is that a requirement of the cuda mem cpy routine ? 🤔 size != 0 ?

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
 
-  if (stream->gpu_index >= cuda_get_number_of_gpus()) {
-    // error code: invalid gpu_index
-    return -2;
-  }
```
IceTDrinker, line 101: why is that check gone ?

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
 
-/// Tries to copy memory to the GPU synchronously
-/// 0: success
-/// -1: error, invalid device pointer
-/// -2: error, gpu index doesn't exist
-/// -3: error, zero copy size
-int cuda_memcpy_to_gpu(void *dest, void *src, uint64_t size) {
+/// Copy memory to the GPU synchronously
+void cuda_memcpy_to_gpu(void *dest, void *src, uint64_t size) {
   if (size == 0) {
```
IceTDrinker, line 98: same question about the size == 0

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
 
-/// Tries to copy memory to the CPU synchronously
-/// 0: success
-/// -1: error, invalid device pointer
-/// -2: error, gpu index doesn't exist
-/// -3: error, zero copy size
-int cuda_memcpy_to_cpu(void *dest, void *src, uint64_t size) {
+/// Copy memory to the CPU synchronously
+void cuda_memcpy_to_cpu(void *dest, void *src, uint64_t size) {
   if (size == 0) {
```
IceTDrinker, line 113: here also

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-/// Tries to copy memory within a GPU asynchronously
-/// 0: success
-/// -1: error, invalid device pointer
-/// -2: error, gpu index doesn't exist
-/// -3: error, zero copy size
-int cuda_memcpy_async_gpu_to_gpu(void *dest, void *src, uint64_t size,
+/// Copy memory within a GPU asynchronously
+void cuda_memcpy_async_gpu_to_gpu(void *dest, void *src, uint64_t size,
                                  cuda_stream_t *stream) {
   if (size == 0) {
```
IceTDrinker, line 129: again

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-    return -3;
+    printf("Cuda error: zero copy size in copy from GPU to GPU.");
+    abort();
   }
 
-  if (stream->gpu_index >= cuda_get_number_of_gpus()) {
-    // error code: invalid gpu_index
-    return -2;
-  }
   cudaPointerAttributes attr_dest;
```
IceTDrinker, line 134: gpu_index check gone ?

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-  return 0;
+void cuda_synchronize_device(uint32_t gpu_index) {
+  check_cuda_error(cudaSetDevice(gpu_index));
+  check_cuda_error(cudaDeviceSynchronize());
 }
 
-int cuda_memset_async(void *dest, uint64_t val, uint64_t size,
+void cuda_memset_async(void *dest, uint64_t val, uint64_t size,
                       cuda_stream_t *stream) {
   if (size == 0) {
```
IceTDrinker, line 166: zero size for mem primitive

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,132 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-    // error code: zero copy size
-    return -3;
+    printf("Cuda error: zero memset size.");
+    abort();
   }
 
-  if (stream->gpu_index >= cuda_get_number_of_gpus()) {
-    // error code: invalid gpu_index
-    return -2;
-  }
```
IceTDrinker, line 223: check index gone ?

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -256,57 +203,43 @@ template void cuda_set_value_async(cudaStream_t *stream, uint64_t *d_array,
...
-/// Tries to copy memory to the GPU asynchronously
-/// 0: success
-/// -1: error, invalid device pointer
-/// -2: error, gpu index doesn't exist
-/// -3: error, zero copy size
-int cuda_memcpy_async_to_cpu(void *dest, const void *src, uint64_t size,
+/// Copy memory to the CPU asynchronously
+void cuda_memcpy_async_to_cpu(void *dest, const void *src, uint64_t size,
                              cuda_stream_t *stream) {
   if (size == 0) {
```
IceTDrinker, line 209: zero size mem primitive

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -256,57 +203,43 @@ template void cuda_set_value_async(cudaStream_t *stream, uint64_t *d_array,
...
-    // error code: zero copy size
-    return -3;
+    printf("Cuda error: zero copy size in async copy to CPU.");
+    abort();
   }
 
-  if (stream->gpu_index >= cuda_get_number_of_gpus()) {
```
IceTDrinker, line 271: index thingy

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -323,18 +256,13 @@ int cuda_drop_async(void *ptr, cuda_stream_t *stream) {
 #else
   check_cuda_error(cudaFree(ptr));
 #endif
-  return 0;
 }
 
 /// Get the maximum size for the shared memory
 int cuda_get_max_shared_memory(uint32_t gpu_index) {
-  if (gpu_index >= cuda_get_number_of_gpus()) {
```
IceTDrinker, line 331: cuda index gone ?

IceTDrinker (commented):
Some comments are just repeat as the same pattern was seen in several changes, looks good overall :)

backends/tfhe-cuda-backend/cuda/include/device.h
agnesLeroy: We do:
```
  if (code != cudaSuccess) {
    fprintf(stderr, "Cuda error: %s %s %d\n", cudaGetErrorString(code), file,
            line);
    if (abort)
      exit(code);
  }
```
We could remove the abort bool argument and just call abort() in the macro, is that what you mean?

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: Yes exactly

backends/tfhe-cuda-backend/cuda/include/device.h
IceTDrinker: yep, I'm feeling like the abort flag might not be useful here ?

IceTDrinker: it might be use I don't know :)

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: No, I'll change it

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: cuda_check_error will detect this situation

IceTDrinker: ok, feel free to close all duplicates for this comment then :)

IceTDrinker:
looks like we have a fail in the integer GPU tests :| sigabort so might be one of the abort/assert that's getting hit, looks like we don't get an output though, don't remember if the test runner captures stdout or stdout and stderr

agnesLeroy:
> looks like we have a fail in the integer GPU tests :| sigabort so might be one of the abort/assert that's getting hit, looks like we don't get an output though, don't remember if the test runner captures stdout or stdout and stderr

Looking into it

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: So actually we should not abort in this case. This is what was causing the tests to fail in the scalar_bitand test, where we set the input to 0 at some point, but if there is the same amount of cleartext blocks than radix blocks the size in the memset is zero. I'll remove these aborts.

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: This error is caught by check_cuda_error

backends/tfhe-cuda-backend/cuda/include/device.h
agnesLeroy: I just removed it, assuming any time we encounter a CudaError we want to abort.

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: I fixed several other compilation warnings in this PR, but I need to make sure we don't let them pass in our PRs in the future.

agnesLeroy: Here's the issue for this: https://github.com/zama-ai/tfhe-rs-internal/issues/434

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -52,184 +50,109 @@ void *cuda_malloc_async(uint64_t size, cuda_stream_t *stream) {
...
-  }
   cudaPointerAttributes attr_dest;
-  cudaPointerGetAttributes(&attr_dest, dest);
+  check_cuda_error(cudaPointerGetAttributes(&attr_dest, dest));
   if (attr_dest.device != stream->gpu_index &&
       attr_dest.type != cudaMemoryTypeDevice) {
-    // error code: invalid device pointer
-    return -1;
+    printf("Cuda error: invalid dest device pointer in copy from GPU to GPU.");
+    abort();
```
tmontaigu, line 121: seeing this get repeated, makes me think we could have panic macro ala Rust
```C++
#include <cstdio>
#include <cstdlib>

#define PANIC(format, ...)                                                     \
  {                                                                            \
    std::fprintf(stderr, "%s::%d::%s: panic.\n" format "\n", __FILE__, __LINE__,    \
            __func__, ##__VA_ARGS__);                                          \
    std::abort();                                                                   \
  }

  int main(int argc, char *argv[])
  {

    PANIC("Hello %s", "world");

    PANIC("oof");
    return 0;
  }
```

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -41,7 +41,9 @@ impl<T: Numeric> CudaVec<T> {
     ///   not be dropped until stream is synchronised
     pub unsafe fn from_async(h_data: &Vec<T>, stream: &CudaStream) -> Self {
         let mut d_data = stream.malloc_async(h_data.len() as u32);
-        stream.copy_to_gpu_async(&mut d_data, h_data.as_slice());
+        if !h_data.is_empty() {
```
agnesLeroy, line 44: I had to add this condition otherwise we trying to copy data from an invalid pointer.

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: That's a good idea, I'll do that :pray:

tfhe/src/core_crypto/gpu/vec.rs
IceTDrinker: ah yes, this is a rust thing when representing slices, if it's empty the pointer can be any well aligned pointer for the type even if there is no memory allocation there

tfhe/src/integer/gpu/server_key/radix/tests.rs
```diff
@@ -32,76 +32,76 @@ macro_rules! create_gpu_parametrized_test{
...
 create_gpu_parametrized_test!(integer_scalar_gt);
 create_gpu_parametrized_test!(integer_scalar_ge);
 create_gpu_parametrized_test!(integer_scalar_lt);
 create_gpu_parametrized_test!(integer_scalar_le);
-create_gpu_parametrized_test!(integer_if_then_else);
-create_gpu_parametrized_test!(integer_max);
-create_gpu_parametrized_test!(integer_min);
+//create_gpu_parametrized_test!(integer_if_then_else);
+//create_gpu_parametrized_test!(integer_max);
+//create_gpu_parametrized_test!(integer_min);
```
IceTDrinker, line 99: commented tests

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -90,12 +90,15 @@ impl CudaStream {
...
     {
         let dest_size = dest.len() * std::mem::size_of::<T>();
-        cuda_memset_async(
-            dest.as_mut_c_ptr(),
-            value.into(),
-            dest_size as u64,
-            self.as_c_ptr(),
-        );
+        // We have to check that dest is not empty, because cuda_memset with size 0 is invalid
+        if !dest.is_empty() {
```
IceTDrinker, line 94: nitpick: we have dest_size already containing the info about copy size so could be dest_size > 0

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -134,12 +141,15 @@ impl CudaStream {
...
-            src.as_c_ptr(),
-            size as u64,
-            self.as_c_ptr(),
-        );
+        // We check that src and dest are not empty to avoid invalid pointers
+        if !src.is_empty() && !dest.is_empty() {
+            cuda_memcpy_async_gpu_to_gpu(
+                dest.as_mut_c_ptr(),
+                src.as_c_ptr(),
+                size as u64,
```
IceTDrinker, line 149: we know that dest.len() > = src.len() and we copy size bytes of data coming from dest (which is potentially longer than source), meaning we can go past the end of src ?

I think the logic is wrong-ish in other functions, what would be safe I believe would be:

either take the min of src.len() and dest.len() but it feels like it could be wrong then, or rather check that the src len is smaller than dest and copy the content from source using the source size ?

tfhe/src/integer/gpu/server_key/radix/tests.rs
agnesLeroy: :scream:

tfhe/src/core_crypto/gpu/mod.rs
agnesLeroy: Yes the logic was not good here: I'll check that dest.len() >= src.len() then use src size for the copy (and check that src_size is > 0)

backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -362,9 +359,12 @@ __host__ void host_integer_mult_radix_kb(
     ch_amount = r / chunk_size;
     dim3 add_grid(ch_amount, num_blocks, 1);
     size_t sm_size = big_lwe_size * sizeof(Torus);
-    cuda_memset_async(new_blocks, 0,
-                      ch_amount * num_blocks * big_lwe_size * sizeof(Torus),
-                      stream);
+    if (ch_amount != 0) {
+      cuda_memset_async(
+          new_blocks, 0,
+          ch_amount * num_blocks * big_lwe_size * sizeof(Torus),
```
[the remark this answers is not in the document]
agnesLeroy: Yeah I know, as num_blocks == 0 should never occur we don't have an early return... same for n = 0. We would have to add those checks in every function basically, right now we just assume we're doing things right and never call those functions with num_blocks == 0 or lwe_dimension == 0.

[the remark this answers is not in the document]
agnesLeroy: I've added one now, but I'm not really satisfied :sweat_smile: I think it would be better to either implement the early return, or assume everywhere num_blocks >0

IceTDrinker (approved):
Looks good for what we discussed, thanks !

PR #832, opened 2024-01-30
chore(ci): use slab action on fast cpu tests workflow

soonum:


IceTDrinker (approved):
This is sooooooooooo nice, also you'll be able to convert CML to this :)

PR #834, opened 2024-01-30
chore(integer): remove deprecated parameter set

IceTDrinker:
Removing experimental parameter sets that were left by mistake


tfhe/src/integer/wopbs/test.rs
```diff
@@ -51,88 +51,6 @@ fn make_basis(message_modulus: usize) -> Vec<u64> {
     }
 }
 
-pub fn wopbs_native_crt() {
```
mayeul-zama, line 54: Why remove these functions?

tfhe/src/integer/wopbs/mod.rs
```diff
@@ -436,13 +436,13 @@ impl WopbsKey {
...
     /// use tfhe::integer::gen_keys_crt;
-    /// use tfhe::integer::parameters::PARAM_4_BITS_5_BLOCKS;
     /// use tfhe::integer::wopbs::WopbsKey;
+    /// use tfhe::shortint::parameters::parameters_wopbs_message_carry::WOPBS_PARAM_MESSAGE_4_CARRY_4_KS_PBS;
     ///
     /// let basis: Vec<u64> = vec![9, 11];
     /// let msg_space: u64 = basis.iter().copied().product();
     ///
-    /// let param = PARAM_4_BITS_5_BLOCKS;
+    /// let param = WOPBS_PARAM_MESSAGE_4_CARRY_4_KS_PBS;
```
mayeul-zama, line 445: Unrelated to this PR, but are the parameters `WOPBS_PARAM_MESSAGE_4_CARRY_4_KS_PBS` optimized for radix wop (without carry I guess)? Is there any advantage of doing that instead of traditional PBS?
In my view Wop is more adapted for CRT representation

tfhe/src/integer/wopbs/test.rs
IceTDrinker: simply unused, those were experiment functions/test functions and not real functions for users

tfhe/src/integer/wopbs/mod.rs
IceTDrinker: I'm just updating the docstrings here, those questions will be tackled during the wop rework, large look ups for radix had some potential advantages through wop but as there was no real alternative it's hard to say

mayeul-zama (approved):

PR #835, opened 2024-01-31
chore(gpu): abort when trying to launch 32 bit multi-bit PBS

agnesLeroy:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/276

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -3,6 +3,7 @@
 
 #include "bootstrap.h"
 #include "bootstrap_multibit.h"
+#include "pbs/bootstrap.cuh"
 #include <cassert>
 #include <cmath>
 #include <functional>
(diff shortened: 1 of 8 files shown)


backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -300,14 +300,16 @@ template <typename Torus> struct int_radix_lut {
...
       // Only 64 bits is supported
-      static_assert(
-          sizeof(Torus) == 8,
-          "Error (GPU multi bit PBS): only 64 bits Torus is supported");
-      scratch_cuda_multi_bit_pbs_64(
-          stream, &pbs_buffer, params.small_lwe_dimension,
-          params.glwe_dimension, params.polynomial_size, params.pbs_level,
-          params.grouping_factor, num_radix_blocks,
-          cuda_get_max_shared_memory(stream->gpu_index), allocate_gpu_memory);
+      if (sizeof(Torus) == sizeof(uint32_t)) {
```
pdroalves, line 303: I think a better solution would be 

```C++
if (sizeof(Torus) == sizeof(uint64_t)) {
// Run the operation
} else {
PANIC("...")
}
```

This way we also handle the possibility of someone running it for `uint8_t` and `uint16_t` (and who knows what else).

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -476,14 +474,18 @@ void scratch_cuda_full_propagation(
...
   if (pbs_type == MULTI_BIT) {
-    uint32_t lwe_chunk_size = get_average_lwe_chunk_size(
-        lwe_dimension, pbs_level, glwe_dimension, num_radix_blocks);
-    // Only 64 bits is supported
-    scratch_cuda_multi_bit_pbs_64(stream, &pbs_buffer, lwe_dimension,
-                                  glwe_dimension, polynomial_size, pbs_level,
-                                  grouping_factor, num_radix_blocks,
-                                  cuda_get_max_shared_memory(stream->gpu_index),
-                                  allocate_gpu_memory, lwe_chunk_size);
+    if (sizeof(Torus) == sizeof(uint32_t)) {
```
pdroalves, line 477: The same applies here. Since we only support 64 bits, we should abort for anything different than that.

pdroalves (commented):
Hey!

pdroalves:
@agnesLeroy do you think we should encapsulate this into a templated method in `bootstrap_multibit.cu`? This way we assert there is nowhere else in the cuda library trying to launch 32-bit operations without going through this `PANIC()` method.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
pdroalves: Just to be sure you didn't miss this one: are we not replacing this `if` by a `switch` as in `execute_scratch_pbs`?

pdroalves (approved):
Great work @agnesLeroy ! I just added a comment about something you might have forgotten, but this PR seems solid to me.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
agnesLeroy: Totally forgot it, thanks :slightly_smiling_face:

PR #836, opened 2024-01-31
Am/chore/add more semver compatible make targets

IceTDrinker:
### PR content/description

Just getting some of the Forward compatibility stuff on main so that the 0.5.x release for forward compat when we get to the 0.6 release

### DIFF
--- a/scripts/integer-tests.sh
+++ b/scripts/integer-tests.sh
@@ -12,6 +12,7 @@ function usage() {
     echo "--signed-only             Run only signed integer tests, by default both signed and unsigned tests are run"
     echo "--cargo-profile           The cargo profile used to build tests"
     echo "--avx512-support          Set to ON to enable avx512"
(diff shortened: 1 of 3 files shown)


scripts/integer-tests.sh
```diff
@@ -12,6 +12,7 @@ function usage() {
     echo "--signed-only             Run only signed integer tests, by default both signed and unsigned tests are run"
     echo "--cargo-profile           The cargo profile used to build tests"
     echo "--avx512-support          Set to ON to enable avx512"
+    echo "--tfhe-package            The package spec like tfhe@0.4.2, default=tfhe"
```
soonum, line 15: Should we put the last tfhe-rs version in the exemple ? :thinking:

soonum (approved):
Just a minor comment, not blocking.

scripts/integer-tests.sh
IceTDrinker: does not really matter, it's an example and we won't update it at each release

PR #837, opened 2024-01-31
chore(ci): add lib symlinking for ease of use of deterministic C API

IceTDrinker:


IceTDrinker:
for 0.5.x, main has another patch coming with that change

IceTDrinker:
actually, will backport https://github.com/zama-ai/tfhe-rs/pull/836 to 0.5 later sorry for the noise

PR #857, opened 2024-02-05
feat(capi): allow cbindgen to generate docs

tmontaigu:
with `documentation=true` cbindgen now properly
generates/copies any rust documentation that
are on `#[no_mangle] pub extern "C" fn` into
their corresponding declaration in the header file.

This will allows to finally start adding some documentation on the CAPI (tfhe.h)

### Check-list:


IceTDrinker (approved):
🎉

we can likely backport this

PR #861, opened 2024-02-06
feat(gpu): cast between unsigned cuda radix ciphertexts

agnesLeroy:
closes: one item of https://github.com/zama-ai/concrete-internal/issues/341

### Check-list:


tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -152,6 +176,63 @@ impl CudaStream {
...
+    {
+        let dest_ptr = dest.as_mut_c_ptr().add(shift as usize);
+        assert!(dest.len() - shift as usize >= src.len());
+        let size = src.len() * std::mem::size_of::<T>();
+        // We check that src is not empty to avoid invalid pointers
+        if size > 0 {
+            cuda_memcpy_async_gpu_to_gpu(dest_ptr, src.as_c_ptr(), size as u64, self.as_c_ptr());
+        }
+    }
+
```
tmontaigu, line 235: Same comments as for `memset_with_ptr_shift_async`

I thing these should be move as method of cuda vec, and use `RangeBouds` to provide a better API.


```rust
impl CudaVec<T> {
	pub unsafe fn copy_range_from_gpu_async<R>(&mut self, range: R, src: &CudaVec<T>) {
		// share code with `memset_with_ptr_shift_async` to get the start and end from a RangeBounds
		let (start, end) = self.range_bounds_to_start_end();

		// check start_end is correct for both self, and src
		// call memcpy
	}
}
```

usage would be `cuda_vec1.copy_range_from_gpu_async(23.., &cuda_vec_2, stream)`

tmontaigu (changes requested):

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -101,6 +101,30 @@ impl CudaStream {
...
+        if dest_shifted_size > 0 {
+            cuda_memset_async(
+                dest_ptr,
+                value.into(),
+                dest_shifted_size as u64,
+                self.as_c_ptr(),
+            );
+        }
+    }
+
```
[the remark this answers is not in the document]
tmontaigu: Later we migh want to create a `CudaSlice` struct but thats not for now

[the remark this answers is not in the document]
agnesLeroy: Thanks for the advice: I implemented this, but still in the CudaStream structure, not in the CudaVec one, as the CudaStream is where all such functions are implemented.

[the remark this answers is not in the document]
agnesLeroy: In the CudaVec we only have the information as to which device the data is stored on. The CudaStream specifies what stream to use to execute the function on GPU.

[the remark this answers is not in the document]
agnesLeroy: That would kind of go against the logic of all the other functions no?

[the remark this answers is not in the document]
tmontaigu: For me its easier if the function is closer to the data is works on, so maybe we would have methods 
on CudaVec<T> that forwards to the free function you defined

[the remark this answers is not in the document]
agnesLeroy: Makes sense, I would have to modify CudaVec to add a CudaStream attribute, and a method to set the CudaStream for a CudaVec. I would need to think a bit more about it though, and this change would concern other functions that are currently exposed in the CudaStream structure. Maybe it's a topic for another PR?

tfhe/src/integer/gpu/ciphertext/mod.rs
```diff
@@ -389,6 +389,66 @@ impl CudaRadixCiphertextInfo {
...
+    }
+
+    pub(crate) fn after_trim_radix_blocks_lsb(&self, num_blocks: usize) -> Self {
+        let mut new_block_info = Self { blocks: Vec::new() };
+        for (i, &b) in self.blocks.iter().enumerate() {
+            if i >= num_blocks {
+                new_block_info.blocks.push(b);
+            }
+        }
+        new_block_info
```
tmontaigu, line 440: Its always good to use Vec::with_capacity when possible I think
```rust
let mut new_block_info = Self { blocks: Vec::with_capacity(self.blocks.len().checked_sub(num_blocks).unwrap_or(0) };
```

Also the loop could be replaced with
```rust
new_block_info.blocks.extend(self.blocks[num_blocks..].iter().cloned())
```

or at least avoid the `if i >= num_blocks` by slicing `self.blocks[num_blocks..]`

tfhe/src/integer/gpu/ciphertext/mod.rs
```diff
@@ -389,6 +389,66 @@ impl CudaRadixCiphertextInfo {
...
+    }
+
+    pub(crate) fn after_trim_radix_blocks_msb(&self, num_blocks: usize) -> Self {
+        let mut new_block_info = Self { blocks: Vec::new() };
+        for (i, &b) in self.blocks.iter().enumerate() {
+            if i < num_blocks {
+                new_block_info.blocks.push(b);
+            }
+        }
+        new_block_info
```
tmontaigu, line 450: same idea as above, we can use Vec::with_capacity, and avoid the if by slicing `self.blocks`

tfhe/src/core_crypto/gpu/mod.rs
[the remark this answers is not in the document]
tmontaigu: No I don't we understand eachother, the method on the CudaVec receives the steam to use as one of the inputs of the method

`vec.fill_range(0..13, 0u64, stream)`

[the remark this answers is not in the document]
agnesLeroy: I'll open another PR for this change, as it affects the other allocation/data copy functions currently in CudaStream.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -66,6 +67,25 @@ impl<T: Numeric> CudaVec<T> {
...
+            Excluded(start) => *start + 1,
+        };
+
+        let end = match range.end_bound() {
+            Unbounded => self.len() - 1,
+            Included(end) => *end,
+            Excluded(end) => *end - 1,
+        };
+
+        (start, end)
```
tmontaigu, line 87: I'm going to nitpick on this:

Generally when given a range as in start, end end is exlusive.

Here you choose that end is inclusive, that's ok, but we should be explicit by making this function return an `std::ops::RangeInclusive` (return `start..=end`), or if you change it to be exclusive return `std::ops::Range` (start..end).

agnesLeroy: Hmm actually looking at it in detail again I think what I did is consistent:
- this function takes a range (start..end) as argument and returns the two usize values for the start and end indexes
- range.end_bound() will return end - 1 if the end bound of the input range is exculsive, and end if it is inclusive: I think this is the expected behavior

agnesLeroy: Thank you for the clarification on the phone :slightly_smiling_face:

tmontaigu (approved):

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -66,6 +67,25 @@ impl<T: Numeric> CudaVec<T> {
...
+            Unbounded => 0usize,
+            Included(start) => *start,
+            Excluded(start) => *start + 1,
+        };
+
+        let end = match range.end_bound() {
+            Unbounded => self.len() - 1,
+            Included(end) => *end,
+            Excluded(end) => *end - 1,
+        };
```
IceTDrinker, line 85: underflows if self.len == 0, the sanity check above (reproduce here) will fail, consider using saturating sub

```rust
    pub unsafe fn copy_dest_range_gpu_to_gpu_async<R, T>(
        &self,
        range: R,
        dest: &mut CudaVec<T>,
        src: &CudaVec<T>,
    ) where
        R: std::ops::RangeBounds<usize>,
        T: Numeric,
    {
        let (start, end) = dest.range_bounds_to_start_end(range).into_inner();
        // size is > 0 thanks to this check
        if end < start 
```

agnesLeroy: Thank you, I've used saturating_sub for the two subtractions in `range_bounds_to_start_end`.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -66,6 +68,25 @@ impl<T: Numeric> CudaVec<T> {
...
+        R: std::ops::RangeBounds<usize>,
+    {
+        let start = match range.start_bound() {
+            Unbounded => 0usize,
+            Included(start) => *start,
+            Excluded(start) => *start + 1,
+        };
+
+        let end = match range.end_bound() {
+            Unbounded => saturating_sub(self.len(), 1),
```
IceTDrinker, line 83: self.len().saturating_sub(1) ? :)

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -66,6 +68,25 @@ impl<T: Numeric> CudaVec<T> {
...
+            Included(end) => *end,
+            Excluded(end) => saturating_sub(*end, 1),
```
IceTDrinker, line 85: same here and you can then remove the import otherwise clippy will be annoying 🙂

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -152,11 +148,66 @@ impl CudaStream {
...
+        range: R,
+        dest: &mut CudaVec<T>,
+        src: &CudaVec<T>,
+    ) where
+        R: std::ops::RangeBounds<usize>,
+        T: Numeric,
+    {
+        let (start, end) = src.range_bounds_to_start_end(range).into_inner();
+        // size is > 0 thanks to this check
+        if end < start {
```
IceTDrinker, line 168: <= if you need size to be > 0 (same elsewhere)

IceTDrinker: nvm missed the formula below

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -152,11 +148,66 @@ impl CudaStream {
...
+    pub unsafe fn copy_src_range_gpu_to_gpu_async<R, T>(
+        &self,
+        range: R,
+        dest: &mut CudaVec<T>,
+        src: &CudaVec<T>,
+    ) where
+        R: std::ops::RangeBounds<usize>,
+        T: Numeric,
+    {
+        let (start, end) = src.range_bounds_to_start_end(range).into_inner();
```
IceTDrinker, line 166: can we maybe use the range.count() method ? I think it should return how many elements are expected to be in the range, and use that invariant afterwards, like if we know the range is empty, skip things and if not we know the range is not empty, I let you judge if it's useful

agnesLeroy: Yes sounds good

agnesLeroy: Hmm actually I don't have a count() method for RangeBounds, only start/end/include, so I left this as it was.

PR #863, opened 2024-02-06
feat(capi): add Cuda support

tmontaigu:
- This adds GPU support in the C API
- Also make ctest (cmake test launcher) print test output when it fails

### Check-list:

### DIFF
--- a/scripts/c_api_tests.sh
+++ b/scripts/c_api_tests.sh
@@ -7,11 +7,12 @@ function usage() {
     echo
     echo "--help                    Print this message"
     echo "--build-only              Pass to only build the tests without running them"
+    echo "--gpu                     Enable GPU support"
     echo
 }
 
 BUILD_ONLY=0
-
+WITH_FEATURE_GPU="OFF"
 while [ -n "$1" ]
 do
    case "$1" in
@@ -24,6 +25,9 @@ do
             BUILD_ONLY=1
             ;;
 
+        "--gpu" )
+            WITH_FEATURE_GPU="ON"
+            ;;
         *)
             echo "Unknown param : $1"
             exit 1
@@ -40,7 +44,7 @@ mkdir -p "${TFHE_BUILD_DIR}"
 
 cd "${TFHE_BUILD_DIR}"
 
-cmake .. -DCMAKE_BUILD_TYPE=RELEASE -DCARGO_PROFILE="${CARGO_PROFILE}"
+cmake .. -DCMAKE_BUILD_TYPE=RELEASE -DCARGO_PROFILE="${CARGO_PROFILE}" -DWITH_FEATURE_GPU="${WITH_FEATURE_GPU}"
 
 make -j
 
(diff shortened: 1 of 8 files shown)


scripts/c_api_tests.sh
```diff
@@ -55,5 +59,8 @@ if [[ $(uname) == "Darwin" ]]; then
     nproc_bin="sysctl -n hw.logicalcpu"
 fi
 
-# Let's go parallel
-ARGS="-j$(${nproc_bin})" make test
+if [ "${WITH_FEATURE_GPU}" == "ON" ]; then
+    ctest --output-on-failure --test-dir "." --parallel "$(${nproc_bin})" --tests-regex ".*cuda.*"
+else
+    ctest --output-on-failure --test-dir "." --parallel "$(${nproc_bin})" --exclude-regex ".*cuda.*"
+fi
```
mayeul-zama, line 66: The Makefile was used before.
We could also use it here
Is there a reason to do this this way now?

tfhe/src/c_api/high_level_api/keys.rs
```diff
@@ -6,18 +6,34 @@ pub struct PublicKey(pub(crate) crate::high_level_api::PublicKey);
 pub struct CompactPublicKey(pub(crate) crate::high_level_api::CompactPublicKey);
 pub struct CompressedCompactPublicKey(pub(crate) crate::high_level_api::CompressedCompactPublicKey);
 pub struct ServerKey(pub(crate) crate::high_level_api::ServerKey);
+/// Compressed version of the ServerKey
+///
+/// Allows to save storage space and transfer time.
+/// Also, the CompressedServerKey is the key format that allows to select
+/// the target hardware of the actual ServerKey when decompressing it.
+pub struct CompressedServerKey(pub(crate) crate::high_level_api::CompressedServerKey);
```
mayeul-zama, line 14: It's not possible to use a `CompressedServerKey` with the C-API on main?

tfhe/src/c_api/high_level_api/keys.rs
```diff
@@ -6,18 +6,34 @@ pub struct PublicKey(pub(crate) crate::high_level_api::PublicKey);
...
+pub struct CudaServerKey(pub(crate) crate::high_level_api::CudaServerKey);
 
 impl_destroy_on_type!(ClientKey);
 impl_destroy_on_type!(PublicKey);
 impl_destroy_on_type!(CompactPublicKey);
 impl_destroy_on_type!(CompressedCompactPublicKey);
 impl_destroy_on_type!(ServerKey);
+impl_destroy_on_type!(CompressedServerKey);
+impl_destroy_on_type!(
+    #[cfg(feature = "gpu")]
```
mayeul-zama, line 27: I guess you could have this guard on the macro directly

scripts/c_api_tests.sh
tmontaigu: This is equivalent than the make file, if you look at it, the makefile just calls ctest, here its 'cleaner' to pass arguments to cmake

```
test:
	@$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --cyan "Running tests..."
	/home/linuxbrew/.linuxbrew/Cellar/cmake/3.24.3/bin/ctest --force-new-ctest-process $(ARGS)
```

tfhe/src/c_api/high_level_api/keys.rs
tmontaigu: No it was missing 😬

tfhe/src/c_api/high_level_api/keys.rs
tmontaigu: Yes, I think I made it like it because I was having some errors about the CudaCompressedServerKey, but they were caused by something else, will fix

mayeul-zama (approved):

mayeul-zama (approved):

PR #865, opened 2024-02-07
feat(gpu): scalar shifts with one wave of pbs

bbarbakadze:
Modify cuda scalar shifts  to use only one wave of pbs


agnesLeroy (approved):
Looks good to me @bbarbakadze, thank you very much! I'd be interested to know what speedup we get with this change on the scalar shifts.

tmontaigu:
You were doing the shift in 2PBS of latency, now it should be only 1

tmontaigu (approved):

PR #866, opened 2024-02-08
chore(tfhe): pin bytemuck temporarily as the 1.14.2 is broken

IceTDrinker:


IceTDrinker:
not the actual root cause, closing

IceTDrinker:
actually let's go with this first as we need to update some crates impacted by the change in bytemuck

agnesLeroy (approved):

tmontaigu:
🙏

PR #868, opened 2024-02-08
feat(gpu): add signed addition

agnesLeroy:

### Check-list:


agnesLeroy:
@tmontaigu here's a first idea to implement signed operations. I haven't seen a difference in the comparison implementations on CPU between signed and unsigned, and for now there are no overflowing ops on GPU, so I'm really not sure this would be a good way to do it.

tmontaigu:
> tmontaigu here's a first idea to implement signed operations. I haven't seen a difference in the comparison implementations on CPU between signed and unsigned, and for now there are no overflowing ops on GPU, so I'm really not sure this would be a good way to do it. I could try to implement the signed casts to see how it looks maybe.

There are differences in comparisons https://github.com/zama-ai/tfhe-rs/blob/f0b4749aca33c38baabc0424a01e275c33fc315a/tfhe/src/integer/server_key/comparator.rs#L422

Also to match the CPU api, we need a different type for signed ciphertexts, also we want tests for every signed operation, no matter if they are the same as unsigned

tmontaigu:
Ok I Saw the CudaSignedRadixCiphertext which is good, but I would still be better to do a few opetations per PR and and the corresponding tests

agnesLeroy:
Thank you very much, I'll do that :slightly_smiling_face:

tmontaigu:
@agnesLeroy this can be closed right ?

PR #870, opened 2024-02-09
Chore/gpu/cuda vec

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -325,8 +325,8 @@ template <typename Torus> struct int_radix_lut {
       for (int i = 0; i < num_radix_blocks; i++)
         h_lwe_indexes[i] = i;
 
-      cuda_memcpy_to_gpu(lwe_indexes, h_lwe_indexes,
-                         num_radix_blocks * sizeof(Torus));
+      cuda_memcpy_async_to_gpu(lwe_indexes, h_lwe_indexes,
+                               num_radix_blocks * sizeof(Torus), stream);
       free(h_lwe_indexes);
 
       // Keyswitch
@@ -375,8 +375,9 @@ template <typename Torus> struct int_radix_lut {
     for (int i = 0; i < num_radix_blocks; i++)
       h_lwe_indexes[i] = i;
 
-    cuda_memcpy_to_gpu(lwe_indexes, h_lwe_indexes,
-                       num_radix_blocks * sizeof(Torus));
+    cuda_memcpy_async_to_gpu(lwe_indexes, h_lwe_indexes,
+                             num_radix_blocks * sizeof(Torus), stream);
+    cuda_synchronize_stream(stream);
     free(h_lwe_indexes);
   }
 
--- a/tfhe/src/core_crypto/gpu/vec.rs
+++ b/tfhe/src/core_crypto/gpu/vec.rs
@@ -3,6 +3,10 @@ use crate::core_crypto::prelude::Numeric;
 use std::collections::Bound::{Excluded, Included, Unbounded};
 use std::ffi::c_void;
 use std::marker::PhantomData;
+use tfhe_cuda_backend::cuda_bind::{
+    cuda_malloc_async, cuda_memcpy_async_gpu_to_gpu, cuda_memcpy_async_to_cpu,
+    cuda_memcpy_async_to_gpu, cuda_memset_async,
+};
 
 /// A contiguous array type stored in the gpu memory.
 ///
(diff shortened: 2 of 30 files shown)


backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -375,8 +375,8 @@ template <typename Torus> struct int_radix_lut {
     for (int i = 0; i < num_radix_blocks; i++)
       h_lwe_indexes[i] = i;
 
-    cuda_memcpy_to_gpu(lwe_indexes, h_lwe_indexes,
-                       num_radix_blocks * sizeof(Torus));
+    cuda_memcpy_async_to_gpu(lwe_indexes, h_lwe_indexes,
+                             num_radix_blocks * sizeof(Torus), stream);
     free(h_lwe_indexes);
```
tmontaigu, line 380: Is that ok ? you launch an async ops, then free.

What if the async ops is not finished before the `free` ?

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,189 @@ pub struct CudaVec<T: Numeric> {
...
+        if end < start {
+            return;
+        }
+        assert!(end < self.len());
+        assert!(end - start < src.len());
+
+        let dest_ptr = self.as_mut_c_ptr().add(start * std::mem::size_of::<T>());
+        let size = (end - start + 1) * std::mem::size_of::<T>();
+        cuda_memcpy_async_gpu_to_gpu(dest_ptr, src.as_c_ptr(), size as u64, stream.as_c_ptr());
+    }
```
tmontaigu, line 190: This is why later we could introduce a `CudaSlice` to make this more rusty

`cuda_vec[0..13].copy_from_gpu(&other_vec[..])`

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,189 @@ pub struct CudaVec<T: Numeric> {
...
+        }
+    }
+
+    /// Copies data between two `CudaVec`
+    ///
+    /// # Safety
+    ///
+    /// - [CudaStream::synchronize] __must__ be called after the copy
+    /// as soon as synchronization is required
+    pub unsafe fn copy_gpu_to_gpu_async(&mut self, src: &Self, stream: &CudaStream)
```
tmontaigu, line 119: The name is `copy_to_gpu_async`, but since ,its  `&mut self`, is should be called  `copy_from_gpu_async`

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,189 @@ pub struct CudaVec<T: Numeric> {
...
+
     /// # Safety
     ///
     /// - `stream` __must__ be synchronized to guarantee computation has finished, and inputs must
     ///   not be dropped until stream is synchronised
-    pub unsafe fn from_async(h_data: &Vec<T>, stream: &CudaStream) -> Self {
-        let mut d_data = stream.malloc_async(h_data.len() as u32);
-        stream.copy_to_gpu_async(&mut d_data, h_data.as_slice());
-        d_data
+    pub unsafe fn from_async(src: &[T], stream: &CudaStream) -> Self {
```
tmontaigu, line 59: from_cpu_async ?

backends/tfhe-cuda-backend/cuda/include/integer.h
agnesLeroy: You're right I'm missing a sync stream here.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,189 @@ pub struct CudaVec<T: Numeric> {
...
+
         Self {
             ptr,
             len,
-            device,
+            device: stream.device,
             _phantom: PhantomData,
         }
     }
+
```
tmontaigu, line 54: Here, does cuda_malloc initialize memory ?

If it does not, I think we should memset it as, uninitialized memory is unsafe

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,189 @@ pub struct CudaVec<T: Numeric> {
...
+    /// # Safety
+    ///
+    /// - [CudaStream::synchronize] __must__ be called as soon as synchronization is
+    /// required
+    pub unsafe fn copy_to_cpu_async(&self, dest: &mut [T], stream: &CudaStream)
+    where
+        T: Numeric,
+    {
+        let src_size = self.len() * std::mem::size_of::<T>();
+        assert!(std::mem::size_of_val(dest) >= src_size);
```
tmontaigu, line 203: Shouldn't we be checking the lenghts here ? and non size_of

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: Makes sense, will do

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: Indeed, thanks for catching it.

agnesLeroy: Now I'm wondering if I should not check that they have the same len(), instead of `dest.len() >= src.len()`

agnesLeroy: @tmontaigu I've checked for equality of len() for `copy_to_gpu_async`,  `copy_from_gpu_async` and `copy_to_cpu_async`.

agnesLeroy: Hmm we're actually doing copies where the target buffer has a size greater than the source. For now it's simpler to keep the original checks.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,190 @@ pub struct CudaVec<T: Numeric> {
...
+        }
+    }
+
+    /// Copies data from slice into `CudaVec`
+    ///
+    /// # Safety
+    ///
+    /// - [CudaStream::synchronize] __must__ be called after the copy
+    /// as soon as synchronization is required
+    pub unsafe fn copy_to_gpu_async(&mut self, src: &[T], stream: &CudaStream)
```
tmontaigu, line 95: This should be named `copy_from_cpu_async` since it `&mut self` that receives data from cpu slice

agnesLeroy: Indeed

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -25,25 +29,190 @@ pub struct CudaVec<T: Numeric> {
...
+    }
     /// # Safety
     ///
-    /// - `ptr` __must__ be a valid device pointer to an array of `len` elements of type `T`
-    pub unsafe fn new(ptr: CudaPtr, len: usize, device: CudaDevice) -> Self {
+    /// - `stream` __must__ be synchronized to guarantee computation has finished
+    pub unsafe fn new_async(len: usize, stream: &CudaStream) -> Self {
+        let size = len as u64 * std::mem::size_of::<T>() as u64;
+        let ptr = CudaPtr {
+            ptr: cuda_malloc_async(size, stream.as_c_ptr()),
```
tmontaigu, line 43: Here, if `len` is `0`, size will be 0

Is it ok for cuda_malloc to allocate 0 bytes ? I've seen on the internet that is returns a null ptr in that case which is ok, can you confirm ?

agnesLeroy: Yes it's ok :+1:

tmontaigu (approved):

PR #872, opened 2024-02-09
Am/feat/keyswitch flexible modulus

IceTDrinker:
### PR content/description

- small update to CiphertextModulus error handling (one of the error messages was not properly printed in the new function)
- provide a convenience From impl for &str from the introduced error type (to be able to use the ? operator when the returned error type is &str)
- update to the keyswitch code to allow to change modulus between input and output
- TODO: manage different input and output Scalar types to benefit from performance improvements
- Decided to wait for the general design around "modulus switching" in keyswitch (and later in the PBS) to dive in a more profound change around Scalar, as this has implications notably for shortint.
- Waiting for actual parmeters for this modulus switching created a test that simulates going from 2^64 to 2^32 as an hardcoded test

### Check-list:

~~* [ ] Docs have been added / updated (for bug fixes / features)~~
~~* [ ] Relevant issues are marked as resolved/closed, related issues are linked in the description~~

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
@@ -5,7 +5,9 @@ use crate::core_crypto::algorithms::misc::divide_ceil;
 use crate::core_crypto::algorithms::slice_algorithms::*;
 use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
 use crate::core_crypto::commons::numeric::UnsignedInteger;
-use crate::core_crypto::commons::parameters::ThreadCount;
+use crate::core_crypto::commons::parameters::{
+    DecompositionBaseLog, DecompositionLevelCount, ThreadCount,
+};
 use crate::core_crypto::commons::traits::*;
 use crate::core_crypto::entities::*;
 use rayon::prelude::*;
@@ -119,24 +121,26 @@ pub fn keyswitch_lwe_ciphertext<Scalar, KSKCont, InputCont, OutputCont>(
         lwe_keyswitch_key.output_key_lwe_dimension(),
         output_lwe_ciphertext.lwe_size().to_lwe_dimension(),
     );
-    assert!(
-        lwe_keyswitch_key.ciphertext_modulus() == input_lwe_ciphertext.ciphertext_modulus(),
-        "Mismatched CiphertextModulus. \
-        LweKeyswitchKey CiphertextModulus: {:?}, input LweCiphertext CiphertextModulus {:?}.",
+
+    let output_ciphertext_modulus = output_lwe_ciphertext.ciphertext_modulus();
+
+    assert_eq!(
         lwe_keyswitch_key.ciphertext_modulus(),
-        input_lwe_ciphertext.ciphertext_modulus()
-    );
-    assert!(
-        lwe_keyswitch_key.ciphertext_modulus() == output_lwe_ciphertext.ciphertext_modulus(),
+        output_ciphertext_modulus,
         "Mismatched CiphertextModulus. \
         LweKeyswitchKey CiphertextModulus: {:?}, output LweCiphertext CiphertextModulus {:?}.",
         lwe_keyswitch_key.ciphertext_modulus(),
-        output_lwe_ciphertext.ciphertext_modulus()
+        output_ciphertext_modulus
     );
     assert!(
-        lwe_keyswitch_key
-            .ciphertext_modulus()
-            .is_compatible_with_native_modulus(),
+        output_ciphertext_modulus.is_compatible_with_native_modulus(),
+        "This operation currently only supports power of 2 moduli"
+    );
+
+    let input_ciphertext_modulus = input_lwe_ciphertext.ciphertext_modulus();
+
+    assert!(
+        input_ciphertext_modulus.is_compatible_with_native_modulus(),
         "This operation currently only supports power of 2 moduli"
     );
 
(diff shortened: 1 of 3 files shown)


IceTDrinker:
ah, the regex fails to validate when there is :: on the first line it seems 🙄

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
```diff
@@ -102,6 +102,45 @@ impl<'de, Scalar: UnsignedInteger> serde::Deserialize<'de> for CiphertextModulus
...
+        match value {
+            CiphertextModulusCreationError::ModulusTooBig => {
+                "Modulus is bigger than the maximum value of the associated Scalar type"
+            }
+            CiphertextModulusCreationError::CustomModuli64BitsOrLessOnly => {
+                "Non power of 2 moduli are not supported for types wider than u64"
+            }
+        }
+    }
+}
```
tmontaigu, line 129: Could we have had  a `const fn err_message(self) -> &str`, and then 
```
impl CiphertextModulusCreationError {
    pub const fn const_panic(self) -> ! {
        panic(self.err_message())
    }
}
```

IceTDrinker: for sure, I was battling const stuff and did not quite think of the design and was hoping you'd have a better idea !

IceTDrinker: I think I struggled with formatting the &str though, I'll check

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
```diff
@@ -146,6 +150,20 @@ pub fn keyswitch_lwe_ciphertext<Scalar, KSKCont, InputCont, OutputCont>(
     // Copy the input body to the output ciphertext
     *output_lwe_ciphertext.get_mut_body().data = *input_lwe_ciphertext.get_body().data;
 
+    // If the moduli are not the same, we need to round the body in the output ciphertext
+    if output_ciphertext_modulus != input_ciphertext_modulus
+        && !output_ciphertext_modulus.is_native_modulus()
```
mayeul-zama, line 155: Isn't there anything to do if `output_ciphertext_modulus != input_ciphertext_modulus` and `output_ciphertext_modulus.is_native_modulus()`?

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
```diff
@@ -146,6 +150,20 @@ pub fn keyswitch_lwe_ciphertext<Scalar, KSKCont, InputCont, OutputCont>(
     // Copy the input body to the output ciphertext
     *output_lwe_ciphertext.get_mut_body().data = *input_lwe_ciphertext.get_body().data;
```
mayeul-zama, line 151: Could be merged with the new following condition

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
```diff
@@ -479,4 +504,18 @@ pub fn par_keyswitch_lwe_ciphertext_with_thread_count<Scalar, KSKCont, InputCont
...
+    {
+        let modulus_bits = output_ciphertext_modulus.get_custom_modulus().ilog2() as usize;
+        let output_decomposer = SignedDecomposer::new(
+            DecompositionBaseLog(modulus_bits),
+            DecompositionLevelCount(1),
+        );
+
+        *output_lwe_ciphertext.get_mut_body().data =
+            output_decomposer.closest_representable(*output_lwe_ciphertext.get_mut_body().data);
+    }
```
mayeul-zama, line 520: Shouldn't this be done before the keyswitch, right after `*output_lwe_ciphertext.get_mut_body().data = *input_lwe_ciphertext.get_body().data;`

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: for now both moduli are power of 2, if the output modulus is the native modulus then no rounding needs to be done, as all the bits are withing the modulus

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: what do you mean ?

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: I think the 32 bits test runs the sequential and parallel code and outputs are bit exact equal, as we are playing with integers everything works ok

IceTDrinker: I'll check that my test actually does that

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: no because we need to copy the body no matter what the modulus is, if I move it we won't copy the body if moduli are equal

mayeul-zama: It could be done in an else condition

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
mayeul-zama: Better to do it in the same place for both versions, even if it's equivalent, don't you think?

IceTDrinker: yes but the parallel thing allocates the output ciphertext itself IIRC, I'll see to tweak it

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: I don't think it's worth it here, we copy the body and round if needed for a single coefficient

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: yep the test runs sequential and parallel keyswitch with the same outputs and checks they are equal

IceTDrinker: I'll see if the code can be made a bit more consistent

tfhe/src/core_crypto/commons/ciphertext_modulus.rs
IceTDrinker: Ok I think I tried to use the from/fmt dbg impleme before meaning that the panic was rejecting the code, now with the const err msg is works as expected, will push the update after I check the rest

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
IceTDrinker: turns out I overcomplicated that part, must have been in between two versions of the code and commited it as is, cleaning up

mayeul-zama (approved):
Thanks!

PR #873, opened 2024-02-09
chore(bench): run keyswitch benchmarks on multi-bit parameters

soonum:


IceTDrinker (approved):
thx ! does it need an update on grafana or will it pick up the new params ?

soonum:
It should pickup the new parameters as well. Let see once merged.

PR #879, opened 2024-02-14
Fixes

tmontaigu:


IceTDrinker (approved):

IceTDrinker:
erm, was a bit fast, maybe a test to check the bit not degree ?

tmontaigu:
what kind of test ?

IceTDrinker:
checking the bit not degree is correct where it was not before for the boolean block ?

tmontaigu:
Ok so quick update the degree problem is actually present for all operations

IceTDrinker:
> Ok so quick update the degree problem is actually present for all operations

Hmm, ok then tests for all boolean blocks

IceTDrinker:
And maybe use the % 2 in all look up tables for boolean blocks LUTs

tmontaigu:
Its not as easy I think because in shortint, boolean ops use a `after_bixor`, `after_bitand` not the degree of a LUT and its that estimation that seems to not be correct

IceTDrinker:
ah yes

IceTDrinker:
but wouldn't we see that in the output of bit ops in integer as well then ?

tmontaigu:
I don't think we really check for the degree, and here we are in a special case where the block only has 1 bit take (Degree(1)) whereas in integer, the degree is always 3 for each block at least (unless its a trivial, but again, we don't really check degree)

I will try to fix after_bitand/bitor/bitxor in this pr add add part of the test hlapo from #850 with added degree checks

tfhe/src/high_level_api/booleans/tests.rs
```diff
@@ -3,21 +3,346 @@
...
+
+fn compact_bool_test_case(setup_fn: impl FnOnce() -> (ClientKey, Device)) {
+    let (cks, sks_device) = setup_fn();
+    let cpk = CompactPublicKey::new(&cks);
+
+    let cttrue = CompactFheBool::encrypt(true, &cpk);
+    let cffalse = CompactFheBool::encrypt(false, &cpk);
+
+    let a = cttrue.expand();
+    let b = cffalse.expand();
```
IceTDrinker, line 324: can we check the degree here as well in the various compressed/compact etc. encryptions ? I believe you have the helper function to do that

I'm guessing PK as well ?

This is getting annoying, sorry :/

tmontaigu: done

IceTDrinker (approved):

PR #880, opened 2024-02-14
chore(ci): set-up a workflow to use the RTX 4090

IceTDrinker:
Add an RTX 4090 CI workflow triggered on label

### DIFF
--- a/.github/workflows/aws_tfhe_gpu_4090_tests.yml
+++ b/.github/workflows/aws_tfhe_gpu_4090_tests.yml
@@ -1,0 +1,70 @@
+# Compile and test tfhe-cuda-backend on an RTX 4090 machine
+name: TFHE Cuda Backend - 4090 full tests
+
+env:
+  CARGO_TERM_COLOR: always
+  ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  RUSTFLAGS: "-C target-cpu=native"
+  RUST_BACKTRACE: "full"
+  SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
+  SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
+  SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
+  SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+on:
+  # Allows you to run this workflow manually from the Actions tab as an alternative.
+  workflow_dispatch:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  cuda-tests-linux:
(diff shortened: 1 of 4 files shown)


.github/workflows/aws_tfhe_gpu_4090_tests.yml
```diff
@@ -0,0 +1,77 @@
+# Compile and test tfhe-cuda-backend on an AWS instance
```
soonum, line 1: Not on an AWS instance. :wink:

.github/workflows/aws_tfhe_gpu_4090_tests.yml
```diff
@@ -0,0 +1,77 @@
...
+
+on:
+  # Allows you to run this workflow manually from the Actions tab as an alternative.
+  workflow_dispatch:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  cuda-tests-linux:
+    name: CUDA tests
```
soonum, line 22: We use the same name in the AWS version. Could we have a way to distinguish between the two flavors ?
Maybe something like `CUDA tests (RTX 4090)`

.github/workflows/aws_tfhe_gpu_4090_tests.yml
```diff
@@ -0,0 +1,77 @@
...
+          labels: 4090_test
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Slack Notification
+        if: ${{ always() }}
+        continue-on-error: true
+        uses: rtCamp/action-slack-notify@b24d75fe0e728a4bf9fc42ee217caa686d141ee8
+        env:
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_MESSAGE: "CUDA AWS tests finished with status: ${{ job.status }}. (${{ env.ACTION_RUN_URL }})"
```
soonum, line 77: Not `CUDA AWS...`

.github/workflows/aws_tfhe_gpu_4090_tests.yml
```diff
@@ -0,0 +1,77 @@
...
+      - cuda-tests-linux
+    if: ${{ always() }}
+    steps:
+      - uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          labels: 4090_test
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Slack Notification
```
soonum, line 71: The location of the notification step seems a bit odd since it's inside `remove_label` job.

IceTDrinker: I copied the M1 mac workflow and it looks to work ok 🤔

soonum (approved):
Now we're talking ! :smiley:

PR #881, opened 2024-02-14
chore(ci): make avx512 enabled by default for benchmarks

IceTDrinker:
- was too error prone when used by other people in the company, no more doubts and we are generally not interested in non avx512 results


soonum (approved):
Nice ! Thanks a lot for removing this source of error.

IceTDrinker:
arf the note needs to be formatted for gitbook... will need to update that

soonum (approved):

PR #884, opened 2024-02-15
chore(doc): display gpu benchmarks results as a showcase

soonum:
This is done to mirror benchmarks results array on CPU. This new array is generated with the latest results available at this date.


IceTDrinker (commented):
Looks ok, I think we should specify the parameters here for the GPU

One thing though is that all new timings are nearly twice as fast as older ones ? 🤔

soonum:
I think that classical PBS parameters have been selected in the former benchmarks. For the newer I've selected multi-bit PBS which explain the gains.

IceTDrinker:
looks fine by me, not approving to avoid launching tests until we get the final word on what to publish

PR #889, opened 2024-02-19
fix(c-api): add missing function on FheBool

tmontaigu:
- safe ser/de
- classical ser/de
- comparisons
- scalar binary fn/comparisons
- compact & compressed fhe bool encryption

### DIFF
--- a/tfhe/src/high_level_api/booleans/base.rs
+++ b/tfhe/src/high_level_api/booleans/base.rs
@@ -1,6 +1,7 @@
 use std::borrow::Borrow;
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign};
 
+use crate::conformance::ParameterSetConformant;
 use crate::high_level_api::booleans::compressed::CompressedFheBool;
 use crate::high_level_api::global_state;
 #[cfg(feature = "gpu")]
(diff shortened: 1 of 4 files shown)


tfhe/src/high_level_api/booleans/base.rs
```diff
@@ -48,6 +50,19 @@ impl Named for FheBool {
     const NAME: &'static str = "high_level_api::FheBool";
 }
 
+impl ParameterSetConformant for FheBool {
+    type ParameterSet = RadixCiphertextConformanceParams;
+
+    fn is_conformant(&self, params: &RadixCiphertextConformanceParams) -> bool {
+        assert_eq!(params.num_blocks_per_integer, 1);
```
IceTDrinker, line 57: shouldn't this return false rather than panic ?

IceTDrinker: ah no I see, you need a specific set of conformant parameter, how about an intermediate type that a RadixCiphertextConformanceParams can be converted into hardcoding the num blocks to 1 ?

tmontaigu: This is so we can keep the same entry point https://github.com/zama-ai/tfhe-rs/blob/ebce4fcfd4d4c9a3dfeaa10a7c39501b0c44198e/tfhe/src/high_level_api/integers/mod.rs#L119

There's the plain CiphertextConformantParameter  but due to the bounds on the function above, it woud mean either we create a different function or somehow change the bounds.

Otherwise, we could change the panic with an error message if the len is not one saying FheBool only expect one block

IceTDrinker: or we hard code the number of blocks here to 1 to check the number of blocks in the boolean blcok directly ? and have a todo to fix later

IceTDrinker: is conformant is not supposed to crash normally it's a safe way to check the data is malformed or well formed

IceTDrinker: remember also that this is a method against adversarial data so we need to check the number of blocks and cannot rely on the HL API constant

IceTDrinker (approved):
Thanks, good for backport, and we'll do the better thing on main in a subsequent PR as discussed

PR #898, opened 2024-02-20
chore(ci): use aws-region as input to stop ec2 instances

soonum:
This is done to handle case where a PR is merged before AWS EC2 instance teardown. If we use profile input in this case, Slab will try to fetch ci/slab.toml on a git reference that doesn't exists anymore thus sending back an error without being able to terminate the instance. By using aws-region Slab won't fetch slab.toml file.


IceTDrinker (approved):
Thanks a lot !

PR #903, opened 2024-02-22
fix(core): fix unsigned noise addition for custom modulus

IceTDrinker:


tfhe/src/core_crypto/commons/numeric/unsigned.rs
```diff
@@ -157,7 +157,10 @@ macro_rules! implement {
                         .wrapping_rem(custom_modulus_u128)
                         .cast_into()
                 } else {
-                    todo!("wrapping_add_custom_mod is not yet implemented for types wider than u64")
+                    if custom_modulus.is_power_of_two() {
+                        return self.wrapping_add(other).wrapping_rem(custom_modulus);
+                    }
+                    todo!("wrapping_add_custom_mod is not yet implemented for types wider than u64 and non power of two moduli")
```
mayeul-zama, line 163: Powers of 2 bigger than 64 do not hit this `todo`
Comment does not reflect that. 

Btw, couldn't we also implement it for non powers of 2 moduli?

Same comment for line 181

IceTDrinker: basically we don't support that for u128 for now as we don't have useful cases for it, so it was left unimplemented for this particular edge case, I don't think we want to support generic u128 moduli, the power of 2 are easy to manage

mayeul-zama: The error message could be interpreted as "not implemented for moduli > 2^64 and for non powers of two modulis"

IceTDrinker: ah got you

IceTDrinker: done

mayeul-zama (approved):

PR #909, opened 2024-02-26
chore(ci): fix missing backslash in rtx benchmark workflow

soonum:
This missing backslash causes the Python command to fail since some input arguments are missing.


agnesLeroy (approved):
Sorry for this!

PR #910, opened 2024-02-26
chore(gpu): add core crypto benches to 4090 bench workflow

agnesLeroy:

### Check-list:

### DIFF
--- a/.github/workflows/gpu_4090_full_benchmark.yml
+++ b/.github/workflows/gpu_4090_full_benchmark.yml
@@ -39,6 +39,8 @@ jobs:
     steps:
       - name: Checkout tfhe-rs
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+        with:
+          fetch-depth: 0
 
       - name: Get benchmark details
         run: |
@@ -116,6 +118,8 @@ jobs:
     steps:
       - name: Checkout tfhe-rs
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+        with:
+          fetch-depth: 0
 
       - name: Get benchmark details
         run: |
--- a/.github/workflows/gpu_4090_full_benchmark.yml
+++ b/.github/workflows/gpu_4090_full_benchmark.yml
@@ -1,5 +1,5 @@
-# Run all integer benchmarks on an RTX 4090 machine and return parsed results to Slab CI bot.
-name: TFHE Cuda Backend - 4090 Integer full benchmarks
+# Run all benchmarks on an RTX 4090 machine and return parsed results to Slab CI bot.
+name: TFHE Cuda Backend - 4090 full benchmarks
 
 env:
   CARGO_TERM_COLOR: always
@@ -25,7 +25,7 @@ jobs:
     name: Cuda integer benchmarks for all operations flavor  (RTX 4090)
     if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || contains(github.event.label.name, '4090_bench') }}
     concurrency:
-      group: ${{ github.workflow }}_${{ github.ref }}
+      group: ${{ github.workflow }}_${{ github.ref }}_cuda_integer_bench
       cancel-in-progress: true
     runs-on: ["self-hosted", "4090-desktop"]
     timeout-minutes: 1440  # 24 hours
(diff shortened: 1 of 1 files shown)


soonum:
> @soonum I should rename the workflow file maybe?

Yes please.

.github/workflows/gpu_4090_full_benchmark.yml
```diff
@@ -95,6 +95,87 @@ jobs:
...
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_MESSAGE: "Integer RTX 4090 full benchmarks failed. (${{ env.ACTION_RUN_URL }})"
+
+  cuda-core-crypto-benchmarks:
+    name: Cuda core crypto benchmarks  (RTX 4090)
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || contains(github.event.label.name, '4090_bench') }}
+    concurrency:
+      group: ${{ github.workflow }}_${{ github.ref }}_cuda_core_crypto_bench
+      cancel-in-progress: true
+      needs: ${{ github.workflow }}_${{ github.ref }}_cuda_integer_bench
```
soonum, line 112: Does the `needs` directive in `concurrency` works ? :thinking: 
In any case, I thinks it's more understandable if `needs` is put right after the `if` (one indentation-level higher).
By using `needs: cuda-integer-benchmarks`, you'll ensure that all sub-jobs in the first step are done.

.github/workflows/gpu_4090_full_benchmark.yml
```diff
@@ -95,6 +95,87 @@ jobs:
...
+    runs-on: ["self-hosted", "4090-desktop"]
+    timeout-minutes: 1440  # 24 hours
+    strategy:
```
soonum, line 115: We don't need a matrix strategy here.

.github/workflows/gpu_4090_full_benchmark.yml
```diff
@@ -95,6 +95,87 @@ jobs:
...
+          curl -v -k \
+          -H "Content-Type: application/json" \
+          -H "X-Slab-Repository: ${{ github.repository }}" \
+          -H "X-Slab-Command: store_data_v2" \
+          -H "X-Hub-Signature-256: sha256=${SIGNATURE}" \
+          -d @${{ env.RESULTS_FILENAME }} \
+          ${{ secrets.SLAB_URL }}
+
       - uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0
         if: ${{ github.event_name == 'pull_request' }}
```
soonum, line 180: We should ensure that the label is removed even in case of failure of the benchmarks.
One solution might be to create an extra job doing just that with the `if: ${{ github.event_name == 'pull_request' }}` at the job level and a `if: always()` at step level.

.github/workflows/gpu_4090_full_benchmark.yml
```diff
@@ -95,16 +95,97 @@ jobs:
...
         if: ${{ always() }}
         continue-on-error: true
         uses: rtCamp/action-slack-notify@b24d75fe0e728a4bf9fc42ee217caa686d141ee8
         env:
           SLACK_COLOR: ${{ job.status }}
-          SLACK_MESSAGE: "Integer RTX 4090 full benchmarks failed. (${{ env.ACTION_RUN_URL }})"
+          SLACK_MESSAGE: "Core crypto RTX 4090 full benchmarks failed. (${{ env.ACTION_RUN_URL }})"
+
+  remove_github_label:
+    name: Remove 4090 bench label
```
soonum, line 185: We need the `if: ${{ github.event_name == 'pull_request' }}` at this level otherwise it would try to run this operation on event other than PR and might fail, or best case be useless.

soonum: I think so yes.
As a matter of fact, you want to add a `needs: [cuda-integer-benchmarks, cuda-core-crypto-benchmarks]` for this job.

agnesLeroy: I've added the `needs: [cuda-integer-benchmarks, cuda-core-crypto-benchmarks]`, does it look good now?

soonum (approved):
Looks good thanks.

soonum (approved):
Small fix added.

PR #911, opened 2024-02-26
chore(ci): lock version of lattice-estimator in workflow

soonum:
Latest version of lattice-estimator produce overflow errors. We force the checkout to the last working version to avoid a red CI.

### DIFF
--- a/.github/workflows/parameters_check.yml
+++ b/.github/workflows/parameters_check.yml
@@ -24,6 +24,7 @@ jobs:
         with:
           repository: malb/lattice-estimator
           path: lattice_estimator
+          ref: '53508253629d3b5d31a2ad110e85dc69391ccb95'
 
       - name: Install Sage
         run: |


IceTDrinker:
> I've tested the fix in CI and it worked.

Thanks, we should try to automatize the update of the lattice estimator and open PRs with the new commits of the lattice estimator and check if things break

IceTDrinker (approved):

IceTDrinker:
we need to take a look at the unattended upgrade making the CSPRNG CI run fail, do we have an issue for it ?

IceTDrinker:
cc @bencrts if you know what could have changed in the lattice estimator that breaks the results ? It seems some values overflow

See failed run here: https://github.com/zama-ai/tfhe-rs/actions/runs/8045852166
And working run here: https://github.com/zama-ai/tfhe-rs/actions/runs/8029173934

Don't know if there are some patches that could be submitted somewhere about it

PR #916, opened 2024-02-26
Am/fix/0 3 compact

IceTDrinker:


tfhe/src/shortint/public_key/compact.rs
```diff
@@ -150,8 +155,8 @@ impl CompactPublicKey {
                     &self.key,
                     &mut ct_list,
                     &plaintext_list,
-                    self.parameters.glwe_modular_std_dev(),
-                    self.parameters.lwe_modular_std_dev(),
+                    encryption_noise,
```
mayeul-zama, line 158: Not sure how it works, but is it expected to have `mask_noise_parameters = body_noise_parameters`?
Then why have separate function arguments?

tfhe/src/core_crypto/fft_impl/fft64/crypto/wop_pbs/tests.rs
mayeul-zama, line 1: Is this already fixed in 0.4 (and 0.5)?

tfhe/src/shortint/public_key/compact.rs
IceTDrinker: Marc's scheme is generic on that but we use the same distribution

dieharder_run.log
```diff
@@ -0,0 +1,10 @@
+seed=Seed(40341300011505446291499474482688097168)
```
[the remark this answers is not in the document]
IceTDrinker: indeed 🤔

tfhe/src/core_crypto/fft_impl/fft64/crypto/wop_pbs/tests.rs
IceTDrinker: yes normally, the dependency that triggered the issue has the fix in 0.4 and 0.5 otherwise we would have tests failing in 0.4 and 0.5

mayeul-zama: I meant the fix `let input_lwe_dimension = lwe_big_sk.lwe_dimension();`
Not sure if this is what you understood

tfhe/src/shortint/public_key/compact.rs
IceTDrinker: you can check, we have the same elsewhere, the idea to reason about it is that the noise is done for the body of size N of a GLWE for example or of size n for an LWE, we have a mask and a body of the same size in the public key, so using the noise for the GLWE in the Big Key case and the LWE noise in the small key case means we are secure in the encryption as we are adding noise in polynomials which have the same size

tfhe/src/core_crypto/fft_impl/fft64/crypto/wop_pbs/tests.rs
IceTDrinker: yes I took the fix from main normally, if you diff the files they should be the same in that place, I'll re check

IceTDrinker: In any case tests are passing and it was only a matter of buffer size so I think we are good :)

mayeul-zama (approved):
Thanks!

IceTDrinker:
@slab-ci cpu_wasm_test

IceTDrinker:
@slab-ci cpu_wasm_test

mayeul-zama (approved):

PR #930, opened 2024-02-28
chore(gpu): extend integer multiplication to N other than 2048

agnesLeroy:
closes:  https://github.com/zama-ai/tfhe-rs-internal/issues/470

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/device.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/device.cu
@@ -142,8 +142,6 @@ __global__ void cuda_set_value_kernel(Torus *array, Torus value, Torus n) {
 template <typename Torus>
 void cuda_set_value_async(cudaStream_t *stream, Torus *d_array, Torus value,
                           Torus n) {
-  if (n == 0)
-    return;
   cudaPointerAttributes attr;
   check_cuda_error(cudaPointerGetAttributes(&attr, d_array));
   if (attr.type != cudaMemoryTypeDevice) {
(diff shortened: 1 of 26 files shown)


bbarbakadze:
can't run tests on my laptop, will have a look

bbarbakadze (approved):

pdroalves:
FYI, yesterday I've fixed the invalid memory access that happened when the PBS runs on `NOSM` mode. However, I found out that there was also a hidden bug when `pbs_level > 1`. I'm on it, but that's why it's taking so long to fix this PR.

pdroalves:
@agnesLeroy Looks like the last issue is an out-of-memory error. I guess that is not a surprise considering that the `lwe_chunk_size` is fixed and `polynomial_size = 8192` consumes a lot of memory. Maybe we can first merge `chore/gpu/lwechunksize_refactor` first and then run again tests. Wdyt?

agnesLeroy:
Getting an out of memory error in the tests :disappointed:

pdroalves:
@slab-ci integer_gpu_full_bench

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -142,6 +142,8 @@ __global__ void cuda_set_value_kernel(Torus *array, Torus value, Torus n) {
 template <typename Torus>
 void cuda_set_value_async(cudaStream_t *stream, Torus *d_array, Torus value,
                           Torus n) {
+  if (n == 0)
+    return;
```
agnesLeroy, line 146: I don't think we need this one

pdroalves: Removed.

pdroalves:
@slab-ci integer_gpu_full_bench

agnesLeroy:
rtx4090 benchmarks on integers are green, I think we're good to merge :slightly_smiling_face:

PR #934, opened 2024-02-29
chore(bench): benchmark 4 bits integer operations

soonum:


tfhe/benches/utilities.rs
```diff
@@ -264,7 +264,7 @@ impl EnvConfig {
             } else if cfg!(feature = "gpu") {
                 BENCH_BIT_SIZES.to_vec()
             } else {
-                vec![8, 16, 32, 40, 64]
+                vec![4, 8, 16, 32, 40, 64]
```
IceTDrinker, line 267: Can we re-organize the if ?

Like the if self.is_fast_bench is the same for everyone, so start with that ?

IceTDrinker: and if needed we have

MULTI_BIT_GPU_SIZES
MULTI_BIT_CPU_SIZES

etc. to make it a bit more clear ? just a thought

IceTDrinker (approved):
Thanks !

PR #937, opened 2024-03-01
test(shortint): add oprf deterministic test

mayeul-zama:


tfhe/src/shortint/oprf.rs
```diff
@@ -117,6 +121,80 @@ pub(crate) mod test {
         a * a
     }
 
+    #[test]
+    // This test is seeded which prevents flakiness
+    // The noise added by the KS and the MS before the PRF LUT evaluation can make this test fail
+    // if the seeded input is close to a boundary between 2 encoded values
+    // Using another KS key can, with a non-neglibgible probability,
```
IceTDrinker, line 128: does it mean that if the KS key generated had different encryption noise we would see different results ?

tfhe/src/shortint/oprf.rs
```diff
@@ -117,6 +121,80 @@ pub(crate) mod test {
...
+    // change the output of the PRF after decoding
+    fn oprf_compare_plain() {
```
IceTDrinker, line 130: add _ci_run_filter at the end otherwise it will not run with our current test filters

tfhe/src/shortint/oprf.rs
mayeul-zama: It's possible, but most of the time, after decryption and decoding, we'd get the same result

tfhe/src/shortint/oprf.rs
mayeul-zama: Done

tfhe/src/shortint/oprf.rs
IceTDrinker: I just want to make sure I understand your comment :)

IceTDrinker: assuming we use the same secret keys, ha I think I know why your key has a hamming weight ratio that is 1 not 1/2, try with 1/2 ?

mayeul-zama: Fixed hamming weight

IceTDrinker (approved):
Thanks !

PR #952, opened 2024-03-05
chore(ci): fix CUDA_PATH bin not being exported in GITHUB_PATH

IceTDrinker:
A small mistake was made while fixing lints in the workflows, restore the cuda path bin to be sent to GITHUB_PATH


soonum (approved):
Thanks for fixing my mistake :pray:

IceTDrinker:
> Thanks for fixing my mistake 🙏

no worries, I missed it during review :)

PR #959, opened 2024-03-07
chore(ci): handle new name format to get pbs throughput values

soonum:
core_crypto benchmark name format has been changed to reflect what's used in other layers. Benchmark result parser was no longer able to compute the right value for the PBS throughput.

### DIFF
--- a/ci/benchmark_parser.py
+++ b/ci/benchmark_parser.py
@@ -108,12 +108,13 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
                     )
                 )
 
+                lowercase_test_name = test_name.lower()
                 # This is a special case where PBS are blasted as vector LWE ciphertext with
                 # variable length to saturate the machine. To get the actual throughput we need to
                 # multiply by the length of the vector.
-                if "PBS_throughput" in test_name and "chunk" in test_name:
+                if "pbs_throughput" in lowercase_test_name and lowercase_test_name.endswith("chunk"):
                     try:
-                        multiplier = int(test_name.split("chunk")[0].split("_")[-1])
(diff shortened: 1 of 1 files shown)


ci/benchmark_parser.py
```diff
@@ -111,9 +111,9 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
                 # This is a special case where PBS are blasted as vector LWE ciphertext with
                 # variable length to saturate the machine. To get the actual throughput we need to
                 # multiply by the length of the vector.
-                if "PBS_throughput" in test_name and "chunk" in test_name:
+                if "pbs_throughput" in test_name and test_name.endswith("chunk"):
```
IceTDrinker, line 171: use test_name.lower() ? to avoid such issues in the future

ci/benchmark_parser.py
```diff
@@ -104,19 +161,20 @@ def recursive_parse(directory, walk_subdirs=False, name_suffix="", compute_throu
...
 
                 # This is a special case where PBS are blasted as vector LWE ciphertext with
                 # variable length to saturate the machine. To get the actual throughput we need to
                 # multiply by the length of the vector.
-                if "PBS_throughput" in test_name and "chunk" in test_name:
+                if "pbs_throughput" in test_name.lower() and test_name.lower().endswith("chunk"):
                     try:
-                        multiplier = int(test_name.split("chunk")[0].split("_")[-1])
+                        multiplier = int(test_name.strip("chunk").split("::")[-1])
                     except ValueError:
```
IceTDrinker, line 174: I'm going to be annoyin line 173 does not have the lower

so let's do

test_name = test_name.lower() and keep the rest as before ?

IceTDrinker (approved):
Thanks a lot 🙂

PR #968, opened 2024-03-08
chore(gpu): rename the low-latency PBS to just PBS and the fast variants to cg

pdroalves:
Solves: https://github.com/zama-ai/tfhe-rs-internal/issues/482

- References to "bootstrap" become "programmable_bootstrap"
- "low_latency" becomes nothing or "classic"
- "fast" as in "low_latency_fast" becomes "cg"
- "multibit_pbs" becomes "multibit_programmable_bootstrap"

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
@@ -457,7 +457,7 @@ __host__ void host_integer_mult_radix_kb(
   cuda_memset_async(block_mul_res, 0, big_lwe_size * sizeof(Torus), stream);
 
   host_addition(stream, radix_lwe_out, vector_result_sb, block_mul_res,
-                big_lwe_size, num_blocks);
+                big_lwe_dimension, num_blocks);
 
   host_propagate_single_carry_low_latency<Torus>(
       stream, radix_lwe_out, mem_ptr->scp_mem, bsk, ksk, num_blocks);


backends/tfhe-cuda-backend/cuda/include/bootstrap.h
```diff
@@ -4,7 +4,7 @@
 #include "device.h"
 #include <cstdint>
 
-enum PBS_TYPE { MULTI_BIT = 0, LOW_LAT = 1, AMORTIZED = 2 };
```
agnesLeroy, line 7: Here it would make more sense to have MULTI_BIT and CLASSICAL only, and AMORTIZED would be a PBS_VARIANT instead.

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -8,7 +8,7 @@
 #include "integer/cmux.cuh"
 #include "integer/negation.cuh"
 #include "integer/scalar_addition.cuh"
-#include "pbs/bootstrap_low_latency.cuh"
```
agnesLeroy, line 11: ?

agnesLeroy (commented):
Wow, looks really good, thanks a lot @pdroalves! The CI is failing just now so I'm not approving right away, just let me know when it's fixed and I'll approve. I only have one comment about PBS_TYPE/VARIANT, you'll see.

backends/tfhe-cuda-backend/cuda/include/bootstrap.h
pdroalves: Indeed!

pdroalves: Thinking again, the amortized implementation is in the limbo. I don't think we need to even mention it as a pbs type or variant. The entry point is completely different and it doesn't need the same buffer structure. It will only be used if the user explictly ask for it, by using its entry point. Thus I will just remove the mention to it.

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
pdroalves: Is there something here?

agnesLeroy: I don't know why this question mark is here sorry :sweat_smile:

pdroalves:
@agnesLeroy I've rebased on main and also replace "bootstrap" references to "programmable_bootstrap".

agnesLeroy (approved):

PR #971, opened 2024-03-11
chore(tfhe): remove last remaining modular_std_dev

IceTDrinker:
- some places were not updated, remove the last non modular std_dev
- the ones to dump parameters are modular so are kept


mayeul-zama (approved):

IceTDrinker:
@mayeul-zama one of the tests was failing, I just pushed the fix, but only after you approved once 😅

PR #975, opened 2024-03-12
feat(integer): make bitnot a trivial operation

tmontaigu:


tfhe/src/shortint/server_key/bitwise_op.rs
```diff
@@ -1217,4 +1218,20 @@ impl ServerKey {
...
         self.smart_evaluate_bivariate_function_assign(ct_left, ct_right, |lhs, rhs| lhs | rhs);
     }
+
+    pub fn bitnot_assign(&self, ct: &mut Ciphertext) {
+        assert!(ct.message_modulus.0.is_power_of_two());
+        if !ct.carry_is_empty() {
+            self.message_extract_assign(ct);
+        }
+        let mut opposite = self.create_trivial(ct.message_modulus.0 as u64 - 1);
+        lwe_ciphertext_sub_assign(&mut opposite.ct, &ct.ct);
```
IceTDrinker, line 1228: how about

```rust
lwe_ciphertext_opposite_assign(&mut ct); // does a wrapping neg on each coeff
let mut body = ct.ct.get_mut_body();
*body.data = (*body.data).wrapping_add(ct.message_modulus.0 as u64 - 1);
```

I know it won't make much of a difference peformance wise but we avoid an allocation and iterate only on the source ciphertext ?

tmontaigu: Yes it is better

IceTDrinker (approved):
Thanks !

IceTDrinker (approved):

PR #979, opened 2024-03-12
replace current fft with cufftdx

bbarbakadze:
### PR content/description
	replace current fft with cufftdx. 
[491](https://github.com/zama-ai/tfhe-rs-internal/issues/491)
### Check-list:


bbarbakadze:
wee need to fix cmake, to automatically set CUDA_ARCH

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ endif()
 
 # Add OpenMP support
 find_package(OpenMP REQUIRED)
+find_package(mathdx REQUIRED COMPONENTS cufftdx CONFIG PATHS "/usr/include/nvidia/mathdx/22.11/")
```
agnesLeroy, line 53: I'm wondering if this is the best solution, this seems platform specific?

backends/tfhe-cuda-backend/cuda/src/integer/integer.cu
```diff
@@ -51,17 +51,17 @@ void cuda_full_propagation_64_inplace(
...
-        bsk, lwe_dimension, glwe_dimension, polynomial_size, ks_base_log,
-        ks_level, pbs_base_log, pbs_level, grouping_factor, num_blocks);
-    break;
+//  case 16384:
+//    host_full_propagate_inplace<uint64_t, int64_t, AmortizedDegree<16384>>(
+//        stream, static_cast<uint64_t *>(input_blocks),
+//        (int_fullprop_buffer<uint64_t> *)mem_ptr, static_cast<uint64_t *>(ksk),
+//        bsk, lwe_dimension, glwe_dimension, polynomial_size, ks_base_log,
+//        ks_level, pbs_base_log, pbs_level, grouping_factor, num_blocks);
+//    break;
```
agnesLeroy, line 60: I think we should delete those lines instead of commenting them out actually :slightly_smiling_face: Code will be cleaner and it's easy to add them back one day.

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
agnesLeroy: @IceTDrinker maybe you have an idea here?

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
```diff
@@ -62,6 +63,11 @@ else()
   set(CMAKE_CUDA_ARCHITECTURES 70)
 endif()
 
+set(CUDA_ARCH 800)
```
agnesLeroy, line 66: We can't set cuda_arch like this, it has to be detected from the platform or set to 700 if nothing is detected. Let me look, see how we can do it.

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
agnesLeroy: @bbarbakadze we should follow the indications from their documentation: https://docs.nvidia.com/cuda/cufftdx/installation.html. They recommend to have a copy of the header files in the `tfhe-cuda-backend` it seems.

agnesLeroy: But we have to be careful with licensing issues. We should leave this PR as a draft for now until this question is addressed.

IceTDrinker: we are already platform dependent through linux and "worse" even Ubuntu officially ? As long as they follow the standard linux filesystem organization (linux standard base or something) I don't think it's too bad potentially, the thing I'm less sure about is the hardcoded version

License questions need to be adressed right away.

I don't see how having the headers in our project is a good idea/practice ? 🤔 could be a way for them to "hijack" code bases

agnesLeroy: @bbarbakadze we should use the 24.01 version which is the latest.

IceTDrinker: I think I saw a thing in the Cmake to download the gtest thing, can it do the same for that lib ?

agnesLeroy: Good idea, I'll look into it

agnesLeroy:
Missing: update the CI to install cufftDx in all workflows.

PR #985, opened 2024-03-13
chore(ci): change slack notifications to be less confusing and more robust

IceTDrinker:
- sometimes the notification will say fail while it did not really fail
- use the generic form which can never be wrong


soonum (approved):
Thanks for tackling that.

soonum (approved):

PR #1004, opened 2024-03-19
feat(gpu): signed scalar shift

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -904,11 +904,19 @@ template <typename Torus> struct int_arithmetic_scalar_shift_buffer {
 
   Torus *tmp_rotated;
 
+  cuda_stream_t *local_stream_1;
+  cuda_stream_t *local_stream_2;
+
   int_arithmetic_scalar_shift_buffer(cuda_stream_t *stream,
                                      SHIFT_TYPE shift_type,
                                      int_radix_params params,
                                      uint32_t num_radix_blocks,
                                      bool allocate_gpu_memory) {
+    // In the arithmetic shift, a PBS has to be applied to the last rotated
+    // block twice: once to shift it, once to compute the padding block to be
+    // copied onto all blocks to the left of the last rotated block
+    local_stream_1 = new cuda_stream_t(stream->gpu_index);
+    local_stream_2 = new cuda_stream_t(stream->gpu_index);
     this->shift_type = shift_type;
     this->params = params;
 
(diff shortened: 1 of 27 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cuh
```diff
@@ -107,4 +109,127 @@ __host__ void host_integer_radix_scalar_shift_kb_inplace(
...
+    uint32_t num_radix_blocks, int_radix_params params, SHIFT_TYPE shift_type,
+    bool allocate_gpu_memory) {
+
+  cudaSetDevice(stream->gpu_index);
+  *mem_ptr = new int_arithmetic_scalar_shift_buffer<Torus>(
+      stream, shift_type, params, num_radix_blocks, allocate_gpu_memory);
+}
+
+template <typename Torus>
+__host__ void host_integer_radix_arithmetic_scalar_shift_kb_inplace(
```
pdroalves, line 124: Do you think it is worth to add a comment explaining the difference between "arithmetic" and "logic"?

backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cuh
```diff
@@ -107,4 +109,127 @@ __host__ void host_integer_radix_scalar_shift_kb_inplace(
...
+  auto lut_univariate_padding_block =
+      mem->lut_buffers_univariate[num_bits_in_block - 1];
+  auto lut_bivariate = mem->lut_buffers_bivariate[shift_within_block - 1];
+
+  if (mem->shift_type == RIGHT_SHIFT) {
+    radix_blocks_rotate_left<<<num_blocks, 256, 0, stream->stream>>>(
+        rotated_buffer, lwe_array, rotations, num_blocks, big_lwe_size);
+    cuda_memcpy_async_gpu_to_gpu(lwe_array, rotated_buffer,
+                                 num_blocks * big_lwe_size_bytes, stream);
+    cuda_synchronize_stream(stream);
```
pdroalves, line 163: I don't think we need this sync.

backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cuh
```diff
@@ -107,4 +109,127 @@ __host__ void host_integer_radix_scalar_shift_kb_inplace(
...
+        for (uint i = 0; i < rotations; i++) {
+          cuda_memcpy_async_gpu_to_gpu(
+              lwe_array + (num_blocks - rotations + i) * big_lwe_size,
+              padding_block, big_lwe_size_bytes, mem->local_stream_1);
+        }
+      }
+#pragma omp section
+      {
+        if (shift_within_block != 0 && rotations != num_blocks) {
+          integer_radix_apply_univariate_lookup_table_kb(
```
pdroalves, line 220: Maybe it would be better to move this conditional outside the `omp` clause so that if it is not satisfied we can avoid the stream synchronization at the cost of maybe increasing redundant code.

backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -868,6 +885,151 @@ template <typename Torus> struct int_shift_buffer {
...
+
+      uint32_t num_bits_in_block = (uint32_t)std::log2(params.message_modulus);
+
+      // LUT
+      // pregenerate lut vector and indexes lut for right shift
+
+      // lut to shift the last block
+      // calculate lut for each 'shift_within_block'
+      // so that in case an application calls scratches only once for a whole
+      // circuit it can reuse memory for different shift values
```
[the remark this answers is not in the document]
agnesLeroy: I think it's alright to keep things as they are, considering that with message 2 carry 2 parameters this is actually only one LUT.

backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cuh
agnesLeroy: Ah I reproduced what 's done in the cmux, where omp is used as well actually.

agnesLeroy: It is actually necessary, otherwise the test fails.

tmontaigu (approved):

agnesLeroy:
@tmontaigu I realize there was an issue with the `create_parametrized_test macro`, as it was redefined in the radix parallel tests.  I've pushed a second commit to the branch to keep only one macro, with classical + multi-bit parameters.

tfhe/src/integer/tests.rs
```diff
@@ -50,5 +64,56 @@ macro_rules! create_parametrized_test{
...
+            $(
+                #[test]
+                #[cfg(not(tarpaulin))]
+                fn [<test_ $name _ $param_no_cover:lower>]() {
+                    $name($param_no_cover)
+                }
+            )*
+        }
+    };
+}
```
tmontaigu, line 118: This macro should call `create_parametrized_test` with the classical PBS params since its bacically the same macro but with different default parameters

tmontaigu (approved):

PR #1007, opened 2024-03-20
chore(ci): create workflow to release tfhe-cuda-backend crate

soonum:


.github/workflows/make_release_cuda.yml
```diff
@@ -0,0 +1,98 @@
...
+        type: boolean
+        default: true
+      push_web_package:
+        description: "Push web js package"
+        type: boolean
+        default: true
+      push_node_package:
+        description: "Push node js package"
+        type: boolean
+        default: true
```
IceTDrinker, line 22: not required for the tfhe backend

.github/workflows/make_release_cuda.yml
```diff
@@ -0,0 +1,98 @@
...
+    steps:
+      - name: Start instance
+        id: start-instance
+        uses: zama-ai/slab-github-runner@8562abbdc96b3619bd5debe1fb934db298f9a044
+        with:
+          mode: start
+          github-token: ${{ secrets.SLAB_ACTION_TOKEN }}
+          slab-url: ${{ secrets.SLAB_BASE_URL }}
+          job-secret: ${{ secrets.JOB_SECRET }}
+          profile: gpu-test
```
IceTDrinker, line 48: does it mean it will start a p3.2xlarge ?

soonum: yes sir

soonum: So that we are sure to build it in the right condition.

IceTDrinker: no shortage on those instances ?

IceTDrinker (approved):
thanks !

.github/workflows/make_release_cuda.yml
soonum: Not anymore on this one. Plus this is a release workflow that is supposed to be run once every quarter.

IceTDrinker: I know, hence why it's important there are no shortages :)

PR #1014, opened 2024-03-21
chore(tfhe): make sure the GPU module is present during doc compilation

IceTDrinker:


backends/tfhe-cuda-backend/build.rs
```diff
@@ -2,6 +2,10 @@ use std::env;
 use std::process::Command;
 
 fn main() {
+    if env::var("RUSTDOC").is_ok() {
+        return;
+    }
```
agnesLeroy, line 7: Hmm it seems this breaks the build with tfhe-cuda-backend?

IceTDrinker:
use https://github.com/rust-lang/docs.rs/issues/147#issuecomment-648844656 to our advantage

backends/tfhe-cuda-backend/build.rs
IceTDrinker: yes that's my bad

agnesLeroy (approved):

PR #1070, opened 2024-04-11
Am/chore/0.6.1 backports

IceTDrinker:
CPU backports

### DIFF
--- a/.github/workflows/m1_tests.yml
+++ b/.github/workflows/m1_tests.yml
@@ -86,6 +86,13 @@ jobs:
         run: |
           make test_boolean
 
+      # Because we do "illegal" things with the build system which Cargo does not seem to like much
+      # we need to clear the cache to make sure the C API is built properly and does not use a stale
+      # cached version
+      - name: Clear build cache
+        run: |
+          cargo clean
+
       - name: Run C API tests
         run: |
           make test_c_api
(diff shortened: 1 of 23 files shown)


.github/workflows/m1_tests.yml
```diff
@@ -86,6 +86,13 @@ jobs:
         run: |
           make test_boolean
 
+      # Because we do "illegal" things with the build system which Cargo does not seem to like much
```
mayeul-zama, line 89: What is it?
Maybe give more info here

tfhe/c_api_tests/test_boolean_keygen.c
```diff
@@ -1,4 +1,4 @@
-#include "tfhe.h"
+#include <tfhe.h>
```
mayeul-zama, line 1: Why this change?
It seems that non system libraries should use ""
https://stackoverflow.com/questions/3162030/difference-between-angle-bracket-and-double-quotes-while-including-heade

.github/workflows/m1_tests.yml
IceTDrinker: was merged in main like that, this is just a backport

tfhe/c_api_tests/test_boolean_keygen.c
IceTDrinker: I took the dominant style in the tests, I have a fix to make for GPU anyways it seems so will see to integrate

.github/workflows/m1_tests.yml
IceTDrinker: I don't have the full explanation except it's using a cached build which is not complete see 2 lines below

mayeul-zama (approved):

PR #1074, opened 2024-04-11
Mz/max noise level woppbs parameters

mayeul-zama:
Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/339


IceTDrinker:
problem is it does not make sense for now to speak of a wopbs noise level, so I don't think this is the right fix, it would more likely mean a WOPBS rework first, so I would spend too much time on that for now

PR #1077, opened 2024-04-12
chore(tfhe): add forward compatibility with 0.6

IceTDrinker:


IceTDrinker:
Not expecting super in depth review given we'll change the way we handle data formats/updates

tfhe/c_api_tests/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ endif()
 set(TFHE_C_API_RELEASE "${CMAKE_CURRENT_SOURCE_DIR}/../../target/${CARGO_PROFILE}")
 
 option(WITH_FEATURE_GPU "Enable if tfhe-rs C API was compiled with the 'gpu' feature activated" OFF)
+option(WITH_FEATURE_GPU "Enable if tfhe-rs C API was compiled with the 'forward_compatibility' feature activated" OFF)
```
mayeul-zama, line 11: Variable name should be `WITH_FORWARD_COMPATIBILITY`

mayeul-zama (commented):
Apart from a small comment, LGTM
Thanks!

tfhe/c_api_tests/CMakeLists.txt
IceTDrinker: thanks indeed 😱

scripts/c_api_tests.sh
```diff
@@ -28,6 +31,12 @@ do
         "--gpu" )
             WITH_FEATURE_GPU="ON"
             ;;
+
+        "--forward-compat" )
+            shift
```
mayeul-zama, line 36: Is `shift` needed here?

IceTDrinker: yes it's taking an ON or OFF value in the Makefile

mayeul-zama (approved):
Thanks!

PR #1079, opened 2024-04-12
fix(gpu): fix a bug that allow pbs_buffer to be initialized with a different max_shared_memory than the one provided by the user

pdroalves:

### Check-list:


agnesLeroy (commented):
Hey @pdroalves! Sorry I don't get how this fixes a bug :sweat_smile: cuda get max shared memory shoud always return  the same value on a given device no?

agnesLeroy:
Hmm, actually I think what we really want is query the device every time we need to know the max shared memory amount: we shouldn't give the possibility to overwrite that value.

pdroalves:
Yes, that would be better. We have an opened issued for some time to remove that parameter from all entry points. Should we do that instead?

agnesLeroy:
Yeah that would be great I think :slightly_smiling_face:

PR #1084, opened 2024-04-15
Mz/ms compression multi bit

mayeul-zama:


tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -336,228 +324,67 @@ pub fn multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>(
...
+}
 
-        // false positive as the mapping function has side effects (thread spawning)
-        #[allow(clippy::needless_collect)]
-        let threads: Vec<_> = (0..thread_count.0)
-            .map(|id| {
-                let tx = tx.clone();
-                s.spawn(move || produce_multi_bit_fourier_ggsw(id, tx))
+impl MultiBitmodulusSwitchedCt {
+    pub fn fromage<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>(
```
IceTDrinker, line 353: 🧀

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1343,8 +1115,6 @@ pub fn std_multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>
 
     assert!(work_queue.len() == lwe_mask.lwe_dimension().0 / grouping_factor.0);
 
-    let work_queue = Mutex::new(work_queue);
```
IceTDrinker, line 1346: I cannot comment directly on it but I see a ggsw_vec comment line 1103 of the new code that could likely be removed

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -289,11 +294,12 @@ pub fn prepare_multi_bit_ggsw_mem_optimized<
 ///     "Multiplication via PBS result is correct! Expected 6, got {pbs_multiplication_result}"
 /// );
 /// ```
-pub fn multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>(
+pub fn multi_bit_ms_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>(
```
IceTDrinker, line 297: This does not match what we have for the PBS, I don't like the fact we mod switch the lwe ciphertext entirely once, as we will be walking the lwe ciphertext once to do the transformation and again afterwards to do the PBS

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -336,228 +324,67 @@ pub fn multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>(
...
-        threads.into_iter().for_each(|t| t.join().unwrap());
-    });
+    }
 }
 
-/// Deterministic version of [`multi_bit_blind_rotate_assign`].
-pub fn multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>(
-    input: &LweCiphertext<InputCont>,
+pub fn multi_bit_blind_rotate_assign<Scalar, OutputCont, KeyCont>(
+    switched_modulus_input: &MultiBitmodulusSwitchedCt,
```
IceTDrinker, line 378: I don't want to change the meaning of the blind rotate

For the modulus switch decompression, the problem lies in the specialization of a blind rotation to manage some "error correcting terms" which I think could be done with something looking like the multi bit PBS modulus switch zipped with the error correction factors or by "just" updating the prepare multi big GGSW by taking an option for error correction, that would be hard coded to None everywhere we don't do the decompression, as it would be hard coded to None the compiler will be able to delete all the code that does not correspond to that case outside of the decompression.

The factorization makes it  much harder to reason about the multi bit PBS now with the callbacks for the determinism

I would prefer keeping the deterministic and non deterministic variants separate as it makes quite a bit of state to remember inside that function

The AtomicUsize is good to keep

IceTDrinker (commented):
Disagree with the refactor of the multi bit PBS as is

IMO the atomic usize is good to keep so we can keep that refactor everywhere.

For the multi bit blind rotation I would have a crate internal function for the FFT deterministic case which can take an optional vector of correction coefficients and adds the correction during the mod switching, the multi bit PBS would hard code the coefficients to None allowing the compiler to optimize the code as before as there is no error correction to be done in that case.

For the decompression then I would have an entry point to decompress a CompressedModSwitchedGroupedLweCiphertext (maybe just MultBitCompressedModSwitchedLweCiphertext ? to match the blind rotation algorithm, but the name makes less sense in that case) that uses the same function providing the error correction factors from the ciphertext as a Some() variant meaning the compiler will also be able to generate optimized code for that case, while sharing the whole code of the deterministic FFT blind rotation and keeping the rest the same

I don't want to have the deterministic code and non deterministic code mixed in a single function, it requires call backs which makes it harder to reason about IMO

tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -121,6 +121,10 @@ impl PolynomialSize {
         );
         FourierPolynomialSize(self.0 / 2)
     }
+
+    pub fn to_ciphertext_modulus_log(&self) -> CiphertextModulusLog {
+        CiphertextModulusLog(self.log2().0 + 1)
+    }
```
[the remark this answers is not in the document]
mayeul-zama: I think the naming could make clear that it applies to the MS only, wdyt?

tfhe/src/shortint/ciphertext/compressed_modulus_switched_ciphertext.rs
```diff
@@ -51,3 +53,20 @@ impl ParameterSetConformant for CompressedModulusSwitchedCiphertext {
...
+
+impl ParameterSetConformant for InternalCompressedModulusSwitchedCiphertext {
+    type ParameterSet = LweCiphertextParameters<u64>;
+
+    fn is_conformant(&self, param: &LweCiphertextParameters<u64>) -> bool {
+        match self {
+            Self::Classic(a) => a.is_conformant(param),
+            Self::MultiBit(a) => a.is_conformant(param),
+        }
+    }
```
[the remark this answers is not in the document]
mayeul-zama: Indeed

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: I could probably remove commit `refactor(core): factorize multi_bit blind_rotate` without changing much else and keeping commit `refactor(core): separate multi_bit modulus_switch and blind_rotate`
What do you think about this?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: What do you don't like about it ?

tfhe/src/core_crypto/commons/parameters.rs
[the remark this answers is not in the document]
IceTDrinker: to_modulus_switch_something ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
IceTDrinker: could be I haven't checked the specific commits while doing the review, I'll take another look

tfhe/src/core_crypto/commons/parameters.rs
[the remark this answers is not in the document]
mayeul-zama: Yes, something like this

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
IceTDrinker: it becomes inconsistent with the existing PBS API and the we need to go through the whole LWE once, which represents ~ 800 * 8 = 6400 bytes of memory, while it could be done during the PBS as it's one today

tfhe/benches/core_crypto/dev_bench.rs
```diff
@@ -145,6 +145,7 @@ fn multi_bit_pbs<Scalar: UnsignedTorus + CastInto<usize> + CastFrom<usize> + Syn
                     &accumulator.as_view(),
                     &multi_bit_bsk,
                     thread_count,
+                    false,
```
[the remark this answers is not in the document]
mayeul-zama: in `ShortintBootstrappingKey::MultiBit`, `deterministic_execution` is a `bool` and IMO we should use the same covention
If we change it, many parameter set will also have to change, so better to do it in another PR. I can open a issue.
Wdyt?

tfhe/src/shortint/server_key/modulus_switched_compression.rs
```diff
@@ -103,40 +124,73 @@ impl ServerKey {
...
                         buffers,
                     );
                 }
-                PBSOrder::BootstrapKeyswitch => {
-                    apply_programmable_bootstrap(
-                        &self.bootstrapping_key,
+                ShortintBootstrappingKey::MultiBit {
+                    fourier_bsk,
+                    thread_count,
+                    deterministic_execution: _,
```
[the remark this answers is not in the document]
IceTDrinker: fair

tfhe/src/core_crypto/algorithms/generate_accumulator.rs
```diff
@@ -0,0 +1,58 @@
+use super::super::prelude::{
+    allocate_and_trivially_encrypt_new_glwe_ciphertext, CastFrom, CastInto, CiphertextModulus,
+    GlweCiphertextOwned, GlweSize, PlaintextList, PolynomialSize, UnsignedTorus,
+};
+
+// Helper function to generate an accumulator for a PBS
```
IceTDrinker, line 6: This needs a docstring

- explain delta allows to scale the message in the MSBs
- explain what the message modulus is here
- explain there is redundancy via "boxes" to reduce the noise during a PBS operation
- explain the accumulator is rotated by half a "box" to manage the negacyclic nature of the PBS
- link to a PBS docstring to show usage

tfhe/src/core_crypto/entities/lwe_multi_bit_bootstrap_key.rs
```diff
@@ -183,6 +183,11 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweMultiBitBootstr
         bsk
     }
 
+    /// Return an iterator over the GGSW ciphertexts composing the key.
+    pub fn ggsw_iter(&self) -> impl DoubleEndedIterator<Item = GgswCiphertext<&'_ [C::Element]>> {
+        self.ggsw_list.iter()
+    }
+
```
IceTDrinker, line 190: this should not be needed as we currently have a deref on the multi bit bsk

so bsk.iter() or &*bsk.iter() should do the trick

IceTDrinker (commented):
Small stuff, will get to the actual compression stuff later

IceTDrinker:
We'll see also for some potential new types for Delta and the message modulus, and naming the primitive at the core level

IceTDrinker:
Do we need the accumulator change for the multi bit ?

tfhe/src/shortint/server_key/modulus_switched_compression.rs
```diff
@@ -14,34 +21,48 @@ impl ServerKey {
...
+                        keyswitch_lwe_ciphertext(
+                            &self.key_switching_key,
+                            &ct.ct,
+                            &mut ciphertext_buffers.buffer_lwe_after_ks,
+                        );
+                    }
+                    PBSOrder::BootstrapKeyswitch => ciphertext_buffers
+                        .buffer_lwe_after_ks
+                        .as_mut_view()
+                        .into_container()
```
IceTDrinker, line 38: .as_mut, check for all .into_container usage in your diff please :)

tfhe/src/shortint/server_key/modulus_switched_compression.rs
```diff
@@ -14,34 +21,48 @@ impl ServerKey {
...
+                        .copy_from_slice(ct.ct.as_view().into_container()),
```
IceTDrinker, line 39: .as_ref

tfhe/src/shortint/server_key/modulus_switched_compression.rs
```diff
@@ -103,40 +124,70 @@ impl ServerKey {
...
-                        buffers,
+                        MonomialDegree(0),
                     );
+                }
+            }
 
+            match self.pbs_order {
+                PBSOrder::KeyswitchBootstrap => output
+                    .as_mut()
+                    .copy_from_slice(ciphertext_buffers.buffer_lwe_after_pbs.into_container()),
```
IceTDrinker, line 189: .as_ref

tfhe/src/core_crypto/algorithms/test/modulus_switch_compression.rs
```diff
@@ -1,5 +1,4 @@
 use super::*;
```
IceTDrinker, line 1: we need the equivalent for the multi bit right ?

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        let body = modulus_switch(*input_lwe_body.data, log_modulus).cast_into();
+
+        let switched_modulus_input_mask_per_group: Vec<_> = input_lwe_mask
+            .as_ref()
+            .chunks_exact(grouping_factor.0)
+            .flat_map(|lwe_mask_elements| {
+                modulus_switch_multi_bit(log_modulus, grouping_factor, lwe_mask_elements)
+            })
+            .collect();
```
IceTDrinker, line 265: this does not compress for large grouping factor right ? where are the correction coefficients for the blind rotation we discussed ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1680,11 +1589,25 @@ pub fn std_multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, Output
...
-                work_queue.iter().skip(thread_id).step_by(thread_count.0)
-            {
-                let (ready_for_consumer_lock, condvar, fourier_ggsw_buffer) =
-                    &fourier_multi_bit_ggsw_buffers[dest_idx];
+
+            #[allow(clippy::type_complexity)]
+            let (ready_for_consumer_lock, condvar, fourier_ggsw_buffer): &(
+                Mutex<bool>,
+                Condvar,
+                Mutex<FourierGgswCiphertext<ABox<[c64]>>>,
```
IceTDrinker, line 1597: why does it need a type annotation ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -6,28 +6,87 @@ use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
...
+}
+
+pub struct StandardMultiBitmodulusSwitchedCt<
+    'a,
+    Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>,
+> {
+    input: &'a LweCiphertext<&'a [Scalar]>,
+    grouping_factor: LweBskGroupingFactor,
+    log_modulus: CiphertextModulusLog,
+}
```
IceTDrinker, line 60: this is public so please use the Container approach here

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -6,28 +6,87 @@ use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
...
+            let mask_position = lwe_mask_elements.len() - (mask_idx + 1);
+            let selection_bit: Scalar = Scalar::cast_from((ggsw_idx >> mask_position) & 1);
+            monomial_degree =
+                monomial_degree.wrapping_add(selection_bit.wrapping_mul(mask_element));
+        }
+        modulus_switch(monomial_degree, ciphertext_modulus_log).cast_into()
+    })
+}
+
+pub trait MultiBitmodulusSwitchedCt: Sync {
```
IceTDrinker, line 45: Modulus with a capital M

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -6,28 +6,87 @@ use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
...
+    fn switched_modulus_input_lwe_body(&self) -> usize {
+        let input_lwe_body = self.input.get_body();
+
+        modulus_switch(*input_lwe_body.data, self.log_modulus).cast_into()
+    }
+    fn switched_modulus_input_mask_per_group(
+        &self,
+        index: usize,
+    ) -> impl Iterator<Item = usize> + '_ {
+        let (_body, mask) = self.input.into_container().split_last().unwrap();
```
IceTDrinker, line 74: self.input.get_mask ?

IceTDrinker (changes requested):
The PR is missing the corrections for the mod switched coefficients right ?

Some minor stuff but since it's missing the corrections I did not read everything

IceTDrinker:
https://github.com/zama-ai/tfhe-rs/actions/runs/8754845373
https://github.com/zama-ai/tfhe-rs/actions/runs/8754845335
https://github.com/zama-ai/tfhe-rs/actions/runs/8754844989
https://github.com/zama-ai/tfhe-rs/actions/runs/8754844951

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Does not compile, with an error about the lifetime of a temporary

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: This object is not meant to be stored, it's a kind of iterator
I don't think it's worth the effort

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
IceTDrinker: I see, you can use .as_ref instead of the into_container though I believe

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
IceTDrinker: It's not about storing, it's about usability without having to call .as_view

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: error[E0282]: type annotations needed for `&(_, _, _)`

IceTDrinker: ok 🤔

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
mayeul-zama: I haven't implemented that yet 
I guess we can do it in another PR

IceTDrinker: I would like to have it in this oneI don't think it will be big, it's about comparing the modulus switched values and storing an additional packed integer vector for the corrections

IceTDrinker: the only potential difficulty is how many bits to store, but that can be determined at compression time worst case instead of a generic formula

IceTDrinker: though for conformance likely better to have a higher bound already computed

tfhe/src/core_crypto/algorithms/test/modulus_switch_compression.rs
mayeul-zama: The decompression does not return a ciphertext
We must go through a multi bit BR here, which is tested in a doctest

IceTDrinker: fair enough but I would like a core test allowing to test the multi bit ms compression/decompression that is a bit more thorough than a doctest

Also taking coefficients sufficiently spaced you should be able to check the decryption nonetheless for the multi bit CT right ? as you store each mask mod switched ?

given you already have the doctest, adapting should be ok I would think ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Done

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Done

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/modulus_switched_compression.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/modulus_switched_compression.rs
mayeul-zama: Done

tfhe/src/shortint/server_key/modulus_switched_compression.rs
mayeul-zama: Done

tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -212,17 +219,17 @@ pub struct ThreadCount(pub usize);
...
+impl MultiBitGroupingFactor {
+    pub fn combined_space_size(&self) -> MultiBitCombinedSpaceSize {
+        MultiBitCombinedSpaceSize(1 << self.0)
     }
 }
 
 /// The number of GGSW ciphertexts required per multi_bit BSK element
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Serialize, Deserialize)]
-pub struct GgswPerLweMultiBitBskElement(pub usize);
+pub struct MultiBitCombinedSpaceSize(pub usize);
```
IceTDrinker, line 232: this naming makes no sense to me 🤔

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1542,27 +1497,17 @@ pub fn std_multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>
...
     multi_bit_bsk: &LweMultiBitBootstrapKey<KeyCont>,
     thread_count: ThreadCount,
 ) where
     // CastInto required for PBS modulus switch which returns a usize
     Scalar: UnsignedTorus + CastInto<usize> + CastFrom<usize> + Sync,
-    InputCont: Container<Element = Scalar> + Sync,
     OutputCont: ContainerMut<Element = Scalar>,
     KeyCont: Container<Element = Scalar> + Sync,
 {
-    assert_eq!(
```
IceTDrinker, line 1557: why remove ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1581,14 +1526,6 @@ pub fn std_multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, Output
         multi_bit_bsk.polynomial_size(),
     );
 
-    assert_eq!(
```
IceTDrinker, line 1584: why remove ?

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1603,42 +1540,31 @@ pub fn std_multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, Output
         "Got thread_count == 0, this is not supported"
     );
 
-    let (lwe_mask, lwe_body) = input.get_mask_and_body();
+    assert!(accumulator
+        .ciphertext_modulus()
+        .is_compatible_with_native_modulus());
```
IceTDrinker, line 1545: an error message please, if we are missing that in other multi bit primitives feel free to update

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1680,11 +1606,25 @@ pub fn std_multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, Output
...
+                Mutex<bool>,
+                Condvar,
+                Mutex<FourierGgswCiphertext<ABox<[c64]>>>,
+            ) = &fourier_multi_bit_ggsw_buffers[dest_idx];
+
+            for i in 0.. {
+                let work_index = thread_id + i * thread_count.0;
+                if work_index >= max_work_index {
+                    break;
+                }
```
IceTDrinker, line 1621: prefer the construction without an infinite range and proper stepping, because we know when the loop is supposed to end

IceTDrinker (commented):
Final nitpicks, let's stick to what was seen in reviews for updates to avoid other review rounds, the objective is to merge the next time around

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -6,28 +6,104 @@ use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
...
+
+// Returns an iterator of booleans (as usize), corresponding to successive mask group elements
+// to indicate if they must be used at the given combined_space_index
+pub(crate) fn selection_bit(
+    grouping_factor: LweBskGroupingFactor,
+    combined_space_index: usize,
+) -> impl Iterator<Item = usize> {
+    debug_assert!(combined_space_index < grouping_factor.ggsw_per_multi_bit_element().0);
+
+    (0..grouping_factor.0).map(move |mask_idx| {
```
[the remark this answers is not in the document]
mayeul-zama: Needed

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -0,0 +1,468 @@
...
+                            .cast_into();
+
+                    sum_modulus_switched = sum_modulus_switched.wrapping_add(modulus_switched);
+                }
+
+                let modulus_switched_sum: usize =
+                    modulus_switch(monomial_degree, log_modulus).cast_into();
+
+                let diff =
+                    modulus_switched_sum.wrapping_sub(sum_modulus_switched % (1 << log_modulus.0));
```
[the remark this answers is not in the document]
mayeul-zama: No it's because the sum may be bigger than `1 << log_modulus`
We want a sum of MSed elements in [0,1 << log_modulus[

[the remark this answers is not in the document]
IceTDrinker: but if you do the sum first and then mod switch then you don't need the mod ?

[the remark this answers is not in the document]
IceTDrinker: my bad misread

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: A while loop would be clearer IMO
Will change

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: The input is already modulus switched (so we don't have it's initial ciphertext_modulus)
And it's ot necessary to have the same input and output moduli

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: I guess we could add a `lwe_dimension` method to the `MultiBitModulusSwitchedCt` trait 
Wdyt?

IceTDrinker: sounds good

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -1303,19 +1277,11 @@ pub fn std_multi_bit_blind_rotate_assign<Scalar, InputCont, OutputCont, KeyCont>
...
     assert_eq!(
-        input.ciphertext_modulus(),
         accumulator.ciphertext_modulus(),
-        "Mismatched CiphertextModulus between input ({:?}) and accumulator ({:?})",
-        input.ciphertext_modulus(),
-        accumulator.ciphertext_modulus(),
-    );
-
-    assert_eq!(
-        input.ciphertext_modulus(),
```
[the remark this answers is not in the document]
mayeul-zama: The input is already modulus switched so we don't have access to its initial ciphertext_modulus

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -637,73 +697,52 @@ pub fn multi_bit_deterministic_blind_rotate_assign<Scalar, InputCont, OutputCont
...
+                Mutex<bool>,
+                Condvar,
+                Mutex<FourierGgswCiphertext<ABox<[c64]>>>,
+            ) = &fourier_multi_bit_ggsw_buffers[dest_idx];
+
+            for i in 0.. {
+                let work_index = thread_id + i * thread_count.0;
+                if work_index >= max_work_index {
+                    break;
+                }
```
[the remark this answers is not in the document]
mayeul-zama: Fixed

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
```diff
@@ -6,28 +6,104 @@ use crate::core_crypto::commons::math::decomposition::SignedDecomposer;
...
+pub(crate) fn modulus_switch_multi_bit<Scalar>(
+    ciphertext_modulus_log: CiphertextModulusLog,
+    grouping_factor: LweBskGroupingFactor,
+    lwe_mask_elements: &[Scalar],
+) -> impl Iterator<Item = usize> + '_
+where
+    Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>,
+{
+    // Start at 1, the first ggsw is not rotated
+    (1..grouping_factor.ggsw_per_multi_bit_element().0).map(move |combined_space_index| {
```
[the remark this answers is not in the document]
mayeul-zama: Renamed power_set_index

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -0,0 +1,468 @@
...
+            let half_needed_space = diffs
+                .iter()
+                .map(|a| *a as isize)
+                .map(|a| {
+                    if a >= 0 {
+                        // number of elements between 0 and a
+                        a + 1
+                    } else {
+                        // number of elements between -1 and a
+                        -a
```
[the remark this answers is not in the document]
mayeul-zama: Added comment

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -0,0 +1,468 @@
...
+                    }
+                })
+                .max()
+                .unwrap() as usize;
+
+            let half_needed_space_ceil_log =
+                half_needed_space.ilog2() + u32::from(!half_needed_space.is_power_of_two());
```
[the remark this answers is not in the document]
mayeul-zama: Fixed

tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -0,0 +1,468 @@
...
+
+        let mut diffs = vec![];
+
+        for lwe_mask_elements in input_lwe_mask.as_ref().chunks_exact(grouping_factor.0) {
+            for ggsw_idx in 1..grouping_factor.ggsw_per_multi_bit_element().0 {
+                if ggsw_idx.count_ones() == 1 {
+                    continue;
+                }
+
+                let mut sum_modulus_switched = 0;
```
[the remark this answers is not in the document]
mayeul-zama: Remaned

tfhe/src/core_crypto/algorithms/test/modulus_switch_compression.rs
```diff
@@ -62,4 +63,79 @@ fn encryption_ms_decryption<Scalar: UnsignedTorus + Sync + Send>(
...
+}
+
+#[test]
+fn test_ms_with_packing() {
+    for ciphertext_modulus in [
+        CiphertextModulus::new(1 << 63),
+        CiphertextModulus::new_native(),
+    ] {
+        for lwe_dimension in 10..1025 {
+            for log_modulus in 10..15 {
```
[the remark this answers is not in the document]
mayeul-zama: Done

tfhe/src/core_crypto/algorithms/test/modulus_switch_compression.rs
```diff
@@ -62,4 +63,79 @@ fn encryption_ms_decryption<Scalar: UnsignedTorus + Sync + Send>(
...
+    }
+}
+
+#[test]
+fn test_ms_with_packing() {
+    for ciphertext_modulus in [
+        CiphertextModulus::new(1 << 63),
+        CiphertextModulus::new_native(),
+    ] {
+        for lwe_dimension in 10..1025 {
```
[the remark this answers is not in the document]
mayeul-zama: Done

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Done

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Added

tfhe/src/core_crypto/algorithms/lwe_multi_bit_programmable_bootstrapping.rs
mayeul-zama: Added

IceTDrinker (approved):
Alright let's go !

Thanks a lot 🙂

PR #1085, opened 2024-04-15
feat(gpu): cast signed/unsigned and signed/signed

agnesLeroy:

### Check-list:


tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub(crate) unsafe fn new(ptr: CudaPtr, len: usize) -> CudaSliceMut<'a, T> {
+        CudaSliceMut {
+            ptr,
+            len,
+            _phantom_1: PhantomData,
+            _phantom_2: PhantomData,
+        }
+    }
+
+    /// # Safety
```
tmontaigu, line 57: We can add 
```
The caller must ensure that the slice outlives the pointer this function returns, or else it will end up pointing to garbage.
```
(Stolen from Rust's stdlib)

tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -0,0 +1,194 @@
...
+    ///
+    /// # Safety
+    ///
+    /// - [CudaStream::synchronize] __must__ be called after the copy
+    /// as soon as synchronization is required
+    pub unsafe fn copy_to_cpu_async(&self, dest: &mut [T], stream: &CudaStream)
+    where
+        T: Numeric,
+    {
+        assert!(self.len() <= dest.len());
```
tmontaigu, line 101: I think we shold test for equality, thus making the user have explicit and correct range for both slices

tmontaigu (commented):
Some comments, haven't yet looked at the indexing stuff

tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -0,0 +1,194 @@
+use crate::core_crypto::gpu::vec::range_bounds_to_start_end;
+use crate::core_crypto::gpu::{CudaPtr, CudaStream};
+use crate::core_crypto::prelude::Numeric;
+use std::ffi::c_void;
+use std::marker::PhantomData;
+use tfhe_cuda_backend::cuda_bind::{cuda_memcpy_async_gpu_to_gpu, cuda_memcpy_async_to_cpu};
+
+#[derive(Debug)]
+pub struct CudaSlice<'a, T: Numeric> {
```
tmontaigu, line 9: This one should also derive Copy and Clone (if the CudaPtr does not derive Copy and Clone we should add it

tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -0,0 +1,194 @@
...
+        let (start, end) = range_bounds_to_start_end(self.len(), range).into_inner();
+
+        // Check the range is compatible with the vec
+        if end <= start || end > self.len - 1 {
+            None
+        } else {
+            // Shift ptr
+            let shifted_ptr: *mut c_void = unsafe {
+                self.ptr
+                    .as_mut_c_ptr()
```
tmontaigu, line 133: Minor thing but we could use [byte_add](https://doc.rust-lang.org/core/primitive.pointer.html#method.byte_add)
`self.ptr.byte_add(start * std::mem::sizeof::<T>())`

agnesLeroy: I get this compilation error:
```
error: current MSRV (Minimum Supported Rust Version) is `1.73.0` but this item is stable since `1.75.0` 
```
Do you know how I could fix it?

tmontaigu: Only way to fix is to bump the minimum rust version that is in the tfhe/Cargo.toml

But that has some implications for users

IceTDrinker: for 0.7 I think it's fine yes, for 0.6 it would be breaking I'm afraid

agnesLeroy: Ok so what I can do is not use `byte_add` in this PR, then I'll do another small PR just to increase the MSRV and use byte_add there. Wdyt?

IceTDrinker: yep if this PR targets 0.6 that sounds good

tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -0,0 +1,192 @@
...
+{
+    /// # Safety
+    pub(crate) unsafe fn new(ptr: CudaPtr, len: usize) -> CudaSlice<'a, T> {
+        CudaSlice {
+            ptr,
+            _len: len,
+            _phantom_1: PhantomData,
+            _phantom_2: PhantomData,
+        }
+    }
```
tmontaigu, line 36: One last comment, we should probably add a Safety comment that says somethinglike `the ptr must be valid for reads for len * std::mem::size_of::<T> bytes on the cuda side` 

and for cuda slice mut `the ptr must be valid for read and write for len * std::mem::size_of::<T> bytes on the cuda side`

agnesLeroy: Thanks a lot!

tmontaigu (approved):

PR #1100, opened 2024-04-23
chore(integer): reduce nb tests for big params

tmontaigu:


tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -29,27 +29,13 @@ use rand::prelude::ThreadRng;
...
-pub(crate) const NB_TESTS_SMALLER: usize = 1;
-
 #[cfg(not(tarpaulin))]
 pub(crate) const NB_CTXT: usize = 4;
 #[cfg(tarpaulin)]
 pub(crate) const NB_CTXT: usize = 2;
 
 #[cfg(not(tarpaulin))]
-pub(crate) const NB_TESTS_UNCHECKED: usize = NB_TESTS;
+pub(crate) const NB_TESTS_UNCHECKED: usize = 30;
```
mayeul-zama, line 38: Can't this be also be replaced by a function?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -59,6 +45,54 @@ pub(crate) const MAX_VEC_LEN: usize = 25;
...
+        if full_modulus >= 1 << 8 {
+            return 5;
+        }
+
+        // >= 6 bits (3_3)
+        if full_modulus >= 1 << 6 {
+            return 15;
+        }
+
+        30
```
mayeul-zama, line 70: Looks much better
How did you come up with these numbers?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -59,6 +45,54 @@ pub(crate) const MAX_VEC_LEN: usize = 25;
...
+/// meant for test where the function tested is more expensive
+pub(crate) const fn nb_tests_smaller_for_params(params: PBSParameters) -> usize {
+    let full_modulus = params.message_modulus().0 * params.carry_modulus().0;
+
+    if cfg!(tarpaulin) {
+        1
+    } else {
+        // >= 8 bits (4_4)
+        if full_modulus >= 1 << 8 {
+            return 2;
```
mayeul-zama, line 84: Why not 1?

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
```diff
@@ -17,6 +17,7 @@ use std::usize;
 /// Outputs are not required to be unique
 /// Input values are not required to span all possible values that
 /// ` ct` could hold.
+#[derive(Debug)]
```
mayeul-zama, line 20: Looks unrelated

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
```diff
@@ -96,7 +96,7 @@ impl ServerKey {
             max_output_value.ceil_ilog2()
         };
         let num_blocks_to_represent_values =
-            num_bits_to_represent_output_value.div_ceil(num_bits_in_message);
+            num_bits_to_represent_output_value.div_ceil(num_bits_in_message) + 1;
```
mayeul-zama, line 100: I don't understand this fix
Maybe add a comment to explainthis `+1`

mayeul-zama (commented):
Thanks for this PR, looks really useful for CI times!

I'm surprised to see many more signed files changed compared to the unsigned.
Is this expected?

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
tmontaigu: In the first commit of this PR, there is a fix, to find the data that triggered the error I added prints of the luts used in the assert in case the thing fails, and kept these in case it fails again later

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
tmontaigu: Actually the +1 should be on the ceil_ilog2 : `ceil_ilog2()` replaced by `ilog2() + 1`

The get the number of bits that allows representing a number  we need the `log2 + 1` or `ceil_ilog2 + u32::from(value.is_power_of_two)`:

with ilog2 + 1:
- value 3 (0b11) -> ilog2 =  1, + 1 = 2 => 2 bits to represent the value
- value 4 (0b100) -> ilog2 =  2, + 1 = 3 => 3 bits to represent the value

with ceil_ilog2 + u32::from(value.is_power_of_two):
- value 3 (0b11) -> ceil_ilog2 =  2, + 0 = 2 => 2 bits to represent the value
- value 4 (0b100) -> ceil_ilog2 =  2, + 1 = 3 => 2 bits to represent the value

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
tmontaigu: Arbitrary choice somewhat based on the timing of the PBS :grin:

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
tmontaigu: I Know some test splits the number of tests, so I wanted a value bigger than one in hope we still have a somewhat ok coverage for 4_4

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
tmontaigu: Yes it probably can, Should it be done in this PR or one after ?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
mayeul-zama: I see
IMO it would be cleaner to have these split tests do a number of tests equal to `2*nb_tests` rather `2*(nb_tests/2)`

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
tmontaigu: I did it in this pr

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
mayeul-zama: Or `(max_output_value+1).ceil_ilog2()` which IMO is more clear (and equivalent I believe)

tmontaigu: with `(max_output_value+1).ceil_ilog2()` we would have to handle the overflow case when max_output == CLEAR::MAX

tmontaigu: unless we change
```rust
let num_bits_to_represent_output_value = if max_output_value == Clear::ZERO {
    1
} else {
    max_output_value.ilog2() + 1
};
```
to 

```rust
let (max_output_value_plus_1, did_overflow) = max_output_value.overflowing_add(1);
let num_bits_to_represent_output_value = if did overflow {
      Clear::BITS
  } else {
      max_output_value_plus_1.ceil_ilog2()
  };
```

mayeul-zama: Or 
```
let num_bits_to_represent_output_value = if max_output_value == Clear::MAX {
    Clear::BITS
} else {
   (max_output_value+1).ceil_ilog2()
};
```

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
mayeul-zama: Btw for tarpaulin, 1 is not ok just above as we may not take some paths

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_vector_find.rs
```diff
@@ -405,15 +407,15 @@ where
...
 
     let mut lut = Vec::with_capacity(modulus as usize);
-    for i in 0..NB_TESTS {
+    for i in 0..nb_tests {
         lut.clear();
 
         let clear = rng.gen_range(0..modulus);
         let num_values = rng.gen_range(1..MAX_VEC_LEN) as usize;
-        let occurrence_count = if i < HALVED_NB_TEST { 0 } else { 1 };
+        let occurrence_count = if i < halved_nb_test { 0 } else { 1 };
```
mayeul-zama, line 418: I propose replacing the by 
```rust
for i in 0..2*nb_tests {

let occurrence_count = if i < nb_tests { 0 } else { 1 };
```

to fix the case nb_tests=1 when tarpaulin is enabled

mayeul-zama (commented):
LGTM

Last thing: `nb_tests=1` for tarpaulin problematic as coverage of some tests (with halved_nb_test) expect `nb_tests>2==0`

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_vector_find.rs
tmontaigu: I don't really like this because it means doubling the test for non tarpaulin cases and these tests are already long (1h and more) so doubling them is just a no go

What we can do is `nb_tests.min(2)` to ensure 2 tests

tmontaigu: And since the code is FHE, there is no branching between the two cases, so even with only one nb_tests all code path will be executed

mayeul-zama: Ok, all good then!

mayeul-zama (approved):
Thanks!

tmontaigu:
When tarpaulin is fixed, we will be to produce coverage and see where we are missing some (either because of nb_tests for tarpauling being 1) or because of the tests themselves

PR #1105, opened 2024-04-25
feat(zk): impl CanonicalSerialize/Deserialize

tmontaigu:
This is to allow specifying whether data should be compressed as compression and validation adds a very signigicant overhead especially in wasm where deserialization goes from 6 min to 450ms

### DIFF
--- a/tfhe-zk-pok/src/curve_api/bls12_381.rs
+++ b/tfhe-zk-pok/src/curve_api/bls12_381.rs
@@ -36,7 +36,17 @@ fn bigint_to_bytes(x: [u64; 6]) -> [u8; 6 * 8] {
 mod g1 {
     use super::*;
 
-    #[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+    #[derive(
+        Copy,
+        Clone,
+        PartialEq,
+        Eq,
+        Serialize,
+        Deserialize,
+        Hash,
+        CanonicalSerialize,
+        CanonicalDeserialize,
+    )]
     #[repr(transparent)]
     pub struct G1 {
         #[serde(serialize_with = "ark_se", deserialize_with = "ark_de")]
@@ -169,7 +179,17 @@ mod g1 {
 mod g2 {
     use super::*;
 
-    #[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+    #[derive(
+        Copy,
+        Clone,
+        PartialEq,
+        Eq,
+        Serialize,
+        Deserialize,
+        Hash,
+        CanonicalSerialize,
+        CanonicalDeserialize,
+    )]
     #[repr(transparent)]
     pub struct G2 {
         #[serde(serialize_with = "ark_se", deserialize_with = "ark_de")]
--- a/tfhe-zk-pok/src/curve_api/bls12_446.rs
+++ b/tfhe-zk-pok/src/curve_api/bls12_446.rs
@@ -36,7 +36,17 @@ fn bigint_to_bytes(x: [u64; 7]) -> [u8; 7 * 8] {
 mod g1 {
     use super::*;
 
-    #[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+    #[derive(
+        Copy,
+        Clone,
+        PartialEq,
+        Eq,
+        Serialize,
+        Deserialize,
+        Hash,
+        CanonicalSerialize,
+        CanonicalDeserialize,
+    )]
     #[repr(transparent)]
     pub struct G1 {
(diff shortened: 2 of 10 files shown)


tmontaigu:
@kc1212 @dd23
Given this, I wonder if we should or should not keep the serde serialization for ZkPok entities since, it would be better sur rely on ark's canonical serialize/deserialize to pass the `Compress` and `Validate` options explicitely and if one want to use serde in a struct that embeds a zk pok think he could use something like ` #[serde(deserialize_with = "deser_function")]` and specify its own default values for `Compress/Validate` in there

kc1212:
Is it possible to use `#[serde(deserialize_with = "deser_function")]` on G1, G2, etc when tfhe-zk-pok is used as a library? Or we'll have to write wrapper types?

tmontaigu:
Yes this should be possible
```rust
#[derive(serde::Deserialize, serde::Serialize)]
struct MyStruct<C: zk-pok::Curve> {
 #[serde(deserialize_with = "deser_function")]
 g1: C::G1
}
```

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -42,7 +75,9 @@ impl<T: ?Sized + IndexMut<usize>> IndexMut<usize> for OneBased<T> {
     }
 }
 
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[derive(
+    Clone, Debug, serde::Serialize, serde::Deserialize, CanonicalSerialize, CanonicalDeserialize,
+)]
```
IceTDrinker, line 80: tfhe wrapper types might need this as well to be able to serialize without compression, as going through only the serialize/deserialize will re use the default serialization mode

tfhe/js_on_wasm_tests/test-hlapi-signed.js
```diff
@@ -467,6 +469,11 @@ test('hlapi_compact_ciphertext_list_with_proof', (t) => {
     let crs = CompactPkeCrs.from_parameters(block_params, 2 + 32 + 1 + 256);
     let public_params = crs.public_params();
 
+    const compress = false; // We do'nt compress as its too slow on wasm
```
tmontaigu, line 472: Typo

tfhe-zk-pok/src/curve_api/bls12_381.rs
```diff
@@ -36,7 +36,17 @@ fn bigint_to_bytes(x: [u64; 6]) -> [u8; 6 * 8] {
...
+        Copy,
+        Clone,
+        PartialEq,
+        Eq,
+        Serialize,
+        Deserialize,
+        Hash,
+        CanonicalSerialize,
+        CanonicalDeserialize,
+    )]
```
IceTDrinker, line 49: is the serde attibute still required on the inner ? I'm really not sure how the ark serialization stuff is supposed to work, if you have the info from the doc e.g. don't hesitate to share

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -36,7 +36,17 @@ fn bigint_to_bytes(x: [u64; 7]) -> [u8; 7 * 8] {
...
+        Eq,
+        Serialize,
+        Deserialize,
+        Hash,
+        CanonicalSerialize,
+        CanonicalDeserialize,
+    )]
     #[repr(transparent)]
     pub struct G1 {
         #[serde(serialize_with = "ark_se", deserialize_with = "ark_de")]
```
IceTDrinker, line 52: is this required ? same question as the other curve file

tfhe-zk-pok/src/proofs/mod.rs
IceTDrinker: is the derive for serde traits required ?

IceTDrinker (commented):
Generally just questions on the required serde/ark bounds, it's not clear to me which structs needs which, otherwise the API changes look good

IceTDrinker:
> The places where a left the serde things are because it makes our life a bit easier by still allowing to use `#[derive(serde)]` on types that uses them (e.g the proofs which are embeded in compact ciphertext list, without this we would have to impl serde manually to call `CannonicalSerialize::serialize_with_mode` on each proof.
> 
> If we think its better to only have Cannonical serialize for zk primitive then this could be a backlog issue
> 
> I benched the serialization/deserialization of compact ciphertext list in wasm (which uses the derived serde, thus Compress::Yes, Validate::Yes) and I get 2ms to serialize and 14ms to deserialize, which is acceptable so keeping this easy solution is possible

Ok let's keep it that way for now and yes please for the backlog issue

IceTDrinker (approved):
Thanks !

IceTDrinker:
go for merge @tmontaigu

PR #1115, opened 2024-04-29
chore(ci): separate pcc and tests for gpu to allow faster merge

soonum:


IceTDrinker (approved):
Looks good, we'll need to update the required stuff once this is merged so that the speed up is indeed available

PR #1116, opened 2024-04-29
chore(ci): update toolchain to latest nightly

nsarlin-zama:
The latest nightly adds a compilation warning with ark's `derive(MontConfig)` macro. The fix is already integrated in their main branch, so the `allow` attribute will be removed when they release a new version.


tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -29,8 +29,8 @@ where
     ///
     /// The ptr must be valid for reads for len * std::mem::size_of::<T> bytes on
     /// the cuda side.
-    pub(crate) unsafe fn new(ptr: CudaPtr, len: usize) -> CudaSlice<'a, T> {
-        CudaSlice {
+    pub(crate) unsafe fn new(ptr: CudaPtr, len: usize) -> Self {
+        Self {
```
agnesLeroy, line 33: It looks good to me, but @tmontaigu you will know better than me if this is really ok :slightly_smiling_face:

tmontaigu: Yes it is

PR #1119, opened 2024-04-29
feat(gpu): add compatibility with cuda from package managers

nsarlin-zama:
### PR content/description
The `build.rs` of the cuda backend uses the default install path from the nvidia runfile to look for libraries. Linux distributions may install the cuda libs in another location. The pkg-config utility should be able to automatically the path of libraries if cuda was installed using a package manager. We can add "best effort" support to other installation methods using this tool.

This has been tested by running the pcc_gpu checks using the cuda provided by the package manager of Archlinux.


agnesLeroy (approved):
Thanks a lot for this @nsarlin-zama! :slightly_smiling_face:

PR #1121, opened 2024-05-02
feat(integer): add an eq_slice function

tmontaigu:
This adds eq_slices functions.

This function, compares to slices of radix ciphertexts and returns true if all pairs of element are equal.

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/vector_comparisons.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/vector_comparisons.rs
@@ -1,0 +1,108 @@
+use crate::integer::{BooleanBlock, IntegerRadixCiphertext, ServerKey};
+
+use rayon::prelude::*;
+
+impl ServerKey {
+    /// Compares two slices containing ciphertexts and returns an encryption of `true` if all
+    /// pairs are equal, otherwise, returns an encryption of `false`.
+    ///
+    /// - If slices do not have the same length, false is returned
+    /// - If at least one  pair (`lhs[i]`, `rhs[i]`) do not have the same number of blocks, false is
+    ///   returned
+    pub fn unchecked_all_eq_slices_parallelized<T>(&self, lhs: &[T], rhs: &[T]) -> BooleanBlock
+    where
+        T: IntegerRadixCiphertext,
+    {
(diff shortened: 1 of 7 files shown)


tfhe/src/integer/server_key/radix_parallel/vector_comparisons.rs
```diff
@@ -0,0 +1,108 @@
...
+            lhs.len(),
+            rhs.len()
+        );
+        if lhs
+            .iter()
+            .zip(rhs.iter())
+            .any(|(l, r)| l.blocks().len() != r.blocks().len())
+        {
+            panic!("All pairs of element in the lhs and rhs slices must have the same number of blocks");
+        }
```
IceTDrinker, line 28: the different length is a trivial case of False IMO

the other loop it could also be a trivial case of false potentially given some pairs are not of the same type/under the same modulus, WDYT ?

IceTDrinker (commented):
One small question but otherwise looks good

tfhe/src/integer/server_key/radix_parallel/vector_comparisons.rs
tmontaigu: Yes non matching len for slices should be  trvial false I agree

For the mismatched len between pairs, I'm not so sure, it depends on how strict we want to be, because technically, two unsigend radix with different could encrypt the same value.

But returning false might be better than panic, so we could return false and document that

IceTDrinker: I think I would document, worst case the HL would not suffer from it as it's going to use "uniform" arrays right ?

so to have something peculiar at the integer level might be kind of ok, not sure if it follows the rest of the design on mismatched sizes though

IceTDrinker (approved):
Thanks !

tmontaigu:
@slab-ci cpu_test

soonum:
@tmontaigu  now you have to use the workflow dispatch method to run tests on your development branch. :wink:

IceTDrinker:
https://github.com/zama-ai/tfhe-rs/pull/1121/checks?check_run_id=24637568068

you should be able to re run the jobs from here

be sure to make "re run all" to have the instance setup

PR #1123, opened 2024-05-02
refactor(tfhe): Allow CompactCiphertextList to store heterogeneous types

tmontaigu:
This refactors the integer's CompactCiphertextList to allow storing usngined, signed (not with necessarily the same number of blocks) and booleans in a single comapct list.

This is better as its more flexible and allows for better compression by not forcing to use a list per data type. This is especially interessing with zero-knowledge proofs as they are expensive to compute.

This also removes 'typed' CompactList and Comapct types from the hlapi.


tfhe/c_api_tests/test_high_level_compact_list.c
```diff
@@ -0,0 +1,129 @@
...
+    compact_ciphertext_list_builder_destroy(builder);
+  }
+
+  // Now we can expand values
+  FheUint32 *a = NULL;
+  FheInt64 *b = NULL;
+  FheBool *c = NULL;
+  FheUint2 *d = NULL;
+  {
+    FheTypes type = Fhe_Bool;
```
IceTDrinker, line 62: might be a tad error prone to have FheBool be the actual type and Fhe_Bool an enum value I would think ?

tfhe/c_api_tests/test_high_level_integers.c
```diff
@@ -625,8 +608,8 @@ int main(void) {
     assert(ok == 0);
     ok = uint8_serialization(client_key);
     assert(ok == 0);
-    ok = uint8_safe_serialization(client_key, server_key);
-    assert(ok == 0);
+    //    ok = uint8_safe_serialization(client_key, server_key);
+    //    assert(ok == 0);
```
IceTDrinker, line 612: keep ? remove ?

tfhe/c_api_tests/test_high_level_integers.c
```diff
@@ -371,9 +371,8 @@ int uint8_public_key(const ClientKey *client_key, const PublicKey *public_key) {
 
 int uint8_safe_serialization(const ClientKey *client_key, const ServerKey *server_key) {
   int ok;
-  CompactFheUint8 *lhs = NULL;
-  CompactFheUint8 *deserialized_lhs = NULL;
```
IceTDrinker, line 375: the notion of compact ciphertext disappears in C ? could it be a thin wrapper around the heterogeneous list perhaps ? Not sure if it's worth it

IceTDrinker (commented):
One question first : do we drop the compact ciphertext type entirely ?

it looks a bit less ergonomic to use, I was thinking that maybe a thin wrapper around the generic list could be of interest

tfhe/c_api_tests/test_high_level_compact_list.c
tmontaigu: Should we have Type_Fhe_Bool ?

tfhe/c_api_tests/test_high_level_integers.c
tmontaigu: It disapears totally from the Lib

tfhe/c_api_tests/test_high_level_compact_list.c
IceTDrinker: What's the usual convention for C API enum names ? this one I'm guessing ?

Then yeah I guess Type_Fhe_Bool or Type_FheBool

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -1,42 +1,380 @@
...
     }
 }
 
 impl CompactCiphertextList {
-    pub fn expand_one<T: IntegerRadixCiphertext>(&self) -> T {
-        let mut blocks = self.ct_list.expand();
-        blocks.truncate(self.num_blocks_per_integer);
-        T::from(blocks)
+    pub fn is_packed(&self) -> bool {
+        self.ct_list.degree.get() > self.ct_list.message_modulus.0
```
IceTDrinker, line 362: properly set at encryption ? I believe some encryption apply the message modulus on the inputs of blocks

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -46,44 +384,52 @@ impl CompactCiphertextList {
...
+
+        Ok(CompactCiphertextListExpander::new(
+            expanded_blocks,
+            self.info.clone(),
+        ))
+    }
+
+    /// This simply expands the list without splitting messages if they were packed
+    /// caller should check using [Self::is_packed] to know if data is packed
+    pub(crate) fn expand_without_key(&self) -> CompactCiphertextListExpander {
```
IceTDrinker, line 430: expand_without_packing ? not sure the name is better than this one

tfhe/src/safe_deserialization.rs
```diff
@@ -159,76 +159,47 @@ mod test_integer {
...
 
         let public_key = CompactPublicKey::new(&client_key);
 
         let msg = [27u8, 10, 3];
 
-        let ct_list = CompactFheUint8List::try_encrypt(&msg, &public_key).unwrap();
+        let ct_list = CompactCiphertextList::builder(&public_key)
+            .push(27u8)
+            .push(10u8)
+            .push(3u8)
```
IceTDrinker, line 180: question, does it try to fit the type in the smallest FheUint type ?

tfhe/src/shortint/public_key/compact.rs
```diff
@@ -27,16 +27,25 @@ pub struct CompactPublicKey {
...
+        encryption_modulus <= full_modulus,
+        "Encryption modulus cannot exceed the plaintext modulus"
+    );
+
     message_iter.map(move |message| {
         //The delta is the one defined by the parameters
-        let delta = (1_u64 << 63) / (message_modulus * carry_modulus);
+        let delta = (1_u64 << 63) / (full_modulus);
 
         //The input is reduced modulus the message_modulus
```
IceTDrinker, line 47: comment is wrong

IceTDrinker (commented):
Haven't gone in the very detailed

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -1,42 +1,380 @@
...
+
+impl<T> Expandable for T
+where
+    T: IntegerRadixCiphertext,
+{
+    fn from_expanded_blocks(blocks: &[Ciphertext], kind: DataKind) -> crate::Result<Self> {
+        match (kind, T::IS_SIGNED) {
+            (DataKind::Unsigned(_), false) | (DataKind::Signed(_), true) => {
+                Ok(T::from_blocks(blocks.to_vec()))
+            }
```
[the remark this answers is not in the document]
tmontaigu: No because the integer API is not typed with regards to the size of the ciphertexts as the HLAPI is.

So the user has to extract the correct kind, and if the circuit really requires specific sizes of radixes, the user
can check the number of blocks in the resulting ciphertext

tfhe/src/high_level_api/integers/signed/tests.rs
```diff
@@ -578,66 +576,6 @@ fn test_compact_public_key_small() {
...
-            crate::shortint::parameters::classic::compact_pk
-            ::PARAM_MESSAGE_2_CARRY_2_COMPACT_PK_PBS_KS,
-            None,
-        )
-        .build();
-    test_compact_public_key_list(config);
-}
-
-fn test_compact_public_key_list(config: Config) {
-    let (client_key, server_key) = generate_keys(config);
```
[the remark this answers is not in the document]
tmontaigu: Yes there is a test where we use the new heterogeneous list

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -0,0 +1,485 @@
...
+        &self,
+        sks: &crate::ServerKey,
+    ) -> crate::Result<CompactCiphertextListExpander> {
+        self.0
+            .expand(sks.key.pbs_key())
+            .map(|inner| CompactCiphertextListExpander { inner })
+    }
+
+    pub fn expand(&self) -> crate::Result<CompactCiphertextListExpander> {
+        if self.0.is_packed() {
```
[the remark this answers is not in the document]
tmontaigu: Yes it should, I will add a to the test 2 FheBool next to another

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -1,42 +1,380 @@
...
+        public_params: &CompactPkePublicParams,
+        load: ZkComputeLoad,
+    ) -> crate::Result<ProvenCompactCiphertextList> {
+        let msg_mod = self.pk.key.message_modulus().0 as u64;
+        let packed_messages = self
+            .messages
+            .chunks(2)
+            .map(|two_values| (two_values[1] * msg_mod) + two_values[0])
+            .collect::<Vec<_>>();
+        let ct_list = self.pk.key.encrypt_and_prove_slice(
```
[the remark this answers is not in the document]
tmontaigu: yep will add one

tfhe/src/integer/ciphertext/compact_list.rs
tmontaigu: Yes it is, In the test in integer::ciphertext::public_key::test we test both the case of building the list packed and unpacked

I will add an explicit call to this function in the test tho

tfhe/src/integer/ciphertext/compact_list.rs
tmontaigu: `expand_without_unpacking` you mean ? It might be better yes

IceTDrinker: well without unpacking reads a bit weird as it could mean "it is packed, but I don't unpack" not sure how to name those things 🙄

tfhe/src/safe_deserialization.rs
IceTDrinker: FheType::Uint32(u32) e.g. ? send the payload via the enum and the vairant gives the type ?

tfhe/src/integer/ciphertext/compact_list.rs
tmontaigu: The name `expand_without_unpacking` seems correct to me because this function will expand the data but never unpack (ie split 1 block into 2), this means that if the blocks are packed this function should not be called, the caller shall use `.is_packed()` before

tfhe/src/safe_deserialization.rs
tmontaigu: Hum it would not solve the problem for things like FheUint{2,4,6,10,12,..} because we don't have u2, u4, etc

Also we already have an enum `FheType` that does not have payload, it is used for example to query the type that is stored in a particular slot (index) of the compact list

And neither options of adding a payload to that enum (which then this payload won't always be used), or adding a different enum that has the payload seems particularly appealing

Also the `FheType` enum may be expanded if we add vectors/tensor types (unless we decide to change the way we describe our types)

tfhe/src/integer/ciphertext/compact_list.rs
IceTDrinker: fair enough

tfhe/src/safe_deserialization.rs
IceTDrinker: indeed, question is do we keep those "bastard" FheTypes though ?

IceTDrinker: ok, could be good to know if we can ditch those types

Understood for the dynamic types as fallback

static types sound like they could work but would likely require some macro to make it work, in any case for BC it's likely not useful as the C API would need an adaptation

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -1,42 +1,380 @@
...
+            .key
+            .encrypt_iter_with_modulus(packed_messaged_iter, msg_mod * msg_mod);
+
+        CompactCiphertextList {
+            ct_list,
+            info: self.info.clone(),
+        }
+    }
+
+    /// For zk proofs, message are always packed
```
IceTDrinker, line 199: this is the docstring I was referring, we have one not packed here with a docstring saying it's always packed, and then the other function later in the file

IceTDrinker: @tmontaigu don't know if you saw this, I think we have both functions and this docstring says zkpok are always packed

tmontaigu: Yep I will fix

PR #1124, opened 2024-05-03
Mz/rename multi bit

mayeul-zama:


IceTDrinker:
let's revisit this next quarter once the forward compat stuff is done

IceTDrinker (commented):
we should revisit this for the v1 as we will probably do a reset of data formats (and so parameter names including this one)

IceTDrinker:
can be closed in the mean time if you added the rename to 2.0 TODO

Issue #1108, opened 2024-04-27
(G)lweCiphertextList::from_fn methods


leonardohn:
**What is the problem you want to solve and can not with the current version?**
There is no straightforward way of arbitrarily initializing (G)lweCiphertextList structs.

**Describe the solution you'd like**
Implement methods named (G)lweCiphertextList::from_fn, that instead of receiving a fixed initialization value for the whole list, receives a closure with an usize parameter (the index) and returns the plaintext initialization value, so the user is able to initialize and encrypt the list arbitrarily, similar to [std::array::from_fn](https://doc.rust-lang.org/std/array/fn.from_fn.html) methods.

**Describe alternatives you've considered**
None.

**Additional context**
None.

IceTDrinker:
hello @leonardohn that's a very good suggestion indeed, thanks :)

IceTDrinker:
We don't have much bandwidth additional on our end so if you want to contribute you are welcome to do so by the way

leonardohn:
Hi! Thanks for the encouragement! I sent a PR addressing this issue, could you please review?

PR #1129, opened 2024-05-05
Implement (G)lweCiphertextList::from_fn/from_elem methods

leonardohn:
closes: #1108

### PR content/description

The first commit adds `from_fn` methods for `(G)lweCiphertextList`, enabling the user to directly initialize ciphertext lists using a closure. This closure must receive an `usize` as argument (the index) and return a scalar value that will initialize the list at the provided index. The closure is the last function parameter so that `rustfmt` can properly format the function call without skipping additional lines if the provided arguments can fit into a single line.

Following the name convention from `from_fn`, the second commit adds `from_elem` methods that behave similarly to the previous `new` methods, except for following a different parameter order to match `from_fn` parameters. Both `new` methods were marked as deprecated and internally invoke `from_elem` methods.

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
+++ b/tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
@@ -53,13 +53,33 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> GlweCiphertextList
     /// let ciphertext_count = GlweCiphertextCount(2);
     /// let ciphertext_modulus = CiphertextModulus::new_native();
     ///
-    /// // Create a new GlweCiphertextList
+    /// // Create a new GlweCiphertextList and fill it using copies of a single element
     /// let glwe_list = GlweCiphertextList::new(
     ///     0u64,
     ///     glwe_size,
     ///     polynomial_size,
     ///     ciphertext_count,
     ///     ciphertext_modulus,
+    /// );
+    ///
+    /// assert_eq!(glwe_list.glwe_size(), glwe_size);
+    /// assert_eq!(glwe_list.polynomial_size(), polynomial_size);
+    /// assert_eq!(glwe_list.glwe_ciphertext_count(), ciphertext_count);
+    /// assert_eq!(glwe_list.ciphertext_modulus(), ciphertext_modulus);
+    ///
+    /// // Alternatively, create a new GlweCiphertextList and fill it using a function
+    /// let glwe_list = GlweCiphertextList::from_fn(
+    ///     glwe_size,
+    ///     polynomial_size,
+    ///     ciphertext_count,
+    ///     ciphertext_modulus,
+    ///     |i, j| {
+    ///         // The `i` value represents the index in the list being filled;
+    ///         // The `j` value represents the index in the ciphertext being filled;
+    ///         // In this example, for every index pair `(i, j)`, we fill the
+    ///         // corresponding value using the formula `i + j`
+    ///         (i + j) as u64
+    ///     },
     /// );
     ///
     /// assert_eq!(glwe_list.glwe_size(), glwe_size);
@@ -174,6 +194,36 @@ impl<Scalar: UnsignedInteger> GlweCiphertextListOwned<Scalar> {
             polynomial_size,
             ciphertext_modulus,
         )
+    }
+
+    /// Allocate memory and create a new owned [`LweCiphertextList`], where each element
+    /// is provided by the `fill_with` function, invoked for each consecutive index.
+    ///
+    /// # Note
+    ///
+    /// This function allocates a vector of the appropriate size and wraps it in the appropriate
+    /// type. If you want to encrypt data in the list you need to use
+    /// [`crate::core_crypto::algorithms::encrypt_glwe_ciphertext_list`] or a variant working on
+    /// a single ciphertext at a time [`crate::core_crypto::algorithms::encrypt_glwe_ciphertext`] on
+    /// the individual ciphertexts in the list.
+    ///
+    /// See [`GlweCiphertextList::from_container`] for usage.
+    pub fn from_fn<F>(
+        glwe_size: GlweSize,
+        polynomial_size: PolynomialSize,
+        ciphertext_count: GlweCiphertextCount,
+        ciphertext_modulus: CiphertextModulus<Scalar>,
+        mut fill_with: F,
+    ) -> Self
+    where
+        F: FnMut(usize, usize) -> Scalar,
+    {
+        let ciphertext_size = glwe_ciphertext_size(glwe_size, polynomial_size);
+        let container: Vec<_> = (0..ciphertext_count.0)
+            .flat_map(move |i| (0..ciphertext_size).map(move |j| (i, j)))
+            .map(|(i, j)| fill_with(i, j))
+            .collect();
+        Self::from_container(container, glwe_size, polynomial_size, ciphertext_modulus)
     }
 }
 
--- a/tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
+++ b/tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
@@ -50,8 +50,26 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCiphertextList<
     /// let lwe_ciphertext_count = LweCiphertextCount(3);
     /// let ciphertext_modulus = CiphertextModulus::new_native();
     ///
-    /// // Create a new LweCiphertextList
+    /// // Create a new LweCiphertextList and fill it using copies of a single element
     /// let lwe_list = LweCiphertextList::new(0u64, lwe_size, lwe_ciphertext_count, ciphertext_modulus);
+    ///
+    /// assert_eq!(lwe_list.lwe_size(), lwe_size);
+    /// assert_eq!(lwe_list.lwe_ciphertext_count(), lwe_ciphertext_count);
+    /// assert_eq!(lwe_list.ciphertext_modulus(), ciphertext_modulus);
+    ///
+    /// // Alternatively, create a new LweCiphertextList and fill it using a function
+    /// let lwe_list = LweCiphertextList::from_fn(
+    ///     lwe_size,
+    ///     lwe_ciphertext_count,
+    ///     ciphertext_modulus,
+    ///     |i, j| {
+    ///         // The `i` value represents the index in the list being filled;
+    ///         // The `j` value represents the index in the ciphertext being filled;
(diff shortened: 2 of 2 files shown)


IceTDrinker:
hello @leonardohn as you can see in this bot comment https://github.com/zama-ai/tfhe-rs/pull/1129#issuecomment-2094777506

you first need to sign a CLA before we can take a look to your PR

Cheers

IceTDrinker:
> @IceTDrinker I already signed the CLA, but the bot did not recognize for some reason.

Let me check on our end then

IceTDrinker:
@cla-bot check

tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
```diff
@@ -150,7 +150,35 @@ pub type GlweCiphertextListView<'data, Scalar> = GlweCiphertextList<&'data [Scal
...
+    /// is a clone of the `fill_with` element.
+    ///
+    /// # Note
+    ///
+    /// This function is deprecated. Use [`GlweCiphertextList::from_elem`] instead.
+    #[inline(always)]
+    #[deprecated(
+        since = "0.6.1",
+        note = "use GlweCiphertextList::from_elem method instead"
+    )]
```
IceTDrinker, line 163: I would keep new, not deprecated it and not add the from_elem method

tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
```diff
@@ -161,19 +189,43 @@ impl<Scalar: UnsignedInteger> GlweCiphertextListOwned<Scalar> {
...
+    ///
+    /// # Note
+    ///
+    /// This function allocates a vector of the appropriate size and wraps it in the appropriate
+    /// type. If you want to encrypt data in the list you need to use
+    /// [`crate::core_crypto::algorithms::encrypt_glwe_ciphertext_list`] or a variant working on
+    /// a single ciphertext at a time [`crate::core_crypto::algorithms::encrypt_glwe_ciphertext`] on
+    /// the individual ciphertexts in the list.
+    ///
+    /// See [`GlweCiphertextList::from_container`] for usage.
```
IceTDrinker, line 215: I don't see a change in the from_container docstring to demonstrate usage ?

tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
```diff
@@ -154,17 +175,41 @@ impl<Scalar: UnsignedInteger> LweCiphertextListOwned<Scalar> {
...
+    ///
+    /// # Note
+    ///
+    /// This function allocates a vector of the appropriate size and wraps it in the appropriate
+    /// type. If you want to encrypt data you need to use
+    /// [`crate::core_crypto::algorithms::encrypt_lwe_ciphertext_list`] or its parallel variant
+    /// [`crate::core_crypto::algorithms::par_encrypt_lwe_ciphertext_list`] using this list as
+    /// output.
+    ///
+    /// See [`LweCiphertextList::from_container`] for usage.
```
IceTDrinker, line 200: I don't see an update to the from_container docstring to demonstrate usage ?

tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
```diff
@@ -143,7 +143,28 @@ pub type LweCiphertextListView<'data, Scalar> = LweCiphertextList<&'data [Scalar
...
+    ///
+    /// # Note
+    ///
+    /// This function is deprecated. Use [`LweCiphertextList::from_elem`] instead.
+    #[inline(always)]
+    #[deprecated(
+        since = "0.6.1",
+        note = "use LweCiphertextList::from_elem method instead"
+    )]
+    pub fn new(
```
IceTDrinker, line 157: I would keep new, not deprecated it and not add the from_elem method

IceTDrinker (changes requested):
Just one small thing, I would keep the new method and not add from_elem, I don't see such a function in the rust std lib, the from_fn looks ok

Deprecation might be a bit overkill for now as we are not stable yet though it is certainly nice attention to detail.

leonardohn:
> Just one small thing, I would keep the new method and not add from_elem, I don't see such a function in the rust std lib, the from_fn looks ok

I thought renaming the function would be a good addition as it describes how the container is being generated -- from a sample element, and it also justifies the parameter order swap without introducing breaking changes.

This functionality is available in the vec! macro, although it doesn't have a name there. Anyway, I added this in a separare commit so it would be easier to revert in that case.

tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
leonardohn: I added an usage example now.

tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
leonardohn: I added an usage example now.

leonardohn:
@IceTDrinker that's right, it currently provides the raw index directly from the container, so the raw mask/body values can be manually set. I'm not sure if that's the best approach for the end user, though. If you agree, I can change `from_fn` to provide the list index only and then the user provides a full (G)lweCiphertext to initialize. What do you prefer?

leonardohn:
In fact, to avoid double allocations I believe it is better to provide a mutable view of a (G)lweCiphertext, right?

IceTDrinker:
For the list type I was thinking about it and likely having the current index inside the list and then the index of the given entity is probably better

IceTDrinker:
> In fact, to avoid double allocations I believe it is better to provide a mutable view of a (G)lweCiphertext, right?

in what part of the code sorry ?

leonardohn:
> in what part of the code sorry ?

There are four possibilities for the closure:

1. Provide the container index and receive a scalar value -- `Fn(usize) -> Scalar`;
    - This is the way it is implemented now and the simplest, not friendly for most of the users;
2. Provide the list and the entity index and receive a scalar value -- `Fn(usize, usize) -> Scalar`;
    - This is the way you proposed, it just requires applying div/mod over the container index;
3. Provide the list index and receive a owned ciphertext -- `Fn(usize) -> (G)lweCiphertextOwned<Scalar>`;
    - The most user friendly option but requires copying and double allocating;
4. Provide the list index and a mutable ciphertext reference -- `Fn(usize, (G)lweCiphertextMutView<Scalar>) -> ()`;
    - This avoids the problems above, but introduces mutability and the values must be default initialized;

Which one do you prefer?

leonardohn:
> do you currently have a use case with lists ?

Yes. It's not doable in the current state of `tfhe-rs`, but I'm planning to upstream all the features I need. Just to contextualize, [this](https://arxiv.org/abs/2403.20190) is what I'm trying to implement.

> 3 and 4 feel like they should be already doable by first creating the list and using the iterators or getters to fill them as the users sees fit ?

That's right, I didn't consider this possibility. So I think we should stick with option 2.

leonardohn:
> Thanks for sharing the use case, did you try Concrete-ML for this maybe ?

We introduced a few cryptographic primitives to achieve those results, so I believe it would be better to implement them over `tfhe-rs`, as they might benefit other general use cases. Despite having "neural networks" in the name, this algorithm does not involve traditional NN building blocks.

IceTDrinker:
> > Thanks for sharing the use case, did you try Concrete-ML for this maybe ?
> 
> We introduced a few cryptographic primitives to achieve those results, so I believe it would be better to implement them over `tfhe-rs`, as they might benefit other general use cases. Despite having "neural networks" in the name, this algorithm does not involve traditional NN building blocks.

Understood !

Not sure about what we would put in TFHE-rs, is there a peer reviewed paper you could share on those ?

leonardohn:
> Not sure about what we would put in TFHE-rs, is there a peer reviewed paper you could share on those ?

We are currently in process of peer-reviewing this, but I believe there are two major features that could benefit `tfhe-rs` in general:

- Secret `(G)lweCiphertextList` indexing based on a binary encoded `GgswCiphertextList`, which is a generalisation of the WoPBS procedure for ciphertext lists;
- Secret single-position `(G)lweCiphertextList` initialization based on a binary encoded `GgswCiphertextList`, using our proposed `CDEMUX` tree procedure -- this technique together with element-wise summation allows secret arbitrary modification of ciphertext lists;

Using these two features together, one is able to manipulate arbitrarily long ciphertext lists using ciphertext indices.

IceTDrinker:
The doctests are failing run:

```
RUSTFLAGS="-C target-cpu=native" cargo +stable test --doc --profile devo         --features=x86_64-unix -p tfhe@0.6.0 -- from_container
```

tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
```diff
@@ -67,6 +67,26 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> GlweCiphertextList
...
+    ///     glwe_size,
+    ///     polynomial_size,
+    ///     ciphertext_count,
+    ///     ciphertext_modulus,
+    ///     |i, j| {
+    ///         // The `i` value represents the index in the list being filled;
+    ///         // The `j` value represents the index in the ciphertext being filled;
+    ///         // In this example, for every index pair `(i, j)`, we fill the
+    ///         // corresponding value using the formula `i + j`
+    ///         (i + j as u64)
```
IceTDrinker, line 81: likely fix is (i + j) as u64

tfhe/src/core_crypto/entities/lwe_ciphertext_list.rs
```diff
@@ -50,13 +50,31 @@ impl<Scalar: UnsignedInteger, C: Container<Element = Scalar>> LweCiphertextList<
...
+    /// let lwe_list = LweCiphertextList::from_fn(
+    ///     lwe_size,
+    ///     lwe_ciphertext_count,
+    ///     ciphertext_modulus,
+    ///     |i, j| {
+    ///         // The `i` value represents the index in the list being filled;
+    ///         // The `j` value represents the index in the ciphertext being filled;
+    ///         // In this example, for every index pair `(i, j)`, we fill the
+    ///         // corresponding value using the formula `i + j`
+    ///         (i + j as u64)
```
IceTDrinker, line 70: same here

tfhe/src/core_crypto/entities/glwe_ciphertext_list.rs
leonardohn: I forgot about the precedence, it should be working now.

IceTDrinker (approved):
Thanks !

PR #1131, opened 2024-05-06
feat(integer): improve scalar_add

tmontaigu:
By using the fact that one input is scalar
we can speed up computation of scalar_add by packing the blocks.

As the main loop as a log2 complexity, it means we just do one less
iteration, but its enough to bring a 15% to 20% improvements.

In other words a 64bits scalar add is as fast as a 32bit add

Same idea is applicable to scalar_sub and overflowing variants


tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
```diff
@@ -255,7 +257,136 @@ impl ServerKey {
...
+                        // That result is shifted by one bit, so the extract_message will need
+                        // to be different
+                        let v = if low_r >= self.message_modulus().0 as u64 {
+                            2
+                        } else if low_r == self.message_modulus().0 as u64 - 1 {
+                            1
+                        } else {
+                            0
+                        };
+
```
IceTDrinker, line 328: not sure I understand what the v does here 🤔

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
```diff
@@ -255,7 +257,136 @@ impl ServerKey {
...
+                .for_each(|(ct_block, new_value)| *ct_block = new_value);
+
+            // Finally add the carries and then, do the message extract
+            ct.blocks_mut()
+                .chunks_mut(2)
+                .zip(carries.iter())
+                .for_each(|(blocks, carry)| {
+                    self.key.unchecked_add_assign(&mut blocks[0], carry);
+                    self.key.unchecked_add_assign(&mut blocks[1], carry);
+                });
```
IceTDrinker, line 369: this confuses me 😵‍💫 let's maybe discuss quickly when you have some time so that I understand the algorithm better ?

tmontaigu:
What changed: The new code was moved into a pub(crate) function to be able to test it (and be sure that is it tested and not skipped because due to num_block/num_threads a different algorithm could be selected)

IceTDrinker (approved):
Thanks !

PR #1138, opened 2024-05-13
feat(gpu): implement update_info for div_rem

bbarbakadze:
### PR content/description

implement `update_info` for cuda backend  div_rem operation


tfhe/src/integer/gpu/ciphertext/info.rs
```diff
@@ -128,6 +128,22 @@ impl CudaRadixCiphertextInfo {
...
+        Self {
+            blocks: self
+                .blocks
+                .iter()
+                .map(|left| CudaBlockInfo {
+                    degree: Degree::new(left.message_modulus.0 - 1),
+                    message_modulus: left.message_modulus,
+                    carry_modulus: left.carry_modulus,
+                    pbs_order: left.pbs_order,
+                    noise_level: left.noise_level + NoiseLevel::NOMINAL,
```
tmontaigu, line 141: After division, if done as in CPU, noise level should be NOMINAL

tmontaigu (approved):

PR #1140, opened 2024-05-13
feat(integer): add contains_sub_slice

tmontaigu:
This method returns a Boolean block that tells whether a slice of radix contains a sub slice.


IceTDrinker (approved):
Thanks !

PR #1142, opened 2024-05-14
chore(ci): fix slack notification in gpu benchmarks

soonum:
Slack notification GitHub action need Docker to be installed on the system before the action is effectively used.
To mitigate this issue, we set the step it's in own job. The previous job takes care of installing Docker along with other dependencies needed to run the benchmarks.

### DIFF
--- a/.github/workflows/core_crypto_gpu_benchmark.yml
+++ b/.github/workflows/core_crypto_gpu_benchmark.yml
@@ -57,7 +57,7 @@ jobs:
           sudo apt install ca-certificates curl
           sudo install -m 0755 -d /etc/apt/keyrings
           sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
-          sudo chmod a+r /etc/apt/keyrings/docker.asc  
+          sudo chmod a+r /etc/apt/keyrings/docker.asc
           echo \
           "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
           $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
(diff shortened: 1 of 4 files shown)


IceTDrinker (commented):
Does that make any difference ? 🤔

.github/workflows/core_crypto_gpu_benchmark.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           sudo apt install ca-certificates curl
           sudo install -m 0755 -d /etc/apt/keyrings
           sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
-          sudo chmod a+r /etc/apt/keyrings/docker.asc  
+          sudo chmod a+r /etc/apt/keyrings/docker.asc
```
IceTDrinker, line 60: is that the current recommended way of installing docker on Linux ?

soonum: It's from the Docker documentation.

soonum:
> Does that make any difference ? 🤔

Yes because the system does not start fresh between jobs, however the workspace does.

.github/workflows/core_crypto_gpu_benchmark.yml
```diff
@@ -161,13 +161,18 @@ jobs:
...
+    steps:
+      - name: Send message
         uses: rtCamp/action-slack-notify@4e5fb42d249be6a45a298f3c9543b111b02f7907
         env:
-          SLACK_COLOR: ${{ job.status }}
-          SLACK_MESSAGE: "PBS GPU benchmarks finished with status: ${{ job.status }}. (${{ env.ACTION_RUN_URL }})"
+          SLACK_COLOR: ${{ needs.cuda-core-crypto-benchmarks.result }}
+          SLACK_MESSAGE: "PBS GPU benchmarks finished with status: ${{ needs.cuda-core-crypto-benchmarks.result }}. (${{ env.ACTION_RUN_URL }})"
 
   teardown-instance:
```
IceTDrinker, line 177: teardown will run at the same time as the slack notifcation here right ?

IceTDrinker: which means it can be randomly run first or after the notification, meaning ~half the time we likely have a termination without notification

soonum: Oh you're totally right. Fixing

IceTDrinker: for all workflows where there is this pattern

IceTDrinker (approved):
Thanks !

PR #1145, opened 2024-05-14
Compute PBS's and KS's on multiple GPUs when possible

agnesLeroy:

### PR content/description
Also computes KS on multiple GPUs for apply_univariate_lut

### Check-list:


backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -121,21 +119,24 @@ void cuda_memcpy_async_gpu_to_gpu(void *dest, void *src, uint64_t size,
...
   check_cuda_error(cudaSetDevice(gpu_index));
+  // if (attr_src.device == attr_dest.device) {
   check_cuda_error(
       cudaMemcpyAsync(dest, src, size, cudaMemcpyDeviceToDevice, stream));
+  //} else {
+  //  printf("src device: %d, dest device: %d\n", attr_src.device,
+  //  attr_dest.device); check_cuda_error(cudaMemcpyPeerAsync(dest,
+  //  attr_dest.device, src,
+  //                                       attr_src.device, size, stream));
+  //}
```
agnesLeroy, line 139: Maybe I could uncomment those :thinking: or I just remove it altogether.

ci/slab.toml
```diff
@@ -56,6 +64,11 @@ workflow = "aws_tfhe_gpu_tests.yml"
 profile = "gpu-test"
 check_run_name = "GPU AWS Tests"
 
+[command.multi_gpu_test]
+workflow = "aws_tfhe_gpu_tests.yml"
+profile = "multi-gpu-test"
+check_run_name = "Multi GPU AWS Tests"
+
```
soonum, line 71: Using this won't work anymore since we spawn the instance directly in the workflow. Instead we should do a kind of matrix build with the single GPU profile and the multi one.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -32,11 +32,14 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
  *
  */
 template <typename Torus>
-__global__ void
-keyswitch(Torus *lwe_array_out, Torus *lwe_output_indexes, Torus *lwe_array_in,
-          Torus *lwe_input_indexes, Torus *ksk, uint32_t lwe_dimension_in,
-          uint32_t lwe_dimension_out, uint32_t base_log, uint32_t level_count,
-          int lwe_lower, int lwe_upper, int cutoff) {
+__global__ void keyswitch(Torus *__restrict__ lwe_array_out,
```
pdroalves, line 35: Why we have ```__restrict__```? Does it really bring any benefit?

backends/tfhe-cuda-backend/cuda/src/device.cu
pdroalves: I like the idea here. I always feel non-intuitive that we need two different methods to copy data within a device or between different devices. I would uncomment these.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -119,12 +119,33 @@ __host__ void integer_radix_apply_univariate_lookup_table_kb(
...
 
-  // Compute Keyswitch-PBS
-  cuda_keyswitch_lwe_ciphertext_vector(
-      streams[0], gpu_indexes[0], lut->tmp_lwe_after_ks,
-      lut->lwe_trivial_indexes, lwe_array_in, lut->lwe_indexes_in, ksk,
-      big_lwe_dimension, small_lwe_dimension, ks_base_log, ks_level,
-      num_radix_blocks);
+  /// If the number of radix blocks is lower than the number of GPUs, not all
+  /// GPUs will be active and there will be 1 input per GPU
+  auto active_gpu_count = std::min(num_radix_blocks, gpu_count);
```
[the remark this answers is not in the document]
pdroalves: A ```execute_multi_gpu_ks_pbs``` method that handles that would be much much better.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -457,7 +478,7 @@ void host_propagate_single_carry(cudaStream_t *streams, uint32_t *gpu_indexes,
         streams, gpu_indexes, gpu_count, cur_blocks, cur_blocks, prev_blocks,
         bsk, ksk, cur_total_blocks, luts_carry_propagation_sum);
 
-    cudaSetDevice(gpu_indexes[0]);
+    cuda_synchronize_stream(streams[0], gpu_indexes[0]);
```
pdroalves, line 481: Why do we need this synchronization?

pdroalves (commented):
Huge PR, but I think we are going in the right direction.

pdroalves:
As it is, if we try to run tests in parallel, we will end up with this error:

```
cuda error: peer access is already enabled /home/pedro/tfhe-rs/backends/tfhe-cuda-backend/cuda/src/device.cu 250
error: test failed, to rerun pass `-p tfhe --lib`
```

I've implemented a new type of error handling function to handle that. It will just ignore expected errors, as ```cudaErrorPeerAccessAlreadyEnabled```.

pdroalves:
Just pushed my changes and I noticed an issue with indexing. 

Suppose we have 4 GPUs and 16 inputs. Each GPU will handle 4 inputs. If input/output indexes are trivial, ```[0, 1, 2...]```, we can use the 4 elements of `lwe_input_indexes` or  `lwe_output_indexes` and we are fine. However, if the index array is anything else that won't work. For instance, with `[7,2,14, ...]` GPU 1 will try to access input 14. Today we are passing to GPU 1 an array with 12 elements, since we shift the input/output arrays addresses, so there is no position 14.

We have to refactor it. For instance, we can pass the same original input/output array to all devices and shift the index array instead.

agnesLeroy:
@pdroalves normally this branch should now be working. I'll trigger the H100 benchmarks to be sure they all get through.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -23,30 +24,72 @@ use tfhe_cuda_backend::cuda_bind::{
...
+
+        Self {
+            ptr: vec![ptr; 1],
+            lengths: vec![len; 1],
+            gpu_indexes: vec![streams.gpu_indexes[gpu_index as usize]; 1],
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn new_multi_gpu(len: usize, streams: &CudaStreams) -> Self {
```
tmontaigu, line 65: Add a docstring saying that this creates a CudaVec that holds memory of `len` elements on _each_ GPU that the `CudaStreams` are on

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -68,69 +111,133 @@ impl<T: Numeric> CudaVec<T> {
...
+                cuda_memcpy_async_to_gpu(
+                    self.get_mut_c_ptr(gpu_index),
+                    src.as_ptr().cast(),
+                    size as u64,
+                    streams.ptr[gpu_index as usize],
+                    streams.gpu_indexes[gpu_index as usize],
+                );
+            }
+        });
+    }
```
tmontaigu, line 219: Since this function only calls memset_async, doing a par_iter seems overkill as the actual setting of values  is not done in the memset_call.

Doing par_iter might only be interesting in the non async version

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -23,30 +24,72 @@ use tfhe_cuda_backend::cuda_bind::{
 /// memory, it is pretty close to a `Vec`. That being said, it only present a very very limited api.
 #[derive(Debug)]
 pub struct CudaVec<T: Numeric> {
-    ptr: CudaPtrMut,
-    len: usize,
-    gpu_index: u32,
+    pub ptr: Vec<*mut c_void>,
+    pub lengths: Vec<usize>,
+    pub gpu_indexes: Vec<u32>,
     _phantom: PhantomData<T>,
```
tmontaigu, line 30: So I get the point of having the CudaVec like this, it allows to easily have the same data (keyswitchkey, bootstrapping key) on all GPUs easily, but since you allow differents lengths it means that each data on each GPU could be totally different, won't it be too easy to misuse ?

agnesLeroy: Actually I'm not sure yet, but it's possible that in the end we will copy the ciphertexts to different GPUs. Since the number of GPUs does not necessarily divide the number of blocks, we would end up having a different work load on each GPU, and hence a different length for each. That's why I defined it this way.

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: No actually I need the par iter for the calls on the GPU to happen asynchronously, even if I'm using different cuda streams. On the C++ side we handle this with open mp.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -68,30 +115,66 @@ impl<T: Numeric> CudaVec<T> {
     ///
     /// - `streams` __must__ be synchronized to guarantee computation has finished, and inputs must
     ///   not be dropped until streams is synchronised
-    pub unsafe fn memset_async(&mut self, value: T, streams: &CudaStreams)
+    pub unsafe fn from_cpu_multi_gpu_async(src: &[T], streams: &CudaStreams) -> Self {
```
pdroalves, line 118: Why now we have a `_multi_gpu_` variant for memset and copy?

pdroalves (approved):
If the benchmarks are good, I'm satisfied with this PR.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -117,33 +118,44 @@ __device__ void mul_ggsw_glwe(Torus *accumulator, double2 *fft,
...
-          lwe_idx, max_shared_memory);
+#pragma omp parallel for num_threads(active_gpu_count)
+      for (uint i = 0; i < active_gpu_count; i++) {
+        int num_inputs_on_gpu =
+            get_num_inputs_on_gpu(input_lwe_ciphertext_count, i, gpu_count);
+        int gpu_offset =
+            get_gpu_offset(input_lwe_ciphertext_count, i, gpu_count);
+        auto d_lut_vector_indexes =
+            lut_vector_indexes + (ptrdiff_t)(gpu_offset);
+        cuda_programmable_bootstrap_lwe_ciphertext_vector_32(
```
pdroalves, line 149: We should be passing `lut_vector_indexes` to the PBS kernel and adjusting the offset in it.

Makefile
```diff
@@ -459,7 +459,7 @@ test_core_crypto_gpu: install_rs_build_toolchain
 .PHONY: test_integer_gpu # Run the tests of the integer module including experimental on the gpu backend
 test_integer_gpu: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
-		--features=$(TARGET_ARCH_FEATURE),integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key::
+		--features=$(TARGET_ARCH_FEATURE),integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key:: --test-threads=2
```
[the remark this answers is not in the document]
agnesLeroy: Have you tried on a p3.8xlarge instance? I restricted it because we were running out of memory... I'll try 16 then.

backends/tfhe-cuda-backend/cuda/src/utils/multi_gpu_helper.cu
```diff
@@ -1,52 +1,98 @@
...
-    int can_access_peer;
-    for (int i = 0; i < num_gpus; i++) {
-      cudaSetDevice(i);
-      for (int j = 0; i < num_gpus; i++) {
-        if (i == j)
-          break;
-        check_cuda_error(cudaDeviceCanAccessPeer(&can_access_peer, i, j));
-        cudaDeviceEnablePeerAccess(j, 0);
+    if (!p2p_enabled) {
+      p2p_enabled = true;
```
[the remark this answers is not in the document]
agnesLeroy: Should I open an issue to do this in a follow up PR? Or would you prefer that I try to do it here?

backends/tfhe-cuda-backend/cuda/src/integer/cmux.cuh
```diff
@@ -62,16 +61,17 @@ __host__ void host_integer_radix_cmux_kb(
 #pragma omp section
     {
       auto mem_true = mem_ptr->zero_if_true_buffer;
-      zero_out_if(&true_stream, gpu_indexes, gpu_count, mem_ptr->tmp_true_ct,
+      zero_out_if(&true_stream, gpu_indexes, 1, mem_ptr->tmp_true_ct,
```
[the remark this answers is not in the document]
agnesLeroy: I will do this in the next PR :slightly_smiling_face: https://github.com/zama-ai/tfhe-rs/pull/1205

backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
```diff
@@ -380,26 +379,26 @@ host_integer_div_rem_kb(cudaStream_t *streams, uint32_t *gpu_indexes,
 #pragma omp section
       {
         // interesting_divisor
-        trim_last_interesting_divisor_bits(&mem_ptr->sub_stream_1,
-                                           &gpu_indexes[0], 1);
+        trim_last_interesting_divisor_bits(&mem_ptr->sub_stream_1, gpu_indexes,
```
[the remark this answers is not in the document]
agnesLeroy: https://github.com/zama-ai/tfhe-rs/pull/1205

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
agnesLeroy: You're going to change this in your PR so I would leave it as it is.

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: The multi gpu variant does the memsest or copy on all GPUs, while with the single GPU one we can select on which GPU we want to copy/memset.

backends/tfhe-cuda-backend/cuda/src/utils/multi_gpu_helper.cu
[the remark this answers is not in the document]
agnesLeroy: Btw it should really not affect performance, this just happens when we create a CudaStreams at the beginning of a computation.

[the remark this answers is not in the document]
agnesLeroy: I've pushed a commit to the branch with the change, let me know if you think it's ok. I've checked benchmark results and they don't seem affected. Mul 64 bits -> 241 ms on 5 V100.

.github/workflows/integer_multi_gpu_full_benchmark.yml
```diff
@@ -0,0 +1,184 @@
+# Run all integer benchmarks on an instance with CUDA and return parsed results to Slab CI bot.
+name: Integer multi GPU full benchmarks
```
agnesLeroy, line 2: @soonum may I ask you to check this workflow please? I think you already checked the one for tests.

backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -414,93 +431,110 @@ template <typename Torus> struct int_radix_lut {
...
+
+      execute_scratch_pbs<Torus>(
+          streams[i], gpu_indexes[i], &gpu_pbs_buffer, params.glwe_dimension,
+          params.small_lwe_dimension, params.polynomial_size, params.pbs_level,
+          params.grouping_factor, num_blocks_on_gpu,
+          cuda_get_max_shared_memory(gpu_indexes[i]), params.pbs_type,
+          allocate_gpu_memory);
+      cuda_synchronize_stream(streams[i], gpu_indexes[i]);
+      buffer.push_back(gpu_pbs_buffer);
+    }
```
bbarbakadze, line 481: don't we need cpu multi-threading here?

bbarbakadze:
This is outstanding work!

backends/tfhe-cuda-backend/cuda/include/helper.h
```diff
@@ -1,10 +1,16 @@
...
 extern "C" {
 int cuda_setup_multi_gpu();
 }
 
-void multi_gpu_checks(uint32_t gpu_count);
+int get_active_gpu_count(int num_inputs, int gpu_count);
+
+int get_num_inputs_on_gpu(int total_num_inputs, int gpu_index, int gpu_count);
+
+int get_gpu_offset(int total_num_inputs, int gpu_index, int gpu_count);
```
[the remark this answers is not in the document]
agnesLeroy: I will rename the helper.h file :+1: 

Regarding your question about the number of threads used for testing, actually I have to restrict it otherwise we run out of memory on the GPU. Since David is working on another PR to change the CI and there he had to restrict it for the CPU I'm using the same value here as will be used later after his changes.

backends/tfhe-cuda-backend/cuda/include/integer.h
agnesLeroy: No if we do we'll push_back in the wrong order.

.github/workflows/integer_multi_gpu_full_benchmark.yml
```diff
@@ -0,0 +1,184 @@
...
+          -H "X-Slab-Repository: ${{ github.repository }}" \
+          -H "X-Slab-Command: store_data_v2" \
+          -H "X-Hub-Signature-256: sha256=${SIGNATURE}" \
+          -d @${{ env.RESULTS_FILENAME }} \
+          ${{ secrets.SLAB_URL }}
+
+  slack-notify:
+    name: Slack Notification
+    needs: [ setup-instance, cuda-integer-full-multi-gpu-benchmarks ]
+    runs-on: ${{ needs.setup-instance.outputs.runner-name }}
```
soonum, line 152: @IceTDrinker noticed that we don't need to runs this job on AWS. You could use `ubuntu-latest` as runner.

.github/workflows/integer_multi_gpu_full_benchmark.yml
```diff
@@ -0,0 +1,184 @@
...
+      - name: Send message
+        uses: rtCamp/action-slack-notify@4e5fb42d249be6a45a298f3c9543b111b02f7907
+        env:
+          SLACK_COLOR: ${{ needs.cuda-integer-full-multi-gpu-benchmarks.result }}
+          SLACK_MESSAGE: "Integer GPU full benchmarks finished with status: ${{ needs.cuda-integer-full-multi-gpu-benchmarks.result }}. (${{ env.ACTION_RUN_URL }})"
+
+  teardown-instance:
+    name: Teardown instance (cuda-integer-full-multi-gpu-benchmarks)
+    if: ${{ always() && needs.setup-instance.result != 'skipped' }}
+    needs: [ setup-instance, cuda-integer-full-multi-gpu-benchmarks, slack-notify ]
```
soonum, line 165: As a consequence of the comment above, you can remove `slack-notify` in the array.

Makefile
```diff
@@ -459,7 +459,7 @@ test_core_crypto_gpu: install_rs_build_toolchain
 .PHONY: test_integer_gpu # Run the tests of the integer module including experimental on the gpu backend
 test_integer_gpu: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
-		--features=$(TARGET_ARCH_FEATURE),integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key::
+		--features=$(TARGET_ARCH_FEATURE),integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key:: --test-threads=6
```
soonum, line 462: This value will be overrode once the GPU tests separation will go to production. (PR in progress)

soonum (commented):
Overall, CI looks good to me. Only small stuff to fix.

agnesLeroy:
@soonum just pushed the changes

Makefile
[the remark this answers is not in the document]
pdroalves: I haven't! That's actually a good point.

soonum (commented):
Thanks, good to me. I let to @tmontaigu the final approval.

Makefile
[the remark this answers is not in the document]
agnesLeroy: That's alright the value I set now works :)

tfhe/src/core_crypto/gpu/vec.rs
tmontaigu: So you mean that I I have 4 streams, 4 operations and 1 thread, I push asynchronously each operation to each stream and then I synchronize the streams but this time in multithreaded, 1 thread per stream, the GPU computations are not going to happen  concurently ?

agnesLeroy: No if we launch in multithreaded 1 operation on 1 stream on each thread the execution will happen asynchronously. But if we don't use different CPU threads for the different streams execution will remain sequential.

agnesLeroy: At least on the C++ side that's what we observed, I haven't checked again here.

tmontaigu: Yes, but my question was about 1 CPU threads but multiple streams, which should be the setup here, unless ` streams.ptr[gpu_index as usize]` may be the same stream

agnesLeroy: Well with 1 CPU thread what we observed is that even using different cudaStreams we were getting a synchronous execution (cc @pdroalves maybe you can confirm, this dates back a bit for me).

pdroalves: The overhead for CUDA calls is far from negligible and it is not async even though the kernel (or memset/memcpy) operation itself will be async of course, and can overlap. Control will only return to the CPU when the kernel's launched.

If you launch two quick asynchronous (like, finish in a few ms) kernels using a single CPU thread, when the second is launched the first will be done and computation will be sequential in practice.  The same happens to memcpy and memset. 

If we launch everything using the same CPU thread we will have a timeline like this:

```
Launch overhead 1 | Execution 1
                               | Launch overhead 2 | Execution 2
                                                                | Launch overhead 3 | Execution 3
...
```

However, if you run each in a different CPU thread:

```
Launch overhead 1 | Execution 1
Launch overhead 2 | Execution 2
Launch overhead 3 | Execution 3
...
```

PR #1147, opened 2024-05-14
chore(ci): run workflows based on file changes

soonum:
This is done to imporve iteration time and feedback for devs.
There is no point to run the full test suite each time. A given development could impact only tfhe-zk-pok for example. In this case only tfhe-zk-pok test would run and thus cutting workflow duration from around 3 hours down to a few minutes.

### DIFF
--- a/.github/workflows/aws_tfhe_tests.yml
+++ b/.github/workflows/aws_tfhe_tests.yml
@@ -10,17 +10,117 @@ env:
   SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
   SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
   SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
 
 on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
     types: [ labeled ]
+  schedule:
+    # Nightly tests @ 1AM after each work day
+    - cron: "0 1 * * MON-FRI"
 
 jobs:
+  should-run:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    outputs:
+      csprng_test: ${{ env.IS_PULL_REQUEST == 'false' || steps.changed-files.outputs.csprng_any_changed }}
+      zk_pok_test: ${{ env.IS_PULL_REQUEST == 'false' || steps.changed-files.outputs.zk_pok_any_changed }}
+      core_crypto_test: ${{ env.IS_PULL_REQUEST == 'false' ||
+        steps.changed-files.outputs.core_crypto_any_changed ||
+        steps.changed-files.outputs.dependencies_any_changed }}
+      boolean_test: ${{ env.IS_PULL_REQUEST == 'false' ||
+        steps.changed-files.outputs.boolean_any_changed ||
+        steps.changed-files.outputs.dependencies_any_changed }}
+      shortint_test: ${{ env.IS_PULL_REQUEST == 'false' ||
+        steps.changed-files.outputs.shortint_any_changed ||
+        steps.changed-files.outputs.dependencies_any_changed }}
+      high_level_api_test: ${{ env.IS_PULL_REQUEST == 'false' ||
+        steps.changed-files.outputs.high_level_api_any_changed ||
+        steps.changed-files.outputs.dependencies_any_changed }}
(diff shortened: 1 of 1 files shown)


IceTDrinker (commented):
Could we avoid the dispatch and have each workflow perform the file checks ?

Basically a first job run on github ubuntu instance generates the changed files that are relevant for the current workflow, if there are changes then continue in the workflow with the instance set up ?

I don't know if the indirection created by the workflow call is a good idea, is that the pattern concrete used ?

soonum:
Concrete use the dispatch pattern indeed. Here's the pros and cons of this approach:
* Pros:  
  * file changes check occurs only in place
* Cons: 
  * need to pass around secrets for each target file inducing lot of duplication
  * need to be exhaustive in file checks and define dependencies chain in conditional expression (`if ( core_crypto_tests == true && shortint_tests == true [...] )`

Regarding your proposal (one file check in each workflow). I think it's a better approach in our case. We define once and for all the file-change check and we carry only one output in the job. Thus we eliminate the need to define long `if` expression to know if we run the  tests or not.

I'll give a shot to your proposal then.

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -38,9 +108,11 @@ jobs:
 
   cpu-tests:
     name: CPU tests
-    needs: setup-instance
+    if: (!cancelled() && github.event_name != 'pull_request') ||
```
IceTDrinker, line 111: I think !cancelled is not required ? I'm not sure

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -57,47 +129,60 @@ jobs:
...
         run: |
           BIG_TESTS_INSTANCE=TRUE make test_shortint_ci
 
       - name: Run high-level API tests
+        if: needs.should-run.outputs.high_level_api_test == 'true'
         run: |
           BIG_TESTS_INSTANCE=TRUE make test_high_level_api
 
       - name: Run example tests
+        if: needs.should-run.outputs.examples_test == 'true'
```
IceTDrinker, line 179: probably any change in core/boolean/shortint/integer would mean a re run

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -57,47 +129,60 @@ jobs:
...
         run: |
           make test_examples
           make dark_market
 
       - name: Run apps tests
+        if: needs.should-run.outputs.apps_test == 'true'
```
IceTDrinker, line 185: same here

tfhe/src/boolean/mod.rs
```diff
@@ -98,7 +98,8 @@ pub(crate) fn random_integer() -> u32 {
     // generate a random u32
     rng.gen::<u32>()
 }
-
+// Simple change to test triggering
+// Second changes
```
IceTDrinker, line 102: to remove ?

IceTDrinker (commented):
I think there is just a tiny bit more work to do to manage file dependency chains, that could be detected in the file chnange step

basically we change core, then we want to re run everything that depends on it

csprng likely means core needs to re run etc.

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -10,17 +10,116 @@ env:
   SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
   SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
   SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
 
 on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
-    types: [ labeled ]
```
mayeul-zama, line 18: Keep this line?

soonum: Do we want to run this only on approval still ?

IceTDrinker: good question, are the "fast tests" still there ?

soonum: For now yes. So I guess we should keep this event types in the workflow.

mayeul-zama (approved):
Thanks!

IceTDrinker:
looks like we started instances but then ran no tests https://github.com/zama-ai/tfhe-rs/actions/runs/9254705023/job/25457030476?pr=1147 ? 

@soonum

IceTDrinker:
probably a check missing to not start instances at all if no file changed

soonum:
> probably a check missing to not start instances at all if no file changed

You're right. Have to fix this.

IceTDrinker (approved):

PR #1148, opened 2024-05-14
feat(c_api): quick 'n' dirty C API for some array fn

tmontaigu:


tfhe/c_api_tests/test_high_level_array.c
```diff
@@ -0,0 +1,123 @@
...
+    assert(ok == 0 && clear_result == true);
+    fhe_bool_destroy(result);
+  }
+
+  // contains sub slice
+  {
+    FheBool *result;
+    bool clear_result;
+
+    // This one is trivial as the length are not the same
```
IceTDrinker, line 93: comment looks unrealted

tfhe/c_api_tests/test_high_level_array.c
```diff
@@ -0,0 +1,123 @@
...
+
+  char const *const sentence = "The quick brown fox jumps over the lazy dog";
+  char const *const pattern_1 = "wn fox ";
+  char const *const pattern_2 = "tfhe-rs";
+
+  size_t sentence_len = strlen(sentence);
+  size_t pattern_1_len = strlen(pattern_1);
+  size_t pattern_2_len = strlen(pattern_2);
+
+  assert(pattern_1_len == pattern_2_len); // We use this later in the tests
```
IceTDrinker, line 54: where is it used ? I might have missed it

IceTDrinker (commented):
Which part is dirty ?

tfhe/c_api_tests/test_high_level_array.c
tmontaigu: This and the other comment `// This one is trivial as the length are not the same` relates to the fact that when calling eq on two slices, if they don't have the same lengths, we trivially know they cant be equal

So one of the test makes sur we go in the path where length ares eq and so we do not trivially know if the slices are equal

IceTDrinker:
> The dirty part is that each layer (CAPI, HLAPI) makes us do a clone of the elements in the FheUint8** array (each array is cloned twice), something that proper Array types solves (or using maybe using a combination of repr(transparent) + transmute would solve but that a maybe)

ok got it

should we have an integer function taking an iterator as input then ? it's a pattern that's becoming more common lately

tfhe/c_api_tests/test_high_level_array.c
tmontaigu: tbf the comment is not really about that

tmontaigu:
It might be ok for CPU implementation but not for other backends/hardware

One downside with CPU implementation is that the functions signature would become more involved eg:
```
    pub fn sum_ciphertexts_parallelized<'a, T, C>(&self, ciphertexts: C) -> Option<T>
    where
        C: IntoIterator<Item = &'a T>,
        T: IntegerRadixCiphertext + 'a,
    {
 ```

and for default / smart operations we actually need to comsume the iterator to check if propagations needs to be done (and in the case of default if propagation is needed a clone needs to happen of each element needs to be made), so while doable, it is a bit annoying

IceTDrinker:
Good point, so I guess we can live with this at least temporarily

IceTDrinker (approved):
Thanks !

PR #1155, opened 2024-05-17
Tm/boolean ifthenelse

### LINKED ISSUE #825: `if_then_else_parallelized(c, a, b)`: Allow `a` and `b` to be `BooleanBlock`
Since the introduction of `BooleanBlock` in `0.5.0`, `k.if_then_else_parallelized(c, a, b)` expects `c` to be a `BooleanBlock` and `a` and `b` to implement trait `IntegerRadixCiphertext`, for an integer radix key `k`.

Sometimes, however, we want to compute an if_then_else on a triple of boolean values (i.e., `a` and `b` being of type `BooleanBlock`). This is not possible currently.

Either allow `a` and `b` of type `BooleanBlock` as an input to `k.if_then_else_parallelized(c, a, b)`, or provide a separate function on the key that permits such inputs.

It is possible to achieve the desired functionality by converting to and from `RadixCiphertext`, but this is cumbersome.

**Additional context**
`tfhe-rs = { version="0.5.0" }`

tmontaigu:
This adds if_then_else/cmux on BoleanBlock for integer API
and FheBool for HLAPI.

To have the same name for both default version of the API
in integer a trait was introduced.

Ideally we would have wanted to do the boolean cmux in one PBS by
packing data like this `(cond * 4) + (if * 2) + else` but this
would not respect the max norm2.

BREAKING CHANGE: integer::ServerKey::if_then_else_parallelized requires the
trait ServerKeyDefaultCMux to be in scope.

Fixes #825

This PR also has the commit that creates the prelude

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/cmux.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/cmux.rs
@@ -3,6 +3,187 @@ use crate::integer::ciphertext::IntegerRadixCiphertext;
 use crate::integer::ServerKey;
 use rayon::prelude::*;
 
+pub trait ServerKeyDefaultCMux<TrueCt, FalseCt> {
+    type Output;
+    fn if_then_else_parallelized(
+        &self,
+        condition: &BooleanBlock,
+        true_ct: TrueCt,
+        false_ct: FalseCt,
+    ) -> Self::Output;
+
+    fn cmux_parallelized(
+        &self,
+        condition: &BooleanBlock,
+        true_ct: TrueCt,
+        false_ct: FalseCt,
+    ) -> Self::Output {
+        self.if_then_else_parallelized(condition, true_ct, false_ct)
+    }
+}
+
+impl<T> ServerKeyDefaultCMux<&T, &T> for ServerKey
+where
+    T: IntegerRadixCiphertext,
+{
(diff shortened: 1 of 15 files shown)


tfhe/src/integer/server_key/radix_parallel/cmux.rs
```diff
@@ -69,7 +130,56 @@ impl ServerKey {
     /// assert_ne!(ct_a, ct_res);
     /// assert_ne!(ct_b, ct_res);
     /// ```
-    pub fn if_then_else_parallelized<T>(
+    fn if_then_else_parallelized(
+        &self,
+        condition: &BooleanBlock,
+        true_ct: &BooleanBlock,
+        false_ct: &BooleanBlock,
```
IceTDrinker, line 137: is this one tested ?

tmontaigu: There is a docstring

tmontaigu: But the docstrings are inverted 🤦

IceTDrinker (approved):
Thanks :)

IceTDrinker (approved):

PR #1167, opened 2024-05-27
Mz/prf sha3

mayeul-zama:


tfhe/src/shortint/oprf.rs
```diff
@@ -85,7 +108,11 @@ impl ServerKey {
             "The number of random bits asked for (={random_bits_count}) is bigger than full_bits_count (={full_bits_count})"
         );
 
-        let ct = self.create_random_from_seed(seed);
+        assert_eq!(self.pbs_order, PBSOrder::KeyswitchBootstrap);
```
IceTDrinker, line 111: why the assert ? We could keyswitch to get back to the right key size for the PBS_KS case right ?

tfhe/src/shortint/oprf.rs
```diff
@@ -205,7 +252,7 @@ pub(crate) mod test {
     fn oprf_test_uniformity_ci_run_filter() {
         let sample_count: usize = 100_000;
 
-        let p_value_limit: f64 = 0.001;
+        let p_value_limit: f64 = 0.000_001;
```
IceTDrinker, line 255: removing the keyswitch makes it work better ?

IceTDrinker (commented):
Looks generally good a few small question

tfhe/src/shortint/oprf.rs
mayeul-zama: No, it's unrelated

tfhe/src/shortint/oprf.rs
mayeul-zama: Yes, looks like it would be ok

tfhe/src/shortint/oprf.rs
IceTDrinker: ok, do you think the limit is better as it was or as it is now ?

mayeul-zama: Better now as this test should fail with `p_value_limit` probability

IceTDrinker: ok did you stress test it to check it did not fail in the CI ?

mayeul-zama: This new bound is less strict (by a huge ratio) than the old one which IIRC only failed one on the CI (p=1/1000) so this should be fine

IceTDrinker: got it, is it still significant enough ?

tfhe/src/shortint/oprf.rs
```diff
@@ -96,20 +122,54 @@ impl ServerKey {
...
+        lwe_ciphertext_plaintext_add_assign(&mut ct, Plaintext((p - 1) * delta / 2));
+
+        let ct = match self.pbs_order {
+            PBSOrder::KeyswitchBootstrap => ct,
+            PBSOrder::BootstrapKeyswitch => {
+                let mut ct_ksed = allocate_and_trivially_encrypt_new_lwe_ciphertext(
+                    in_lwe_size,
+                    Plaintext(0),
+                    self.ciphertext_modulus,
+                );
```
IceTDrinker, line 148: let mut ct_ksed = LweCiphertext::new(0u64, in_lwe_size, self.ciphertext_modulus);

tfhe/src/shortint/oprf.rs
```diff
@@ -96,20 +122,54 @@ impl ServerKey {
         let acc: LookupTableOwned =
             self.generate_lookup_table_no_encode(|x| (2 * (x / poly_delta) + 1) * delta / 2);
 
-        let mut ct = self.apply_lookup_table(&ct, &acc);
+        let out_lwe_size = self.bootstrapping_key.output_lwe_dimension().to_lwe_size();
 
-        lwe_ciphertext_plaintext_add_assign(&mut ct.ct, Plaintext((p - 1) * delta / 2));
+        let mut ct = allocate_and_trivially_encrypt_new_lwe_ciphertext(
```
IceTDrinker, line 127: same here you can allocate directly a new 0ed ciphertext

tfhe/src/shortint/oprf.rs
mayeul-zama: Moved this change to https://github.com/zama-ai/tfhe-rs/pull/1179

tfhe/src/shortint/oprf.rs
mayeul-zama: Done

tfhe/src/shortint/oprf.rs
mayeul-zama: Done

IceTDrinker (approved):
thx

PR #1170, opened 2024-05-28
chore(bench): set fast bench precision to 64 bits for cuda backend

soonum:


agnesLeroy (commented):
@IceTDrinker for the CPU do you want to keep fast benches <-> 32 bit benches? Wouldn't it make more sense to bench the 64 bit integers there as well?

IceTDrinker:
> @IceTDrinker for the CPU do you want to keep fast benches <-> 32 bit benches? Wouldn't it make more sense to bench the 64 bit integers there as well?

good point, we can probably go for 64 bits yes

agnesLeroy (approved):

PR #1172, opened 2024-05-28
doc: use from_config() with correct parameter in zk-pok doc

dd23:
This PR fixes an oversight in the zk-pok documentation by switching to a different way of initializing the CRS parameters and supplying the right parameters. It also adds small documentation to the used function and adds details on correctly running the code sample in release mode.


IceTDrinker (approved):
thanks

PR #1174, opened 2024-05-28
Erc20 bench

tmontaigu:
This adds benchmarks of both latency and throughput
of 4 variants of the erc20 transfer.

It also prints the PBS count of each versions.

- "whitepaper" is the variant written in the fhevm whitepaper
- "no_cmux" is similar to whitepaper, but uses a "boolean multiplication"
  instead of a cmux
- "overflow" uses an overflowing_sub to remove the need for comparison
- "safe" use both overflowing_sub and overflowing_add to make sure both
  then sender has enough money and the that the transfer won't overflow
  the receiver's money

"overflow" has the lowest latency, and second best throughput
"no_cmux" has the second lowest latenc and the best throughput

### DIFF
--- a/.github/workflows/benchmark_erc20.yml
+++ b/.github/workflows/benchmark_erc20.yml
@@ -1,0 +1,144 @@
+# Run all ERC20 benchmarks on an AWS instance and return parsed results to Slab CI bot.
+name: ERC20 benchmarks
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Weekly benchmarks will be triggered each Saturday at 5a.m.
+    - cron: '0 5 * * 6'
+
+env:
+  CARGO_TERM_COLOR: always
+  RESULTS_FILENAME: parsed_benchmark_results_${{ github.sha }}.json
+  ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  RUST_BACKTRACE: "full"
+  RUST_MIN_STACK: "8388608"
+  SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
+  SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
+  SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
+  SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+jobs:
+  setup-instance:
+    name: Setup instance (erc20-benchmarks)
+    runs-on: ubuntu-latest
+    outputs:
+      runner-name: ${{ steps.start-instance.outputs.label }}
+    steps:
+      - name: Start instance
+        id: start-instance
+        uses: zama-ai/slab-github-runner@c0e7168795bd78f61f61146951ed9d0c73c9b701
+        with:
+          mode: start
+          github-token: ${{ secrets.SLAB_ACTION_TOKEN }}
+          slab-url: ${{ secrets.SLAB_BASE_URL }}
+          job-secret: ${{ secrets.JOB_SECRET }}
+          backend: aws
+          profile: bench
+
+  erc20-benchmarks:
+    name: Execute ERC20 benchmarks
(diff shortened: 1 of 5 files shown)


soonum (commented):
There is no target in the Makefile.
Are these benchmark cannot be run on a local machine ?

tfhe/benches/high_level_api/erc20.rs
```diff
@@ -0,0 +1,301 @@
...
+    let amount = FheType::encrypt(rng.gen::<u64>(), client_key);
+
+    tfhe::reset_pbs_count();
+    let (_, _) = transfer_func(&from_amount, &to_amount, &amount);
+    let count = tfhe::get_pbs_count();
+
+    println!("ERC20 transfer/{fn_name}::{type_name}: {count} PBS");
+}
+
+fn bench_transfer_latency<FheType, F>(
```
soonum, line 128: Are these benchmarks never meant to be stored in database ?

tmontaigu: hum I think we want to store them in a database but they are not integer bench so we likely need a different database/schema and we would need to store both the throughput and latency (and pbs count if posssible)

tmontaigu:
> Are these benchmark cannot be run on a local machine
What do you mean by local ? You can run them on your laptop if you want but its not really going to be meaningfull especially for the thourhgput benches

tfhe/benches/high_level_api/erc20.rs
soonum: Ok so we need to discuss how to add them.

soonum:
> What do you mean by local ? You can run them on your laptop if you want but its not really going to be meaningfull especially for the thourhgput benches

No problem, since it's just a target missing in the Makefile. I was wondering if it needed a more sophisticated infra than just one machine.

.github/workflows/benchmark_erc20.yml
```diff
@@ -0,0 +1,144 @@
...
+          --branch ${{ github.ref_name }} \
+          --commit-date "${{ env.COMMIT_DATE }}" \
+          --bench-date "${{ env.BENCH_DATE }}" \
+          --walk-subdirs \
+          --name-suffix avx512
+
+      - name: Parse PBS counts
+        run: |
+          python3 ./ci/benchmark_parser.py tfhe/erc20_pbs_count.csv ${{ env.RESULTS_FILENAME }} \
+          --key-sizes \
```
IceTDrinker, line 99: key sizes ?

IceTDrinker (commented):
Looks generally ok minor comments, did we do a run of the bench workflow to check it works properly @soonum ?

.github/workflows/benchmark_erc20.yml
soonum: I think I should change this flag name since we are not only measuring key sizes now.
Maybe `--object-sizes` would be more suitable.

tfhe/src/high_level_api/keys/client.rs
```diff
@@ -70,6 +70,10 @@ impl ClientKey {
         }
     }
 
+    pub fn parameters(&self) -> crate::shortint::PBSParameters {
+        self.key.block_parameters()
+    }
```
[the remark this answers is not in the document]
soonum: @tmontaigu I'll you judge on that matter.

soonum:
> Looks generally ok minor comments, did we do a run of the bench workflow to check it works properly @soonum ?

I can do that, I just have to use the placeholder workflow to test it.

.github/workflows/benchmark_erc20.yml
IceTDrinker: no need to do it in that PR, backlog issue

tfhe/src/high_level_api/keys/client.rs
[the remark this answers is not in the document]
tmontaigu: I think I would go for `compuration_parameters`

soonum:
Workflow tested with success.

IceTDrinker (approved):

IceTDrinker (approved):

PR #1184, opened 2024-05-30
chore(ci): use only pull_request event in workflow

soonum:
Using pull_request_target event to handle PR from forks was clashing with pull_request event. It would launch double amount of actions and moreover leads to cancellation in jobs due to the concurrency directive.

### DIFF
--- a/.github/workflows/aws_tfhe_tests.yml
+++ b/.github/workflows/aws_tfhe_tests.yml
@@ -17,8 +17,6 @@ on:
   workflow_dispatch:
   pull_request:
     types: [ labeled ]
-  pull_request_target:
-    types: [ labeled ]
   schedule:
     # Nightly tests @ 1AM after each work day
     - cron: "0 1 * * MON-FRI"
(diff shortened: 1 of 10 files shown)


.github/workflows/aws_tfhe_tests.yml
```diff
@@ -141,9 +138,8 @@ jobs:
 
   cpu-tests:
     name: CPU tests
-    if: (github.event_name != 'pull_request' && github.event_name != 'pull_request_target') ||
-      ((github.event_name == 'pull_request' || github.event_name != 'pull_request_target') &&
-      needs.setup-instance.result != 'skipped')
+    if: (github.event_name != 'pull_request' ||
+      (github.event_name == 'pull_request' && needs.setup-instance.result != 'skipped')
```
IceTDrinker, line 142: looks broken

IceTDrinker: picked up by CI lint

IceTDrinker (approved):
Thanks, merge as soon as this is mergeable

PR #1185, opened 2024-05-31
chore(tfhe): alias if_then_else to select

tmontaigu:
Closes https://github.com/zama-ai/tfhe-rs-internal/issues/594


tmontaigu:
@IceTDrinker Should we also update the docs to use select instead of if_then_else ?

IceTDrinker:
> @IceTDrinker Should we also update the docs to use select instead of if_then_else ?

yes it was a request as well

IceTDrinker (approved):
Thanks !

PR #1188, opened 2024-05-31
chore(gpu): parallelize keyswitch further

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
@@ -3,8 +3,10 @@
 
 #include "device.h"
 #include "gadget.cuh"
+#include "polynomial/functions.cuh"
 #include "polynomial/polynomial_math.cuh"
 #include "torus.cuh"
+#include "utils/kernel_dimensions.cuh"
 #include <thread>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


agnesLeroy (approved):

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -76,15 +101,16 @@ __host__ void cuda_keyswitch_lwe_ciphertext_vector(
     uint32_t base_log, uint32_t level_count, uint32_t num_samples) {
 
   cudaSetDevice(gpu_index);
-  constexpr int ideal_threads = 1024;
-  if (lwe_dimension_out + 1 > ideal_threads)
-    PANIC("Cuda error (keyswitch): lwe dimension size out should be greater "
-          "or equal to the number of threads per block")
+
+  constexpr int num_threads_x = 16;
+  constexpr int num_threads_y = 32;
```
pdroalves, line 106: Do we really need to hard code the block dimension?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -76,15 +101,16 @@ __host__ void cuda_keyswitch_lwe_ciphertext_vector(
...
 
   int lwe_size = lwe_dimension_out + 1;
-  int shared_mem = sizeof(Torus) * lwe_size;
-  dim3 grid(num_samples, 1, 1);
-  dim3 threads(ideal_threads, 1, 1);
+  int num_blocks = (lwe_size + num_threads_x - 1) / num_threads_x;
```
pdroalves, line 109: Let's use `getNumBlocksAndThreads()` in `utils/kernel_dimensions.cuh` to calculate the number of blocks.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
pdroalves: If so, why so small values for each dimension?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -31,43 +31,68 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+        local_lwe_out -= (Torus)ksk_block[tid] * decomposed;
       }
     }
-    block_lwe_array_out[tid] = local_lwe_array_out[tid];
+
+    lwe_acc_out[shmem_index] = local_lwe_out;
+  }
+
+  if (tid <= lwe_dimension_out) {
+    for (int offset = blockDim.y / 2; offset > 0 && threadIdx.y < offset;
```
pdroalves, line 85: We may want to run such a reduction in other places in the future. Can you generalize and move this to its own function? That would also make this piece of code easier to read.

pdroalves (changes requested):
Seems quite good. Thanks @guillermo-oyarzun !
I just added a few suggestions for improvement.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -31,43 +31,68 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+
+    if (tid == lwe_dimension_out && threadIdx.y == 0) {
+      local_lwe_out = block_lwe_array_in[lwe_dimension_in];
     }
+    const Torus mask_mod_b = (1ll << base_log) - 1ll;
 
-    for (int i = 0; i < lwe_dimension_in; i++) {
+    const int pack_size = (lwe_dimension_in + blockDim.y - 1) / blockDim.y;
+    const int start_i = pack_size * threadIdx.y;
+    const int end_i = pack_size * (threadIdx.y + 1) <= lwe_dimension_in
```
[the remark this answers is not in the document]
guillermo-oyarzun: added!

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: You have to consider that it is a 2D block, so the total number of threads in the block is 16x32 = 512. 
The number of threads in the y direction is hardcoded since it is the one that follows the reduction logic.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: I saw that function only applies to unidimensional blocks in which each threads produces one output. I created a similar function that works for 2D blocks for which you set the number of threads in y that are going to participate in the calculation of one output.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: This is a very specific reduction pattern, in the sense that we only want to perform the reduction on the threads in y, and only within the block. I think what you suggest could be another ticket in which we could generalise it to different configurations or different operations.

pdroalves: Indeed!

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
pdroalves: I think it should work as it is indeed. I remember seeing parameter sets with N=256 but don't think they are in use nowadays.

pdroalves (approved):

PR #1193, opened 2024-06-03
fix(gpu): ensure `single carry propagation` returns carry

bbarbakadze:
cuda backend `propagate_single_carry` now returns `carry_out` instead of only processing input carries.

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -242,9 +242,12 @@ void scratch_cuda_propagate_single_carry_kb_64_inplace(
     uint32_t num_blocks, uint32_t message_modulus, uint32_t carry_modulus,
     PBS_TYPE pbs_type, bool allocate_gpu_memory);
 
-void cuda_propagate_single_carry_kb_64_inplace(
-    void **streams, uint32_t *gpu_indexes, uint32_t gpu_count, void *lwe_array,
-    int8_t *mem_ptr, void *bsk, void *ksk, uint32_t num_blocks);
+void cuda_propagate_single_carry_kb_64_inplace(void **streams,
+                                               uint32_t *gpu_indexes,
+                                               uint32_t gpu_count,
+                                               void *lwe_array, void *carry_out,
+                                               int8_t *mem_ptr, void *bsk,
+                                               void *ksk, uint32_t num_blocks);
 
 void cleanup_cuda_propagate_single_carry(void *stream, uint32_t gpu_index,
                                          int8_t **mem_ptr_void);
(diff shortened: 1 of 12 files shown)


backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -244,7 +244,7 @@ void scratch_cuda_propagate_single_carry_kb_64_inplace(
 
 void cuda_propagate_single_carry_kb_64_inplace(
     void **streams, uint32_t *gpu_indexes, uint32_t gpu_count, void *lwe_array,
-    int8_t *mem_ptr, void *bsk, void *ksk, uint32_t num_blocks);
+    void *carry, int8_t *mem_ptr, void *bsk, void *ksk, uint32_t num_blocks);
```
agnesLeroy, line 247: Maybe `carry_out` for the name?

backends/tfhe-cuda-backend/cuda/src/integer/integer.cu
```diff
@@ -136,10 +136,10 @@ void scratch_cuda_propagate_single_carry_kb_64_inplace(
 
 void cuda_propagate_single_carry_kb_64_inplace(
     void **streams, uint32_t *gpu_indexes, uint32_t gpu_count, void *lwe_array,
-    int8_t *mem_ptr, void *bsk, void *ksk, uint32_t num_blocks) {
+    void *carry, int8_t *mem_ptr, void *bsk, void *ksk, uint32_t num_blocks) {
```
agnesLeroy, line 139: Same here

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -410,8 +410,8 @@ void scratch_cuda_propagate_single_carry_kb_inplace(
 template <typename Torus>
 void host_propagate_single_carry(cudaStream_t *streams, uint32_t *gpu_indexes,
                                  uint32_t gpu_count, Torus *lwe_array,
-                                 int_sc_prop_memory<Torus> *mem, void *bsk,
-                                 Torus *ksk, uint32_t num_blocks) {
+                                 Torus *carry, int_sc_prop_memory<Torus> *mem,
```
agnesLeroy, line 413: Same here

agnesLeroy (commented):
Hi @bbarbakadze! Thanks a lot for this PR, I think you may also need to modify the places on the Rust side where single carry propagation is called to do something like:
```
let _carry = self.propagate_single_carry_parallelized_low_latency(ct.blocks_mut());
```
instead of 
```
 self.propagate_single_carry_assign_async(ct, streams);
```

agnesLeroy (approved):

PR #1199, opened 2024-06-04
feat(core): add keyswitch that changes the scalar type from input to output

IceTDrinker:
refs: https://github.com/zama-ai/tfhe-rs-internal/issues/526

### PR content/description

Add a separate function for switching Scalar types between input and ouptut while keyswitching.

I have preferred this approach for now as it was making the bounds and the code way more convoluted in the existing keyswitch.

But if we think it's better to put it in the same function I can.

A follow-up PR will do that for the PBS but should be much easier to write as the two types do not interact at all (as everything happens in the fourier domain), here we need to cast to make computations with both types

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
@@ -2,6 +2,7 @@
 //! keyswitch`](`LweKeyswitchKey#lwe-keyswitch`).
 
 use crate::core_crypto::algorithms::slice_algorithms::*;
+use crate::core_crypto::commons::ciphertext_modulus::CiphertextModulusKind;
 use crate::core_crypto::commons::math::decomposition::{
     SignedDecomposer, SignedDecomposerNonNative,
 };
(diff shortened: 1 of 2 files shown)


tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
```diff
@@ -315,6 +316,121 @@ pub fn keyswitch_lwe_ciphertext_other_mod<Scalar, KSKCont, InputCont, OutputCont
...
 }
 
+/// Keyswitch an [`LWE ciphertext`](`LweCiphertext`) with a certain InputScalar type to represent
+/// data encrypted under an [`LWE secret key`](`LweSecretKey`) to another [`LWE secret
+/// key`](`LweSecretKey`) using a different OutputScalar type to represent data.
+///
+/// # Notes
+///
+/// This function only supports power of 2 moduli and going from a large InputScalar with
+/// `input_bits` to a a smaller OutputScalar with `output_bits` and `output_bits` < `input_bits`.
```
mayeul-zama, line 326: Why this constraint?
In my understanding, we only need `level*base_log<=input_bits` (and arguably `base_log < output_bits`)

IceTDrinker: It was to make things easier when managing casting from input to output

Basically what we are interested in at the moment is going from 64 bits to 32 bits during the keyswitch

When I was writing the code it looked like there would be more and more possible branches in the implementation if I made the most generic version so as a first version of that I thought of simplifying the function to match our need with the intent to add more functionality later if required.

What do you think ?

You are right on the required level * base_log < output_bits that I did not capture here.

IceTDrinker: but then we likely have the same input scalar and people should use the other implem

tfhe/src/core_crypto/algorithms/lwe_keyswitch.rs
```diff
@@ -315,6 +316,137 @@ pub fn keyswitch_lwe_ciphertext_other_mod<Scalar, KSKCont, InputCont, OutputCont
...
+{
+    assert!(
+        InputScalar::BITS > OutputScalar::BITS,
+        "This operation only supports going from a large InputScalar type \
+        to a strictly smaller OutputScalar type."
+    );
+    assert!(
+        lwe_keyswitch_key.decomposition_base_log().0
+            * lwe_keyswitch_key.decomposition_level_count().0
+            <= OutputScalar::BITS,
```
mayeul-zama, line 357: I think this should be the `InputScalar` here

IceTDrinker: if it holds for the OutputScalar which is smaller than InputScalar then it also holds for the InputScalar, and it should hold for both :)

mayeul-zama: Indeed, but it's more restrictive than it could be I think
Why should it hold for both?

IceTDrinker: The current decomposer formulas only work if level * base_log < Bits, the decomposer does an approximation over level * base_log bits, so it means that we need the approximated bits to fit in the OutputScalar so approximated bits counts <= OutputScalar::BITS

mayeul-zama (approved):
Thanks

PR #1214, opened 2024-06-10
refactor(tfhe): create associated CompactPrivateKey and prepare casting

IceTDrinker:

### PR content/description

- for casting from the CompacthPublicKey parameter we need to add the notion of a kind on the CompactCiphertextList, where one kind will need to be cast thanks to an auxiliary keyswitching key and the other kind can just be expanded as before
- to avoid weird situations/corner cases we remove the ability to encrypt a "normal" ciphertext from a CompactPublicKey (which consisted in expanding right after encryption)

### Check-list:

### DIFF
--- a/tfhe/src/integer/ciphertext/compact_list.rs
+++ b/tfhe/src/integer/ciphertext/compact_list.rs
@@ -179,7 +179,7 @@ impl CompactCiphertextListBuilder {
     }
 
     pub fn build_packed(&self) -> crate::Result<CompactCiphertextList> {
-        if self.pk.key.parameters.carry_modulus().0 < self.pk.key.parameters.message_modulus().0 {
+        if self.pk.key.parameters.carry_modulus.0 < self.pk.key.parameters.message_modulus.0 {
             return Err(crate::Error::new("In order to build a packed compact ciphertext list, parameters must have CarryModulus >= MessageModulus".to_string()));
         }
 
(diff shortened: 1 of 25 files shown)


tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -432,6 +436,10 @@ impl CompactCiphertextList {
         ))
     }
 
+    pub fn expand_with_casting(&self) {
+        todo!()
+    }
+
```
tmontaigu, line 442: Just to be sure, this is a todo for the next PR ?

tfhe/src/shortint/parameters/compact_public_key_only.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        Ok(parameters)
+    }
+
+    pub const fn is_valid(&self) -> bool {
+        self.encryption_lwe_dimension.0.is_power_of_two()
+    }
+
+    /// This should be used while defining static parameters to verify they comply with the
+    /// requirements of ompact public key encryption.
```
tmontaigu, line 63:
```suggestion
    /// requirements of compact public key encryption.
```

tfhe/src/integer/ciphertext/compact_list.rs
IceTDrinker: I can remove it from here, at first I thought of doing everything at once but it seemed too much, so yes will be done in the next PR, I'll remove this to keep a consistent code base

tmontaigu (approved):

PR #1218, opened 2024-06-11
feat(tfhe): add FheUint2048

tmontaigu:
Adds FheUint2048 to 0.6 release

### DIFF
--- a/tfhe/src/c_api/high_level_api/u2048.rs
+++ b/tfhe/src/c_api/high_level_api/u2048.rs
@@ -1,0 +1,88 @@
+use crate::c_api::utils::*;
+use std::os::raw::c_int;
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct U2048 {
+    words: [u64; 32],
+}
+
+impl From<crate::integer::bigint::U2048> for U2048 {
+    fn from(value: crate::integer::bigint::U2048) -> Self {
+        Self { words: value.0 }
+    }
+}
+
+impl From<U2048> for crate::integer::bigint::U2048 {
+    fn from(value: U2048) -> Self {
+        Self(value.words)
+    }
+}
+
+/// Creates a U2048 from little endian bytes
+///
+/// len must be 256
+#[no_mangle]
+pub unsafe extern "C" fn U2048_from_little_endian_bytes(
+    input: *const u8,
+    len: usize,
+    result: *mut U2048,
+) -> c_int {
+    catch_panic(|| {
+        let mut inner = crate::integer::bigint::U2048::default();
+
+        let input = std::slice::from_raw_parts(input, len);
+        inner.copy_from_le_byte_slice(input);
+
+        *result = U2048::from(inner);
+    })
+}
+
+/// Creates a U2048 from big endian bytes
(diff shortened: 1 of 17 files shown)


tfhe/src/integer/bigint/mod.rs
```diff
@@ -6,6 +6,11 @@ pub mod static_unsigned;
 pub mod u256;
 pub mod u512;
 
+pub use static_unsigned::StaticUnsignedBigInt;
+
+pub type U2048 = static_unsigned::StaticUnsignedBigInt<32>;
+pub type U4096 = static_unsigned::StaticUnsignedBigInt<64>;
```
IceTDrinker, line 12: question to maybe use bits instead of 64 bits as a reference for the number of blocks ?

and maybe restrain to multiple of 64 ? unsure but this looks error prone

tfhe/src/c_api/high_level_api/utils.rs
```diff
@@ -41,6 +41,7 @@ impl_c_api_integer_type!(i64);
 impl_c_api_integer_type!(crate::c_api::high_level_api::u128::U128 => u128);
 impl_c_api_integer_type!(crate::c_api::high_level_api::i128::I128 => i128);
 impl_c_api_integer_type!(crate::c_api::high_level_api::u256::U256 => crate::integer::U256);
+impl_c_api_integer_type!(crate::c_api::high_level_api::u2048::U2048 => crate::integer::bigint::U2048);
```
IceTDrinker, line 44: below I256 perhaps ?

complete OCD nitpick

tfhe/src/c_api/high_level_api/u2048.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl From<U2048> for crate::integer::bigint::U2048 {
+    fn from(value: U2048) -> Self {
+        Self(value.words)
+    }
+}
+
+/// Creates a U2048 from little endian bytes
+///
+/// len must be 256
```
IceTDrinker, line 24: when/where is it checked ?

IceTDrinker (commented):
Looks generally good small questions and nitpicks

thanks !

tfhe/src/c_api/high_level_api/u2048.rs
tmontaigu: deep down in the call satck https://github.com/zama-ai/tfhe-rs/blob/6014968655bb3f283d6caaca5ac90500a1d5ae78/tfhe/src/integer/bigint/algorithms.rs#L9

IceTDrinker: alright, those could likely use some nice error message, but it's not required for this

tfhe/src/integer/bigint/mod.rs
tmontaigu: It's not possible, rust const generic are too limited we cannot do const operations with them

IceTDrinker: yeah, I seemed to recall that thanks

IceTDrinker (approved):

PR #1222, opened 2024-06-11
chore(shortint): update default parameters

soonum:
The default parameters are now offering a security level of 132 bits and uses a p-fail of 2**-64.


IceTDrinker (approved):
Thanks looks good !

IceTDrinker:
hmm the commit message is all weird ?

PR #1204, opened 2024-06-06
Am/chore/update wasm bench

IceTDrinker:
closes: _please link all relevant issues_

### Check-list:

* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Relevant issues are marked as resolved/closed, related issues are linked in the description
* [ ] Check for breaking changes (including serialization changes) and add them to commit message following the conventional commit [specification][conventional-breaking]

[conventional-breaking]: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer

PR #1224, opened 2024-06-11
Am/feat/cpk cast

IceTDrinker:
First step for the casting after compact public key encryption.

second step is here : https://github.com/zama-ai/tfhe-rs/pull/1239

### Check-list:


tfhe/src/shortint/key_switching_key/mod.rs
```diff
@@ -250,9 +378,119 @@ impl KeySwitchingKey {
...
+                .dest_server_key
+                .bootstrapping_key
+                .input_lwe_dimension()
+                .to_lwe_size(),
+        };
+        let mut keyswitched = self
+            .dest_server_key
+            .unchecked_create_trivial_with_lwe_size(0, output_lwe_size);
+
+        match self.cast_rshift {
```
tmontaigu, line 398: nitpick, to avoid the _ => unreachble, maybe we could do

```rust
match self.cast_rshift.cmp(0) {
 Ordering::Eq => {}
 Ordering::Less => {}
 Ordering::Greater => {}
 
}
```

IceTDrinker: was from the previous code that's a good point

tfhe/src/integer/ciphertext/compact_list.rs
```diff
@@ -486,64 +522,80 @@ impl ProvenCompactCiphertextList {
...
+                unpacking_mode,
+                IntegerCompactCiphertextListUnpackingMode::NoUnpacking
+            )
+        {
+            return Err(crate::Error::new(String::from(
+                "Cannot expand a CompactCiphertextList that requires unpacking without \
+                a server key, please provide a shortint::ServerKey passing it with the \
+                enum variant CompactCiphertextListUnpackingMode::UnpackIfNecessary \
+                as unpacking_mode.",
+            )));
```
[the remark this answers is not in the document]
IceTDrinker: Yep

IceTDrinker:
I will update the PR with the second HL commit and fix the first comments

tfhe/src/integer/client_key/secret_encryption_key.rs
```diff
@@ -0,0 +1,7 @@
+use crate::core_crypto::commons::traits::Container;
+use crate::shortint::client_key::secret_encryption_key::SecretEncryptionKey as ShortintSecretEncryptionKey;
+
+#[derive(Clone)]
+pub struct SecretEncryptionKey<KeyCont: Container<Element = u64>> {
```
mayeul-zama, line 5: Do we want generics in the integer API (and even in the shortint one)?

mayeul-zama (commented):
A few comments after a first read

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -31,26 +34,48 @@ impl CompactCiphertextList {
...
+        // For WASM
+        if !self.0.is_packed() && !self.0.needs_casting() {
+            // No ServerKey required, shortcircuit to avoid the global state call
+            return Ok(CompactCiphertextListExpander {
+                inner: self.0.expand(
+                    IntegerCompactCiphertextListUnpackingMode::NoUnpacking,
+                    IntegerCompactCiphertextListCastingMode::NoCasting,
+                )?,
+            });
         }
```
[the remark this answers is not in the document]
IceTDrinker: I find easier to read here by getting the special case out of the way first, I tried something that looks what you wrote, made the comprehension very hard

tfhe/src/integer/client_key/secret_encryption_key.rs
IceTDrinker: Well if we don't we have to manually create types for views, seems rather wasteful don't you think ?

mayeul-zama: Not sure to understand what you mean, but I haven't looked as the usage
If this type is used with more than generic in the code, it's fine to have it generic I guess

IceTDrinker: actually looks like I only use views, so I'll change it to be views only

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -31,26 +34,48 @@ impl CompactCiphertextList {
...
+
+        global_state::try_with_internal_keys(|maybe_keys| match maybe_keys {
+            None => Err(crate::high_level_api::errors::UninitializedServerKey.into()),
```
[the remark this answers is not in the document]
IceTDrinker: not sure I follow what you mean ?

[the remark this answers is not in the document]
mayeul-zama: In this branch, we don't have a server_key, but if there's no unpacking, we don't need it
So we could do the expand instead of returning an error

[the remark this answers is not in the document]
IceTDrinker: isn't it managed at the beginning of the functions ? If there is no need for either, I don't touch the keys, then if some key is needed I enter the with_internal_keys section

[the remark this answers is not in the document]
IceTDrinker: might have missed it, I don't see the start of the function in the diff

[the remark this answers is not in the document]
mayeul-zama: I guess it's because the todo for the casting will need the server key, which is not the case yet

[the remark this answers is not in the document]
IceTDrinker: ah it's likely yes, sorry, I'll push the new commit on top so it's clearer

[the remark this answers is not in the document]
IceTDrinker: let me know if the latest push makes things clearer

IceTDrinker:
hmm, looks like there were push to main and I need to resolve conflicts, the previous content in the PR is still in the first commit, the second one has the HL update

tfhe/src/high_level_api/config.rs
```diff
@@ -53,13 +62,21 @@ impl ConfigBuilder {
     pub fn with_custom_parameters<P>(
         block_parameters: P,
         wopbs_block_parameters: Option<crate::shortint::WopbsParameters>,
+        dedicated_compact_public_key_parameters: Option<(
+            crate::shortint::parameters::CompactPublicKeyEncryptionParameters,
+            crate::shortint::parameters::ShortintKeySwitchingParameters,
+        )>,
```
tmontaigu, line 68: I going to do the same remark as for the compression PR,

Is it possible not to add this `dedicated_compact_public_key_parameter` so that the API of this function is not broken
and have something that looks like 

```
Builder::with_custom_param(params, None)
  .with_dedicated_compact_pk_params(...)
```

IceTDrinker: yep just saw your comment on the other PR, I was hesitating, and decided to chameleon my way as the wopbs params, but 100% agree with the proposal

tmontaigu: Yeah wopbs params are weird, I think I made it that way for wopbs because when creating the Wopbs sks you need the classical and wopbs params at the same time but ideally wopbs should also have been in a dedicated .enable/with

IceTDrinker:
the PR will need a rebase on https://github.com/zama-ai/tfhe-rs/pull/1204 to make sense, will do that tomorrow as it's nearly all green

tfhe/src/shortint/ciphertext/compact_list.rs
```diff
@@ -81,29 +91,56 @@ impl CompactCiphertextList {
...
-                    self.ct_list.ciphertext_modulus(),
-                );
-                Ciphertext {
-                    ct,
-                    degree: self.degree,
-                    message_modulus: self.message_modulus,
-                    carry_modulus: self.carry_modulus,
-                    pbs_order,
-                    noise_level: self.noise_level,
+        match self.expansion_kind {
```
tmontaigu, line 94: Minor suggestiong to avoid nested matches

```rust
match (self.expansion_kind, casting_mode) {
	(
		CompactCiphertextListExpansionKind::NoCasting(pbs_order),
		_
	) => {
		// without casting
	}
	(
		CompactCiphertextListExpansionKind::RequiresCasting,
		ShortintCompactCiphertextListCastingMode::CastIfNecessary(casting_key)
	) => {
		// no casting
	}
	(
		CompactCiphertextListExpansionKind::RequiresCasting,
		ShortintCompactCiphertextListCastingMode::NoCasting
	) => {
		// error
	}
}
```

IceTDrinker: I feared the double match would complicate the number of cases to keep track, so I manage the error case, do you think the nested match hurt readability ?

tmontaigu: For me yes, but its a tiny bit

tmontaigu (approved):

PR #1227, opened 2024-06-12
Multi GPU: dispatch/gather inputs and outputs to the ks/pbs on all GPUs

agnesLeroy:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/598

### Check-list:


backends/tfhe-cuda-backend/cuda/src/integer/cmux.cuh
```diff
@@ -89,6 +89,17 @@ __host__ void host_integer_radix_cmux_kb(
...
 
+  auto message_extract_lut_f = [params](Torus x) -> Torus {
+    return x % params.message_modulus;
+  };
+  generate_device_accumulator<Torus>(
+      streams[0], gpu_indexes[0],
+      mem_ptr->message_extract_lut->get_lut(gpu_indexes[0], 0),
+      params.glwe_dimension, params.polynomial_size, params.message_modulus,
+      params.carry_modulus, message_extract_lut_f);
+  mem_ptr->message_extract_lut->broadcast_lut(streams, gpu_indexes,
```
pdroalves, line 100: The number of synchronizations in `broadcast_lut()` can impact performance here. Maybe we can implement a lazy sync by introducing a `sync_streams` parameter in that function. In cases like this we could synchronize all the streams only once inside `integer_radix_apply_univariate_lookup_table_kb`. Wdyt?

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -153,28 +155,69 @@ __host__ void integer_radix_apply_univariate_lookup_table_kb(
...
+  /// outputs
+  std::vector<Torus *> lwe_array_in_vec = lut->lwe_array_in_vec;
+  std::vector<Torus *> lwe_after_ks_vec = lut->lwe_after_ks_vec;
+  std::vector<Torus *> lwe_after_pbs_vec = lut->lwe_after_pbs_vec;
+  std::vector<Torus *> lwe_trivial_indexes_vec = lut->lwe_trivial_indexes_vec;
+
   auto active_gpu_count = get_active_gpu_count(num_radix_blocks, gpu_count);
-  for (uint i = 0; i < active_gpu_count; i++) {
-    cuda_synchronize_stream(streams[i], gpu_indexes[i]);
+  if (active_gpu_count == 1) {
```
pdroalves, line 166: Ideally we should not need this conditional. We should have generic functions that would silently skip gather/scatter if `active_gpu_count == 1`. That would simplify the code and thus improve readability. However, I don't think this is a priority and I want to merge this PR ASAP. Should we open an issue for our future selfs?

backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -224,11 +226,12 @@ __host__ void host_integer_sum_ciphertexts_vec_kb(
   auto message_modulus = mem_ptr->params.message_modulus;
   auto carry_modulus = mem_ptr->params.carry_modulus;
   auto num_blocks = num_blocks_in_radix;
-  auto big_lwe_size = mem_ptr->params.big_lwe_dimension + 1;
+  auto big_lwe_dimension = mem_ptr->params.big_lwe_dimension;
+  auto big_lwe_size = big_lwe_dimension + 1;
   auto glwe_dimension = mem_ptr->params.glwe_dimension;
   auto polynomial_size = mem_ptr->params.polynomial_size;
   auto lwe_dimension = mem_ptr->params.small_lwe_dimension;
```
pdroalves, line 233: Probably we should rename this to `small_lwe_dimension` for consistency.

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
```diff
@@ -0,0 +1,183 @@
...
+      cuda_synchronize_stream(streams[i], gpu_indexes[i]);
+}
+template <typename Torus>
+void multi_gpu_release_lwe(cudaStream_t *streams, uint32_t *gpu_indexes,
+                           std::vector<Torus *> &vec,
+                           bool sync_threads = true) {
+
+  for (uint i = 0; i < vec.size(); i++) {
+    cuda_drop_async(vec[i], streams[i], gpu_indexes[i]);
+    if (sync_threads)
```
pdroalves, line 177: We probably want to launch all asynchronous memory release and then have a second loop synchronizing. Doing that we may have released some arrays before we get to the related synchronization call.

Of course, performance impact would not be huge but it's a small sand grain.

tfhe/benches/integer/bench.rs
```diff
@@ -1983,32 +1983,32 @@ mod cuda {
 
     criterion_group!(
         default_cuda_ops,
-        cuda_neg,
-        cuda_sub,
-        cuda_unsigned_overflowing_sub,
-        cuda_unsigned_overflowing_add,
+        //cuda_neg,
```
pdroalves, line 1986: What's the motivation here? Don't we care about these benchmarks?

pdroalves (approved):
I like what I see. I have a few questions and a optimization idea, but feel free to merge as is as long as the CI allows you :-)

tfhe/benches/integer/bench.rs
agnesLeroy: Oh wow good catch I forgot to remove this commit in the rebase. Will revert this change.

backends/tfhe-cuda-backend/cuda/src/integer/cmux.cuh
agnesLeroy: Actually do we need to synchronize streams in broadcast_lut?

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
agnesLeroy: I don't understand what you mean: here we release memory then synchronize, for each GPU. It seems to me this is equivalent to doing:
```
  for (uint i = 0; i < vec.size(); i++) {
      cuda_drop_async(vec[i], streams[i], gpu_indexes[i]);
  }
  if (sync_threads) {
      for (uint i = 0; i < vec.size(); i++) {
          cuda_synchronize_stream(streams[i], gpu_indexes[i]);
      }
  }
  vec.clear();
```
Am I wrong?

backends/tfhe-cuda-backend/cuda/src/integer/cmux.cuh
agnesLeroy: Ok it seems that we do need it :sweat_smile:

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
agnesLeroy: Hmm ok I see what you mean. Indeed it should have a very small impact on performance, I'll make the change anyway.

agnesLeroy: Actually, is it legitimate to call vec.clear() if we don't synchronize the threads? Isn't it going to remove some pointers before they're dropped?

backends/tfhe-cuda-backend/cuda/src/integer/cmux.cuh
pdroalves: Good point. Probably not since any kernel dependency will be added to the same stream as the copy. We have to test to be sure, but we probably would be safe.

agnesLeroy: I'll open an issue to do that in another PR, I'd like to merge this one asap :sweat_smile:

agnesLeroy: https://github.com/zama-ai/tfhe-rs-internal/issues/664

PR #1234, opened 2024-06-13
feat(gpu): Implement unsigned_overflowing_scalar_add

bbarbakadze:
### PR content/description
Implement overflowing scalar add
### Check-list:

### DIFF
--- a/tfhe/benches/integer/bench.rs
+++ b/tfhe/benches/integer/bench.rs
@@ -1701,6 +1701,12 @@ mod cuda {
         rng_func: div_scalar
     );
 
+    define_cuda_server_key_bench_clean_input_scalar_fn!(
+        method_name: unchecked_unsigned_overflowing_scalar_add,
+        display_name: unsigned_overflowing_scalar_add,
+        rng_func: default_scalar
+    );
+
     //===========================================
     // Default
     //===========================================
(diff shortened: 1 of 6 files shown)


agnesLeroy (commented):
We should also add the overflowing entry points to the hl api

tfhe/benches/integer/bench.rs
```diff
@@ -1723,6 +1728,11 @@ mod cuda {
         display_name: overflowing_add
     );
 
+    define_cuda_server_key_bench_clean_input_fn!(
```
agnesLeroy, line 1731: Same here

agnesLeroy (commented):
Thanks Beka, looks good to me but @tmontaigu will know better than me for the first commit that modifies the noise level.

tmontaigu:
I don't think the first commit is right with regards to how the GPU works compared to CPU

On the CPU, when the input to a PBS is a trivial ciphertext, we actually skip the PBS and apply the LUT directly via clear operations, so the output is still a trivial ciphertext without noise
https://github.com/zama-ai/tfhe-rs/blob/e825277219cc142bf13959b486192ddc97711445/tfhe/src/shortint/server_key/mod.rs#L750

I don't think you do that on GPU, meaning that you apply PBS on trivial inputs meaning the output has noise and is no longer a trivial

agnesLeroy:
Ok so we'll remove the first commit and in case the gpu feature is activated we won't check if NoiseLevel is zero in case a trivial ciphertext is given as input to unsigned_overflowing_scalar_add

agnesLeroy:
We just pushed the change

tmontaigu:
Hum ok I looked at the wrong think, I check again

tmontaigu (approved):

PR #1235, opened 2024-06-13
feat(hlapi): bind match_value/match_value_or

tmontaigu:
### Check-list:


tfhe/src/high_level_api/integers/unsigned/base.rs
```diff
@@ -477,6 +479,126 @@ where
...
+        OutId: FheUintId,
+    {
+        global_state::with_internal_keys(|key| match key {
+            InternalServerKey::Cpu(cpu_key) => {
+                let (result, matched) = cpu_key
+                    .pbs_key()
+                    .match_value_parallelized(&self.ciphertext.on_cpu(), matches);
+                let result = cpu_key
+                    .pbs_key()
+                    .cast_to_unsigned(result, OutId::num_blocks(cpu_key.message_modulus()));
```
mayeul-zama, line 534: Why do the cast in this operation?
A user could do it after, right?

tmontaigu: The number of blocks of the output of the interger function depends on the max possible output value that is in the MatchValue map allowing to map 16blocks value to values with less or more blocks

Since the HLAPI API types are not as dynamic, we have to let the user cast to the wanted type

tfhe/src/integer/server_key/radix_parallel/vector_find.rs
```diff
@@ -112,7 +114,7 @@ impl ServerKey {
             // Thus, in that case, the returned value is always 0 regardless of ct's value,
             // but we still have to see if the input matched something
             (
-                self.create_trivial_radix(0, ct.blocks.len()),
+                self.create_trivial_radix(0, num_blocks_to_represent_values),
                 BooleanBlock::new_unchecked(
```
IceTDrinker, line 118: maybe a check in tests to verify this ?

tmontaigu: what check ?

IceTDrinker: the number of blocks of the output is what we expect

IceTDrinker: and not the number of blocks of the input

tmontaigu: Ah, I did not see the "in tests" part 🥶

IceTDrinker (approved):

IceTDrinker:
fail on a noise check it seems

tmontaigu:
Another case of writing tests is harder than the function itself

We check that blocks have nominal noise level, but the function may append trivial  blocks, which have a zero noise level, which is not nominal

IceTDrinker:
sheesh, at most nominal noise then I guess

IceTDrinker (approved):

PR #1238, opened 2024-06-14
chore(ci): reduce number of keys generated in keycache

soonum:
The keys/ folder was containing lots of unused keys in tests and thus was eating around 20GB of disk space.


IceTDrinker (approved):
Thanks a lot !

PR #1241, opened 2024-06-15
chore(ci): manage the memory issues we are seeing

IceTDrinker:


soonum (approved):
Thanks, hope the CI will be more stable/

agnesLeroy (commented):
Should we try to increase again the number of threads used in the tests? Second line of this code block:
```
if [[ "${BIG_TESTS_INSTANCE}" == TRUE ]]; then
    test_threads="$((num_cpu_threads * 1 / 4))"
    doctest_threads="${num_cpu_threads}"
else
    test_threads="${small_instance_n_threads}"
    doctest_threads="${num_cpu_threads}"
fi
```
in scripts/integer-tests.sh 

We could maybe put it back to `test_threads="$((num_cpu_threads * 1 / 2))"` but for the GPU tests we can't go further otherwise we run out of memory on the GPU.

IceTDrinker:
ah is it at 1/4 currently ?

IceTDrinker:
I would probably merge this to unblock other PRs and then re increase CPU count to 1/2 if needed

agnesLeroy:
Ok fine by me!

PR #1250, opened 2024-06-17
chore(ci): increase integer test thread count

IceTDrinker:


scripts/integer-tests.sh
```diff
@@ -100,7 +100,8 @@ else
 fi
 
 if [[ "${BIG_TESTS_INSTANCE}" == TRUE ]]; then
-    test_threads="$((num_cpu_threads * 1 / 4))"
+    # Big test instances have 512 GB of RAM, 4.7 GB keys, keep some headroom
+    test_threads=100
```
agnesLeroy, line 104: This won't work for the GPU tests, we'll run out of memory. It used to be 
```
test_threads="$((num_cpu_threads * 1 / 2))"
```
and we know this works. Otherwise we could have a profile to set the number of threads for GPU workflows specifically.

IceTDrinker: ah right sorry, hmm, yeah then I guess we need to have a difference for GPU and CPU, given we already have special cases IIRC in the filters for GPU, it's not critical, I'll see how to update

IceTDrinker:
still some sigkills anyways

IceTDrinker:
solution will be to remove the mandatory 4_4 tests in PRs and have them as nightly

PR #1252, opened 2024-06-17
feat(all): add `safe_serialize_versioned`

nsarlin-zama:

### PR content/description

Adds a few methods for versioned serializations:
- `safe_serialize_versioned`
- `safe_deserialize_versioned`
- `safe_deserialize_conformant_versioned`

To avoid conflict with the `SERIALIZATION_VERSION`, a different number `VERSIONING_VERSION` is used to record the version of the versioning scheme that has been used. A specific header type has been created that will be embed inside the serialized packet and records if the message is versioned or not. To be compatible with other message serialized with previous 0.6, this header is not used for messages that are not versioned.

However this might be useless in the future since the SERIALIZATION_VERSION should not evolve anymore when versioning will be used by default.

### Check-list:

### DIFF
--- a/tfhe/c_api_tests/test_high_level_integers.c
+++ b/tfhe/c_api_tests/test_high_level_integers.c
@@ -430,6 +430,67 @@ int uint8_safe_serialization(const ClientKey *client_key, const ServerKey *serve
   return ok;
 }
 
+int uint8_safe_serialization_versioned(const ClientKey *client_key, const ServerKey *server_key) {
+  int ok;
+  CompactFheUint8 *lhs = NULL;
+  CompactFheUint8 *deserialized_lhs = NULL;
+  CompactFheUint8 *result = NULL;
+  DynamicBuffer value_buffer = {.pointer = NULL, .length = 0, .destructor = NULL};
+  DynamicBuffer cks_buffer = {.pointer = NULL, .length = 0, .destructor = NULL};
+  DynamicBufferView deser_view = {.pointer = NULL, .length = 0};
+  ClientKey *deserialized_client_key = NULL;
+
+  const uint64_t max_serialization_size = UINT64_C(1) << UINT64_C(20);
+
+  uint8_t lhs_clear = 123;
+
+  ok = client_key_serialize(client_key, &cks_buffer);
+  assert(ok == 0);
+
+  deser_view.pointer = cks_buffer.pointer;
+  deser_view.length = cks_buffer.length;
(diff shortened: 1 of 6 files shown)


tfhe/src/safe_deserialization.rs
```diff
@@ -15,13 +18,73 @@ use serde::Serialize;
...
+    /// Checks the validity of a versioned message
+    fn check_versioned<T: Named>(&self) -> Result<(), String> {
+        if self.mode != SerializationMode::Versioned {
+            return Err(
+                "On deserialization, expected versioned type but got unversioned one".to_string(),
+            );
+        }
+
+        if self.version != VERSIONING_VERSION {
+            return Err(format!(
```
IceTDrinker, line 57: add a comment + why it's currently like this, but in the future we will want to load old formats and update it to the new system

tfhe/src/safe_deserialization.rs
```diff
@@ -150,14 +285,47 @@ mod test_shortint {
         let dec = ck.decrypt(&ct2);
         assert_eq!(msg, dec);
     }
+
+    #[test]
+    fn safe_desererialization_ct_versioned() {
```
IceTDrinker, line 290: typo in safe_deserialize name

IceTDrinker (commented):
small comments

tfhe/src/safe_deserialization.rs
nsarlin-zama: done

tfhe/src/safe_deserialization.rs
nsarlin-zama: fixed

tfhe/c_api_tests/test_high_level_integers.c
```diff
@@ -430,6 +430,67 @@ int uint8_safe_serialization(const ClientKey *client_key, const ServerKey *serve
...
+
+  uint8_t clear;
+  ok = fhe_uint8_decrypt(expanded, deserialized_client_key, &clear);
+  assert(ok == 0);
+
+  assert(clear == lhs_clear);
+
+  if (value_buffer.pointer != NULL) {
+    destroy_dynamic_buffer(&value_buffer);
+  }
```
IceTDrinker, line 485: normally destroy_dynamic_buffer does not destroy if the pointer is NULL

IceTDrinker (approved):
thanks !

PR #1260, opened 2024-06-19
fix(shortint): count PBS in trivial_many_lut

tmontaigu:
We forgot increase the number of PBSes when doing
a many_lut PBS on a trivial input.

This fixes that, and changes the many_lut test to
test the result on trivial ciphertext and also test that the PBS count is correct.


IceTDrinker (approved):
Thanks !

PR #1265, opened 2024-06-20
chore(ci): update gpu ec2 ami with rustup snap package removed

soonum:
Rustup was installed using snap and could clash in the workflow notably on cargo-clippy calls.


IceTDrinker (approved):
Thanks !

PR #1284, opened 2024-06-24
chore(ci): run build workflow on large instances

soonum:


IceTDrinker:
workflow lint not happy, runner labels likely need to be added to the whitelist

IceTDrinker (approved):

PR #1285, opened 2024-06-24
fix clippy doctests

mayeul-zama:


Makefile
```diff
@@ -266,6 +266,13 @@ clippy: install_rs_check_toolchain
 		--features=$(TARGET_ARCH_FEATURE),boolean,shortint,integer \
 		-p $(TFHE_SPEC) -- --no-deps -D warnings
 
+.PHONY: rustdoc_clippy # Run clippy lints enabling the boolean, shortint, intege
```
IceTDrinker, line 269: the comment here is cut before the end

scripts/clippy_driver.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -eu
+
```
IceTDrinker, line 4: this is not required anymore right ?

IceTDrinker (approved):
Thanks, did some minor tweaks after the toolchain update was merged

PR #1286, opened 2024-06-24
Update toolchain

mayeul-zama:


tfhe/Cargo.toml
```diff
@@ -324,3 +330,10 @@ required-features = ["integer", "internal-keycache"]
 
 [lib]
 crate-type = ["lib", "staticlib", "cdylib"]
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(__profiling)',
```
mayeul-zama, line 336: Not sure about this one
There's a feature `__profiling` which may be misused
@sarah-ek do you know if the pattern `#[cfg_attr(__profiling, inline(never))]` works as expected?

tfhe/Cargo.toml
```diff
@@ -324,3 +330,10 @@ required-features = ["integer", "internal-keycache"]
...
+    'cfg(bench)',
```
IceTDrinker, line 337: bench is not whitelisted by default ? 🤔

backends/tfhe-cuda-backend/src/cuda_bind.rs
```diff
@@ -364,16 +346,15 @@ extern "C" {
...
-    /// `(input_lwe_dimension + 1) * input_lwe_ciphertext_count` that should have been allocated on
-    /// the GPU before calling this function, and that will hold the result of the computation.
-    /// - `lwe_array_in` is the LWE ciphertext vector used as input, it should have been
-    /// allocated and initialized before calling this function. It has the same size as the output
-    /// array.
+    /// - `lwe_array_out` is an array of size `(input_lwe_dimension + 1) *
+    ///   input_lwe_ciphertext_count` that should have been allocated on the GPU before calling this
+    ///   function, and that will hold the result of the computation.
+    /// - `lwe_array_in` is the LWE ciphertext vector used as input, it should have been allocated
+    ///   and initialized before calling this function. It has the same size as the output array.
```
IceTDrinker, line 353: @mayeul-zama does rust analyzer automatically or does rustfmt automatically format this ?

tfhe/src/core_crypto/algorithms/misc.rs
```diff
@@ -218,6 +218,7 @@ mod test {
         for _ in 0..NB_TESTS {
             let num: f64 = rng.gen();
             let mut denom = 0.0f64;
+            #[allow(clippy::while_float)]
```
IceTDrinker, line 221: what's this ?

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -345,6 +345,7 @@ impl<T: Numeric> CudaVec<T> {
         }
     }
 
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
IceTDrinker, line 348: feels like a false positive ?

tfhe/src/high_level_api/integers/signed/inner.rs
```diff
@@ -207,7 +208,7 @@ impl RadixCiphertext {
     }
 
     #[inline]
-    #[allow(clippy::unused_self)]
+    #[allow(clippy::unused_self, clippy::needless_pass_by_ref_mut)]
     pub(crate) fn move_to_device_of_server_key_if_set(&mut self) {
         #[cfg(feature = "gpu")]
         if let Some(device) = global_state::device_of_internal_keys() {
```
IceTDrinker, line 214: does this require the cfg(feature) actually ?

tfhe/src/js_on_wasm_api/js_high_level_api/integers.rs
```diff
@@ -1022,7 +1022,7 @@ define_expander_get_method!(
 #[wasm_bindgen]
 impl CompactCiphertextListExpander {
     #[wasm_bindgen]
-    pub fn get_bool(&mut self, index: usize) -> Result<FheBool, JsError> {
+    pub fn get_bool(&self, index: usize) -> Result<FheBool, JsError> {
         catch_panic_result(|| {
             self.0
                 .get::<crate::FheBool>(index)
```
IceTDrinker, line 1028: maybe a better fix is required here as we have an unwrap ?

backends/tfhe-cuda-backend/src/cuda_bind.rs
mayeul-zama: IIRC I tried `cargo clippy --fix` and it didn't work

tfhe/src/core_crypto/algorithms/misc.rs
mayeul-zama: A float comparison in a loop
https://rust-lang.github.io/rust-clippy/master/#/while_float

tfhe/src/core_crypto/gpu/vec.rs
mayeul-zama: I think it's because we use mut raw pointer which we don't need a &mut reference to access

tfhe/src/high_level_api/integers/signed/inner.rs
mayeul-zama: Features defined in Cargo.toml are know to cargo so are marked as expected

tfhe/src/js_on_wasm_api/js_high_level_api/integers.rs
mayeul-zama: Added a fix

backends/tfhe-cuda-backend/src/cuda_bind.rs
IceTDrinker: rust fmt by any chance ? and so does it auto indent with rust analyze ?

otherwise I'm inclined to ignore that lint as it will just be a nuisance

tfhe/src/core_crypto/algorithms/misc.rs
IceTDrinker: weird lint given how we use it

tfhe/src/core_crypto/gpu/vec.rs
IceTDrinker: ok but it's better to have the &mut still here to enforce a single mutable access, thanks

tfhe/src/high_level_api/integers/signed/inner.rs
IceTDrinker: what I mean is, does the move to device require the GPU feature here ?

tfhe/Cargo.toml
IceTDrinker: can you see if there has been an update to cfgs for bench ? otherwise I would report this one as a false positive to the rust/clippy projects

Maybe some people already reported this

backends/tfhe-cuda-backend/src/cuda_bind.rs
mayeul-zama: No

I think it's more readable now
I don't think it will cause significant friction, only 19 files are affected

IceTDrinker: does not change anything in docs.rs and if we have to do it manually it's going to be super annoying

IceTDrinker: how did you fix those then, manually ?

mayeul-zama: Yes

The time to format new doc should be much smaller than the one to write it

tfhe/src/high_level_api/integers/signed/inner.rs
mayeul-zama: Removed

tfhe/Cargo.toml
mayeul-zama: What do you mean by an update?

tfhe/Cargo.toml
sarah-ek: no. it should be feature = "__profiling"
nice catch

tfhe/Cargo.toml
IceTDrinker: i.e. that bench is/was a legit cfg but it's now missing, has it been complitely removed or is it some sort of mistake on the clippy side of things ?

backends/tfhe-cuda-backend/src/cuda_bind.rs
IceTDrinker: I'll see when using it, but wasting time on that is really unnecessary as it gets processed into an HTML anyways, hopefully they include some auto fix for clippy of cargo fmt

IceTDrinker (approved):
Thanks !

PR #1287, opened 2024-06-25
chore(gpu): allocate pinned host memory when possible

agnesLeroy:
This speeds up the copy to the GPU

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -530,7 +530,8 @@ template <typename Torus> struct int_radix_lut {
           num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0]);
       lwe_trivial_indexes = (Torus *)cuda_malloc_async(
           num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0]);
-      auto h_lwe_indexes = (Torus *)malloc(num_radix_blocks * sizeof(Torus));
+      Torus *h_lwe_indexes;
+      cudaMallocHost((void **)&h_lwe_indexes, num_radix_blocks * sizeof(Torus));
 
       for (int i = 0; i < num_radix_blocks; i++)
         h_lwe_indexes[i] = i;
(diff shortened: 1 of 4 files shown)


backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -525,7 +525,8 @@ template <typename Torus> struct int_radix_lut {
           num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0]);
       lwe_trivial_indexes = (Torus *)cuda_malloc_async(
           num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0]);
-      auto h_lwe_indexes = (Torus *)malloc(num_radix_blocks * sizeof(Torus));
+      Torus *h_lwe_indexes;
+      cudaMallocHost((void **)h_lwe_indexes, num_radix_blocks * sizeof(Torus));
```
guillermo-oyarzun, line 529: I think you are missing an & in all the calls
cudaMallocHost((void **)&h_lwe_indexes, num_radix_blocks * sizeof(Torus));

agnesLeroy: Indeed, sorry for this! :sweat_smile: I'll make the CI green and let you know when it's ok :slightly_smiling_face:

guillermo-oyarzun (approved):

pdroalves (commented):
Do we know pinned memory will bring any advantage? AFAIK it is more expensive than `malloc` and only really useful if you repeatedly run copies using the pinned pointer, which I don't think happens often in our case.

agnesLeroy:
> Do we know pinned memory will bring any advantage? AFAIK it is more expensive than `malloc` and only really useful if you repeatedly run copies using the pinned pointer, which I don't think happens often in our case.

My intention was to run the benchmarks to see the effect. I've rebased on main and triggered the multi-bit 64 bit ones we should get the results soon :slightly_smiling_face:

agnesLeroy:
@pdroalves so this doesn't seem to have any effect on performance: https://zama.grafana.net/d/b1d1e27b-a00c-46ed-b188-3dbdea9db83a/operation-performance-evolution-multi-branches?var-hardware=n3-H100x1&var-operation=add&var-operation=div_mod&var-operation=mul&var-branch=main&var-branch=al%2Fpinned_host_memory&var-parameters=PARAM_GPU_MULTI_BIT_MESSAGE_2_CARRY_2_GROUP_3_KS_PBS&var-bit_size=64

agnesLeroy:
I' running complementary benchmarks. Actually I think this will have some effect when using multiple GPUs, based on this: https://medium.com/gpgpu/multi-gpu-programming-6768eeb42e2c

agnesLeroy:
@pdroalves so I've checked single & multi-GPU performance on this branch compared to main and there is no improvement. If anything we seem to loose 1ms in the multi-gpu benchmarks for all precisions & operations. I guess we don't want to merge this.
single-gpu run: https://github.com/zama-ai/tfhe-rs/actions/runs/9693223295/job/26748315432
multi-gpu run: https://github.com/zama-ai/tfhe-rs/actions/runs/9695743021
(Benchmarks on `main` are still ongoing, they haven't been uploaded to Grafana with the fix in the FFT yet).

pdroalves:
Yeah, don't think our case benefits from pinned memory.

PR #1288, opened 2024-06-25
chore(gpu): use byte add, update rust msrv

agnesLeroy:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/549
### Check-list:

### DIFF
--- a/tfhe/Cargo.toml
+++ b/tfhe/Cargo.toml
@@ -17,7 +17,7 @@ exclude = [
     "/js_on_wasm_tests/",
     "/web_wasm_parallel_tests/",
 ]
-rust-version = "1.73"
+rust-version = "1.75"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/tfhe/src/core_crypto/gpu/slice.rs
+++ b/tfhe/src/core_crypto/gpu/slice.rs
@@ -158,12 +158,8 @@ where
             None
         } else {
             // Shift ptr
-            let shifted_ptr: *mut c_void = unsafe {
-                self.ptrs[gpu_index as usize]
-                    .cast::<u8>()
-                    .add(start * std::mem::size_of::<T>())
-                    .cast()
-            };
+            let shifted_ptr: *mut c_void =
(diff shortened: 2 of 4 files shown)


tfhe/Cargo.toml
```diff
@@ -17,7 +17,7 @@ exclude = [
     "/js_on_wasm_tests/",
     "/web_wasm_parallel_tests/",
 ]
-rust-version = "1.73"
+rust-version = "1.75"
```
IceTDrinker, line 20: we will likely take the most recent rust when we agree we go for v1, this is good for now 👍

tfhe/src/core_crypto/gpu/slice.rs
```diff
@@ -158,12 +158,8 @@ where
...
         } else {
             // Shift ptr
-            let shifted_ptr: *mut c_void = unsafe {
-                self.ptrs[gpu_index as usize]
-                    .cast::<u8>()
-                    .add(start * std::mem::size_of::<T>())
-                    .cast()
-            };
+            let shifted_ptr: *mut c_void =
+                unsafe { self.ptrs[gpu_index as usize].byte_add(start * std::mem::size_of::<T>()) };
```
IceTDrinker, line 162: wouldn't https://doc.rust-lang.org/std/primitive.pointer.html#method.add-1 be better to avoid having to manually multiplying by some constant ?

IceTDrinker (commented):
Just a question on which pointer primitive to add, it seems the add primitive does what you want not in byte size but in T counts and does not require a new MSRV if I read things correctly

tmontaigu:
We could use https://doc.rust-lang.org/std/primitive.pointer.html#method.wrapping_byte_add-1 which does the same thing but is not unsafe

tfhe/src/core_crypto/gpu/slice.rs
IceTDrinker: ah damn missed that

agnesLeroy:
@tmontaigu I just pushed the change;

IceTDrinker (approved):
Looks good 👍

tmontaigu (approved):

PR #1297, opened 2024-06-25
Am/chore/container metadata

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/194

### PR content/description

Metadata for container implementations were tuple structs but it could be unclear what value should be provided, changed all tuple metadata to structs with named fields.


mayeul-zama (approved):
Thanks!

PR #1301, opened 2024-06-26
chore(ci): run full multi-bit gpu benchmarks on demand

soonum:


agnesLeroy:
Interesting, how will this run we can configure the run when triggering the workflow?

soonum:
> Interesting, how will this run we can configure the run when triggering the workflow?

I don't quite get your question.

agnesLeroy (approved):

PR #1325, opened 2024-06-28
chore(ci): optimize triggering of gpu workflows

soonum:


.github/workflows/aws_tfhe_integer_gpu_tests.yml
```diff
@@ -16,10 +16,17 @@ on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
+  push:
+    branches:
+      - main
```
IceTDrinker, line 21: we run GPU test on push to main ?

.github/workflows/aws_tfhe_multi_gpu_tests.yml
```diff
@@ -85,9 +87,10 @@ jobs:
             echo "HOME=/home/ubuntu";
           } >> "${GITHUB_ENV}"
 
-      - name: Run core crypto, integer and internal CUDA backend tests
+      # No need to test core_crypto and classic PBS in integer since it's already tested on single GPU.
+      - name: Run multi-bit CUDA integer tests
         run: |
-          make test_gpu
```
IceTDrinker, line 90: we remove some tests ?

.github/workflows/aws_tfhe_signed_integer_gpu_tests.yml
```diff
@@ -16,10 +16,17 @@ on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
+  push:
+    branches:
+      - main
```
IceTDrinker, line 21: test on push to main ?

IceTDrinker: why ?

.github/workflows/aws_tfhe_multi_gpu_tests.yml
soonum: Yes, seen with @agnesLeroy

.github/workflows/aws_tfhe_integer_gpu_tests.yml
soonum: It's not mandatory as long we have a nightly run I guess.

.github/workflows/aws_tfhe_multi_gpu_tests.yml
agnesLeroy: Yes there is no multi-GPU support in core-crypto and in the cuda backend, only at the integer level. We only need to test the multi-bit integer ops, since the multi-gpu logic is implemented at the integer level if the classical PBS breaks we'll see it in single GPU tests.

.github/workflows/aws_tfhe_integer_gpu_tests.yml
agnesLeroy: I think those we want to run on approve, we have the H100 tests for signed & unsigned operations that should run on push to the PRs

.github/workflows/aws_tfhe_signed_integer_gpu_tests.yml
agnesLeroy: I think those we want to run on approve, we have the H100 tests for signed & unsigned operations that should run on push to the PRs

.github/workflows/aws_tfhe_integer_gpu_tests.yml
soonum: So if I understand correctly, there is no need to trigger the workflow on push to `main`, only in PR on approval ?

agnesLeroy: What I mean to say if I think I prefer to have them run on approve instead of nightly or push to main I think.

.github/workflows/aws_tfhe_multi_gpu_tests.yml
IceTDrinker: ah indeed

agnesLeroy (approved):

agnesLeroy (approved):

IceTDrinker:
I saw a weird WASM test failure, it did not quite look like a flaky though

IceTDrinker (approved):

IceTDrinker:
is this ready soon ? I see the cuda backend tests failed still with a CUDA out of memory error, so I'm guessing test threads is still too large ?

IceTDrinker:
also needs a rebase for the wasm fix if needed

IceTDrinker (approved):
Let's see if tests run to completion for GPU

PR #1349, opened 2024-07-10
chore(ci): check labels to launch only on approval

soonum:
Any other label added other than "approved" would trigger these workflow which is not desired.


IceTDrinker (approved):
Thanks !

PR #1363, opened 2024-07-15
chore(ci): fix the no internal test patch

IceTDrinker:
- condition was not precise enough and we were still running tests on push to internal main


IceTDrinker:
I need to fix something before this can be approved

IceTDrinker:
good to review @soonum

soonum (approved):

PR #1365, opened 2024-07-16
chore(ci): fix test filtering for gpu multi-bit parameters set

soonum:


scripts/test_filtering.py
```diff
@@ -98,8 +98,12 @@
 def filter_integer_tests(input_args):
     multi_bit_filter = "_multi_bit" if input_args.multi_bit else ""
     backend_filter = ""
+    gpu_params_filter = ""
```
IceTDrinker, line 101: could we maybe update the multibit filter depending on the gpu input param ?

soonum: You're right!

IceTDrinker (approved):

PR #1366, opened 2024-07-16
Am/chore/prepare alpha

IceTDrinker:


soonum (approved):

IceTDrinker:
this does not work because of missing version in the workspace dep I believe

nsarlin-zama (approved):

PR #1369, opened 2024-07-17
chore(gpu): clean overflowing sub memory

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer.h
@@ -1157,7 +1157,7 @@ template <typename Torus> struct int_sc_prop_memory {
   }
 };
 
-template <typename Torus> struct int_single_borrow_prop_memory {
+template <typename Torus> struct int_overflowing_sub_memory {
   Torus *generates_or_propagates;
   Torus *step_output;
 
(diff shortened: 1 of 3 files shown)


backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -1157,7 +1157,7 @@ template <typename Torus> struct int_sc_prop_memory {
   }
 };
 
-template <typename Torus> struct int_single_borrow_prop_memory {
+template <typename Torus> struct int_overflowing_sub_memory {
```
bbarbakadze, line 1160: why  name it as `int_overflowing_sub_memory`, this structure  is for a function `unsigned_overflowing_propagate_subtraction_borrow`, old name is really not good, so we had to update it but maybe a different name to specify that it is for `unsigned_overflowing_propagate_subtraction_borrow`

backends/tfhe-cuda-backend/cuda/include/integer.h
```diff
@@ -1169,10 +1169,10 @@ template <typename Torus> struct int_single_borrow_prop_memory {
 
   int_radix_params params;
 
-  int_single_borrow_prop_memory(cudaStream_t *streams, uint32_t *gpu_indexes,
-                                uint32_t gpu_count, int_radix_params params,
-                                uint32_t num_radix_blocks,
-                                bool allocate_gpu_memory) {
+  int_overflowing_sub_memory(cudaStream_t *streams, uint32_t *gpu_indexes,
```
bbarbakadze, line 1172: same about naming

backends/tfhe-cuda-backend/cuda/include/integer.h
bbarbakadze: also I don't know any other function that's using `unsigned_overflowing_propagate_subtraction_borrow` and can't imagine what else can use it, so maybe it`s fine to name it as `int_overflowing_sub_memory`

bbarbakadze (commented):
Thank you, looks good to me, maybe think a bit more about naming.

backends/tfhe-cuda-backend/cuda/include/integer.h
agnesLeroy: Yeah this is really the only operation where this is used, so I know it's a bit inconsistent but it seemed easier to read like this to me.

bbarbakadze (approved):

PR #1398, opened 2024-07-22
refactor(core): make GGSW encryption consistent

IceTDrinker:
- functions take un-encoded values, reflect that by taking Cleartext instead of Plaintext


mayeul-zama (approved):
Thanks!

PR #1418, opened 2024-07-25
Mz/enable all doctest warnings

mayeul-zama:


tfhe/docs/fundamentals/encrypt-data.md
```diff
@@ -13,12 +13,12 @@ use tfhe::{generate_keys, ConfigBuilder, FheUint8};
...
-    let (client_key, server_key) = generate_keys(config);
+    let (client_key, _server_key) = generate_keys(config);
 
     let clear_a = 27u8;
     let clear_b = 128u8;
 
-    let a = FheUint8::encrypt(clear_a, &client_key);
-    let b = FheUint8::encrypt(clear_b, &client_key);
+    let _a = FheUint8::encrypt(clear_a, &client_key);
+    let _b = FheUint8::encrypt(clear_b, &client_key);
```
IceTDrinker, line 22: let's keep the allow dead code in the tfhe/src/test_user_docs.rs otherwise the outside facing documentation will look very weird

tfhe/src/lib.rs
```diff
@@ -2,6 +2,9 @@
 //!
 //! TFHE-rs is a fully homomorphic encryption (FHE) library that implements Zama's variant of TFHE.
 
+// Enable all warnings in doctests
+// https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#showing-warnings-in-doctests
+#![doc(test(attr(warn(unused))))]
```
IceTDrinker, line 7: so as I said somewhere else, let's not have that in the user facing docs, otherwise it just looks ridiculous, I'm not convinced for doctests in general either, it looks very bad to have to essentially to prefix everything with _, but I'm guessing we are missing the unused imports if we allow that in that case ?

IceTDrinker (changes requested):
Keep the changes that are relevant but we don't want to have to rename our variables for doc tests and our md docs

tfhe/docs/references/fine-grained-apis/boolean/parameters.md
```diff
@@ -25,8 +25,9 @@ You can also create your own set of parameters. This is an `unsafe` operation as
 use tfhe::boolean::prelude::*;
 
 fn main() {
-// WARNING: might be insecure and/or incorrect
-// You can create your own set of parameters
+    // WARNING: might be insecure and/or incorrect
+    // You can create your own set of parameters
+    #[allow(unused_unsafe)]
     let parameters = unsafe {
         BooleanParameters::new(
```
IceTDrinker, line 32: I think we can remove the unsafe here, the function is marked as safe

tfhe/docs/references/fine-grained-apis/shortint/parameters.md
```diff
@@ -55,6 +55,9 @@ use tfhe::shortint::prelude::*;
 use tfhe::shortint::parameters::DynamicDistribution;
 
 fn main() {
+    // WARNING: might be insecure and/or incorrect
+    // You can create your own set of parameters
+    #[allow(unused_unsafe)]
```
IceTDrinker, line 60: same remark no need for unsafe just a bad habit from the concrete core days

tfhe/src/high_level_api/prelude.rs
```diff
@@ -3,6 +3,7 @@
 //!
 //! It is meant to be glob imported:
 //! ```
+//! #[allow(unused_imports)]
```
IceTDrinker, line 6: can you add a `# ` (with the sapce) in front ? IIRC it's a way to hide the warnings in docstrings

https://blog.guillaume-gomez.fr/articles/2020-03-12+Guide+on+how+to+write+documentation+for+a+Rust+crate#Hiding-lines

tfhe/docs/getting_started/quick_start.md
```diff
@@ -59,5 +59,6 @@ Refer to the [installation documentation](installation.md) for configuration opt
 The `prelude` pattern provides a convenient way to globally import all important **TFHE-rs** traits at once. This approach saves time and avoids confusion.
 
 ```rust
+# #[allow(unused_imports)]
```
IceTDrinker, line 62: does this show up in the gitbook doc ?

unfortunately in the .md files we should not have allow attributes otherwise people reading the doc will see those appear in the HTML most likely

mayeul-zama: Apparently not
I tried a few different ways but it didn't work

IceTDrinker: can it be added in modules except docs ? like in high_level, integer, shortint, core ? or is it a always a lib attribute ?

IceTDrinker: it's an attribute so unless we do it on a per module basis I don't think we can undo it, what a lame way to enable stuff

mayeul-zama: I can try

mayeul-zama: Error says "error: this attribute can only be applied at the crate level"

IceTDrinker: 🙄 what a weird design

IceTDrinker (approved):

PR #1423, opened 2024-07-25
refactor(shortint): refactor the shortint keyswitching code

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/670

### PR content/description

- this manages better all the cases we encouter, we force a refresh PBS in all cases for now which is less optimal in certain cases but allows to be safe in cases where keyswitches might be chained

cc @titouantanguy to check the cases which were failing previously


IceTDrinker:
I'll push a second commit tomorrow that will have an edge case test which this code fixes

titouantanguy:
Seems to be working now! (based on 1 test). Running a more extensive test suit but it'll take some time.

IceTDrinker:
> Seems to be working now! (based on 1 test). Running a more extensive test suit but it'll take some time.

Added a test with something more similar to what you do, test passes, so should be fine hoepfully

tmontaigu (approved):

IceTDrinker:
I'll let you guys merge this, I'll likely be off before I can, please include it in the alpha.2 of monday @nsarlin-zama

PR #1425, opened 2024-07-25
refactor(gpu): fix sample extraction when nth > 0 and keep input unchanged

pdroalves:
This PR:

- introduces a rust test for GPU's sample extraction,
- fixes it when `nth > 0`,
- refactors it so the input (the glwe) is not modified by the method.

We might also observe another performance improvement.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/671

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
@@ -191,65 +191,55 @@ __device__ void add_to_torus(double2 *m_values, Torus *result,
 
 // Extracts the body of the nth-LWE in a GLWE.
 template <typename Torus, class params>
-__device__ void sample_extract_body(Torus *lwe_array_out, Torus *accumulator,
+__device__ void sample_extract_body(Torus *lwe_array_out, Torus *glwe,
                                     uint32_t glwe_dimension, uint32_t nth = 0) {
-  // Set first coefficient of the accumulator as the body of the LWE sample
+  // Set first coefficient of the glwe as the body of the LWE sample
   lwe_array_out[glwe_dimension * params::degree] =
-      accumulator[glwe_dimension * params::degree + nth];
+      glwe[glwe_dimension * params::degree + nth];
 }
 
 // Extracts the mask from the nth-LWE in a GLWE.
 template <typename Torus, class params>
-__device__ void sample_extract_mask(Torus *lwe_array_out, Torus *accumulator,
-                                    uint32_t num_poly = 1, uint32_t nth = 0) {
-  for (int z = 0; z < num_poly; z++) {
+__device__ void sample_extract_mask(Torus *lwe_array_out, Torus *glwe,
+                                    uint32_t glwe_dimension = 1,
+                                    uint32_t nth = 0) {
+  for (int z = 0; z < glwe_dimension; z++) {
     Torus *lwe_array_out_slice =
         (Torus *)lwe_array_out + (ptrdiff_t)(z * params::degree);
-    Torus *accumulator_slice =
-        (Torus *)accumulator + (ptrdiff_t)(z * params::degree);
+    Torus *glwe_slice = (Torus *)glwe + (ptrdiff_t)(z * params::degree);
 
     synchronize_threads_in_block();
-    // Reverse the accumulator
+    // Reverse the glwe
+    // Set ACC = -ACC
     int tid = threadIdx.x;
     Torus result[params::opt];
 #pragma unroll
     for (int i = 0; i < params::opt; i++) {
-      result[i] = accumulator_slice[params::degree - tid - 1];
-      tid = tid + params::degree / params::opt;
-    }
-    synchronize_threads_in_block();
-
-    // Set ACC = -ACC
-    tid = threadIdx.x;
-#pragma unroll
-    for (int i = 0; i < params::opt; i++) {
-      accumulator_slice[tid] =
-          SEL(-result[i], result[i], tid >= params::degree - nth);
+      auto x = glwe_slice[params::degree - tid - 1];
+      result[i] = SEL(-x, x, tid >= params::degree - nth);
       tid = tid + params::degree / params::opt;
     }
     synchronize_threads_in_block();
 
     // Perform ACC * X
     // (equivalent to multiply_by_monomial_negacyclic_inplace(1))
-    tid = threadIdx.x;
-    result[params::opt];
-    for (int i = 0; i < params::opt; i++) {
-      // if (tid < 1)
-      //  result[i] = -accumulator_slice[tid - 1 + params::degree];
-      // else
-      //  result[i] = accumulator_slice[tid - 1];
-      int x = tid - 1 + SEL(0, params::degree - nth, tid < 1);
-      result[i] = SEL(1, -1, tid < 1) * accumulator_slice[x];
-      tid += params::degree / params::opt;
-    }
-    synchronize_threads_in_block();
-
     // Copy to the mask of the LWE sample
     tid = threadIdx.x;
-#pragma unroll
     for (int i = 0; i < params::opt; i++) {
-      lwe_array_out_slice[tid] = result[i];
-      tid = tid + params::degree / params::opt;
+      // if (tid < 1)
+      //  result[i] = -glwe_slice[tid - 1 + params::degree];
+      // else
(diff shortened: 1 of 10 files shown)


backends/tfhe-cuda-backend/cuda/include/functions.h
```diff
@@ -0,0 +1,16 @@
...
+
+extern "C" {
+void cuda_glwe_sample_extract_64(void **streams, uint32_t *gpu_indexes,
+                                 uint32_t gpu_count, void *lwe_array_out,
+                                 void *glwe_array_in, uint32_t *nth_array,
+                                 uint32_t num_samples, uint32_t glwe_dimension,
+                                 uint32_t polynomial_size);
+}
+
+#endif
```
agnesLeroy, line 16: Maybe this could go to ciphertext.h?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cu
```diff
@@ -0,0 +1,56 @@
...
+    host_sample_extract<uint64_t, AmortizedDegree<16384>>(
+        (cudaStream_t *)(streams), (uint64_t *)lwe_array_out,
+        (uint64_t *)glwe_in, (uint32_t *)nth_array, num_samples,
+        glwe_dimension);
+    break;
+  default:
+    PANIC("Cuda error: unsupported polynomial size. Supported "
+          "N's are powers of two in the interval [256..16384].")
+  }
+}
```
agnesLeroy, line 56: Maybe this could go to core_crypto/ciphertext.cu?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -191,67 +191,86 @@ __device__ void add_to_torus(double2 *m_values, Torus *result,
...
+                                  uint32_t num_samples,
+                                  uint32_t glwe_dimension) {
+
+  dim3 grid(num_samples);
+  dim3 thds(params::degree / params::opt);
+  apply_sample_extract<Torus, params><<<grid, thds, 0, streams[0]>>>(
+      lwe_array_out, glwe_in, nth_array, glwe_dimension);
+  check_cuda_error(cudaGetLastError());
+}
+
```
agnesLeroy, line 275: Hmm ok this was in functions.cuh already. Well what do you think of moving it to ciphertext.cuh?

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -345,6 +345,20 @@ impl<T: Numeric> CudaVec<T> {
         }
     }
 
+    /// Returns a copy of the vector
+    ///
+    /// # Safety
+    ///
+    /// - [CudaStreams::synchronize] __must__ be called as soon as synchronization is required
```
agnesLeroy, line 352: This function should not be unsafe and the comment above can be removed.

tfhe/src/core_crypto/gpu/entities/lwe_ciphertext.rs
```diff
@@ -0,0 +1,147 @@
...
+        let d_vec = unsafe { self.0.d_vec.duplicate(streams, 0) };
+
+        let cuda_lwe_list = CudaLweList {
+            d_vec,
+            lwe_ciphertext_count: LweCiphertextCount(1),
+            lwe_dimension,
+            ciphertext_modulus,
+        };
+        Self(cuda_lwe_list)
+    }
```
agnesLeroy, line 138: We only had the CudaLweCiphertextList and CudaGlweCiphertextList types, and now I'm wondering if you really need to add those: can't you apply sample extraction to an array of input GLWEs and output  an array of LWEs? This is likely to be more useful for the compression of vectors of arrays, which is what we ultimately want to do I think.

agnesLeroy (commented):
Hey @pdroalves! Thank you very much for this. I have some comments on the PR but mainly a question: shouldn't we apply sample extraction to a list of GLWE ciphertexts and output a list of LWE ciphertexts? That would be more efficient for the compression of a vector of inputs, which is what we'll need if I'm correct. Then you don't need to add types for CudaGlweCiphertext and CudaLweCiphertext, you can reuse the existing List types.

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
pdroalves: I think it's a great idea. Sample extraction relates directly to GLWE and LWE ciphertexts.

tfhe/src/core_crypto/gpu/entities/lwe_ciphertext.rs
pdroalves: I could, but that does that make sense in this context? Compression takes a list of LWEs and outputs a single GLWE. To do what you suggest we would be running multiple compressions in parallel. Is that a use case?

agnesLeroy: Actually this is the use case that's covered by the CPU version, if you llok at how it's done in shortint: tfhe/src/shortint/list_compression/compression.rs:19

backends/tfhe-cuda-backend/cuda/include/functions.h
pdroalves: Done.

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cu
pdroalves: Done.

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
pdroalves: Done.

tfhe/src/core_crypto/gpu/entities/lwe_ciphertext.rs
pdroalves: Oh I haven't noticed `modulus_switched_glwe_ciphertext_list` was a vector. I only saw the "Glwe" word there.

I don't see a big problem on introducing single-element versions of those ciphertext types, though. But yeah, sample extraction (and consequently, compression) would be more efficient by working with lists only. I will refactor.

tfhe/src/core_crypto/gpu/vec.rs
pdroalves: Fixed!

tfhe/src/core_crypto/gpu/entities/lwe_ciphertext.rs
pdroalves: Done.

I removed `CudaGlweCiphertext` and `CudaLweCiphertext` and refactored the sample extraction rust API to take a list of GLWEs.

pdroalves: @agnesLeroy actually, I'm not sure that's the use case considered in the CPU version.

Compression will indeed output a list of GLWEs. However, the unpack method used for decompression at line 120 will received a single index from the user to decompress a single ciphertext from a particular GLWE, and that's where sample extraction is used. Probably that's why there is no `glwe_sample_extraction_list.rs` in `core_crypto/algorithms`.

However, since I already did the refactor, we may now offer the capability of unpacking many ciphertexts in parallel in the GPU backend. This could be an new feature offered by the backend, but our version of the sample extraction API will need to be modified. Take a look and tell me what you think

pdroalves: And now the test is broken but I will have to stop for today. Monday I fix it.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -345,6 +345,18 @@ impl<T: Numeric> CudaVec<T> {
         }
     }
 
+    /// Returns a copy of the vector
+    pub fn duplicate(&self, streams: &CudaStreams, gpu_index: u32) -> Self {
```
agnesLeroy, line 349: I don't think we need to add this entry point, it doesn't seem to be used.

agnesLeroy (commented):
Hey @pdroalves! Thanks a lot for this PR :slightly_smiling_face: I just have some small comments about it, after that it'll be good to go!

agnesLeroy (approved):
All good! Thanks!

PR #1430, opened 2024-07-26
feat(integer): improve {overflowing_}scalar_add/sub

tmontaigu:
- [x] check benches
- [x] handles trivial oveflow (scalar > ciphertext max)
- [x] modifiy test to use non fixed number of blocks
- [x] remove commented code
- [x] add some docs

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -856,21 +856,6 @@ impl ServerKey {
         // with the parameters we provide
         let grouping_size = (num_bits_in_block as usize).min(self.key.max_noise_level.get());
 
-        let num_groupings = num_blocks.div_ceil(grouping_size);
-
-        let num_carry_to_resolve = num_groupings - 1;
-
-        let sequential_depth =
-            (num_carry_to_resolve.saturating_sub(1) as u32) / (grouping_size as u32 - 1);
-        let hillis_steel_depth = if num_carry_to_resolve == 0 {
-            0
-        } else {
-            num_carry_to_resolve.ceil_ilog2()
-        };
-
-        let use_sequential_algorithm_to_resolved_grouping_carries =
-            sequential_depth <= hillis_steel_depth;
-
         let mut output_flag = None;
 
         // First step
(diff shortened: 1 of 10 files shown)


tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
```diff
@@ -269,12 +276,560 @@ impl ServerKey {
...
+                    let result = packed_block + u64::from(packed_scalar_block);
+                    let state = if result >= packed_modulus {
+                        2 // Generate
+                    } else if result == packed_modulus - 1 {
+                        1 // Propagate
+                    } else {
+                        0 // Nothing
+                    };
+
+                    state << index_in_grouping
```
nsarlin-zama, line 627: Maybe I'm missing something but it looks like the exact same LUT as the one used in the first group

nsarlin-zama (commented):
Left some small comments, but awesome work !

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
```diff
@@ -269,12 +276,560 @@ impl ServerKey {
...
+            .chain(std::iter::repeat(if scalar < Scalar::ZERO {
+                (self.message_modulus().0 - 1) as u8
+            } else {
+                0
+            }))
+            .take(ct.blocks().len())
+            .collect();
+
+        const COMPUTE_OVERFLOW: bool = false;
+        const INPUT_CARRY: bool = false;
```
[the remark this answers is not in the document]
tmontaigu: I used const over let, just because they are compile time constants, and I prefer let/const over direcly using the value in the argument as its more readable

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -1180,6 +1006,195 @@ impl ServerKey {
...
+                            1 << ((grouping_index - 1) % (grouping_size - 1)),
+                        );
+                    } else {
+                        self.key.unchecked_scalar_add_assign(cum_sum_block, 1);
+                    }
+                    cum_sum_block.degree = Degree::new(message_modulus as usize - 1);
+                }
+            });
+
+        let num_groupings = num_blocks / grouping_size;
```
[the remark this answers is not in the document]
tmontaigu: yes indeed

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
tmontaigu: Yes the lut is the same, its the captured value of index_in_grouping that changes
I extrated the logic of this in a local closure

nsarlin-zama: But since we are `< grouping_size` in the first group it should yield the same value anyways ?

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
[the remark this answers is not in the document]
nsarlin-zama: Ok, I tend to use const only for globals since rust vars are const by default, but that's fine too. I think it compiles to the same code in release anyways.

tfhe/src/integer/server_key/radix_parallel/scalar_add.rs
tmontaigu: Ah yes indeed that should work, I've changed the way these are written so many times that some things wre not properly changed

tmontaigu: Well actually no there is still a problem

nsarlin-zama: Yes you're right ! I've missed the `.get(1..).enumerate()` which is different than `.enumerate().get(1..)`.

nsarlin-zama: Maybe that's worthy of a comment ! To avoid an aggressive refactor by someone going through this code in a few years :)
You currently state that the first group is handled differently but not really how

tmontaigu: I added one

nsarlin-zama (approved):

tmontaigu:
@cla-bot check

PR #1431, opened 2024-07-29
refactor(integer): add compression key types

mayeul-zama:


mayeul-zama:
To merge after https://github.com/zama-ai/tfhe-rs/pull/1437
Will need to add a new intermediate versioned type

IceTDrinker:
some GPU test crash related to the change at the moment it seems

nsarlin-zama (commented):
Apart from the CI break, the code looks good to me.

Maybe we should add a client key with compression to the data tests because I am not sure that this is properly tested right now (the client key comes from 0.6 and does not have compression)

IceTDrinker:
can you check if there are test cases for compression in the data repo ?

tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs
```diff
@@ -158,7 +158,7 @@ mod tests {
         );
 
         let (cuda_compression_key, cuda_decompression_key) =
-            radix_cks.new_cuda_compression_decompression_keys(&private_compression_key, &streams);
+            radix_cks.new_cuda_compression_decompression_keys(&private_compression_key.0, &streams);
```
IceTDrinker, line 161: I would move the .0 in the cuda code and make sure this interface uses an integer key

tfhe/src/integer/compression_keys.rs
```diff
@@ -0,0 +1,56 @@
+use super::ClientKey;
+use crate::integer::backward_compatibility::list_compression::*;
+use serde::{Deserialize, Serialize};
+use tfhe_versionable::Versionize;
+
+#[derive(Clone, Debug, Serialize, Deserialize, Versionize)]
+#[versionize(CompressionPrivateKeysVersions)]
+pub struct CompressionPrivateKeys(pub crate::shortint::list_compression::CompressionPrivateKeys);
```
IceTDrinker, line 8: this is not consistent with other keys, they are struct with named fields, see e.g. tfhe/src/integer/server_key/mod.rs::36

IceTDrinker (commented):
Small things

tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs
mayeul-zama: Done

tfhe/src/integer/compression_keys.rs
mayeul-zama: Replaced by a field named `key`

nsarlin-zama (approved):

PR #1435, opened 2024-07-31
chore(versionable): add Send and Sync marker traits to Err type

kc1212:
It was not possible to use `unversionize(...)?` in an async block.


nsarlin-zama (approved):
thanks !

PR #1437, opened 2024-07-31
fix(hlapi): fix Client/Server Key versionning

mayeul-zama:


tfhe/src/high_level_api/backward_compatibility/keys.rs
```diff
@@ -95,23 +116,48 @@ pub struct IntegerServerKeyV0 {
...
 
+impl Upgrade<IntegerServerKey> for IntegerServerKeyV1 {
+    type Error = Infallible;
+
+    fn upgrade(self) -> Result<IntegerServerKey, Self::Error> {
+        Ok(IntegerServerKey {
+            key: self.key,
+            cpk_key_switching_key_material: self.cpk_key_switching_key_material,
+            compression_key: self.compression_key,
+            decompression_key: self.decompression_key,
```
nsarlin-zama, line 151: What should we do if the loaded key has a wopbs_key field ? Here we just ignore it be we could also return an error if we think that's what should be done.

nsarlin-zama (approved):
thanks !

PR #1441, opened 2024-07-31
impl Versionize for HashSet/HashMap and Box<[T]>/ABox<[T]>

nsarlin-zama:

### PR content/description
The implem converts to `Vec<T::Versioned>` for HashSet and `Vec<(K::Versioned, V::Versioned)>` for HashMap to remove the bound on Eq/Hash for the versioned type. That shouldn't matter since the versioned type is not supposed to be modified before serialization.

This PR also implements Versionize for `Box<[T]>` and `ABox<[T]>` (boxed slices)


mayeul-zama (approved):
Thanks!

kc1212 (approved):

PR #1442, opened 2024-08-01
chore(CI): update LE commit

bencrts:
WIP

### DIFF
--- a/.github/workflows/parameters_check.yml
+++ b/.github/workflows/parameters_check.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           repository: malb/lattice-estimator
           path: lattice_estimator
-          ref: '53508253629d3b5d31a2ad110e85dc69391ccb95'
+          ref: 'e80ec6bbbba212428b0e92d0467c18629cf9ed67'
 
       - name: Install Sage
         run: |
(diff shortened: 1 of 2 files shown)


bencrts:
@IceTDrinker I think this is ready for merging -- the commit ID has been updated and the CI check passes in 38min.

.github/workflows/parameters_check.yml
```diff
@@ -24,7 +24,6 @@ jobs:
         with:
           repository: malb/lattice-estimator
           path: lattice_estimator
-          ref: '53508253629d3b5d31a2ad110e85dc69391ccb95'
```
IceTDrinker, line 27: could we pin the more recent reference ? it is to avoid random breakage in the CI and means we have to dedicate some time from time to time to update the commit

.github/workflows/parameters_check.yml
```diff
@@ -37,7 +36,15 @@ jobs:
...
-          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 0 "boolean_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 1 "boolean_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 2 "boolean_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 0 "shortint_classic_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 1 "shortint_classic_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 2 "shortint_classic_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 0 "shortint_multi_bit_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 1 "shortint_multi_bit_parameters_lattice_estimator"
+          PYTHONPATH=lattice_estimator sage ci/lattice_estimator.sage 2 "shortint_multi_bit_parameters_lattice_estimator"
```
IceTDrinker, line 47: this is likely unnecessary by auto partitioning in the sage script ?

also if this is an issue we can just change the runs-on of the job to e.g. large_ubuntu_16 see line 17, this way the change to the sage script may not be required anymore

IceTDrinker (commented):
if you need more details as to what I have in mind don't hesitate to ping :)

.github/workflows/parameters_check.yml
bencrts: Ok, I'll add the current commit back.

.github/workflows/parameters_check.yml
bencrts: The problem is that releasing memory from a single run of the script seems to be a bit problematic -- the CI machine runs out of RAM and crashes. Splitting up the task in this way with some command line args is a bit of an easy way around it

.github/workflows/parameters_check.yml
bencrts: Added back on L27

ci/lattice_estimator.sage.py
```diff
@@ -0,0 +1,131 @@
+"""
+lattice_estimator
+-----------------
+
+Test cryptographic parameters set against several attacks to estimate their security level.
+"""
+
+
+# This file was *autogenerated* from the file ci/lattice_estimator.sage
```
IceTDrinker, line 9: careful you committed an auto generated file

PR #1446, opened 2024-08-01
fix(gpu): fix multi-gpu error in division

agnesLeroy:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/675

### PR content/description

I don't know how this bug was not detected before, it exists since we merged the dispatch/gather logic.
With small numbers of inputs there were no local cuda streams on some GPUs but we were trying to use them in the scalar shifts, which happen on 2*num_blocks inputs.

### Check-list:


pdroalves (approved):
Very subtle issue. Probably I didn't notice it because we were using less gpus than needed, no?

agnesLeroy:
> Very subtle issue. Probably I didn't notice it because we were using less gpus than needed, no?

It was hard to catch, but what's strange is that the CI didn't catch it immediately... benchmarks with 2 blocks should have crashed right away. But then :shrug:

PR #1450, opened 2024-08-02
Impl versionize for statically sized arrays

nsarlin-zama:
Add versionize support for statically sized arrays (`[T; N]`) and `std::num::Wrapping<T>`.

Static arrays are converted to slice/vec before versioning since serde does not support const generics in arrays: https://github.com/serde-rs/serde/issues/1937


nsarlin-zama:
I added a test with a struct containing all these types

tmontaigu (approved):

PR #1453, opened 2024-08-02
feat(integer): Adds bitslice operation

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/640

### PR content/description
Adds a bitslice operation to the integer API. This operation extracts a range of bits from an integer. On native rust integers, this is roughly equivalent to the following operation: 
```rust
fn bitslice(val: u64, start_bit: usize, end_bit: usize)  -> u64 {
    (val % (1 << end_bit)) >> start_bit
}
```
On FHE radix integers, we have different cases based on the alignment of the slice bounds w.r.t the block size:
- If start_bit is block aligned, we can simply copy full blocks until we are left with fewer bits than the block size
- If start_bit is not aligned, we have to create blocks by "merging" higher bits from a block and lower bits from its successor. This is done using PBS with bivariate LUTs (one per block).
- In both cases, if the slice size is not block aligned, we have to do one last PBS to create the last block.

The returned ciphertext is filled with 0-trivial ciphertexts to match the input ciphertext size.

### Check-list:

### DIFF
--- a/tfhe/src/error.rs
+++ b/tfhe/src/error.rs
@@ -3,6 +3,8 @@ use std::fmt::{Debug, Display, Formatter};
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum ErrorKind {
     Message(String),
+    /// The provide range for a slicing operation was invalid
+    InvalidRange(InvalidRangeError),
     /// The zero knowledge proof and the content it is supposed to prove
     /// failed to correctly prove
     #[cfg(feature = "zk-pok")]
@@ -34,6 +36,7 @@ impl Display for Error {
             ErrorKind::InvalidZkProof => {
                 write!(f, "The zero knowledge proof and the content it is supposed to prove were not valid")
             }
+            ErrorKind::InvalidRange(err) => write!(f, "Invalid range: {err}"),
         }
     }
 }
@@ -56,6 +59,13 @@ impl From<String> for Error {
     }
 }
 
+impl From<InvalidRangeError> for Error {
+    fn from(value: InvalidRangeError) -> Self {
+        let kind = ErrorKind::InvalidRange(value);
+        Self { kind }
+    }
+}
+
 impl std::error::Error for Error {}
 
 // This is useful to use infallible conversions as well as fallible ones in certain parts of the lib
@@ -65,3 +75,28 @@ impl From<std::convert::Infallible> for Error {
         unreachable!()
     }
 }
+
+/// Error returned when the provided range for a slice is invalid
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum InvalidRangeError {
+    /// The upper bound of the range is greater than the size of the integer
+    SliceTooBig,
+    /// The upper gound is smaller than the lower bound
+    WrongOrder,
+}
+
+impl Display for InvalidRangeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::SliceTooBig => write!(
+                f,
+                "The upper bound of the range is greater than the size of the integer"
+            ),
+            Self::WrongOrder => {
+                write!(f, "The upper gound is smaller than the lower bound")
+            }
+        }
+    }
+}
+
+impl std::error::Error for InvalidRangeError {}
--- a/tfhe/src/integer/server_key/radix/slice.rs
+++ b/tfhe/src/integer/server_key/radix/slice.rs
@@ -1,0 +1,556 @@
+use std::ops::{Bound, Range, RangeBounds};
+
+use crate::error::InvalidRangeError;
+use crate::integer::{RadixCiphertext, ServerKey};
+use crate::prelude::{CastFrom, CastInto};
+use crate::shortint;
+
+/// Normalize a rust RangeBound object into an Exclusive Range, and check that it is valid for the
+/// source integer.
+pub(crate) fn normalize_range<B, R>(
+    range: &R,
+    nb_bits: usize,
+) -> Result<Range<usize>, InvalidRangeError>
+where
+    R: RangeBounds<B>,
+    B: CastFrom<usize> + CastInto<usize> + Copy,
+{
+    let start = match range.start_bound() {
+        Bound::Included(inc) => (*inc).cast_into(),
+        Bound::Excluded(excl) => (*excl).cast_into() + 1,
+        Bound::Unbounded => 0,
+    };
+
+    let end = match range.end_bound() {
+        Bound::Included(inc) => (*inc).cast_into() + 1,
+        Bound::Excluded(excl) => (*excl).cast_into(),
+        Bound::Unbounded => nb_bits,
+    };
+
+    if end > nb_bits {
+        Err(InvalidRangeError::SliceTooBig)
+    } else if start > end {
+        Err(InvalidRangeError::WrongOrder)
+    } else {
+        Ok(Range { start, end })
+    }
+}
+
+/// This is the operation to extract a non-aligned block, on the clear.
+/// For example, with a 2x4bits integer: |abcd|efgh|, extracting the block
+/// at offset 2 will return |cdef|. This function should be used inside a LUT.
+pub(in crate::integer) fn slice_oneblock_clear_unaligned(
+    cur_block: u64,
+    next_block: u64,
+    offset: usize,
+    block_size: usize,
+) -> u64 {
+    cur_block >> (offset) | ((next_block << (block_size - offset)) % (1 << block_size))
+}
+
+impl ServerKey {
+    /// Extract a slice of blocks from a ciphertext.
+    ///
+    /// The result is returned as a new ciphertext.
+    ///
+    /// # Example
+    ///
+    /// ```rust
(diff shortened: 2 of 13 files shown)


tmontaigu:
> * The operations are called `scalar_bitslice` because the range parameters are cleartexts, and we could technically add support for encrypted ranges later. Does it make sense ? Or should I call them just `bitslice` because there is no encrypted version ?

scalar_bitslice is good as as you said, we may add the encrypted version later

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_slice.rs
```diff
@@ -0,0 +1,443 @@
...
+    B: RangeBounds<T>,
+    T: CastFrom<usize> + Copy,
+    usize: CastFrom<T>,
+{
+    let (start, end) = parse_bounds(&bounds, modulus as usize).unwrap();
+
+    let bin: String = format!("{value:064b}").chars().rev().collect();
+
+    let out_bin: String = bin[start..end].chars().rev().collect();
+    u64::from_str_radix(&out_bin, 2).unwrap_or_default()
```
tmontaigu, line 37: We can to this without intermediate strings: 

```rust
    // This removes the bits >= end
    let result = value << (u64::BITS - end);
    // This removes the bits < start
    result >> (start + (u64::BITS - end))
```

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_slice.rs
tmontaigu, line 1: For smart and default tests, we should add some values to the ciphertext so that the test checks that carry propagations checks are correctly implemented 

```
 let mut ct = cks.encrypt(clear);
 let offset = random_non_zero_value(&mut rng, modulus);
         
sks.unchecked_scalar_add_assign(&mut ct, offset);

let clear = overflowing_add_under_modulus(clear, offset, modulus);

// executor.execute
```

tfhe/src/integer/server_key/radix_parallel/slice.rs
```diff
@@ -0,0 +1,361 @@
...
+use crate::error::InvalidRangeError;
+use crate::integer::server_key::radix::slice::{parse_bounds, slice_oneblock_clear_unaligned};
+use crate::integer::{RadixCiphertext, ServerKey};
+use crate::prelude::{CastFrom, CastInto};
+
+impl ServerKey {
+    /// Extract a slice from a ciphertext. The size of the slice is a multiple of the block
+    /// size but it is not aligned on block boundaries, so we need to mix block n and (n+1) to
+    /// create a new block, using the lut function `slice_oneblock_clear_unaligned`.
+    fn blockslice_unaligned_parallelized(
```
tmontaigu, line 14: Shouldn't this be called bitslice_unaligned ?

tfhe/src/integer/server_key/radix/slice.rs
```diff
@@ -0,0 +1,533 @@
...
+    range: &R,
+    nb_bits: usize,
+) -> Result<(usize, usize), InvalidRangeError>
+where
+    R: RangeBounds<B>,
+    B: CastFrom<usize> + CastInto<usize> + Copy,
+{
+    let start = match range.start_bound() {
+        Bound::Included(inc) => (*inc).cast_into(),
+        Bound::Excluded(excl) => (*excl).cast_into() - 1,
```
tmontaigu, line 19: Shouldn't this be +1 here ?

If my range starts with `]x`, then the first index i include is x+1, not x - 1

tfhe/src/integer/server_key/radix/slice.rs
```diff
@@ -0,0 +1,533 @@
...
+    ///
+    /// // Decrypt:
+    /// let clear = cks.decrypt(&ct_res);
+    /// assert_eq!((msg % (1 << end_bit)) >> start_bit, clear);
+    /// ```
+    pub fn blockslice(
+        &self,
+        ctxt: &RadixCiphertext,
+        start_block: usize,
+        end_block: usize,
```
tmontaigu, line 86: I feel like this could also take a range type as input and be called scalar_blockslice

tfhe/src/error.rs
```diff
@@ -65,3 +65,12 @@ impl From<std::convert::Infallible> for Error {
         unreachable!()
     }
 }
+
+/// Error returned when the provided range for a slice is invalid
+#[derive(Debug)]
+pub enum InvalidRangeError {
```
tmontaigu, line 71: Since this is meant to be an error, I think it should impl std::error::Error, and maybe also Into<tfhe::Error> to allow  using `?`

tfhe/src/integer/server_key/radix/slice.rs
nsarlin-zama: yes indeed !

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_slice.rs
nsarlin-zama: This does not work if `start == end`, which I think should return 0 as "the empty slice".

But I think the string implem is closer to the definition which is why it is used in tests. But I can switch to an arithmetic implementation if you think it's better.

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_slice.rs
nsarlin-zama: done

tfhe/src/integer/server_key/radix/slice.rs
nsarlin-zama: done

tfhe/src/error.rs
nsarlin-zama: done

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_slice.rs
tmontaigu: Yeah for start == end a `if` would be needed

The string version is ok for tests as its not a performance sensitive thing

tfhe/src/integer/server_key/radix/slice.rs
```diff
@@ -0,0 +1,549 @@
+use std::ops::{Bound, RangeBounds};
+
+use crate::error::InvalidRangeError;
+use crate::integer::{RadixCiphertext, ServerKey};
+use crate::prelude::{CastFrom, CastInto};
+use crate::shortint;
+
+/// Normalize a rust bound object, and check that it is valid for the source integer
+pub(crate) fn parse_bounds<B, R>(
```
tmontaigu, line 9: Here I would return either a `Range` if the bound object is normalized to a exclusive range (caller then has to use .start(), .end()), or a `RangeInclusive` if the bound object is normalized to an inlusive range (and caller then uses .into_inner())

tfhe/src/integer/server_key/radix_parallel/slice.rs
```diff
@@ -0,0 +1,362 @@
...
+impl ServerKey {
+    /// Extract a slice from a ciphertext. The size of the slice is a multiple of the block
+    /// size but it is not aligned on block boundaries, so we need to mix block n and (n+1) to
+    /// create a new block, using the lut function `slice_oneblock_clear_unaligned`.
+    fn scalar_blockslice_unaligned_parallelized(
+        &self,
+        ctxt: &RadixCiphertext,
+        start_block: usize,
+        block_count: usize,
+        offset: usize,
```
tmontaigu, line 19: I'm wondering how much do we want to do pre-emptive checks on these, to make sure the code panics is start, block count would yield invalid indexes later in the parallel code. and also check offset < message_modulus.ilog2

tfhe/src/integer/server_key/radix/slice.rs
nsarlin-zama: good idea !

nsarlin-zama: done

tfhe/src/integer/server_key/radix_parallel/slice.rs
nsarlin-zama: done

tfhe/src/integer/server_key/radix_parallel/slice.rs
```diff
@@ -0,0 +1,371 @@
...
+            if slice_width % block_width != 0 {
+                let last_block = self.bitslice_remainder(
+                    ctxt,
+                    range.end / block_width,
+                    slice_width % block_width,
+                );
+                sliced.blocks.push(last_block);
+            }
+
+            sliced
```
tmontaigu, line 133: The computations inside the `if` should be done in a rayon::join/rayon::scope, otherwise, when `slice_width % block_width != 0`, the latency is 2 PBS, with rayon::join/rayon::scope we could keep the latency at 1 PBS

nsarlin-zama: done

tmontaigu (approved):

PR #1465, opened 2024-08-09
chore(all): use a builder pattern for safe serialization API

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/646

### PR content/description
The goal of this PR is to simplify the safe_serialization API. With the introduction of data versioning, there are many different serialization possibilities:
- safe or not,
- conformant or not (only for deserialization),
- versioned or not

This can be difficult for us to maintain and confusing to the user since the function they should use by default have the most complicated names (`safe_deserialize_conformant_versioned`). The idea is to introduce a simple builder pattern to configure the serialization, with the safest defaults and with a possibility to disable functionalities for advanced users.

This PR introduces 2 objects, `SerializationConfig` and `DeserializationConfig`. The idea for serialization is to create this object, eventually modify it to disable some functionalities, and serialize an object with the `serialize_into` method:

- default config (safe + conformance + versioning):
```rust
// Serialization is "safe" by default, so we have to give the max size to the config
SerializationConfig::new(1 << 20).serialize_into(&ct, &mut buffer).unwrap();
    

// Deserialization is also "conformant" by default, so we also have to give some conformance parameters
DeserializationConfig::new(1 << 20)
    .deserialize_from::<FheUint8>(buffer.as_slice(), &conformance_params)
    .unwrap();
```

- no versioning:
```rust
// We use the `new` constructor and disable the functionality
SerializationConfig::new(1 << 20).disable_versioning().serialize_into(&ct, &mut buffer).unwrap();
    
// No need to configure that the data is not versioned, this is detected during deserialization
DeserializationConfig::new(1 << 20)
    .deserialize_from::<FheUint8>(buffer.as_slice())
    .unwrap();
```
- no conformance:
```rust
SerializationConfig::new(1 << 20).serialize_into(&ct, &mut buffer).unwrap();
    
// Here we use the `new_without_conformance` so we don't have to give conformance params that won't be used
DeserializationConfig::new(1 << 20)
    .disable_conformance()
    .deserialize_from::<FheUint8>(buffer.as_slice())
    .unwrap();
```

The `safe_serialize`/`safe_serialize` have been kept as defaults for users that don't need customization. They include versioning.

Other changes:
- The versioning state (enabled or not) is encoded in the serialized data and is detected automatically during deserialization. This mean that `DeserializationConfig::new(...).deserialize_from(...)` handles both versioned and unversioned data.
- If the data is unversioned, the tfhe-rs version is stored in the binary file, and we raise an error if the version is different than the current one (this check can be disabled if the user is sure that the types are compatible).
- In the C API, the only safe_serialization method is versioned. There are also `serialize/deserialize` (without safe) that are only wrappers around serde.
- In the JS API  the only safe_serialization method is versioned.

### Check-list:

* [x] Docs have been added / updated (for bug fixes / features) -> update serialization doc


tfhe/src/high_level_api/integers/signed/tests.rs
```diff
@@ -648,11 +648,14 @@ fn test_safe_deserialize_conformant_fhe_int32() {
     let clear_a = random::<i32>();
     let a = FheInt32::encrypt(clear_a, &client_key);
     let mut serialized = vec![];
-    assert!(crate::safe_serialize(&a, &mut serialized, 1 << 20).is_ok());
+    assert!(SerializationConfig::new(1 << 20)
+        .serialize_into(&a, &mut serialized)
+        .is_ok());
```
mayeul-zama, line 653: I think all `assert!(_.is_ok())` can be replaced by `unwrap`
Also applies to other files

tfhe/src/zk.rs
```diff
@@ -17,11 +17,22 @@ impl Named for CompactPkeProof {
...
 }
 
+// If we call `CompactPkePublicParams::compress` we end up with a
+// `SerializableCompactPkePublicParams` that should also impl Named to be serializable with
+// `safe_serialization`. Since the `CompactPkePublicParams` is transformed into a
+// `SerializableCompactPkePublicParams` anyways before serialization, their impl of `Named` should
+// return the same string.
+impl Named for SerializableCompactPkePublicParams {
+    const NAME: &'static str = CompactPkePublicParams::NAME;
+}
```
mayeul-zama, line 34: Could be part of the first commit

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,607 @@
...
+
+/// Header with global metadata about the serialized object. This help checking that we are not
+/// deserializing data that we can't handle.
+#[derive(Serialize, Deserialize)]
+// This type should not be versioned because it is part of a wrapper of versioned messages.
+#[cfg_attr(tfhe_lints, allow(tfhe_lints::serialize_without_versionize))]
+struct SerializationHeader {
+    header_version: Cow<'static, str>,
+    versioning_mode: SerializationVersioningMode,
+    versioning_version: Cow<'static, str>,
```
mayeul-zama, line 62: Maybe better to have  `versioning_version` in `versioning_mode` as its meaning  depends on its value

We could have 
```
enum SerializationVersioningMode {
    /// Serialize with type versioning for backward compatibility
    Versioned { version: Cow<'static, str>, },
    /// Serialize the type without versioning information
    Unversioned  { crate_version: Cow<'static, str>, },
}
```

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,607 @@
...
+
+        let ct = ck.encrypt(msg);
+
+        let mut buffer = vec![];
+
+        SerializationConfig::new(1 << 20)
+            .serialize_into(&ct, &mut buffer)
+            .unwrap();
+
+        assert!(DeserializationConfig::new(1 << 20,)
```
mayeul-zama, line 424: `,` can be removed

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,607 @@
...
+pub fn safe_deserialize_conformant<
+    T: DeserializeOwned + Unversionize + Named + ParameterSetConformant,
+>(
+    reader: impl std::io::Read,
+    serialized_size_limit: u64,
+    parameter_set: &T::ParameterSet,
+) -> Result<T, String> {
+    DeserializationConfig::new(serialized_size_limit).deserialize_from(reader, parameter_set)
+}
+
```
mayeul-zama, line 368: Do we still want these functions?
They have been removed from many places (if not all)

nsarlin-zama: It was asked by BC users to keep those in this release to simplify their migration, but it can be removed in the next release. I can add a deprecation attribute if we want to. 

However we might also decide that they serve as a slightly easier shortcut for users that don't want to do any configuration.

tfhe/src/zk.rs
nsarlin-zama: likely a rebase error, I will fix it !

mayeul-zama (approved):
Thanks!

tfhe/src/safe_serialization.rs
```diff
@@ -0,0 +1,645 @@
...
+        let ct = ck.encrypt(msg);
+
+        let mut buffer = vec![];
+
+        SerializationConfig::new(1 << 20)
+            .disable_versioning()
+            .serialize_into(&ct, &mut buffer)
+            .unwrap();
+
+        assert!(DeserializationConfig::new(1 << 20)
```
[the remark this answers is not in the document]
nsarlin-zama: What would be a good default value ? We don't know the kind of device where this will run nor the concrete type that is deserialized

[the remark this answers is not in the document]
tmontaigu: In all our example we use 1 << 20, which is roughly 1GB, and I bet most users are just going to copy paste that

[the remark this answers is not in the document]
nsarlin-zama: Are you ok if I add this as an issue to not rush this before release ?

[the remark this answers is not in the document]
IceTDrinker: 1<<20 = 1 MB IIRC

1 << 30 = 1GB

tmontaigu (approved):

PR #1468, opened 2024-08-12
feat(hlapi): add tag system

tmontaigu:
Tag

The `Tag` allows to store bytes alongside of entities (keys, and ciphertext) the main purpose of this system is to `tag` / identify ciphertext with their keys.

* When encrypted, a ciphertext gets the tag of the key used to encrypt it.
* Ciphertexts resulting from operations (add, sub, etc.) get the tag from the ServerKey used
* PublicKey gets its tag from the ClientKey that was used to create it
* ServerKey gets its tag from the ClientKey that was used to create it

User can change the tag of any entities at any point.

BREAKING CHANGE: Many of the into_raw_parts and from_raw_parts changed to accommodate the addition of the `tag``

### Check-list:

### DIFF
--- a/tfhe/src/high_level_api/backward_compatibility/tag.rs
+++ b/tfhe/src/high_level_api/backward_compatibility/tag.rs
@@ -1,0 +1,7 @@
+use crate::high_level_api::tag::Tag;
+use tfhe_versionable::VersionsDispatch;
+
+#[derive(VersionsDispatch)]
+pub enum TagVersions {
+    V0(Tag),
+}
--- a/tfhe/src/high_level_api/compact_list.rs
+++ b/tfhe/src/high_level_api/compact_list.rs
@@ -6,6 +6,7 @@ use crate::core_crypto::commons::math::random::{Deserialize, Serialize};
 use crate::core_crypto::prelude::Numeric;
 use crate::high_level_api::global_state;
 use crate::high_level_api::keys::InternalServerKey;
+use crate::high_level_api::traits::Tagged;
 use crate::integer::ciphertext::{Compactable, DataKind, Expandable};
 use crate::integer::encryption::KnowsMessageModulus;
 use crate::integer::parameters::{
@@ -15,8 +16,11 @@ use crate::integer::parameters::{
 use crate::named::Named;
 use crate::shortint::MessageModulus;
 #[cfg(feature = "zk-pok")]
+pub use zk::ProvenCompactCiphertextList;
+
+#[cfg(feature = "zk-pok")]
 use crate::zk::{CompactPkePublicParams, ZkComputeLoad};
-use crate::CompactPublicKey;
+use crate::{CompactPublicKey, Tag};
 
 impl crate::FheTypes {
     fn from_data_kind(data_kind: DataKind, message_modulus: MessageModulus) -> Option<Self> {
@@ -71,7 +75,10 @@ impl crate::FheTypes {
 
 #[derive(Clone, Serialize, Deserialize, Versionize)]
 #[versionize(CompactCiphertextListVersions)]
-pub struct CompactCiphertextList(pub(crate) crate::integer::ciphertext::CompactCiphertextList);
+pub struct CompactCiphertextList {
+    pub(crate) inner: crate::integer::ciphertext::CompactCiphertextList,
+    pub(crate) tag: Tag,
+}
 
 impl Named for CompactCiphertextList {
     const NAME: &'static str = "high_level_api::CompactCiphertextList";
@@ -83,7 +90,7 @@ impl CompactCiphertextList {
     }
 
     pub fn len(&self) -> usize {
-        self.0.len()
+        self.inner.len()
     }
 
     pub fn is_empty(&self) -> bool {
@@ -91,8 +98,8 @@ impl CompactCiphertextList {
     }
 
     pub fn get_kind_of(&self, index: usize) -> Option<crate::FheTypes> {
-        self.0.get_kind_of(index).and_then(|data_kind| {
-            crate::FheTypes::from_data_kind(data_kind, self.0.ct_list.message_modulus)
+        self.inner.get_kind_of(index).and_then(|data_kind| {
+            crate::FheTypes::from_data_kind(data_kind, self.inner.ct_list.message_modulus)
         })
     }
 
@@ -100,36 +107,40 @@ impl CompactCiphertextList {
         &self,
         sks: &crate::ServerKey,
     ) -> crate::Result<CompactCiphertextListExpander> {
-        self.0
+        self.inner
             .expand(
                 IntegerCompactCiphertextListUnpackingMode::UnpackIfNecessary(sks.key.pbs_key()),
                 IntegerCompactCiphertextListCastingMode::NoCasting,
             )
-            .map(|inner| CompactCiphertextListExpander { inner })
+            .map(|inner| CompactCiphertextListExpander {
+                inner,
+                tag: self.tag.clone(),
+            })
     }
 
     pub fn expand(&self) -> crate::Result<CompactCiphertextListExpander> {
         // For WASM
-        if !self.0.is_packed() && !self.0.needs_casting() {
-            // No ServerKey required, shortcircuit to avoid the global state call
+        if !self.inner.is_packed() && !self.inner.needs_casting() {
+            // No ServerKey required, short-circuit to avoid the global state call
             return Ok(CompactCiphertextListExpander {
-                inner: self.0.expand(
+                inner: self.inner.expand(
                     IntegerCompactCiphertextListUnpackingMode::NoUnpacking,
                     IntegerCompactCiphertextListCastingMode::NoCasting,
                 )?,
+                tag: self.tag.clone(),
             });
         }
 
         global_state::try_with_internal_keys(|maybe_keys| match maybe_keys {
             None => Err(crate::high_level_api::errors::UninitializedServerKey.into()),
             Some(InternalServerKey::Cpu(cpu_key)) => {
-                let unpacking_mode = if self.0.is_packed() {
+                let unpacking_mode = if self.inner.is_packed() {
                     IntegerCompactCiphertextListUnpackingMode::UnpackIfNecessary(cpu_key.pbs_key())
                 } else {
                     IntegerCompactCiphertextListUnpackingMode::NoUnpacking
                 };
 
-                let casting_mode = if self.0.needs_casting() {
+                let casting_mode = if self.inner.needs_casting() {
                     IntegerCompactCiphertextListCastingMode::CastIfNecessary(
                         cpu_key.cpk_casting_key().ok_or_else(|| {
                             crate::Error::new(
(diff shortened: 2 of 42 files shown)


tfhe/src/high_level_api/tests/tags_on_entities.rs
```diff
@@ -0,0 +1,254 @@
...
+            let list: CompactCiphertextList = bincode::deserialize(&data).unwrap();
+            assert_eq!(list.tag(), cks.tag());
+            expand_and_check_tags(&list, &cks);
+        }
+
+        // #[cfg(feature = "zk-pok")]
+        // {
+        //     let list_packed = builder.build_with_proof_packed();
+        //     assert_eq!(list_packed.tag(), cks.tag());
+        // }
```
tmontaigu, line 252: Todo 😉

IceTDrinker:
tests are red at the moment

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,331 @@
+use crate::high_level_api::backward_compatibility::tag::{SmallVecVersions, TagVersions};
+use tfhe_versionable::Versionize;
+
+/// Simple short optimized vec, where if the data is small enough
+/// (<= std::mem::size_of::<Vec<u8>>() - 1) the data will be stored on the stack
+///
+/// Once a true heap allocated Vec was needed, it won't be deallocated in favor
+/// of stack data.
+#[derive(Clone, Debug, Versionize)]
```
nsarlin-zama, line 9: When someone call `small_vec.versionize().serialize()`, the custom impl of `Serialize` for SmallVec won't be called because the call to `serialize` will be made on type `SmallVec::Versioned`.

However, if the idea is to convert this to a plain slice before serialization, this struct does not need to be versioned in itself because it will never be stored itself (it is only a runtime type). So I believe you can directly store the array without versioning info.
This can be done using the `from/into` attribute of the macro (https://github.com/zama-ai/tfhe-rs/blob/main/utils/tfhe-versionable/examples/convert.rs) or using a manual impl of `Versionize/VersionizeOwned/Unversionize` (for example with `type Versioned<'vers> = &'vers [u8]`).

tmontaigu: But we would still need the custom serialize impl as the `versionize()` thing is opt-in rather than opt-out (and even if it were opt-out we would still need it)

nsarlin-zama: yes ! The goal is not to remove the Serialize impl but to have a consistent serialization behavior. If you want to factorize the conversion code between serialize and versionize you can use from/into (it's [also supported by serde](https://serde.rs/container-attrs.html#from)), but to be fair I'm not sure it will work with versionize in your case.

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,331 @@
...
+                Self::Heap(r_vec),
+            ) => &l_bytes[..usize::from(*l_len)] == r_vec,
+        }
+    }
+}
+
+impl Eq for SmallVec {}
+
+impl SmallVec {
+    pub fn data(&self) -> &[u8] {
```
nsarlin-zama, line 64: Maybe call this `as_slice` to mirror the Vec API ?

tfhe/src/high_level_api/tag.rs
tmontaigu: One small thing is that serde's from/into would require to clone the data that is inside the small vec into a new vec so it can be serialized, which is a bit less efficient than the custom impl where i can `serialize_bytes(self.data())`

Also, that means the `Tag` also do not need to be versionned

nsarlin-zama: You're right the custom impl should be more efficient !

I think the Tag needs to be versioned because it is not guaranteed to always be represented as a wrapper around an array-like type. However the `SmallVec` is an optimization that only makes sense in a running application and will always be equivalent to an array in the serialization data model.

tmontaigu: yes it makes sense

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,331 @@
...
+///
+/// The `Tag` allows to store bytes alongside of entities (keys, and ciphertext)
+/// the main purpose of this system is to `tag` / identify ciphertext with their keys.
+///
+/// * When encrypted, a ciphertext gets the tag of the key used to encrypt it.
+/// * Ciphertexts resulting from operations (add, sub, etc.) get the tag from the ServerKey used
+/// * PublicKey gets its tag from the ClientKey that was used to create it
+/// * ServerKey gets its tag from the ClientKey that was used to create it
+///
+/// User can change the tag of any entities at any point.
```
nsarlin-zama, line 247: Maybe we can document that tag compatibility is not checked by tfhe-rs and that it is up to the user to do so. And that by default the tag is empty (and an example on how to set it on a freshly generated ClientKey ?)

nsarlin-zama (commented):
nice API !

tfhe/src/high_level_api/tag.rs
tmontaigu: Yep, we might need to add that into the actual docs but I don't know if we want to do that yet

IceTDrinker: I don't think we should say anything about compatibility ?

here the tag is just a piece of added metadata, the BC will happen to check it

tmontaigu: I think that's what he proposed to say in the doc, that the tag is just some data storage and tfhe-rs does not do any kind of check or interpretation it only stores it and passes it around

tfhe/src/high_level_api/global_state.rs
```diff
@@ -122,10 +122,23 @@ pub(in crate::high_level_api) fn device_of_internal_keys() -> Option<crate::Devi
...
 
+/// This returns the [tag](crate::Tag) stored in the internal server key
+#[inline]
+pub(in crate::high_level_api) fn tag_of_internal_server_key() -> crate::Result<crate::Tag> {
+    INTERNAL_KEYS.with(|keys| {
+        let cell = keys.borrow();
+        Ok(match cell.as_ref().ok_or(UninitializedServerKey)? {
+            InternalServerKey::Cpu(cpu_key) => cpu_key.tag.clone(),
+            #[cfg(feature = "gpu")]
+            InternalServerKey::Cuda(cuda_key) => cuda_key.tag.clone(),
```
IceTDrinker, line 133: how does this work properly for GPU ? 🤔 we have two values that could be returned here ?

also what do you think about trying to use if cfg!(feature = "gpu") instead of the attributes to manage inline feature stuff ?

IceTDrinker: nevermind did not see it was a match 😵‍💫

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,420 @@
...
+
+const STACK_ARRAY_SIZE: usize = std::mem::size_of::<Vec<u8>>() - 1;
+
+/// Simple short optimized vec, where if the data is small enough
+/// (<= std::mem::size_of::<Vec<u8>>() - 1) the data will be stored on the stack
+///
+/// Once a true heap allocated Vec was needed, it won't be deallocated in favor
+/// of stack data.
+#[derive(Clone, Debug)]
+pub(in crate::high_level_api) enum SmallVec {
```
IceTDrinker, line 12: does the size of the enum == the size of Vec<u8> ?

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,420 @@
+use crate::high_level_api::backward_compatibility::tag::TagVersions;
+use tfhe_versionable::{Unversionize, UnversionizeError, Versionize, VersionizeOwned};
+
+const STACK_ARRAY_SIZE: usize = std::mem::size_of::<Vec<u8>>() - 1;
```
IceTDrinker, line 4: this may not be constant across compiler versions and perhaps even platforms/hw ?

could be an issue with WASM as well ?

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,420 @@
...
+/// assert_eq!(a.tag(), cks.tag());
+/// ```
+#[derive(
+    Default, Clone, Debug, serde::Serialize, serde::Deserialize, Versionize, PartialEq, Eq,
+)]
+#[versionize(TagVersions)]
+pub struct Tag {
+    // We don't want the enum to be public
+    inner: SmallVec,
+}
```
IceTDrinker, line 280: this has a cost even if the Tag is empty given the default SmallVec is an array of bytes

tfhe/src/high_level_api/tag.rs
```diff
@@ -0,0 +1,420 @@
+use crate::high_level_api::backward_compatibility::tag::TagVersions;
+use tfhe_versionable::{Unversionize, UnversionizeError, Versionize, VersionizeOwned};
+
+const STACK_ARRAY_SIZE: usize = std::mem::size_of::<Vec<u8>>() - 1;
+
+/// Simple short optimized vec, where if the data is small enough
```
IceTDrinker, line 6: I feel like the SmallVec optimization might be premature and has other implications that cause more trouble at the moment than it solves

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -78,26 +87,54 @@ impl CompactCiphertextList {
...
 #[cfg(feature = "zk-pok")]
 #[derive(Clone, Serialize, Deserialize)]
-pub struct ProvenCompactCiphertextList(crate::integer::ciphertext::ProvenCompactCiphertextList);
+pub struct ProvenCompactCiphertextList {
+    inner: crate::integer::ciphertext::ProvenCompactCiphertextList,
+    tag: Tag,
+}
+
+#[cfg(feature = "zk-pok")]
+impl Tagged for ProvenCompactCiphertextList {
```
IceTDrinker, line 129: should we create a zk mod here and have a feature gate on it + a pub re export ?

allows to limit the number of #cfg and I find it less error prone

tfhe/src/high_level_api/integers/unsigned/compressed.rs
```diff
@@ -49,26 +50,57 @@ where
...
+    pub fn tag(&self) -> &Tag {
+        &self.tag
     }
 
-    pub fn from_raw_parts(ciphertext: CompressedRadixCiphertext, id: Id) -> Self {
-        Self { ciphertext, id }
+    pub fn tag_mut(&mut self) -> &mut Tag {
+        &mut self.tag
     }
 }
```
IceTDrinker, line 105: not in the Tagged trait impl ? probably worth a check for all the fn tag(&self) functions

tfhe/src/high_level_api/integers/unsigned/scalar_ops.rs
```diff
@@ -545,7 +544,8 @@ macro_rules! generic_integer_impl_scalar_operation {
 
                     fn $rust_trait_method(self, rhs: $scalar_type) -> Self::Output {
                         let inner_result = $closure(self, rhs);
-                        <$concrete_type>::new(inner_result)
+                        let tag = global_state::tag_of_internal_server_key().unwrap_display();
```
IceTDrinker, line 547: what's the unwrap_display here ?

tfhe/src/high_level_api/tag.rs
tmontaigu: What other implications ?

IceTDrinker: the size, compatibility of compiler versions, the fact the default case copies an array of 0s around

tfhe/src/high_level_api/tag.rs
tmontaigu: This would just mean that if for example sizeof::Vec is 3*32 = 96, then, a 128 bit number would be stored in a Vec rather than a the stack.

So when creating, no problem, when deserializing no problem either, and if a 64 bit CPU deserialize that, as it would fit on stack, the 128 bits would live on the stack

tfhe/src/high_level_api/tag.rs
tmontaigu: No the save of the enum is likely slightly bigger as the enum handles the tag.

To have it have the same size would would have to do it ourselves with a union and a u8 tag that serves as both tag/len of data in stack array, but that requires unsafe code so that were it start not to be worth to actuall y do

tfhe/src/high_level_api/tag.rs
tmontaigu: It has the  ~same cost as if we just had a Vec<T>, as the enum maybe like be just one byte more than an empty Vec<T>

IceTDrinker: right

tfhe/src/high_level_api/tag.rs
tmontaigu: just like with just a Vec, the compiler would copy the bytes of the internal data (ptr, size, capacity) of the vec

IceTDrinker: yep my bad

tmontaigu: The SmallVec size is 8 bytes (likely sizeof::<usize>) bigger than a plain Vec<T>

tfhe/src/high_level_api/tag.rs
IceTDrinker: all right I need to check the details of the small vec stuff then

nsarlin-zama: That's why it's important to convert it to a plain array before serialization (as you did)

tfhe/src/high_level_api/tag.rs
tmontaigu: I do agree that the SmallVec thing might be premature, but the complexity of it is not that high I think

Also, the vec of the Tag which only gets and return bytes or simple u64/u128 and the fact that the serialization/versioning returns bytes protects us in a sense.

tfhe/src/high_level_api/tests/tags_on_entities.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+#[test]
+#[cfg(feature = "zk-pok")]
+fn test_tag_propagation_zk_pok() {
+    use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64;
+    use crate::ProvenCompactCiphertextList;
+
+    let config =
+        ConfigBuilder::with_custom_parameters(PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64).build();
```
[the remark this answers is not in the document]
tmontaigu: I will try to add it, but since the tag is for the whole key set it won't change how it behaves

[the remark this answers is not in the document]
IceTDrinker: It’s to make sure it gets from the public key through the keyswitch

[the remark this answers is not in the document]
IceTDrinker: what I mean is for the BC use case we have

PKE -> KS -> PBS

and we want to make sure the tagging works well in that scenario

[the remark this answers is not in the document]
tmontaigu: ok i see

tfhe/src/high_level_api/backward_compatibility/tag.rs
```diff
@@ -1,12 +1,6 @@
-use crate::high_level_api::tag::{SmallVec, Tag};
+use crate::high_level_api::tag::Tag;
 use tfhe_versionable::VersionsDispatch;
 
-#[derive(VersionsDispatch)]
-pub(in crate::high_level_api) enum SmallVecVersions {
-    #[allow(unused)] // Unused because V1 does not exists yet
-    V0(SmallVec),
-}
-
```
IceTDrinker: no Versions for the SmallVec ?

tfhe/src/high_level_api/tag.rs
```diff
@@ -182,15 +265,40 @@ impl<'de> serde::Deserialize<'de> for SmallVec {
 
 /// Tag
 ///
-/// The `Tag` allows to store bytes alongside of entities (keys, and ciphertext)
+/// The `Tag` allows to store bytes alongside entities (keys, and ciphertexts)
 /// the main purpose of this system is to `tag` / identify ciphertext with their keys.
 ///
+/// TFHE-RS does not interpret or check this data, it only stores it and passes it around
```
IceTDrinker, line 271: TFHE-rs

IceTDrinker: missing only this I believe and go for merge

IceTDrinker (approved):
Thanks a lot !

PR #1469, opened 2024-08-12
fix(integer): remove double carry prop in sub

tmontaigu:
The subtraction is done via addition of the negation, the negation is done via unchecked_neg, this will make the first block have a carry.
Then we called add_assign_with_carry_parallelized which did a carry propagation on the rhs which here is the negated value, meaning the subtraction would do 2 carry propagation.

To fix that we directly call the lower function.


IceTDrinker (approved):
Just curious how did the previous code do the propagation twice ?

PR #1470, opened 2024-08-13
Am/chore/ci fixes

IceTDrinker:
several CI fixes/changes, each commit is its own fix to make reviewing easier


IceTDrinker:
> Looks like the m1 is running all cargo-builds tests, is it expected ?

The new macos latest are M1 CPUs on github and I was afraid it would fail, but I can try with the basic macos-latest, with our M1 we would have been faster for sure, I'll update

nsarlin-zama (approved):

PR #1473, opened 2024-08-13
Am/chore/ci tfhe versionable

IceTDrinker:
added a workflow and bumped version to 0.2.1


nsarlin-zama (approved):
thanks !

nsarlin-zama:
Maybe also update the dep in `tfhe/Cargo.toml` ?

nsarlin-zama (approved):

PR #1474, opened 2024-08-14
feat(integer): implement sub_assign_with borrow

tmontaigu:
To get the same kind of speed ups for unsigned_overflow as we got in previous commits that changed the carry propagation algorithm


tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+        lhs.clone_from_slice(&prepared_blocks);
+
+        if compute_overflow {
+            Some(BooleanBlock::new_unchecked(overflow_block))
+        } else {
+            None
+        }
+    }
+
+    /// blocks must e the result of (left_block - right_block + message_modulus)
```
IceTDrinker, line 567: mus e -> must be

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+            .key
+            .generate_lookup_table(|x| if x < modulus { 1 } else { 0 });
+
+        let mut borrow = input_borrow.map_or_else(|| self.key.create_trivial(0), |b| b.0.clone());
+        for (lhs_block, rhs_block) in lhs.blocks.iter_mut().zip(rhs.blocks.iter()) {
+            self.key.unchecked_sub_assign(lhs_block, rhs_block);
+            // Here unchecked_sub_assign does not give correct result, we don't want
+            // the correcting term to be used
+            // -> This is ok as the value returned by unchecked_sub is in range
+            // 1..(message_mod * 2)
```
IceTDrinker, line 717: not sure I'm clear on why this is ok

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+        for mut blocks in tmp {
+            assert_eq!(blocks.len(), 2);
+            shifted_blocks.push(blocks.pop().unwrap());
+            block_states.push(blocks.pop().unwrap());
+        }
+
+        (shifted_blocks, block_states)
+    }
+
+    pub(crate) fn advanced_sub_assign_with_borrow_sequential(
```
IceTDrinker, line 687: obligatory: this function has its own dedicated somehow to make sure we are testing it ?

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+                    if block < message_modulus {
+                        2 << 1 // Generates a borrow
+                    } else {
+                        0 // Nothing
+                    }
+                };
+                self.key
+                    .generate_many_lookup_table(&[&first_block_state_fn, &shift_block_fn])
+            } else {
+                first_grouping_luts[2].clone()
```
IceTDrinker, line 652: if needed we can introduce view types to avoid these issues

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+        // The propagation state of the last block will be used to determine
+        // if overflow occurs (i.e is there an output borrow)
+        let mut overflow_block = block_states.pop().unwrap();
+
+        let block_modulus = self.message_modulus().0 * self.carry_modulus().0;
+        let num_bits_in_block = block_modulus.ilog2();
+
+        // Just in case we compare with max noise level, but it should always be num_bits_in_blocks
+        // with the parameters we provide
+        let grouping_size = (num_bits_in_block as usize).min(self.key.max_noise_level.get());
```
IceTDrinker, line 475: don't recall if this "safety" is in the add as well

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+                    // simulator may have either of these value
+                    // '2' if the block is borrowed from
+                    // '1' if the block will be borrowed from if the group it belongs to receive a
+                    //     borrow
+                    // '0' if the block will absorb any potential borrow
+                    //
+                    // What we do is we subtract this value from the block, as it's a borrow, not a
+                    // carry, and we add one, this means:
+                    //
+                    // '- (2 + 1) ==  -1' We remove one if the block was meant to receive a borrow
```
IceTDrinker, line 496: wrong sign in the parens or parens not in the right place ? same for all three

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+                    // '- (1 + 1) ==  -0' The block won't change, which means that when subtracting
+                    // the borrow (value: 1 or 0) that the group receives, its correctly applied
+                    // i.e the propagation simulation will be correctly done
+                    // '- (0 + 1) ==  +1' we add one, meaning that if the block receives a borrow,
+                    // we would remove one from the block, which would be absorbed by the 1 we just
+                    // added
```
IceTDrinker, line 502: not super clear here on what that means

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+        let block_modulus = self.message_modulus().0 * self.carry_modulus().0;
+        let num_bits_in_block = block_modulus.ilog2();
+
+        let grouping_size = num_bits_in_block as usize;
+
+        let shift_block_fn = |block| {
+            let overflow_guard = message_modulus;
+            let block = block % message_modulus;
+            (overflow_guard | block) << 1
+        };
```
IceTDrinker, line 586: this works because the correcting term does not go beyon 2*msg_mod ?

IceTDrinker (commented):
Some comments, as per tradition

tfhe/src/integer/server_key/radix_parallel/sub.rs
```diff
@@ -402,6 +413,329 @@ impl ServerKey {
...
+    ///
+    /// - Parameters must have at least 2 bits of message, 2 bits of carry
+    /// - blocks of lhs and rhs must be clean (no carries)
+    /// - lhs and rhs must have the same length
+    pub(crate) fn advanced_sub_assign_with_borrow_parallelized_at_least_4_bits(
+        &self,
+        lhs: &mut RadixCiphertext,
+        rhs: &RadixCiphertext,
+        input_borrow: Option<&BooleanBlock>,
+        compute_overflow: bool,
```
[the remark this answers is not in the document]
tmontaigu: We do compute the overflow, but only the unsigned meaning (borrow/carry flag) we don't support computing the signed version of it (overflow flag)

The names are very confusing I took those name from https://www.felixcloutier.com/x86/sbb#description
> The SBB instruction does not distinguish between signed or unsigned operands. Instead, the processor evaluates the result for both data types and sets the OF and CF flags to indicate a borrow in the signed or unsigned result, respectively. The SF flag indicates the sign of the signed result.

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Ah right it should be

```
-2 + 1 = -1
-1 + 1 = 0
0 + =  1
```

tfhe/src/integer/server_key/radix_parallel/sub.rs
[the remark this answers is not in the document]
IceTDrinker: ok!

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: I remove the correcting term by doing ` block % message_modulus`, so add and overflow guard, to protect when ` block % message_modulus == 0`, as subtracting 1 would overflow the block and trash the padding bit

But actually, doing `block << 1` should also work as the block is the result from (`a -b + msg_mod`) so is already in a state where we can safely subtract a 1 with any provlem of overflow (same idea as in tfhe/src/integer/server_key/radix_parallel/sub.rs line 717, after `a - b + msg_mod` the result is in range `[1..=(2*msg_mod) -1]`


high bound: a=3,b=0,msg_mod=4 => (3-0 + 4) = 7 = (2*4) -1
low bound: a = 0, b=3,msg_mod=4 => (0 -3 + 4) = 1

So the lowest value after shortint_sub_assign is `1`, which means we can then safely subtract a borrow that may come from previous blocks

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: sub assign is (x - y + msg_mod), so with (3-0 + 4) = 7, and (0 - 3 + 4) = 1, the resulting ciphertext is in range 1..msg_mod

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Yes

tfhe/src/integer/server_key/radix_parallel/sub.rs
tmontaigu: Whe have a shifted block that looks like `0xx0`, to put the 'propagation simulator block', the block that simulates how the input borrow of the group is propagated (or not) to the block in question, we don't do `0xx0 + 00pp`, we do `0xx0 - 00pp + 0001`.

So if `00pp == 0000` (meanins any borrow is the group receives is actually absorbed an so it does not reach the block
This translates in us doing o `0xx0 - 0000 + 0001 = 0xx1`, so if I the group was to be borrowed from, we need to sub 1 from it, `0xx1 - 0001 = 0xx0` , the borrow has been absorbed, and the block is untouched by it as it should have been

IceTDrinker (approved):
🔥

Issue #1477, opened 2024-08-16
Broken link to example in README


peterspbr:
The link to "Find an example with more explanations in [this part of the documentation](https://docs.zama.ai/tfhe-rs/getting-started/quick_start)" is broken and should be replaced with https://docs.zama.ai/tfhe-rs/get-started/quick_start

IceTDrinker:
Thanks for the report @peterspbr we’ll fix it !

PR #1479, opened 2024-08-19
chore(docs): fix README link to getting started

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs/issues/1477


nsarlin-zama:
It seems that the "not found" page returns a HTTP code 200 so with remote url check it still wouldn't have detected it...

nsarlin-zama (approved):

PR #1481, opened 2024-08-19
fix(gpu): fix compression bench name

pdroalves:

### Check-list:

### DIFF
--- a/tfhe/benches/integer/bench.rs
+++ b/tfhe/benches/integer/bench.rs
@@ -1126,15 +1126,11 @@ define_server_key_bench_default_fn!(
 #[cfg(feature = "gpu")]
 mod cuda {
     use super::*;
-    use criterion::{black_box, criterion_group};
+    use criterion::criterion_group;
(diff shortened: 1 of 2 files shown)


IceTDrinker:
could be good to check the benchmarks run before approving/reviewing this PR (maybe only running the compression temporarily to check it works fine)

pdroalves:
Ok, I think this needs attention from someone more familiar to integer benchmarking and slab. It seems that we need to generate parameter files in `tfhe/benchmarks_parameters`for these new benchmarks, and I think these may not be the last modifications we need to do. 

I will rollback to the original strategy of having all compression benchmarks on a separate file, just as the CPU does, so we fix GPU's integer benchmarking. When @agnesLeroy or @soonum are back we return to this topic and (maybe) refactor both benchmark sets.

pdroalves:
@IceTDrinker Now it works: https://github.com/zama-ai/tfhe-rs/actions/runs/10473783028

IceTDrinker (approved):

PR #1482, opened 2024-08-19
Ns/versionize zk

nsarlin-zama:

This PR is based on this one #1296 and will need to be rebased.

### PR content/description
Versionize the ProvenCompactCiphertextList.

#### Arkworks serialization
For this I changed a bit the way the serialization of arkworks data is handled.
Arkworks use a custom serialization framework (not serde). The reason they state for this is that they want to be able to compress the elliptic curve data during serialization.

However, their serialization scheme is "code oriented" (whereas serde is more "data oriented" because the code serialization is generated automatically based only on type definitions). It means that for backward compatibility we need to check not only the types but also the serialization functions, where subtle breaking changes can occur.

This PR changes this by defining serializable counterparts to arkworks types. These types then simply derive "serialize/deserialize" so they can be easily versioned.
The compression is still optionally supported. the "SerializableAffine" type is defined as an enum:
```rust
pub enum SerializableAffine<F> {
    Infinity,
    Compressed { x: F, take_largest_y: bool },
    Uncompressed { x: F, y: F },
}
```
The "compression" happens when this type is created from arkowrk's `Affine` type. Using an enum is easier for versioning than the bitflags system used by arkworks, and is directly compatible with serde.

The serialization methods defined for arkworks are still needed only because they are required by the `SwCurveConfig`.

#### Versioning
Some upgrades were needed to tfhe-versionable:
- handling of Vec<(T, U)>
- better handling of custom bounds in the code generated by the proc macro

### Check-list:

* [ ] Adds a test the data repo for ProvenCompactCiphertextList


nsarlin-zama:
The commit adding pke_v2 has been removed

IceTDrinker:
backward tests failing ?

nsarlin-zama:
> backward tests failing ?

Yes because the data were generated for pke_v2, I just need to regenerate them

tfhe/tests/backward_compatibility/shortint.rs
```diff
@@ -52,6 +49,17 @@ pub fn load_params(test_params: &TestParameterSet) -> ClassicPBSParameters {
...
+    match value {
+        TestDistribution::Gaussian { stddev } => {
+            DynamicDistribution::new_gaussian_from_std_dev(StandardDev(*stddev))
+        }
+        TestDistribution::TUniform { bound_log2 } => {
+            DynamicDistribution::TUniform(TUniform::new(*bound_log2))
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
nsarlin-zama: No it can't because `TestDistribution` comes from the crate `tfhe-backward-compat-data` and `DynamicDistribution` from `tfhe-rs`, but since I am inside an integration test the orphan rule applies.

tmontaigu (approved):

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -1,44 +1,47 @@
...
 #[repr(transparent)]
-struct OneBased<T: ?Sized>(T);
+pub(crate) struct OneBased<T: ?Sized>(T);
 
-impl<T: Valid> Valid for OneBased<T> {
-    fn check(&self) -> Result<(), SerializationError> {
-        self.0.check()
-    }
-}
+// TODO: these impl could be removed by adding support for `repr(transparent)` in tfhe-versionable
```
IceTDrinker, line 15: backlog issue then

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -136,6 +138,34 @@ impl<G: Curve> GroupElements<G> {
...
+    type UncompressError = InvalidSerializedGroupElementsError;
+
+    fn compress(&self) -> Self::Compressed {
+        let mut g_list = Vec::new();
+        let mut g_hat_list = Vec::new();
+        for idx in 0..self.message_len {
+            g_list.push(self.g_list[(idx * 2) + 1].compress());
+            g_list.push(self.g_list[(idx * 2) + 2].compress());
+            g_hat_list.push(self.g_hat_list[idx + 1].compress())
+        }
```
IceTDrinker, line 159: we skip every second element ? I'm not sure I understand

tfhe/Cargo.toml
```diff
@@ -46,7 +46,7 @@ hex = "0.4.3"
 # End regex-engine deps
 # Used for backward compatibility test metadata
 ron = "0.8"
-tfhe-backward-compat-data = { git = "https://github.com/zama-ai/tfhe-backward-compat-data.git", branch = "v0.1", default-features = false, features = [
+tfhe-backward-compat-data = { git = "https://github.com/zama-ai/tfhe-backward-compat-data.git", branch = "v0.2", default-features = false, features = [
```
IceTDrinker, line 49: what's the branch change for ?

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -366,27 +370,27 @@ pub struct CompactCiphertextListExpander {
...
 
-    pub fn get_kind_of(&self, index: usize) -> Option<crate::FheTypes> {
+    fn get_kind_of(&self, index: usize) -> Option<crate::FheTypes> {
         self.inner.get_kind_of(index).and_then(|data_kind| {
             crate::FheTypes::from_data_kind(data_kind, self.inner.message_modulus())
         })
     }
 
-    pub fn get<T>(&self, index: usize) -> Option<crate::Result<T>>
+    fn get<T>(&self, index: usize) -> crate::Result<Option<T>>
```
IceTDrinker, line 388: can't we collapse that into a single Result ?

Maybe not for this release as it would change a few callsites, but a bit later, dunno

tfhe/src/high_level_api/compressed_ciphertext_list.rs
```diff
@@ -97,16 +97,16 @@ impl Tagged for CompressedCiphertextList {
     }
 }
 
-impl CompressedCiphertextList {
-    pub fn len(&self) -> usize {
+impl CiphertextList for CompressedCiphertextList {
```
IceTDrinker, line 100: where is the trait used btw ?

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -366,27 +370,27 @@ pub struct CompactCiphertextListExpander {
     tag: Tag,
 }
 
-impl CompactCiphertextListExpander {
-    pub fn len(&self) -> usize {
+impl CiphertextList for CompactCiphertextListExpander {
```
IceTDrinker, line 373: I'm not a big fan of having to load prelude to be able to use structs, but on the other hand it would lead to code duplication (having the code on the struct and a trait forwarding to those), I guess we just have to tell people to always import the prelude for the HL stuff, which is what we do I believe in the docs

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1,30 +1,182 @@
...
-    hash_w: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash_t: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash_agg: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash_lmap: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash_z: [u8; HASH_METADATA_LEN_BYTES],
+    pub(crate) hash_w: [u8; HASH_METADATA_LEN_BYTES],
+}
+
+// Manual impl of Versionize because TryFrom + generics is currently badly handled by the proc macro
```
IceTDrinker, line 50: backlog issue ?

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1116,24 +1342,26 @@ mod tests {
...
+            }
+        };
 
         // To check management of bigger k_max from CRS during test
         let crs_k = k + 1 + (rng.gen::<usize>() % (d - k));
 
         let original_public_param = crs_gen::<Curve>(d, crs_k, b_i, q, t, rng);
         let public_param_that_was_compressed =
-            serialize_then_deserialize(&original_public_param, Compress::No).unwrap();
+            serialize_then_deserialize(&original_public_param, true).unwrap();
```
IceTDrinker, line 1362: could we have a new type like Compress ? makes it easier to know what the boolean mean

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1209,4 +1437,153 @@ mod tests {
...
+                        c[i + j] = c[i + j].wrapping_add(a[i].wrapping_mul(b[d - j - 1]));
+                    } else {
+                        c[i + j - d] = c[i + j - d].wrapping_sub(a[i].wrapping_mul(b[d - j - 1]));
+                    }
+                }
+            }
+
+            c
+        };
+
```
IceTDrinker, line 1473: factorize out from the other tests as well please

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2270,4 +2556,153 @@ mod tests {
...
+                    if i + j < d {
+                        c[i + j] = c[i + j].wrapping_add(a[i].wrapping_mul(b[d - j - 1]));
+                    } else {
+                        c[i + j - d] = c[i + j - d].wrapping_sub(a[i].wrapping_mul(b[d - j - 1]));
+                    }
+                }
+            }
+
+            c
+        };
```
IceTDrinker, line 2591: factorize out of the test code

IceTDrinker (commented):
Some comments here and there nothing major, nothing critical for the release can be done later

agnesLeroy:
@nsarlin-zama the multi-gpu failure is not related to your PR, we'll investigate it.

tfhe/Cargo.toml
nsarlin-zama: because I updated some test metadata types in `tfhe-backward-compat-data` so it is not backward compatible, meaning that if we don't update the branch, we won't be able to reproduce tests on previous commits in tfhe-rs.

tfhe/src/high_level_api/compact_list.rs
nsarlin-zama: I think @mayeul-zama also had this in mind.

Personally I think they have a slightly different semantic, and having a "None" meaning "not found" makes it in line with the stdlib. But I don't have a strong opinion about that, and I agree that it can simplify the handling for the caller.

tfhe/src/high_level_api/compressed_ciphertext_list.rs
nsarlin-zama: here: https://github.com/zama-ai/tfhe-rs/pull/1482/commits/71523277f4e25ff5e8ef75706e8f45a5148f09bf#diff-629705e126043a5a46affa3b1c72fb1bb635d620b3515a2a895dfec783ff4313L295

With this trait I can de-duplicate some code but I can do without if you think that's not worth having a trait.

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: Actually I wonder if this is necessary since we have a dedicated test for compression.

agnesLeroy:
@nsarlin-zama compression should be fixed on main :)

nsarlin-zama:
> @nsarlin-zama compression should be fixed on main :)

thanks ! :)

tfhe/src/shortint/ciphertext/zk.rs
```diff
@@ -27,13 +32,16 @@ impl CompactPkeCrs {
         // Our plaintext modulus does not take into account the bit of padding
         plaintext_modulus *= 2;
 
+        // 1 padding bit for the PBS
+        // Note that if we want to we can prove carry bits are 0 should we need it
```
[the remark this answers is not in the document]
nsarlin-zama: It's from this pr: https://github.com/zama-ai/tfhe-rs/pull/1570, I didn't set the target branch

tmontaigu (approved):

tmontaigu (approved):

tmontaigu (approved):

PR #1483, opened 2024-08-20
chore(gpu): add debug/release modes

pdroalves:
Introduces a cmake flag, `CMAKE_BUILD_TYPE`, to compile in Release mode (`-O3`) or not (`-O0 -G -g`).

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/570

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/CMakeLists.txt
+++ b/backends/tfhe-cuda-backend/cuda/CMakeLists.txt
@@ -67,9 +67,21 @@ endif()
 
 add_compile_definitions(CUDA_ARCH=${CUDA_ARCH})
 
+# Check if the DEBUG flag is defined
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  # Debug mode
+  message("Compiling in Debug mode")
+  add_definitions(-DDEBUG)
+  set(OPTIMIZATION_FLAGS "${OPTIMIZATION_FLAGS} -O0 -G -g")
+else()
+  # Release mode
+  message("Compiling in Release mode")
+  set(OPTIMIZATION_FLAGS "${OPTIMIZATION_FLAGS} -O3")
+endif()
+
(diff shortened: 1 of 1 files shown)


backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -13,13 +13,18 @@ extern "C" {
...
 inline void cuda_error(cudaError_t code, const char *file, int line) {
   if (code != cudaSuccess) {
     std::fprintf(stderr, "Cuda error: %s %s %d\n", cudaGetErrorString(code),
                  file, line);
     std::abort();
   }
 }
+#else
+inline void cuda_error(cudaError_t code, const char *file, int line) {}
+#endif
```
tmontaigu, line 27: So in release mode we would just let the program continue even when an error occured ? isn't that a bit dangerous ?

pdroalves: Yes,in release mode we wouldn't be  explicitly checking for exceptions during kernel launches (as invalid configurations) or memory operations (as copies to device).
The motivation is to avoid possible overheads.

Not sure about dangers. Would that be bad? Maybe we should rethink @agnesLeroy

tmontaigu: Was the overhead measured ?

Are we sure the tests we run do check that all possible error case are properly checked and that when running in release mode, we would not skip an error and continue in an invalid state ?

If I'm correct the CI tests run in release mode, so the CI would not trigger errors

pdroalves: Ok, this change does not impact performance so I will remove it, since it may also impact correctness / security. 

The debug/release modes during compiling are useful so I will keep the PR only with that.

agnesLeroy (approved):

PR #1484, opened 2024-08-21
chore(integer): addition test based on trivial inputs

tmontaigu:
This adds `overflowing_add` and `add` tests that
are on trivial inputs. As these are faster to run they can be more extensive than on true encryptions

This also binds the advanced_add_assign functions tests to include overflow computation

On a standard laptop with 1 test thread it takes ~7 minutes to run these trivial tests

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_add.rs
@@ -2,7 +2,8 @@ use super::{
     nb_tests_for_params, nb_tests_smaller_for_params, overflowing_add_under_modulus,
     panic_if_any_block_info_exceeds_max_degree_or_noise, panic_if_any_block_is_not_clean,
     panic_if_any_block_values_exceeds_its_degree, random_non_zero_value, unsigned_modulus,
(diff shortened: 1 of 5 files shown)


tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_add.rs
```diff
@@ -69,24 +90,67 @@ where
...
+        let mut result = lhs.clone();
+        let overflowed = sks
+            .advanced_add_assign_with_carry_at_least_4_bits(
+                &mut result.blocks,
+                &rhs.blocks,
+                None,
+                OutputFlag::Carry,
+            )
+            .unwrap();
+        (result, overflowed)
```
IceTDrinker, line 150: no carry management here ? I see it done in other test functions

IceTDrinker (commented):
small question otherwise looks good

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_add.rs
tmontaigu: Aded them just for future proofing if we later change the extensive_trivial test to also have non clean inputs

IceTDrinker (approved):
🚀

PR #1486, opened 2024-08-22
chore(integer): make remaining non-parallel test use defined test cases

tmontaigu:
This makes the remaining non-parallel ops implementation use the same test cases that are used for parallel implementations.

There are still some test that do not share the test case but its either because they do not have a parallel impl (not interesting to have) or when its tests about encryption/decryption

Closes https://github.com/zama-ai/tfhe-rs-internal/issues/265


IceTDrinker (approved):
thanks !

PR #1488, opened 2024-08-23
feat(integer): add is_even/is_odd functions

tmontaigu:
Closes https://github.com/zama-ai/tfhe-rs-internal/issues/639


tfhe/src/integer/gpu/server_key/radix/mod.rs
```diff
@@ -699,6 +700,108 @@ impl CudaServerKey {
...
+                    d_bsk.decomp_base_log,
+                    num_ct_blocks,
+                    self.message_modulus,
+                    self.carry_modulus,
+                    PBSType::Classical,
+                    LweBskGroupingFactor(0),
+                );
+            }
+            CudaBootstrappingKey::MultiBit(d_multibit_bsk) => {
+                apply_univariate_lut_kb_async(
```
IceTDrinker, line 774: is there also bootstrap then keyswitch ? I don't know the cuda code

tfhe/src/integer/server_key/radix/even_odd.rs
```diff
@@ -0,0 +1,92 @@
...
+        )
+    }
+
+    /// Returns an encryption of true if the value is even
+    pub fn smart_is_even_parallelized<T>(&self, ct: &mut T) -> BooleanBlock
+    where
+        T: IntegerRadixCiphertext,
+    {
+        // Even though we don't need to worry about carries,
+        // we still clean them
```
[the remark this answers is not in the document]
tmontaigu: Yes, but i did that because if we don't clean, we would not need the `&mut` in the input so as options we
* can add an #[allow]
* or make the smart take a `&`,
* or have no smart version
* or change the check in the if be block.carries_are_full()

tfhe/src/integer/gpu/server_key/radix/mod.rs
tmontaigu: I think they only do ks_pbs not pbs_ks

IceTDrinker: that's weird it seems there is a PBS order in the struct

tmontaigu: Yep its in the struct, but its not really used, a quick grep showed me that the inly place where its use is in the gpu's create_trivial_radix function

IceTDrinker: Can be a different pr to check the logic with the gpu team

tfhe/src/integer/gpu/server_key/radix/even_odd.rs
```diff
@@ -0,0 +1,143 @@
...
+        ct: &T,
+        streams: &CudaStreams,
+    ) -> CudaBooleanBlock
+    where
+        T: CudaIntegerRadixCiphertext,
+    {
+        let radix = ct.as_ref();
+        let lut = self.generate_lookup_table(|block| u64::from((block & 1) == 0));
+        let mut single_block = CudaRadixCiphertext {
+            d_blocks: CudaLweCiphertextList::new(
```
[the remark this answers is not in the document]
tmontaigu: No its sync

tfhe/src/integer/server_key/radix/even_odd.rs
```diff
@@ -63,7 +63,11 @@ impl ServerKey {
     {
         // Even though we don't need to worry about carries,
         // we still clean them
-        if !ct.block_carries_are_empty() {
+        if ct
+            .blocks()
+            .iter()
+            .any(|block| self.key.max_degree.validate(block.degree).is_err())
```
IceTDrinker, line 69: any reason for this change ?

tmontaigu: its related to this comment https://github.com/zama-ai/tfhe-rs/pull/1488#discussion_r1731437886

this function does not care about carries, so a smart variant does no make sense unless there is some carry cleaning (or we accept to put a #[allow] of some sort

IceTDrinker: ah right, I think we have some unused &mut on smart variants where cleaning is not done we end up with an allow IIRC

IceTDrinker:
good to squash, I'll approve the squashed commits

IceTDrinker (commented):
Good like this go for squash, I'll approve then

IceTDrinker (approved):

PR #1494, opened 2024-08-27
chore(hl): a bad merge led to two server key generation

IceTDrinker:
probably what led to some wasm bench to be slower

cc @nsarlin-zama when you are back


IceTDrinker:
will need a backport in 0.7

tmontaigu (approved):

PR #1496, opened 2024-08-28
feat(wasm): bind CompactListExpander::get_kind_of

tmontaigu:
And other things to allow wasm users to explore
what kind of data is stored in the compact list.


tmontaigu:
Technically, the get_kind_of API could be implemented on the `CompactCiphertextList` and `ProvenCompactCiphertextList` types

This would allow users to chek the data kinds at each slot without before having to expand or verify the zk proof, is it something that we want to have ?

IceTDrinker:
For BC we need to be able to do it without expanding

IceTDrinker (commented):
As discussed it would likely be better to be able to call those functions on the objects themselves as the issue is to be able to fetch metadata without expanding (as if there is packing currently it's not possible to run PBSes in wasm to unpack the blocks)

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -18,6 +18,57 @@ use crate::shortint::MessageModulus;
...
+                    8 => Self::Int8,
+                    10 => Self::Int10,
+                    12 => Self::Int12,
+                    14 => Self::Int14,
+                    16 => Self::Int16,
+                    32 => Self::Int32,
+                    64 => Self::Int64,
+                    128 => Self::Int128,
+                    160 => Self::Int160,
+                    256 => Self::Int256,
```
IceTDrinker, line 63: no larger signed values I'm guessing ?

tfhe/src/high_level_api/compact_list.rs
```diff
@@ -31,6 +82,20 @@ impl CompactCiphertextList {
...
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn get_kind_of(&self, index: usize) -> Option<crate::FheTypes> {
+        self.0.get_kind_of(index).and_then(|data_kind| {
+            crate::FheTypes::from_data_kind(data_kind, self.0.ct_list.message_modulus)
+        })
+    }
```
IceTDrinker, line 97: does not work for packed lists ? the modulus would be message_modulus * carry_modulus ?

IceTDrinker:
damn it may not be possible to properly return the data kind without expansion if we have a block ending at a boundary in a ciphertext when things are packed ?

tfhe/src/high_level_api/compact_list.rs
tmontaigu: Yep, 512, 1024, and 2048 are available as uint as the blockchain requested them, but its not needed for signed ints

tfhe/src/high_level_api/compact_list.rs
tmontaigu: Yes, it work because the metada is metadata, it does not care whether the actual block are packed or not

The integer metadata keeps string of the bool or signed/unsigned kind with the number of blocks or the original input, the hlapi list just require the message modulus to be able to know how many bits a block encrypts to then be able to translate a num blocks in a num bits

IceTDrinker (approved):
Thanks for the clarifications !

tmontaigu:
The test of get_kind_of used on ProvenCompactCiphertextList (which iirc is packed) should prove that things are ok even when data is packed

PR #1505, opened 2024-09-02
refactor(gpu): avoid synchronizations in the keybundle

guillermo-oyarzun:
closes: [#667](https://github.com/orgs/zama-ai/projects/21?pane=issue&itemId=74014357)

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
@@ -18,9 +18,9 @@
 #include <vector>
 
 template <typename Torus, class params>
-__device__ Torus calculates_monomial_degree(const Torus *lwe_array_group,
-                                            uint32_t ggsw_idx,
-                                            uint32_t grouping_factor) {
+__device__ uint32_t calculates_monomial_degree(const Torus *lwe_array_group,
+                                               uint32_t ggsw_idx,
+                                               uint32_t grouping_factor) {
   Torus x = 0;
   for (int i = 0; i < grouping_factor; i++) {
     uint32_t mask_position = grouping_factor - (i + 1);
@@ -29,6 +29,13 @@ __device__ Torus calculates_monomial_degree(const Torus *lwe_array_group,
   }
 
   return modulus_switch(x, params::log2_degree + 1);
+}
+
+__device__ __forceinline__ int
+get_start_ith_ggsw_offset(uint32_t polynomial_size, int glwe_dimension,
+                          uint32_t level_count) {
+  return polynomial_size * (glwe_dimension + 1) * (glwe_dimension + 1) *
+         level_count;
 }
 
 template <typename Torus, class params, sharedMemDegree SMD>
@@ -60,8 +67,6 @@ __global__ void device_multi_bit_programmable_bootstrap_keybundle(
   uint32_t input_idx = blockIdx.x / lwe_chunk_size;
 
   if (lwe_iteration < (lwe_dimension / grouping_factor)) {
-    //
-    Torus *accumulator = (Torus *)selected_memory;
 
     const Torus *block_lwe_array_in =
         &lwe_array_in[lwe_input_indexes[input_idx] * (lwe_dimension + 1)];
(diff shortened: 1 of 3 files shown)


agnesLeroy (commented):
Hey @guillermo-oyarzun! This PR looks good to me, thanks a lot! I just have a small comment about a possible simplification, but very minor. I don't fully understand why the __syncthreads() at line 119 is needed, you wrote it's to avoid having the monomial degrees overwritten but I don't understand how they could be overwritten at that stage.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
```diff
@@ -81,57 +86,52 @@ __global__ void device_multi_bit_programmable_bootstrap_keybundle(
...
-      temp[i].y /= (double)std::numeric_limits<Torus>::max();
-      tid += params::degree / params::opt;
+      uint32_t monomial_degree = monomial_degrees[g];
+
+      const Torus *bsk_poly = bsk_poly_ini + g * offset;
+      // Multiply by the bsk element
+      polynomial_product_accumulate_by_monomial_nosync<Torus, params>(
+          reg_acc, bsk_poly, monomial_degree);
     }
+    __syncthreads(); // need to make sure that the monomial_degrees are not
```
agnesLeroy, line 119: > Here I don't understand why we need to synctreads again

guillermo-oyarzun: I was being extra careful, if you see in the following lines i reuse the shared memory to store the fft elements. I have seen that the compiler at low level adds a sync when I re-cast the shared memory, but i was afraid that this might not be the case for all architectures, so I thought it was better to sync explicitly there. I re-run the benchmarks after adding this sync (which was not there in the beginning) and the performance was not affected (probably because it was already done implicitly).

agnesLeroy: Oh ok, maybe it would be worth updating the comment then to reflect this. It would mean that if the FFT takes data from the register directly we could avoid this synchronization?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
```diff
@@ -81,57 +86,52 @@ __global__ void device_multi_bit_programmable_bootstrap_keybundle(
...
-    // Accumulate the other terms
-    for (int g = 1; g < (1 << grouping_factor); g++) {
+    copy_polynomial_in_regs<Torus, params::opt, params::degree / params::opt>(
+        bsk_poly_ini, reg_acc);
 
-      const Torus *bsk_slice = get_multi_bit_ith_lwe_gth_group_kth_block(
-          bootstrapping_key, g, rev_lwe_iteration, glwe_id, level_id,
-          grouping_factor, 2 * polynomial_size, glwe_dimension, level_count);
-      const Torus *bsk_poly = bsk_slice + poly_id * params::degree;
+    int offset = get_start_ith_ggsw_offset(2 * polynomial_size, glwe_dimension,
```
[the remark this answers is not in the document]
agnesLeroy: Yes sounds good, I'll open an issue to refactor get_ith_mask_kth_block.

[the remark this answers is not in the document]
agnesLeroy: https://github.com/zama-ai/tfhe-rs-internal/issues/696

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
guillermo-oyarzun: we would transfer this synchronization up to the calculation of the first fft level, but that requires to have a specific fft_direct that receives the data from registers, because in the other parts that call it the fft elements are expected in shared memory already (pbs or keygen).

agnesLeroy: Yes totally, @bbarbakadze maybe something to look at with you next week at the office :slightly_smiling_face:

PR #1507, opened 2024-09-03
chore(ci): avoid running integer tests on push to internal repo

soonum:
Closes: zama-ai/tfhe-rs-internal#695

### DIFF
--- a/.github/workflows/aws_tfhe_integer_tests.yml
+++ b/.github/workflows/aws_tfhe_integer_tests.yml
@@ -26,6 +26,11 @@ on:
 
 jobs:
   should-run:
+    if:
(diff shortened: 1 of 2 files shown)


.github/workflows/aws_tfhe_integer_tests.yml
```diff
@@ -26,6 +26,7 @@ on:
 
 jobs:
   should-run:
+    if: github.event_name == 'push' && github.repository == 'zama-ai/tfhe-rs'
```
IceTDrinker, line 29: too restrictive on the event_name, it's probably the same condition as the setup instance that is required

IceTDrinker (approved):
Thanks !

PR #1512, opened 2024-09-04
fix(integer): do sum by safe chunk sizes

tmontaigu:
Parameters are made with with assumptions on the number of leveled add/sub/scalar_mul operations are made, so that the noise level before doing a PBS has a correct level and everything is safe, secure and correct.

So the lib implementation has to uphold these assumptions in order to keep the error probability failure correct.

In the comparisons, at some point we had a vector of ciphertexts with a degree == 1, so we greedily summed them (e.g with 2_2 params we summed them by chunks of 15), while it is correct with regards to the carry and message space it is however less correct with regards to the noise level.

Noise wise, doing this huge sum is correct as long as the noise of each ciphertext is independent from the others in the same chunk.

While it may generally be the case we are in, its not guaranteed, and since we do not track that information we have to take the safer approach of assuming the worst case: all noise are dependent.

So to fix the issue we compute the correct size of sum chunk by also taking into account the max noise level.

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
@@ -74,7 +74,7 @@ __host__ void are_all_comparisons_block_true(
   auto tmp_out = are_all_block_true_buffer->tmp_out;
 
   uint32_t total_modulus = message_modulus * carry_modulus;
-  uint32_t max_value = total_modulus - 1;
+  uint32_t max_value = (total_modulus - 1) / (message_modulus - 1);
 
   cuda_memcpy_async_gpu_to_gpu(tmp_out, lwe_array_in,
                                num_radix_blocks * (big_lwe_dimension + 1) *
@@ -173,7 +173,7 @@ __host__ void is_at_least_one_comparisons_block_true(
   auto buffer = mem_ptr->eq_buffer->are_all_block_true_buffer;
 
   uint32_t total_modulus = message_modulus * carry_modulus;
-  uint32_t max_value = total_modulus - 1;
+  uint32_t max_value = (total_modulus - 1) / (message_modulus - 1);
 
   cuda_memcpy_async_gpu_to_gpu(mem_ptr->tmp_lwe_array_out, lwe_array_in,
                                num_radix_blocks * (big_lwe_dimension + 1) *
--- a/tfhe/src/integer/server_key/mod.rs
+++ b/tfhe/src/integer/server_key/mod.rs
@@ -9,7 +9,7 @@ pub(crate) mod radix;
 pub(crate) mod radix_parallel;
 
 use crate::integer::client_key::ClientKey;
-use crate::shortint::ciphertext::MaxDegree;
+use crate::shortint::ciphertext::{Degree, MaxDegree};
 use serde::{Deserialize, Serialize};
 use tfhe_versionable::Versionize;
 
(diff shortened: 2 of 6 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -174,7 +175,7 @@ __host__ void is_at_least_one_comparisons_block_true(
   auto buffer = mem_ptr->eq_buffer->are_all_block_true_buffer;
 
   uint32_t total_modulus = message_modulus * carry_modulus;
-  uint32_t max_value = total_modulus - 1;
+  uint32_t max_value = ()total_modulus - 1) / (message_modulus - 1);
```
agnesLeroy, line 178: Typo here with the parenthesis?

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
```diff
@@ -14,7 +15,7 @@ impl ServerKey {
     /// - Returns None if the scalar is in the range of values that the ciphertext can represent
     ///
     /// - Returns Some(ordering) when the scalar is out of representable range of the ciphertext.
-    ///     - Equal will never be returned
+    ///     - Equal will neve\r be returned
```
agnesLeroy, line 18: Another typo

tfhe/src/integer/server_key/mod.rs
```diff
@@ -231,6 +231,25 @@ impl ServerKey {
...
+    /// Returns how many ciphertext can be summed at once
+    ///
+    /// The number of ciphertext that can be added together depends on the degree
+    /// (in order not to go beyond the carry space and keep results correct) but also
+    /// on the noise level (in order to have the correct error probability and so correctness and
+    /// security)
+    ///
+    /// - `degree` is expected degree of all elements to be summed
+    pub(crate) fn max_sum_size(&self, degree: Degree) -> usize {
+        let max_degree = (self.message_modulus().0 * self.carry_modulus().0) - 1;
```
IceTDrinker, line 244: there is a function MaxDegree::from_msg_carry_modulus that does just that

tfhe/src/integer/server_key/mod.rs
```diff
@@ -231,6 +231,26 @@ impl ServerKey {
...
+    ///
+    /// - `degree` is expected degree of all elements to be summed
+    pub(crate) fn max_sum_size(&self, degree: Degree) -> usize {
+        let max_degree =
+            MaxDegree::from_msg_carry_modulus(self.message_modulus(), self.carry_modulus());
+        let max_sum_to_full_carry = max_degree.get() / degree.get();
+
+        // We want to compare with the true max noise level, that is the one
+        // were we don't need enough room to accept a carry
+        let true_max_noise_level = max_degree.get() / (self.message_modulus().0 - 1);
```
IceTDrinker, line 250: self.key.max_noise_level I believe has the value you are looking for

tmontaigu: No because its made to be able to receive a carry, when this is for when we want to be completely full

IceTDrinker: not sure I follow, 15 / 3 = 5 and we have 5 in the key ?

tmontaigu: No, the max degree in the key is computed via

```
  pub(crate) fn integer_radix_server_key(
        message_modulus: MessageModulus,
        carry_modulus: CarryModulus,
    ) -> Self {
        let full_max_degree = message_modulus.0 * carry_modulus.0 - 1;

        let carry_max_degree = carry_modulus.0 - 1;

        // We want to be have a margin to add a carry from another block
        Self::new(full_max_degree - carry_max_degree)
    }
```

IceTDrinker: what I mean is that the noise is dependeant on the parameters, the max degree you already re compute it

tmontaigu: I yeah bit TRUE, I misread sorry

IceTDrinker: no worries :)

IceTDrinker:
integer test failure https://github.com/zama-ai/tfhe-rs/actions/runs/10813176643/job/29996656992

tfhe/src/integer/server_key/radix_parallel/comparison.rs
```diff
@@ -50,34 +50,8 @@ impl ServerKey {
...
-                .collect_into_vec(&mut block_comparisons_2);
-            std::mem::swap(&mut block_comparisons_2, &mut block_comparisons);
-        }
-
-        BooleanBlock::new_unchecked(
-            block_comparisons
-                .into_iter()
-                .next()
-                .unwrap_or_else(|| self.key.create_trivial(0)),
-        )
```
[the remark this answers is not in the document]
IceTDrinker: agree, should the bit mask be removed ?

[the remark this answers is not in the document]
tmontaigu: I did

PR #1514, opened 2024-09-06
chore(gpu): pass over all cuda bind

agnesLeroy:
closes:
### PR content/description
I rewrote all the binding to be sure it's consistent with the c++ headers.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
+++ b/backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
@@ -8,7 +8,7 @@ extern std::mutex m;
 extern bool p2p_enabled;
 
 extern "C" {
-int cuda_setup_multi_gpu();
+int32_t cuda_setup_multi_gpu();
 }
 
 // Define a variant type that can be either a vector or a single pointer
--- a/backends/tfhe-cuda-backend/src/cuda_bind.rs
+++ b/backends/tfhe-cuda-backend/src/cuda_bind.rs
@@ -3,27 +3,18 @@ use std::ffi::c_void;
 #[link(name = "tfhe_cuda_backend", kind = "static")]
 extern "C" {
 
-    /// Create a new Cuda stream on GPU `gpu_index`
     pub fn cuda_create_stream(gpu_index: u32) -> *mut c_void;
 
-    /// Destroy the Cuda stream `v_stream`
     pub fn cuda_destroy_stream(stream: *mut c_void, gpu_index: u32);
 
-    /// Allocate `size` memory on GPU `gpu_index` asynchronously
+    pub fn cuda_synchronize_stream(stream: *mut c_void, gpu_index: u32);
+
+    pub fn cuda_malloc(size: u64, gpu_index: u32) -> *mut c_void;
+
     pub fn cuda_malloc_async(size: u64, stream: *mut c_void, gpu_index: u32) -> *mut c_void;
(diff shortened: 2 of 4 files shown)


backends/tfhe-cuda-backend/src/cuda_bind.rs
```diff
@@ -3,104 +3,58 @@ use std::ffi::c_void;
 #[link(name = "tfhe_cuda_backend", kind = "static")]
 extern "C" {
```
IceTDrinker, line 4: "stupid" question are we sure the C ABI is respected when the CUDA backend is compiled ? in the sense that C++ does not exactly have the C ABI and I don't know how both relate 🤔

IceTDrinker: (related to the potential launch failure stuff)

agnesLeroy: Hmm I'm not totally sure no

IceTDrinker: this maybe https://isocpp.org/wiki/faq/mixing-c-and-cpp#call-c

IceTDrinker: this way you make sure in the C++ code base that you expose the C function as extern "C"

agnesLeroy: We should be alright, as all the entry points in the binding are actually "C compatible". Nothing C++-like is passed through the interface, only raw pointer types.

IceTDrinker: while I agree, I would not be against the "ceinture et bretelles" approach here :)

agnesLeroy: And we define all functions that go through the interface with "extern C" in the header files.

IceTDrinker: so then we are good on the ABI declaration, nothing else to do on the C++ side, just need to make sure definitions agree, do you want someone to double check function by function ?

agnesLeroy: I think this would be very troublesome and not so efficient: I already went through it thoroughly yesterday. But really at some point we should spend some time to figure out how to generate this API automatically, at least part of it.

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
```diff
@@ -8,7 +8,7 @@ extern std::mutex m;
 extern bool p2p_enabled;
 
 extern "C" {
-int cuda_setup_multi_gpu();
+int32_t cuda_setup_multi_gpu();
```
pdroalves, line 11: For curiosity, why do we want to have `int32_t` instead of `int` here?

pdroalves (approved):
Looks good. Just moving functions and removing unnecessary comments in the binding file helps a lot.

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
agnesLeroy: Just to be sure with the types on any architecture, I prefer to always specify the size of the integer (as we do on the Rust side as well).

PR #1516, opened 2024-09-06
chore(ci): use python script to send benchmark results

soonum:
Using this script simplify writing of corresponding workflow step. Moreover, now when an upload fails it translate into a workflow failure.

This needs to wait for new release of Slab to shipped to production before merging.


IceTDrinker (commented):
Looks good, is there a way to do a workflow dispatch on this PR to see if it works as expected ?

soonum:
I've just tried on boolean benchmarks and it worked (after having fixed the name of the script)

PR #1523, opened 2024-09-10
feat(zk): manage D as an upper bound as in the report

IceTDrinker:
- allows to prove less slots than what the CRS can handle

### DIFF
--- a/tfhe-zk-pok/src/proofs/pke_v2.rs
+++ b/tfhe-zk-pok/src/proofs/pke_v2.rs
@@ -300,10 +300,10 @@ pub fn prove<G: Curve>(
     let (
         &PublicParams {
             ref g_lists,
-            D,
+            D: D_max,
             n,
             d,
-            k,
+            k: k_max,
             B,
             B_r: _,
             B_bound,
@@ -328,8 +328,13 @@ pub fn prove<G: Curve>(
 
     let PrivateCommit { r, e1, m, e2, .. } = private_commit;
 
-    assert!(c2.len() <= k);
-    let k = k.min(c2.len());
+    let k = c2.len();
+    assert!(k <= k_max);
+
+    // Recompute the D for our case if k is smaller than the k max
+    // formula in Prove_pp: 2.
+    let D = d + k * t_input.ilog2() as usize;
+    assert!(D <= D_max);
 
     // FIXME: div_round
     let delta = {
@@ -1456,10 +1461,10 @@ pub fn verify<G: Curve>(
 
     let &PublicParams {
         ref g_lists,
-        D,
+        D: D_max,
         n,
         d,
-        k,
+        k: k_max,
         B,
         B_r: _,
         B_bound: _,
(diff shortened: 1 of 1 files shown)


tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -328,8 +328,14 @@ pub fn prove<G: Curve>(
 
     let PrivateCommit { r, e1, m, e2, .. } = private_commit;
 
-    assert!(c2.len() <= k);
-    let k = k.min(c2.len());
+    assert!(c2.len() <= k_max);
+    let k = k_max.min(c2.len());
```
nsarlin-zama, line 332: with the above assert this can be just `let k = c2.len()`

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -1487,10 +1493,18 @@ pub fn verify<G: Curve>(
     };
 
     let PublicCommit { a, b, c1, c2, .. } = public.1;
-    if c2.len() > k {
+    if c2.len() > k_max {
+        return Err(());
+    }
+    let k = k_max.min(c2.len());
```
nsarlin-zama, line 1499: same, considering the above check this is equivalent to `let k = c2.len();`

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -328,8 +328,14 @@ pub fn prove<G: Curve>(
...
-    assert!(c2.len() <= k);
-    let k = k.min(c2.len());
+    assert!(c2.len() <= k_max);
+    let k = k_max.min(c2.len());
+
+    // Recompute the D for our case if k is smaller than the k max
+    // formula in Prove_pp: 2.
+    let D = d + k * t_input.ilog2() as usize;
+
+    assert!(D <= D_max);
```
nsarlin-zama, line 338: this API use a result on one side and asserts on the other side, this could be uniformized (but that's maybe out of the scope of this pr)

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2128,7 +2142,7 @@ mod tests {
                 PublicParams::deserialize_with_mode(data.as_slice(), compress, Validate::No)
             };
 
-        let original_public_param = crs_gen_ghl::<Curve>(d, k, B, q, t, rng);
+        let original_public_param = crs_gen_ghl::<Curve>(d, 2 * k, B, q, t, rng);
```
nsarlin-zama, line 2145: Maybe we should use `crs_gen_cs` since that's the one used in the impl, or simply `crs_gen`

IceTDrinker: yep you are right here

tfhe-zk-pok/src/proofs/pke_v2.rs
IceTDrinker: indeed, we could do k = c2.len(); k <= k_max

tfhe-zk-pok/src/proofs/pke_v2.rs
IceTDrinker: changed the call to have crs_gen here instead, changed the 2 * k to something giving a random value bigger than k and <= to d

nsarlin-zama (approved):

PR #1526, opened 2024-09-11
refactor(shortint): move parameters set to their own directory

soonum:
This is done to ease automatic parameters updates.


IceTDrinker (approved):
Thanks !

PR #1530, opened 2024-09-11
Am/chore/remove lint reason

IceTDrinker:


IceTDrinker:
Actually @soonum I'll split this and do the toolchain change at a different time to avoid breakage before the alpha.5 for the BC

soonum (approved):

PR #1546, opened 2024-09-17
fix: expand_with_key was not providing the safest set of modes

IceTDrinker:
- it meant that lists needing unpacking could crash during expand

added non reg test


nsarlin-zama (approved):
thanks for the comments :)

PR #1549, opened 2024-09-18
chore(ci): remove support for slab calls with issue comments

soonum:
Now all workflows use Slab GitHub Action and thus can be launched directly with a workflow_dispatch event.


IceTDrinker (approved):
🔥

PR #1551, opened 2024-09-18
Ns/fix/versionable bad bounds

nsarlin-zama:
needed by: #1482 

### PR content/description
Fix the bounds added in the derived traits for the Versionize macro. These bounds were too restrictive and there were cases where they were not satisfiable, so the `versionize` method on a type was actually not callable. This was the case for example when a type was defined with a generic that is not itself used, but only through associated types. For example:

```rust
trait WithAssoc {
  type Assoc;
}

struct MyStruct<T: WithAssoc> {
  inner: T::Assoc
}
```

Using the `#[derive(Versionize)]` on this type before the PR would require `T: Versionize` which is actually not necessary and sometimes not possible. This was caused by the proc macro trying to apply bounds directly on the generics of the type.

The main idea behind this PR is to add bounds closer to what is actually used, and not necessarily to the generics. For example, If I have 3 traits:
```rust
trait TraitA {}

trait TraitB {
  fn method_b(&self);
}

trait TraitC {
  fn method_c(&self);
}
```

and these 2 types:
```rust
struct Type1<T> {
  value: T
}

impl<T: TraitA> TraitB for Type1<T> {
  fn method_b(&self) {}
}

struct Type2<T> {
  inner: Type1<T>
}

impl<T: ??> TraitC for Type2<T> where ?? {
  fn method_c(&self) {
    self.inner.method_b();
  }
}
```
I have 2 ways of expressing the bound:
- `impl<T: TraitA> TraitC for Type2<T>`
- `impl<T> TraitC for Type2<T> where Type1<T>: TraitB`

This PR works by using the second method instead of the first one. The advantage is that theses bounds are easier to generate by looking only at `Type2`, without knowing the internals of `Type1`.

As a side effect, the `#[versionize(..., bound = "...")]` attribute for the proc macro is not necessary anymore to express the bounds present in tfhe-rs, so it have been removed. This is a **breaking change**. There might still be specific corner cases that will require it, but as I don't have any example to test it I preferred to removed it.

This PR also adds the versioning of vector of small tuples of arbitrary versioned types.

### Check-list:

### DIFF
--- a/utils/tfhe-versionable/README.md
+++ b/utils/tfhe-versionable/README.md
@@ -1,9 +1,15 @@
 # TFHE-versionable
-This crate provides type level versioning for serialized data. It offers a way to add backward
-compatibility on any data type. The versioning scheme works recursively and is independant of the
-chosen serialization backend.
+This crate provides type level versioning for serialized data. It offers a way
+to add backward compatibility on any data type. The versioning scheme works
+recursively and is independant of the chosen serialized file format. It uses the
+`serde` framework.
 
-To use it, simply define an enum that have a variant for each version of your target type.
+The crate will convert any type into an equivalent packed with versions
+information. This "versioned" type is then serializable using any format
+compatible with `serde`.
+
+To use it, simply define an enum that have a variant for each version of your
+target type.
 
 For example, if you have defined an internal type:
 ```rust
@@ -19,7 +25,9 @@ enum MyStructVersions {
 }
 ```
 
-If at a subsequent point in time you want to add a field to this struct, the idea is to copy the previously defined version of the struct and create a new one with the added field. This mostly becomes:
+If at a subsequent point in time you want to add a field to this struct, the
+idea is to copy the previously defined version of the struct and create a new
+one with the added field. This mostly becomes:
 ```rust
 struct MyStruct {
 	val: u32,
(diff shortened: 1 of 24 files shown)


nsarlin-zama:
For reference, the rust std lib and serde have the same problem this PR is trying to solve: https://github.com/rust-lang/rust/issues/26925.
Serde even implemented it in 0.7 and reverted in 0.8. dtolnay explained the 3 cases where this approach did not work at the time: https://github.com/dtolnay/syn/issues/370#issuecomment-371881224:
1. **Private in public**: when a pub type has a private field, the bound expose this private type
2. **Overflow evaluating requirements**: when there are recursions in the type definitions, the trait solver fails to resolve the bounds
3. **Lifetime deduplication inference is broken**:

nsarlin-zama:
I have done a few tests on these issues:
1. is fixed
2. still fails to compile due to the circular dependency in the type. If we ever use this kind of type we will need to do a manual impl
3. fails to compile but I think the problem is not only caused by the rust compiler, lifetime generics are not very well handled by the proc macro. Since we avoid them in tfhe-rs I have not properly checked them but that is something that I should fix when I have more time.

utils/tfhe-versionable/README.md
```diff
@@ -1,9 +1,14 @@
 # TFHE-versionable
-This crate provides type level versioning for serialized data. It offers a way to add backward
-compatibility on any data type. The versioning scheme works recursively and is independant of the
-chosen serialization backend.
+This crate provides type level versioning for serialized data. It offers a way
+to add backward compatibility on any data type. The versioning scheme works
+recursively and is independant of the chosen serialization backend.
 
-To use it, simply define an enum that have a variant for each version of your target type.
+The crate will convert any type into an equivalent packad with versions
```
IceTDrinker, line 6: packad ?

utils/tfhe-versionable/README.md
```diff
@@ -1,9 +1,14 @@
 # TFHE-versionable
-This crate provides type level versioning for serialized data. It offers a way to add backward
-compatibility on any data type. The versioning scheme works recursively and is independant of the
-chosen serialization backend.
+This crate provides type level versioning for serialized data. It offers a way
+to add backward compatibility on any data type. The versioning scheme works
+recursively and is independant of the chosen serialization backend.
```
IceTDrinker, line 4: Maybe worth indicating it uses the serde framework, even though it's almost a given in rust

utils/tfhe-versionable/src/lib.rs
```diff
@@ -632,6 +632,7 @@ impl Unversionize for () {
 
 impl NotVersioned for () {}
 
+// TODO: use a macro for more tuple sizes
```
mayeul-zama, line 635: I guess you planned to implement that only if we need it at some point

utils/tfhe-versionable/README.md
mayeul-zama, line 6: packad: typo

IceTDrinker (commented):
I won't pretend I understand everything that's going on, but the assurance it works with the test runs gives reassurance on that end (and the fact our lib and others are using this successfully)

some small stuff I saw

utils/tfhe-versionable/README.md
nsarlin-zama: I think I meant "packed"

utils/tfhe-versionable/src/lib.rs
nsarlin-zama: yes since we don't need it I didn't make it a priority

mayeul-zama (approved):
Thanks!

PR #1553, opened 2024-09-18
fix(gpu): remove all resettings of shared memory size

agnesLeroy:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -116,12 +116,6 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
   switch (polynomial_size) {
   case 256:
     if (shared_memory_size <= cuda_get_max_shared_memory(0)) {
-      check_cuda_error(cudaFuncSetAttribute(
-          batch_NSMFFT<FFTDegree<AmortizedDegree<256>, ForwardFFT>, FULLSM>,
-          cudaFuncAttributeMaxDynamicSharedMemorySize, shared_memory_size));
-      check_cuda_error(cudaFuncSetCacheConfig(
```
pdroalves, line 122: Maybe we could keep this one since it's always setting the same value `cudaFuncCachePreferShared`.

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -447,6 +447,8 @@ impl<T: Numeric> Drop for CudaVec<T> {
     /// Free memory for pointer `ptr` synchronously
     fn drop(&mut self) {
         for &gpu_index in self.gpu_indexes.iter() {
+            // Synchronizes the device to be sure no stream is still using this pointer
+            synchronize_device(gpu_index);
```
pdroalves, line 451: Are we mixing PRs?

pdroalves (approved):
My only doubt is regarding setting `cudaFuncCachePreferShared`. Since it's the same value I don't think this is a race condition. We could keep it.

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: It's based on a previous PR sorry, only the second commit should be looked at here.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
agnesLeroy: Does it do anything actually? :thinking:

pdroalves: Not sure. Probably none on Hopper, but could affect low end GPUs I think. Or maybe is a legacy thing that was only relevant on older archs. What if we bench this commit to see the impact if we remove it?

agnesLeroy: Bench running: https://github.com/zama-ai/tfhe-rs/actions/runs/10995585669

PR #1570, opened 2024-09-20
Am/feat/zk v1 padding

IceTDrinker:


nsarlin-zama (approved):
fixed a missing modulus in the tests, but everything else looks good

PR #1580, opened 2024-09-24
Ns/clippy versionable

nsarlin-zama:
The idea is to run examples as tests, as it is not done by default by `cargo test` (even with `cargo test --examples`). Adding the `#[test]` to the main worked with `cargo test --examples` but not in the other cases.


IceTDrinker:
as long as we are sure tests are running then fine by me, if you can check and share the log link here

IceTDrinker (commented):
Log link to test run, other wise this looks good !

IceTDrinker (approved):
Thanks for the links !

PR #1597, opened 2024-09-27
Ns/refactor zk tests

nsarlin-zama:
Removes code duplication in zk tests

### DIFF
--- a/tfhe-zk-pok/src/proofs/mod.rs
+++ b/tfhe-zk-pok/src/proofs/mod.rs
@@ -146,3 +146,217 @@ pub mod pke;
 pub mod pke_v2;
 pub mod range;
 pub mod rlwe;
+
+#[cfg(test)]
+mod test {
+    #![allow(non_snake_case)]
+    use std::fmt::Display;
+
+    use bincode::ErrorKind;
+    use rand::rngs::StdRng;
+    use rand::Rng;
+    use serde::{Deserialize, Serialize};
+
+    use crate::curve_api::Compressible;
+
+    // One of our usecases uses 320 bits of additional metadata
+    pub(super) const METADATA_LEN: usize = (320 / u8::BITS) as usize;
+
+    pub(super) enum Compress {
+        Yes,
+        No,
+    }
+
+    pub(super) fn serialize_then_deserialize<
+        Params: Compressible + Serialize + for<'de> Deserialize<'de>,
+    >(
+        public_params: &Params,
+        compress: Compress,
+    ) -> bincode::Result<Params>
+    where
+        <Params as Compressible>::Compressed: Serialize + for<'de> Deserialize<'de>,
+        <Params as Compressible>::UncompressError: Display,
+    {
+        match compress {
+            Compress::Yes => Params::uncompress(bincode::deserialize(&bincode::serialize(
+                &public_params.compress(),
+            )?)?)
+            .map_err(|e| Box::new(ErrorKind::Custom(format!("Failed to uncompress: {}", e)))),
+            Compress::No => bincode::deserialize(&bincode::serialize(&public_params)?),
+        }
+    }
+
+    pub(super) fn polymul_rev(a: &[i64], b: &[i64]) -> Vec<i64> {
+        assert_eq!(a.len(), b.len());
+        let d = a.len();
+        let mut c = vec![0i64; d];
+
+        for i in 0..d {
+            for j in 0..d {
+                if i + j < d {
+                    c[i + j] = c[i + j].wrapping_add(a[i].wrapping_mul(b[d - j - 1]));
+                } else {
+                    c[i + j - d] = c[i + j - d].wrapping_sub(a[i].wrapping_mul(b[d - j - 1]));
+                }
+            }
+        }
+
+        c
+    }
+
+    /// Parameters needed for a PKE zk proof test
+    #[derive(Copy, Clone)]
+    pub(super) struct PkeTestParameters {
+        pub(super) d: usize,
+        pub(super) k: usize,
+        pub(super) B: u64,
+        pub(super) q: u64,
+        pub(super) t: u64,
+        pub(super) msbs_zero_padding_bit_count: u64,
+    }
+
+    /// An encrypted PKE ciphertext
+    pub struct PkeTestCiphertext {
+        pub(super) c1: Vec<i64>,
+        pub(super) c2: Vec<i64>,
+    }
+
+    /// A randomly generated testcase of pke encryption
+    pub(super) struct PkeTestcase {
+        pub(super) a: Vec<i64>,
+        pub(super) e1: Vec<i64>,
+        pub(super) e2: Vec<i64>,
+        pub(super) r: Vec<i64>,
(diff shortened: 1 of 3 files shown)


tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2357,165 +2357,64 @@ pub fn verify<G: Curve>(
...
         let fake_e1 = (0..d)
-            .map(|_| (rng.gen::<u64>() % (2 * B)) as i64 - B as i64)
-            .collect::<Vec<_>>();
-        let e2 = (0..k)
-            .map(|_| (rng.gen::<u64>() % (2 * B)) as i64 - B as i64)
+            .map(|_| (rng.gen::<u64>() % (2 * b_i)) as i64 - b_i as i64)
             .collect::<Vec<_>>();
         let fake_e2 = (0..k)
-            .map(|_| (rng.gen::<u64>() % (2 * B)) as i64 - B as i64)
+            .map(|_| (rng.gen::<u64>() % (2 * b_i)) as i64 - b_i as i64)
```
IceTDrinker, line 2393: are we sure about the B vs b_i ?

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1678,123 +1478,30 @@ mod tests {
     fn test_proof_compression() {
         let d = 2048;
         let k = 320;
-        let big_b = 1048576;
+        let b_i = 512;
```
IceTDrinker, line 1481: why the change ?

nsarlin-zama: I think it was a mistake in the previous version, since the other tests for pke v1 use `b_i = 512`

tfhe-zk-pok/src/proofs/pke_v2.rs
nsarlin-zama: in both proofs they play the same role (noise bound), so I think we should give them the same name in pke and pke_v2. Before the pr this is called B in pke and b_i in pke_v2.

IceTDrinker: I think it's linked to the notations in the technical notes, they are not strictly the same IIRC as we need to derive an upper bound of the std dev in v2

nsarlin-zama: I used B everywhere and used the value from the TUniform params

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -174,3 +174,202 @@ pub mod pke;
 pub mod pke_v2;
 pub mod range;
 pub mod rlwe;
+
+#[cfg(test)]
+mod test {
+    #![allow(non_snake_case)]
```
IceTDrinker, line 180: what for ?

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -174,3 +174,202 @@ pub mod pke;
...
+        pub(super) d: usize,
+        pub(super) k: usize,
+        pub(super) B: u64,
+        pub(super) q: u64,
+        pub(super) t: u64,
+        pub(super) msbs_zero_padding_bit_count: u64,
+    }
+
+    pub(super) const DEFAULT_TEST_PKE_PARAMS: PkeTestParameters = PkeTestParameters {
+        d: 2048,
```
IceTDrinker, line 246: not the correct d, you want the one from tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs:9

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -174,3 +174,202 @@ pub mod pke;
...
+        k: 320,
+        B: 131072, // 2^17, taken from TUniform 2^-64 parameters
+        q: 0,
+        t: 1024,
```
IceTDrinker, line 250: this t is not correct either if we want BC like params

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2758,119 +2579,28 @@ mod tests {
...
-        let effective_cleartext_t = t >> msbs_zero_padding_bit_count;
-
-        let delta = {
-            let q = if q == 0 { 1i128 << 64 } else { q as i128 };
-            (q / t as i128) as u64
-        };
+        let PkeTestParameters {
+            d,
+            k,
+            B,
```
IceTDrinker, line 2585: we don't recompute the upper bound of the std dev with the formula I gave you for TUniform stuff ?

tfhe-zk-pok/src/proofs/mod.rs
nsarlin-zama: because of the var named 'B'

tfhe-zk-pok/src/proofs/pke_v2.rs
nsarlin-zama: If I understand correctly the formula is to convert from a gaussian stddev, but here we already have a TUniform value

IceTDrinker: actually the fomula should give an equivalent std dev for the TUniform that can safely be used in places where a gaussian is expected, not sure it is the exact std dev for a TUniform distribution however, guess we could check that formula

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1457,56 +1369,30 @@ mod tests {
...
+            r,
+            m: _m,
+            b,
+            c1,
+            c2,
+            metadata,
+        } = PkeTestProofInputs::gen(rng, PKEV1_TEST_PARAMS);
 
         // Generate messages with padding set to fail verification
         let m = {
```
IceTDrinker, line 1398: this is not correct, here this m is equivalent to a "fake m" what we want for the padding bit test is :

generate a legit encryption but with m that has at least one slot with a padding bit set and with that proven data we will check that the verification fails, indicating we are able to properly detect when a padding bit is set

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2545,41 +2464,30 @@ mod tests {
...
+            r,
+            m: _m,
+            b,
+            c1,
+            c2,
+            metadata,
+        } = PkeTestProofInputs::gen(rng, PKEV2_TEST_PARAMS);
 
         // Generate messages with padding set to fail verification
         let m = {
```
IceTDrinker, line 2493: same thing for the padding bit test, this is a fake m test

IceTDrinker (changes requested):
Some tests are not equivalent and we need to be rock solid on the padding bit thing :) otherwise looks good !

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: You're right, good catch !

IceTDrinker (approved):

PR #1599, opened 2024-09-27
chore(gpu): add gpu compression in the hl api

agnesLeroy:

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/gpu/vec.rs
+++ b/tfhe/src/core_crypto/gpu/vec.rs
@@ -5,8 +5,9 @@ use std::collections::Bound::{Excluded, Included, Unbounded};
 use std::ffi::c_void;
 use std::marker::PhantomData;
 use tfhe_cuda_backend::cuda_bind::{
-    cuda_drop, cuda_malloc_async, cuda_memcpy_async_gpu_to_gpu, cuda_memcpy_async_to_cpu,
-    cuda_memcpy_async_to_gpu, cuda_memset_async,
+    cuda_drop, cuda_malloc, cuda_malloc_async, cuda_memcpy_async_gpu_to_gpu,
+    cuda_memcpy_async_to_cpu, cuda_memcpy_async_to_gpu, cuda_memcpy_gpu_to_gpu, cuda_memset_async,
+    cuda_synchronize_device,
 };
 
 /// A contiguous array type stored in the gpu memory.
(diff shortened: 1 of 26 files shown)


tmontaigu (approved):

agnesLeroy:
@nsarlin-zama you might want to check the versionning part of this PR. It's what we did earlier together, but well :sweat_smile:

tfhe/src/high_level_api/backward_compatibility/compressed_ciphertext_list.rs
```diff
@@ -6,12 +6,29 @@ use crate::{CompressedCiphertextList, Tag};
...
+}
+
+impl Upgrade<CompressedCiphertextList> for CompressedCiphertextListV1 {
     type Error = Infallible;
 
     fn upgrade(self) -> Result<CompressedCiphertextList, Self::Error> {
         Ok(CompressedCiphertextList {
-            inner: self.0,
+            inner: crate::high_level_api::compressed_ciphertext_list::InnerCompressedCiphertextList::Cpu(self.inner),
             tag: Tag::default(),
```
nsarlin-zama, line 32: if you come from a CompressedCiphertextListV1, I think you should reuse the tag from the previous version here:
```
   tag: self.tag
```

nsarlin-zama (commented):
a small thing to fix

tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -29,6 +30,33 @@ pub struct CudaVec<T: Numeric> {
     _phantom: PhantomData<T>,
 }
 
+impl<T: Numeric> Clone for CudaVec<T> {
+    fn clone(&self) -> Self {
+        let size = self.len as u64 * std::mem::size_of::<T>() as u64;
+        let mut cloned_vec = Vec::with_capacity(self.ptr.len());
+        for &index in self.gpu_indexes.iter() {
+            unsafe {
+                cuda_synchronize_device(index);
```
pdroalves, line 39: I agree synchronization is needed before copy, but not at this point. 

Since all the entire multi-gpu logic happens only at cuda side and everything is synchronized before returning to rust, we should synchronize `self.gpu_indexes[0]` only and before this loop.

pdroalves (approved):
Thanks for this @agnesLeroy . This PR takes GPU compression much closer to the CPU API and gives the rusty touch it didn't have.

I'm not able to review the HL API part, but I think you fixed most of the issues with the integer API.

tfhe/src/core_crypto/gpu/vec.rs
agnesLeroy: I can't pass any stream to clone, that's why I'm using the same logic than for Drop. I can't tell what happens on which GPU so it's safer to synchronize all of them.

pdroalves: I don't get it. The rust side takes the stream at `gpu_indexes[0]` as the main stream, and the CUDA-sided methods with multi-gpu logic will synchronize all devices before returning. We shouldn't have anything happening on other devices than the first.

agnesLeroy: We can't prevent someone from doing something on gpu_indexes[1] on the Rust side, so in this kind of method (Drop, Clone) we should act on all pointers in the CudaVec. I don't think this will impact performance significantly anyway, the Clone method is only there because we need to have an implementation of Clone for CudaVec in the HL API. We can always call duplicate on a ptr in the CudaVec as we usually do.

PR #1614, opened 2024-10-02
feat(cuda): modify double to torus cast and normalization for bsk

bbarbakadze:
### PR content/description

Normalization for bsk is changed to divide on 2^bits_in_torus instead of TORUS_MAX. cast from double to torus is modified to avoid casting into in128

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
@@ -1,9 +1,15 @@
 #ifndef CNCRT_TORUS_CUH
 #define CNCRT_TORUS_CUH
 
+#include "polynomial/parameters.cuh"
 #include "types/int128.cuh"
 #include "utils/kernel_dimensions.cuh"
 #include <limits>
+
+template <typename T>
+__host__ __device__ __forceinline__ constexpr double get_two_pow_torus_bits() {
+  return (sizeof(T) == 4) ? 4294967296.0 : 18446744073709551616.0;
+}
 
 template <typename T>
 __device__ inline void typecast_double_to_torus(double x, T &r) {
@@ -25,6 +31,16 @@ __device__ inline void typecast_double_to_torus<uint64_t>(double x,
   uint128 nnnn = make_uint128_from_float(x);
   uint64_t lll = nnnn.lo_;
   r = lll;
+}
+
+template <typename T>
+__device__ inline void typecast_double_round_to_torus(double x, T &r) {
+  constexpr double mx = get_two_pow_torus_bits<T>();
+  // floor must be used here because round has an issue with rounding .5,
+  // as it rounds away from zero.
+  double frac = x - floor(x);
+  frac *= mx;
+  typecast_double_to_torus(round(frac), r);
 }
 
 template <typename T>
--- a/backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
@@ -3,6 +3,7 @@
 
 #include "crypto/torus.cuh"
 #include "device.h"
+#include "parameters.cuh"
 
 // Return A if C == 0 and B if C == 1
 #define SEL(A, B, C) ((-(C) & ((A) ^ (B))) ^ (A))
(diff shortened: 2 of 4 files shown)


backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
```diff
@@ -1,6 +1,7 @@
 #ifndef CNCRT_TORUS_CUH
 #define CNCRT_TORUS_CUH
 
+#include "polynomial/parameters.cuh"
```
agnesLeroy, line 4: Do we really need this new import?

backends/tfhe-cuda-backend/cuda/src/types/complex/operations.cuh
```diff
@@ -49,16 +49,9 @@ __device__ inline double2 operator-(const double2 a, const double2 b) {
...
-  double yx = a.y * b.x;
-  double yy = a.y * b.y;
-
   double2 res;
-  // asm volatile("fma.rn.f64 %0, %1, %2, %3;": "=d"(res.x) : "d"(a.x),
-  // "d"(b.x), "d"(yy));
-  res.x = xx - yy;
-  res.y = xy + yx;
+  res.x = (a.y * -b.y) + (a.x * b.x);
+  res.y = (a.x * b.y) + (a.y * b.x);
```
agnesLeroy, line 54: Oh you made that change as well, thanks!

agnesLeroy (approved):
Thanks a lot for this @bbarbakadze! Looks good to me now :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
bbarbakadze: we need it for `get_two_pow_torus_bits` also `parameters.cuh` might not be the best place for this function so we can move it somewhere else.

agnesLeroy: Oh maybe having it in torus.cuh would make more sense?

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
```diff
@@ -27,6 +28,14 @@ __device__ inline void typecast_double_to_torus<uint64_t>(double x,
   r = lll;
 }
 
+template <typename T>
+__device__ inline void typecast_double_round_to_torus(double x, T &r) {
+  constexpr double mx = get_two_pow_torus_bits<T>();
+  double frac = x - floor(x);
```
IceTDrinker, line 34: this is a round not a floor in the CPU implementation

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -162,28 +163,17 @@ __device__ void round_to_closest_multiple_inplace(T *rotated_acc, int base_log,
...
-
-    frac = v2 - floor(v2);
-    frac *= mx;
-    carry = frac - floor(v2);
-    frac += (carry >= 0.5);
+    typecast_double_round_to_torus<Torus>(v1, V1);
 
     Torus V2 = 0;
-    typecast_double_to_torus<Torus>(frac, V2);
+    typecast_double_round_to_torus<Torus>(v2, V2);
```
IceTDrinker, line 176: may I suggest more readable variable names ?

IceTDrinker (commented):
small comments

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
agnesLeroy: Do you have something in mind? Like `cast_double_to_torus_with_rounding`? We use the word `Torus` in the code base to refer to the ciphertext modulus size (u32 or u64).

IceTDrinker: I meant the v1 and V1

agnesLeroy: Oh indeed, could be `double_v1` and `torus_v1` for example @bbarbakadze.

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
bbarbakadze: @IceTDrinker  is there a specific reason why round can work better than floor here? Are there any benefits?

bbarbakadze: also round for us does not work tests are failing

IceTDrinker: I think here it may not have that big of an impact as it would essentially flip the sign I believe, as frac should end up between -0.5 and 0.5, in some cases I recall rounding towards 0 or rounding towards infinity as having a stabilizing effect on float computations (but then the ranges are really not the same)

IceTDrinker: really ? interesting, which tests are failing ?

bbarbakadze: pbs tests

IceTDrinker: ok so it means there is something about correctness with the proper rounding mode, I'll check for rust as the rounding mode may be a bit peculiar for floats, please add a comment about the correctness of floor so we don't hesitate in the future, I'll add an issue for the CPU implem and check if floor would be preferable to round in the implem

bbarbakadze: weird actually it should not affected it floor on minus values should be same as round on positive ones maybe  
`.5` is not acting as expected

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
bbarbakadze: what about `double_real` `double_imag`  `torus_real` `torus_imag` v1 and v2 are still confusing

agnesLeroy: Yep looks much better!

agnesLeroy (approved):
:fire:

PR #1626, opened 2024-10-04
chore(hlapi): remove useless flags

mayeul-zama:


tfhe/src/high_level_api/integers/signed/base.rs
```diff
@@ -33,7 +33,6 @@ pub trait FheIntId: IntegerId {}
 ///
 /// [FheInt8]: crate::high_level_api::FheUint8
 /// [FheInt16]: crate::high_level_api::FheInt16
-#[cfg_attr(all(doc, not(doctest)), doc(cfg(feature = "integer")))]
```
IceTDrinker, line 36: no change in doc output ?

hopefully doc_auto_cfg takes care of that for us now

mayeul-zama: It still prints `Available on crate feature integer only.`

IceTDrinker (approved):
Thanks !

PR #1627, opened 2024-10-04
chore(all): enforce polynomial size to be a power of 2

mayeul-zama:


IceTDrinker (changes requested):
No we may have other cyclotomics coming, I think there are more urgent backlog issues

IceTDrinker:
Documenting the function to indicate it’s the floor log2 like the rust std lib is simpler here, I don’t think an assert would make sense, until we have other polynomial rings I don’t think this change is worthwhile

tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -137,6 +137,7 @@ pub struct PolynomialSize(pub usize);
 
 impl PolynomialSize {
     /// Return the associated [`PolynomialSizeLog`].
+    /// If the polynomial size is not a power of 2, returns the ceil of its log2
     pub fn log2(&self) -> PolynomialSizeLog {
         PolynomialSizeLog((self.0 as f64).log2().ceil() as usize)
     }
```
IceTDrinker, line 143: 🤔 let's update this function to return "just" the log2 and have a ceil_log2 that returns the ceiling log2

also ceil_ilog2 should be written similar to the UnsignedInteger trait:

```rust
    fn ceil_ilog2(self) -> u32 {
        // ilog2 returns the rounded down log2
        self.ilog2() + u32::from(!self.is_power_of_two())
    }
```

tfhe/src/core_crypto/commons/numeric/unsigned.rs:94 and following

mayeul-zama: Let's not have `ceil_ilog2` at all then

IceTDrinker (approved):

PR #1630, opened 2024-10-04
add assert on compression

mayeul-zama:


tfhe/src/shortint/list_compression/compression.rs
```diff
@@ -57,6 +57,8 @@ impl CompressionKey {
                 let mut list: Vec<_> = vec![];
 
                 for ct in ct_list {
+                    assert!(ct.noise_level() == NoiseLevel::NOMINAL);
+
                     assert!(ct.carry_is_empty());
```
IceTDrinker, line 62: could we have a nicer error message ? Also we may want to convert those asserts to errors at some point

tfhe/src/shortint/list_compression/compression.rs
```diff
@@ -57,6 +57,11 @@ impl CompressionKey {
                 let mut list: Vec<_> = vec![];
 
                 for ct in ct_list {
+                    assert!(
+                        ct.noise_level() == NoiseLevel::NOMINAL,
+                        "Ciphertexts must have a nominal (post PBS) noise to be compressed"
+                    );
+
                     assert!(ct.carry_is_empty());
```
IceTDrinker, line 65: here as well

mayeul-zama: Done

IceTDrinker (approved):

PR #1631, opened 2024-10-04
regroup optionals in proofs

mayeul-zama:


tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -240,9 +240,18 @@ pub struct Proof<G: Curve> {
...
-    pi_kzg: Option<G::G1>,
+    optional_fields: Option<ProofOptionalFields<G>>,
+}
+
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(bound(
+    deserialize = "G: Curve, G::G1: serde::Deserialize<'de>, G::G2: serde::Deserialize<'de>",
+    serialize = "G: Curve, G::G1: serde::Serialize, G::G2: serde::Serialize"
+))]
+struct ProofOptionalFields<G: Curve> {
```
IceTDrinker, line 251: ComputeLoadProofFields, field name compute_load_proof_fields, maybe with a comment to indicate that those are only populated in the case of a proof with ComputeLoad::Proof

would better reflect what this is, I would wait for the https://github.com/zama-ai/tfhe-rs/pull/1597 PR to be merged first

this will need some upgrade code to make compatbile

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -301,28 +329,29 @@ where
...
             pi: G::G1::uncompress(pi)?,
-            c_hat_t: c_hat_t.map(G::G2::uncompress).transpose()?,
-            c_h: c_h.map(G::G1::uncompress).transpose()?,
-            pi_kzg: pi_kzg.map(G::G1::uncompress).transpose()?,
-        })
-    }
-}
 
-impl<G: Curve> Proof<G> {
-    pub fn content_is_usable(&self) -> bool {
```
IceTDrinker, line 321: we probably need a check that the points are properly on the elliptic curve, cc @nsarlin-zama, this was a function checking "integrity" of the proof in the sense that the raw data is valid EC data

nsarlin-zama: I don't think this was done here but I agree this should be part of the conformance for the zk. It can be done by calling `if self.is_on_curve() && self.is_in_correct_subgroup_assuming_on_curve()` on the Affine representation

Since I think this was quite heavy maybe it should be an option in the conformance params

nsarlin-zama: But it might be worthy of a dedicated PR

IceTDrinker: agreed, I'm not sure I agree it should be an option for conformance however, as a wrong point during verification is likely a very bad idea

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -232,18 +261,24 @@ where
             c_hat,
             c_y,
             pi,
-            c_hat_t,
-            c_h,
-            pi_kzg,
+            compute_load_proof_fields: optional_fields,
```
mayeul-zama, line 264: `optional_fields` could be renamed `compute_load_proof_fields`
Applies to other places

tfhe-zk-pok/src/proofs/pke.rs
mayeul-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/809

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -186,14 +189,26 @@ impl<G: Curve> PublicParams<G> {
     deserialize = "G: Curve, G::G1: serde::Deserialize<'de>, G::G2: serde::Deserialize<'de>",
     serialize = "G: Curve, G::G1: serde::Serialize, G::G2: serde::Serialize"
 ))]
-#[versionize(PKEv1ProofVersions)]
+#[versionize(ProofVersions)]
```
mayeul-zama, line 192: Why rename?

tfhe-zk-pok/src/proofs/rlwe.rs
```diff
@@ -81,9 +81,14 @@ pub struct Proof<G: Curve> {
...
     c_y: G::G1,
     pi: G::G1,
-    c_hat_t: Option<G::G2>,
-    c_h: Option<G::G1>,
-    pi_kzg: Option<G::G1>,
+    optional_fields: Option<ProofOptionalFields<G>>,
+}
+
+#[derive(Clone, Debug)]
+struct ProofOptionalFields<G: Curve> {
```
mayeul-zama, line 88: Any better name for this struct and the associated field?

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: https://github.com/zama-ai/tfhe-rs-internal/issues/744

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: Before everything was in the `backward_compatibility/mod.rs`, so I I have to name the types differently to avoid conflicts. 

But for the old versions this lead to names such as `PKEv1ProofV0` and `PKEv2ProofV0`, which are a bit confusing. So I split the `backward_compatibility/mod.rs` into `backward_compatibility/pke.rs` and `backward_compatibility/pke_v2.rs` to mirror what is done in the main module.

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: done !

tfhe-zk-pok/src/proofs/rlwe.rs
nsarlin-zama: done !

mayeul-zama (commented):
I can't approve, but it looks good to me, thanks for the update
There's just a pcc issue

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -935,14 +971,13 @@ pub fn verify<G: Curve>(
...
+    let Proof {
         c_hat,
         c_y,
         pi,
-        c_hat_t,
-        c_h,
-        pi_kzg,
-    } = proof;
+        compute_load_proof_fields,
+    } = proof.clone();
```
IceTDrinker, line 979: need to clone ?

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -229,52 +230,69 @@ impl<G: Curve> PublicParams<G> {
...
+    pub(crate) C_R: G::G1,
+    pub(crate) C_hat_bin: G::G2,
+    pub(crate) C_y: G::G1,
+    pub(crate) C_h1: G::G1,
+    pub(crate) C_h2: G::G1,
+    pub(crate) C_hat_t: G::G2,
+    pub(crate) pi: G::G1,
+    pub(crate) pi_kzg: G::G1,
+
+    pub(crate) compute_load_proof_fields: Option<ComputeLoadProofFields<G>>,
```
IceTDrinker, line 247: pub(in crate::bacwkard_...::pke_v2) or something like that ?

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -229,52 +230,69 @@ impl<G: Curve> PublicParams<G> {
...
+    pub(crate) C_e: CompressedG1<G>,
+    pub(crate) C_r_tilde: CompressedG1<G>,
+    pub(crate) C_R: CompressedG1<G>,
+    pub(crate) C_hat_bin: CompressedG2<G>,
+    pub(crate) C_y: CompressedG1<G>,
+    pub(crate) C_h1: CompressedG1<G>,
+    pub(crate) C_h2: CompressedG1<G>,
+    pub(crate) C_hat_t: CompressedG2<G>,
+    pub(crate) pi: CompressedG1<G>,
+    pub(crate) pi_kzg: CompressedG1<G>,
```
IceTDrinker, line 279: same comment

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -1747,11 +1786,11 @@ pub fn verify<G: Curve>(
...
         C_h2,
         C_hat_t,
-        C_hat_h3,
-        C_hat_w,
         pi,
         pi_kzg,
-    } = proof;
+        compute_load_proof_fields,
+    } = proof.clone();
+
```
IceTDrinker, line 1793: needed ?

tfhe-zk-pok/src/proofs/rlwe.rs
```diff
@@ -615,14 +620,13 @@ pub fn verify<G: Curve>(
...
+    let Proof {
         c_hat,
         c_y,
         pi,
-        c_hat_t,
-        c_h,
-        pi_kzg,
-    } = proof;
+        compute_load_proof_fields,
+    } = proof.clone();
```
IceTDrinker, line 628: clone ?

IceTDrinker (commented):
Some details

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: Before this PR the code was silently cloned because all fields implement copy. The `ComputeLoadProof` does not implement copy. I need to "own" the values because a lot of the functions used later take their arguments by value.

I can derive Copy for CoputeLoadProof if you prefer.

PR #1632, opened 2024-10-04
fix(integer): fix max_noise_level to keep space for carry

mayeul-zama:


IceTDrinker (changes requested):
I don’t think this is safe to do as integer algorithms often fill the noise and don’t need to propagate carries all the time, to be checked with @tmontaigu

tmontaigu:
This would'break' `max_sum_size` as it needs to use the 'true' max degree (i.e the one without the space left for a carry addition)

IceTDrinker:
this need a rewrite

PR #1633, opened 2024-10-04
versionize: rework of the from/into/convert feature to allow conversion with generics

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/748

### PR content/description
The `Versionize` proc macro accepts some arguments that will affect the implementation of the traits. From example, the `#[versionize(SerializableMyTypeVersions, from = SerializableMyType, into = SerializableMyType)]` will apply a conversion method before the versioning, like serde does: https://serde.rs/container-attrs.html#from. However before this PR, this parameter did not work if the conversion type had generics, or if the conversion method needed specific bounds.

On a side note, the syntax was a bit updated. The `from`/`try_from`/`into` keywords were used to mimic serde, for user convenience. However in Versionize we actually don't support having a different value for from and into. This is now checked during the argument parsing. New shortcut arguments have also been added: `convert`/`try_convert`. They work the same way except that you only need one argument: `#[versionize(SerializableMyTypeVersions, convert = SerializableMyType)]`.

On the same topic, before this pr the attribute required the to state the name of the dispatch enum in that case: `#[versionize(SerializableMyTypeVersions, from = SerializableMyType, into = SerializableMyType)]`. This is now automatically detected when we use a conversion, which means that the attribute can be simplified as `#[versionize(convert = SerializableMyType)]`

### Check-list:

### DIFF
--- a/utils/tfhe-versionable/examples/manual_impl.rs
+++ b/utils/tfhe-versionable/examples/manual_impl.rs
@@ -2,11 +2,10 @@
 
 use std::convert::Infallible;
 
-use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use tfhe_versionable::{Unversionize, UnversionizeError, Upgrade, Versionize, VersionizeOwned};
 
-struct MyStruct<T: Default> {
+struct MyStruct<T> {
     attr: T,
     builtin: u32,
 }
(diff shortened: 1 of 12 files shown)


utils/tfhe-versionable/examples/manual_impl.rs
```diff
@@ -2,11 +2,10 @@
 
 use std::convert::Infallible;
 
-use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use tfhe_versionable::{Unversionize, UnversionizeError, Upgrade, Versionize, VersionizeOwned};
 
-struct MyStruct<T: Default> {
+struct MyStruct<T> {
```
tmontaigu, line 8: is default removed because is causes problem or it was not needed / its a style change ?

nsarlin-zama: It was not needed, but it was before. The change that allowed this was done in this pr: https://github.com/zama-ai/tfhe-rs/pull/1551.

In that case it doesn't change anything but sometimes you don't want to have a too restrictive bound, you want to be able to create MyStruct with any T but only be able to unversionize/upgrade if `T: Default`. 

A more realistic example is in tfhe-rs with container, we might require `C: ContainerMut` for some upgrade implementation (and so needed for `Unversionize`) but `C: Container` is sufficient for `Versionize`

PR #1639, opened 2024-10-07
chore(ci): add firefox support for wasm tests and benchmarks

soonum:


.github/workflows/benchmark_wasm_client.yml
```diff
@@ -106,12 +106,12 @@ jobs:
...
           make install_node
-          make install_chrome_browser
-          make install_chrome_web_driver
+          make install_firefox_browser
+          make install_firefox_web_driver
 
       - name: Run benchmarks
         run: |
-          make bench_web_js_api_parallel_chrome_ci
+          make bench_web_js_api_parallel_firefox_ci
```
IceTDrinker, line 114: could we have both benchmarks ? at least initially to be able to compare ?

IceTDrinker (commented):
just a small thing

.github/workflows/benchmark_wasm_client.yml
soonum: I've put a mechanism to do so.

IceTDrinker (approved):
Alright let's try this, do you have a backlog issue for updating Selenium and reducing the stdout spam ?

PR #1640, opened 2024-10-07
chore(ci): remove symlinking of the dynamic buffer API

IceTDrinker:


IceTDrinker:
some leftover of the previous way we had to build the C API, was forgotten for whatever reason

soonum (approved):
Less is more :+1:

PR #1647, opened 2024-10-07
feat(integer): improve shift/rotate by encrypted amount

tmontaigu:
This commit does a few things:
* Changes the BitExtractor to use many_lut to reduce number of PBS done
* Add blocks rotation/shift operation
* Implement a new algorithm for bit shift/rotation by encrypted amounts
* Add support bit shift/rotation for 1_1 parameters (as result of adding block shift/rotation)

The gist of the new bit shift/rotation is to use the same idea as the scalar version where we first shift blocks between adjacent blocks, then use a rotation of blocks.

Doing this requires to do a division and modulo operation:
```rust
let (shift_within_blocks, block_rotations) =
  (amount / bits_per_block, amount % bits_per_block)
```
When `amount` is clear this operation is simple, when `amount` is encrypted then is harder (`bits_per_block` is always clear). However, when bits_per_block is a power of 2 (e.g 1, 2, 4) `/` and `%` can be made by shifting and bit-masking, which are simple operations.

This means the new algorithm is only compatible with 1_1, 2_2, 4_4 but not 3_3.
The new algorithm improves the latency as well as the throughput as it requires less PBS in total

### Check-list:

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/bit_extractor.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/bit_extractor.rs
@@ -1,18 +1,45 @@
 use crate::integer::ServerKey;
-use crate::shortint::server_key::LookupTableOwned;
+use crate::shortint::server_key::ManyLookupTableOwned;
 use crate::shortint::Ciphertext;
-use itertools::iproduct;
 use rayon::prelude::*;
+use std::collections::VecDeque;
 
+/// Extracts bits from a slice of shortint ciphertext
+///
+/// * Relies on many-lut to do less PBSes
+/// * Position of the extracted bit is customizable
+/// * Has a buffer system that allows to pre-generate extracted bits such that the next call to
+///   `next()` is free, giving better control as to where computations happen
 pub(crate) struct BitExtractor<'a> {
-    bit_extract_luts: Vec<LookupTableOwned>,
+    input_blocks: std::slice::Iter<'a, Ciphertext>,
+    bit_extract_luts: Option<ManyLookupTableOwned>,
     bits_per_block: usize,
     server_key: &'a ServerKey,
+    buffer: VecDeque<Ciphertext>,
+}
+
+impl<'a> Iterator for BitExtractor<'a> {
+    type Item = Ciphertext;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let maybe_bit_block = self.buffer.pop_front();
+        if maybe_bit_block.is_some() {
+            return maybe_bit_block;
+        }
+
+        self.prepare_next_batch();
+
+        self.buffer.pop_front()
+    }
 }
 
 impl<'a> BitExtractor<'a> {
-    pub(crate) fn new(server_key: &'a ServerKey, bits_per_block: usize) -> Self {
-        Self::with_final_offset(server_key, bits_per_block, 0)
+    pub(crate) fn new(
+        input: &'a [Ciphertext],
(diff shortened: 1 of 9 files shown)


tfhe/src/integer/server_key/radix_parallel/block_shift.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+impl ServerKey {
+    /// Implementation of the barrel shift to shift/rotate blocks
+    ///
+    /// ct: The ciphertext for which blocks will be shifted/rotated
+    /// shift_bits_extractor: Must be configured such that shift bits blocks returned
+    ///     are in the first carry bit
+    /// d_range: Range of bits to process
+    ///     Note that starting from something else that 0
```
tmontaigu, line 46: that -> than

tfhe/src/integer/server_key/radix_parallel/shift.rs
```diff
@@ -296,6 +296,374 @@ impl ServerKey {
...
+
+        let radix = T::from_blocks(messages);
+
+        let num_bit_that_tells_shift_within_blocks = message_bits_per_block.ilog2();
+        let num_bits_already_done = num_bit_that_tells_shift_within_blocks + 1;
+        self.block_barrel_shifter_impl(
+            &radix,
+            &mut shift_bit_extractor,
+            // We already did the first block rotation so we start at 1
+            // And do + 1 and the range is exclusive
```
tmontaigu, line 661:
```suggestion
            // And do + 1 as the range is exclusive
```

tfhe/src/integer/server_key/radix_parallel/bit_extractor.rs
```diff
@@ -23,46 +50,125 @@ impl<'a> BitExtractor<'a> {
...
+    pub(crate) fn extract_all_bits(&mut self) -> Vec<Ciphertext> {
+        let num_blocks = self.input_blocks.len();
+        self.extract_n_bits(num_blocks * self.bits_per_block)
+    }
 
+    /// Extract the remaining `n` bits in parallel from the current source blocks
+    ///
+    /// # Panics
+    ///
+    /// Panics if the current slice of blocks does not have less than n bits available
```
nsarlin-zama, line 162: `/// Panics if the current slice of blocks have less than n bits available`
or
`/// Panics if the current slice of blocks does not have n bits available`

tfhe/src/integer/server_key/radix_parallel/block_shift.rs
```diff
@@ -0,0 +1,410 @@
...
+        if d_range.is_empty() {
+            return ct.clone();
+        }
+
+        assert!(
+            self.key
+                .max_noise_level
+                .validate(NoiseLevel::NOMINAL * 3usize)
+                .is_ok(),
+            "Parameters must support 3 additions before a PBS"
```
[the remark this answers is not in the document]
tmontaigu: No its checks that the addition of 3 inputs with nominal noise is possible: Nominal + Nominal + Nominal = 3 * Nominal

Which is something all parameters supported have (1_1, 2_2, 4_4), so there is no need to do something else if parameters don't allow that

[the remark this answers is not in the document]
tmontaigu: Ok Yes I understand, its 2 addition but 3 ciphertext 🗿

nsarlin-zama (approved):
Good for me !

PR #1648, opened 2024-10-08
chore(ci): add dry run mode to tfhe-versionable releases

nsarlin-zama:


.github/workflows/make_release_tfhe_versionable.yml
```diff
@@ -19,9 +24,10 @@ jobs:
       - name: Publish crate.io package
         env:
           CRATES_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+          DRY_RUN: ${{ inputs.dry_run && '--dry-run' || '' }}
         run: |
-          cargo publish -p tfhe-versionable-derive --token ${{ env.CRATES_TOKEN }}
-          cargo publish -p tfhe-versionable --token ${{ env.CRATES_TOKEN }}
+          cargo publish -p tfhe-versionable-derive --token ${{ env.CRATES_TOKEN }} ${{ env.DRY_RUN }}
+          cargo publish -p tfhe-versionable --token ${{ env.CRATES_TOKEN }} ${{ env.DRY_RUN }}
```
IceTDrinker, line 30: That would not work as the derive crate would not be on crates right ?

IceTDrinker (commented):
Don't think that works

.github/workflows/make_release_tfhe_versionable.yml
IceTDrinker: it's either splitting that workflow or no dry run

IceTDrinker: at least I don't see another way for now

IceTDrinker (approved):

PR #1650, opened 2024-10-08
feat(integer): construct proven ct list conformance from another source

IceTDrinker:
this was already added on 0.8.x as it was needed quickly


nsarlin-zama:
I approved but maybe for main we should rework the CompactPkeCrs wrapper before ?

IceTDrinker:
after to avoid having divergence between 0.8 and main

PR #1651, opened 2024-10-08
fix(gpu): fix sample extract many lut use case

guillermo-oyarzun:

### Check-list:


pdroalves (approved):
Not it seems correct.
Do we have tests that cover this when `glwe_dimension > 1`?

guillermo-oyarzun:
> Not it seems correct. Do we have tests that cover this when `glwe_dimension > 1`?

I guess you meant "Now", we don't have a test for that case (neither for the apply lut end point), there is only a test in the doc tests, i guess if i use another parameter set there it could test the glwe_dimension > 1 use case too

guillermo-oyarzun:
> Sorry, it was "Now" indeed.
> 
> It would be good if we had at least one test with `glwe_dimension > 1`.

I added 3_3 params in the doc test, that case works with glwe_dimension=2

PR #1653, opened 2024-10-09
chore(ci): remove wasm test that is not relevant

soonum:


IceTDrinker:
I'm ok with this for the general idea, I need to check with cleaning task of removing the old PublicKey from the code base from shortint and higher

IceTDrinker (changes requested):
Waiting on whether the legacy public key gets the boot

IceTDrinker:
looks like the legacy public key has not been removed so I don't think we should kill this test

soonum:
Not needed for the v1 of the library

PR #1654, opened 2024-10-09
chore(ci): fix browser identification on wasm benchmarks

soonum:
Rust script wasn't able to parse results if the browser name was suffixed after parameters name.

### DIFF
--- a/ci/webdriver.py
+++ b/ci/webdriver.py
@@ -362,7 +362,8 @@ def dump_benchmark_results(results, browser_kind):
     """
     if results:
         results = {
-            "_".join((key, browser_kind.name)): val for key, val in results.items()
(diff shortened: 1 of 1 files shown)


ci/webdriver.py
```diff
@@ -362,7 +362,8 @@ def dump_benchmark_results(results, browser_kind):
     """
     if results:
         results = {
-            "_".join((key, browser_kind.name)): val for key, val in results.items()
+            key.replace("mean", "_".join((browser_kind.name, "mean"))): val
+            for key, val in results.items()
```
IceTDrinker, line 366: checked locally ?

soonum: yes

PR #1678, opened 2024-10-14
fix(ci): fix secret on CI checkout

mayeul-zama:


IceTDrinker (commented):
check the least access rule here when possible

soonum:
> check the least access rule here when possible

I have an idea regarding that.
1. replacing all usage of `FHE_ACTIONS_TOKEN` for downloading Slab repo to `SLAB_ACTIONS_TOKEN`
2. reducing permissions to read-only whenever possible for `FHE_ACTIONS_TOKEN`
3. creating a new token for the releases since it needs a write permission in the repo

The second step will need careful analysis to avoid side-effect that could pop-up later.

IceTDrinker:
Ok, should this be merged and then what you propose done later ?

IceTDrinker:
Ok, should this be merged and then what you propose done later ?

soonum (approved):

PR #1679, opened 2024-10-14
chore(ci): enforce commit signature for PRs

IceTDrinker:
- uses 1Password actions


IceTDrinker:
as you can see I tried with a non signed commit @soonum and I got a reject from the action

soonum:
That's awesome! Thanks for pushing this !

PR #1683, opened 2024-10-15
chore(ci): fix firefox wasm benchmarks with new aws ami

soonum:
Some libs were missing to be able to run Firefox out of the box. Besides, action runner is now installed as ubuntu user since Firefox is not able to run as root.


IceTDrinker (approved):
We may have a failure with the wasm tests because of the docker run with the pkg being owned by root, so this looks good for the benchmarks but we may need to fix the CPU WASM tests

soonum:
> We may have a failure with the wasm tests because of the docker run with the pkg being owned by root, so this looks good for the benchmarks but we may need to fix the CPU WASM tests

Issues are fixed now :tada:

IceTDrinker (approved):
🔥

PR #1684, opened 2024-10-15
chore: fix some overflow related panics

tmontaigu:
Some overflow panics were occurring when
overflow-check=true

Most of them were expected/accepted, so this commit only makes changes so that its now explicit that overflow is accepted.

Related to https://github.com/zama-ai/tfhe-rs-internal/issues/85

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/add.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/add.rs
@@ -1110,7 +1110,7 @@ impl ServerKey {
                 } else {
                     // u64::MAX is -1 in two's complement
                     // We apply the modulus including the padding bit
-                    u64::MAX % (1 << (block_modulus + 1))
+                    u64::MAX % (block_modulus * 2)
                 }
             })]
         };
--- a/tfhe/src/integer/server_key/radix_parallel/comparison.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/comparison.rs
@@ -544,7 +544,7 @@ impl ServerKey {
                 } else {
                     // u64::MAX is -1 in tow's complement
                     // We apply the modulus including the padding bit
-                    u64::MAX % (1 << (block_modulus + 1))
+                    u64::MAX % (block_modulus * 2)
                 }
             })]
         };
--- a/tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
@@ -881,8 +881,8 @@ impl ServerKey {
         let block_states = {
             for i in 1..grouping_size {
                 let state_fn = |block| {
-                    let r = (u64::MAX * u64::from(block != 0)) % (packed_modulus * 2);
-                    r << (i - 1)
+                    let r = u64::MAX * u64::from(block != 0);
+                    (r << (i - 1)) % (packed_modulus * 2)
                 };
                 first_grouping_luts.push(self.key.generate_lookup_table(state_fn));
             }
@@ -890,8 +890,8 @@ impl ServerKey {
             let other_block_state_luts = (0..grouping_size)
                 .map(|i| {
                     let state_fn = |block| {
-                        let r = (u64::MAX * u64::from(block != 0)) % (packed_modulus * 2);
-                        r << i
+                        let r = u64::MAX * u64::from(block != 0);
+                        (r << i) % (packed_modulus * 2)
                     };
                     self.key.generate_lookup_table(state_fn)
                 })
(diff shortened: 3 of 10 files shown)


tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
```diff
@@ -881,17 +881,17 @@ impl ServerKey {
         let block_states = {
             for i in 1..grouping_size {
                 let state_fn = |block| {
-                    let r = (u64::MAX * u64::from(block != 0)) % (packed_modulus * 2);
-                    r << (i - 1)
+                    let r = u64::MAX * u64::from(block != 0);
+                    (r << (i - 1)) % (packed_modulus * 2)
```
IceTDrinker, line 885: this is not stricticly equivalent here is it ?

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
```diff
@@ -881,17 +881,17 @@ impl ServerKey {
...
                 first_grouping_luts.push(self.key.generate_lookup_table(state_fn));
             }
 
             let other_block_state_luts = (0..grouping_size)
                 .map(|i| {
                     let state_fn = |block| {
-                        let r = (u64::MAX * u64::from(block != 0)) % (packed_modulus * 2);
-                        r << i
+                        let r = u64::MAX * u64::from(block != 0);
+                        (r << i) % (packed_modulus * 2)
```
IceTDrinker, line 894: same as above

tfhe/src/integer/server_key/radix_parallel/comparison.rs
```diff
@@ -544,7 +544,7 @@ impl ServerKey {
                 } else {
                     // u64::MAX is -1 in tow's complement
                     // We apply the modulus including the padding bit
-                    u64::MAX % (1 << (block_modulus + 1))
+                    u64::MAX % (block_modulus * 2)
```
IceTDrinker, line 547: this can't be equivalent either right ?

tfhe/src/integer/server_key/radix_parallel/add.rs
```diff
@@ -1110,7 +1110,7 @@ impl ServerKey {
                 } else {
                     // u64::MAX is -1 in two's complement
                     // We apply the modulus including the padding bit
-                    u64::MAX % (1 << (block_modulus + 1))
+                    u64::MAX % (block_modulus * 2)
```
IceTDrinker, line 1113: does not look equivalent

IceTDrinker (commented):
Some changes do not look correct to me from afar, but maybe we had bugs we were not aware of

tfhe/src/integer/server_key/radix_parallel/comparison.rs
tmontaigu: They are not equivalent, but the form `(1 << (block_modulus + 1))` was incorrect

tfhe/src/integer/server_key/radix_parallel/add.rs
tmontaigu: They are not equivalent, but the form (1 << (block_modulus + 1)) was incorrect

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
tmontaigu: The new version is the more correct one, as the modulus needs to be applied after

IceTDrinker: same question is there a way to detect this with a test ?

IceTDrinker: and pinging the GPU

tfhe/src/integer/server_key/radix_parallel/comparison.rs
tmontaigu: These are a bit hard to detect because the modulus would implicitly be done when generating the lut in `f(i) * delta`

I think this particular case created a problem on GPU but only for 3_3, using hillis steele

IceTDrinker: ok worth pinging the GPU then

tmontaigu: Guillermo encoutered it today :D

tfhe/src/integer/server_key/radix_parallel/scalar_comparison.rs
IceTDrinker: ok

IceTDrinker (approved):

PR #1691, opened 2024-10-16
Ns/versionable unsupported

nsarlin-zama:

### PR content/description
Adds a trait and a type to mark deprecation of older versions of a type that are no longer supported.

Namely, there is a trait `Deprecable` that is used on a user facing type to tell that older versions of this type can be deprecated. This trait mainly defines some str associated constants that will be used to provide a friendlier error message to the user. In the example below, a user that tries to deserialize a deprecated version, would get the following error:
> Deprecated MyStruct found in serialized data, minimal supported version is my_app v2

(In the context of tfhe-rs, this could be something like `Deprecated ServerKey found in serialized data, minimal supported version is tfhe-rs v0.8`)

The type `Deprecated<MyStruct>` should be used in the dispatch enum and replaces the deprecated versions. As it replaces them, it allows to remove all the code associated to defining, loading and upgrading them.

Since the remaining valid versions will still keep their place in the dispatch enum, any data that has not been explicitly deprecated will still be compatible.

Example:
```rust
use tfhe_versionable::deprecation::{Deprecable, Deprecated};
use tfhe_versionable::{Versionize, VersionsDispatch};
#[derive(Versionize)]
#[versionize(MyStructVersions)]
struct MyStruct;

impl Deprecable for MyStruct {
    const TYPE_NAME: &'static str = "MyStruct";
    const MIN_SUPPORTED_APP_VERSION: &'static str = "my_app v2";
}

#[derive(VersionsDispatch)]
#[allow(unused)]
pub enum MyStructVersions {
    V0(Deprecated<MyStruct>),
    V1(Deprecated<MyStruct>),
    V2(MyStruct),
}
```
We could also define a macro in tfhe-rs for the Deprecable trait implementation. (Something like `deprecable!(ServerKey)`)

### Check-list:

### DIFF
--- a/utils/tfhe-versionable/src/deprecation.rs
+++ b/utils/tfhe-versionable/src/deprecation.rs
@@ -1,0 +1,140 @@
+//! Handle the deprecation of older versions of some types
+
+use std::error::Error;
+use std::fmt::Display;
+use std::marker::PhantomData;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{UnversionizeError, Upgrade, Version};
+
+/// This trait should be implemented for types that have deprecated versions. You can then use them
+/// inside the dispatch enum by wrapping them into the [`Deprecated`] type.
+pub trait Deprecable {
+    const TYPE_NAME: &'static str;
+    const MIN_SUPPORTED_APP_VERSION: &'static str;
+
+    fn error() -> DeprecatedVersionError {
+        DeprecatedVersionError {
+            type_name: Self::TYPE_NAME.to_string(),
+            min_supported_app_version: Self::MIN_SUPPORTED_APP_VERSION.to_string(),
+        }
+    }
+}
+
+/// An error returned when trying to interact (unserialize or unversionize) with a deprecated type.
+#[derive(Debug)]
+pub struct DeprecatedVersionError {
+    type_name: String,
+    min_supported_app_version: String,
+}
+
+impl Display for DeprecatedVersionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Deprecated {} found in serialized data, minimal supported version is {}",
+            self.type_name, self.min_supported_app_version
+        )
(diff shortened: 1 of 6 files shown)


utils/tfhe-versionable/src/deprecation.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use std::error::Error;
+use std::fmt::Display;
+use std::marker::PhantomData;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{UnversionizeError, Upgrade, Version};
+
+/// This trait should be implemented for types that have deprecated versions. You can them use them
```
IceTDrinker, line 11: typo them -> then

utils/tfhe-versionable/src/deprecation.rs
```diff
@@ -0,0 +1,139 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Err(<D::Error as serde::de::Error>::custom(T::error()))
+    }
+}
+
+impl<T: Deprecable> Version for Deprecated<T> {
+    type Ref<'vers>
+        = DeprecatedVersion<T>
```
IceTDrinker, line 102: isn't Ref type supposed to have a ref somehow ?

IceTDrinker (commented):
Looks good small things I saw

utils/tfhe-versionable/src/deprecation.rs
nsarlin-zama: No it "can" have a reference to avoid clones if possible, as opposed to the owned that "must" own its data. We already store an owned value in some cases, for example when we do a from/into conversion before the versioning.

Here since the type is a ZST it does not really make sense to make a reference on it in my opinion, but I can change it to `&'vers DeprecatedVersion<T>` if you think that it's less confusing

IceTDrinker: just a comment to indicate why this is fine

IceTDrinker (approved):

PR #1694, opened 2024-10-17
chore(ci): make build script regenerate bindings at all times

IceTDrinker:
- git diff shows no diff

@guillermo-oyarzun try this and check if it manages to update your bindings.rs


backends/tfhe-cuda-backend/build.rs
```diff
@@ -60,31 +60,21 @@ fn main() {
...
+            .header(header_path)
+            .clang_arg("-x")
+            .clang_arg("c++")
+            .clang_arg("-std=c++17")
+            .clang_arg("-I/usr/include")
+            .clang_arg("-I/usr/local/include")
+            .ctypes_prefix("ffi")
+            .raw_line("use crate::ffi;")
+            .generate()
+            .expect("Unable to generate bindings");
```
guillermo-oyarzun, line 73: This will generate the bindings every compilation? wouldn't that affect compiling times?

IceTDrinker: it only runs if the build script is run, if it needs to run it either means you are building the cuda code (slow) or you are regenerating the binding and I don't think it's that slow, to be checked, it's a proposal to avoid missing binding regen

IceTDrinker: however for whatever reason the git from the pcc checks is failing, but I don't get the issue locally

IceTDrinker: https://github.com/zama-ai/tfhe-rs/actions/runs/11387881080/job/31683686190?pr=1694#step:9:683

IceTDrinker: ah looks like it does not output the same rust file format

IceTDrinker: it's a step before the check that modifies the file, very weird

guillermo-oyarzun: if I run locally make pcc_gpu, I get : Generated bindings have changed! Please run 'git add backends/tfhe-cuda-backend/src/bindings.rs' and commit the changes.

IceTDrinker: what's the diff ?

guillermo-oyarzun: not sure what are all those for

IceTDrinker: unfurtonately for you it's a change linked to the headers on your system and the export of macros by the binding generator

IceTDrinker:
@guillermo-oyarzun try this iteration, should be much more stable

guillermo-oyarzun (approved):

PR #1701, opened 2024-10-21
Ns/backport/fix serialized size

nsarlin-zama:


tfhe/src/safe_serialization.rs
```diff
@@ -508,7 +608,7 @@ mod test_shortint {
 
         let mut buffer = vec![];
 
-        let config = SerializationConfig::new(1 << 20).disable_versioning();
+        let config = SerializationConfig::new_with_unlimited_size().disable_versioning();
```
mayeul-zama, line 611: Why have unlimited size here?

nsarlin-zama: the goal of the test is just to check the deserialization when I give exactly the size of the buffer as a limit (which was failing before), so I don't really care about the serialization size limit. 

Since it does not really matter I could have left it with the original size anyways, but since it is a backport I think we should stick with what has been merged to main.

mayeul-zama (approved):
Thanks!

PR #1702, opened 2024-10-21
chore(bench): add packing keyswtich benchmarks

soonum:


tfhe/benches/core_crypto/ks_bench.rs
```diff
@@ -175,6 +175,101 @@ fn keyswitch<Scalar: UnsignedTorus + CastInto<usize> + Serialize>(
...
+    let mut secret_generator =
+        SecretRandomGenerator::<ActivatedRandomGenerator>::new(seeder.seed());
+
+    for (name, params) in parameters.iter() {
+        let lwe_dimension = params.lwe_dimension.unwrap();
+        let glwe_dimension = params.glwe_dimension.unwrap();
+        let polynomial_size = params.polynomial_size.unwrap();
+        let ks_decomp_base_log = params.ks_base_log.unwrap();
+        let ks_decomp_level_count = params.ks_level.unwrap();
+        let ciphertext_modulus = params.ciphertext_modulus.unwrap();
```
IceTDrinker, line 208: this is not correct, you could use the compression parameters from shortint to use here as they have packing keyswitch parameters

IceTDrinker (changes requested):
packing KS are not the same as for the simple KS parameters as they have different noise constraints, you could use the compression parameters for TUniform for the packing

tfhe/benches/core_crypto/ks_bench.rs
soonum: Alright, but then the doctest example for the packing function is wrong no ? Since it other noise constraints (gaussian instead of t-uniform).

IceTDrinker: we'll discuss that, not sure what this is about

tfhe/benches/core_crypto/ks_bench.rs
```diff
@@ -402,15 +514,52 @@ mod cuda {
 
 #[cfg(feature = "gpu")]
 use cuda::cuda_keyswitch_group;
+use tfhe::shortint::parameters::{
+    COMP_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64, PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64,
+};
```
IceTDrinker, line 519: why is the import here ?

tfhe/benches/core_crypto/ks_bench.rs
```diff
@@ -402,15 +514,52 @@ mod cuda {
...
+        "par_packing_keyswitch",
+        &benchmark_compression_parameters(),
+        par_keyswitch_lwe_ciphertext_list_and_pack_in_glwe_ciphertext,
+    );
+}
+
 #[cfg(not(feature = "gpu"))]
-criterion_main!(keyswitch_group);
+criterion_main!(packing_keyswitch_group); // DEBUG
+                                          // criterion_main!(keyswitch_group, packing_keyswitch_group);
```
IceTDrinker, line 563: to change

IceTDrinker (changes requested):
leftover debug to remove and a weird location for an import, otherwise should be good

tfhe/benches/core_crypto/ks_bench.rs
```diff
@@ -402,15 +514,52 @@ mod cuda {
...
+        "par_packing_keyswitch",
+        &benchmark_compression_parameters(),
+        par_keyswitch_lwe_ciphertext_list_and_pack_in_glwe_ciphertext,
+    );
+    packing_keyswitch(
+        &mut criterion,
+        "par_packing_keyswitch",
+        &benchmark_compression_parameters(),
+        par_keyswitch_lwe_ciphertext_list_and_pack_in_glwe_ciphertext,
+    );
```
IceTDrinker, line 558: wbhy are the functions written twice here ?

soonum: Leftovers from `u64` and `u32` versions.

IceTDrinker:
looks like the change for the parameters conversion is breaking a bench, maybe the previous conversion should be kept and add the new one on top

IceTDrinker:
see failed cargo build

IceTDrinker (approved):
Thanks a lot for the hard work !

PR #1704, opened 2024-10-22
Am/chore/data breaking change

IceTDrinker:
the backward compatibility for shortint server key needs to be checked the code that was in 0.8 is not exactly the one from main


nsarlin-zama (approved):

IceTDrinker:
will wait for tests to pass here, as the release 0.9 branch has been updated with latest 0.8 github sees conflicts (though it should be able to manage them), then I'll do the rebase

PR #1706, opened 2024-10-22
feat(wasm): export safe_deserialize for CompactPkePublicParams

nsarlin-zama:


IceTDrinker (approved):
Looks good, a note I don't seem to see an "enable versioning" function, not really critical but if people want to be explicit about the config it is currently missing, could be the same for conformance

nsarlin-zama:
> Looks good, a note I don't seem to see an "enable versioning" function, not really critical but if people want to be explicit about the config it is currently missing, could be the same for conformance

Yes you're right, I will make an issue

IceTDrinker (approved):

PR #1710, opened 2024-10-22
chore(backward): use the Deprecated type from tfhe-versionable

nsarlin-zama:
### PR content/description
Use the `Deprecated` type from `tfhe-versionable` for all the deprecated versions. This removes a bit of code in the backward_compat folders and will improve the error messages because most errors should happen at deserialization, before the call to the "upgrade" method.

Sadly the `Deprecable` trait for the types that have deprecated versions is a bit hard to implement with a declarative macro, because the macro needs to handle types with generics (especially with "nested" generics like `<Scalar: UnsignedInteger, C: Container<Element = Scalar>>`). I think it would require a proc-macro, but I don't know if its worth the price for a few impl (And I think it would be the same thing with impl of `Named`)

### DIFF
--- a/tfhe/src/shortint/backward_compatibility/server_key/mod.rs
+++ b/tfhe/src/shortint/backward_compatibility/server_key/mod.rs
@@ -1,4 +1,5 @@
-use tfhe_versionable::{Upgrade, Version, VersionsDispatch};
+use tfhe_versionable::deprecation::{Deprecable, Deprecated};
+use tfhe_versionable::VersionsDispatch;
 
 use crate::core_crypto::prelude::Container;
 use crate::shortint::server_key::*;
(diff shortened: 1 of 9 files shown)


IceTDrinker:
I see failed cargo builds

nsarlin-zama:
> I see failed cargo builds

fixed !

tfhe/src/shortint/backward_compatibility/server_key/mod.rs
```diff
@@ -18,23 +19,13 @@ pub enum ShortintCompressedBootstrappingKeyVersions {
...
-    fn upgrade(self) -> Result<CompressedServerKey, Self::Error> {
-        Err(crate::Error::new(
-            "Unable to load CompressedServerKey, \
-            this format is unsupported by this TFHE-rs version."
-                .to_string(),
-        ))
-    }
+impl Deprecable for CompressedServerKey {
+    const TYPE_NAME: &'static str = "CompressedServerKey";
+    const MIN_SUPPORTED_APP_VERSION: &'static str = "TFHE-rs v0.8";
```
IceTDrinker, line 24: 0.9 I believe as we break things in 0.9 and some things cannot be loaded from 0.8 ?

IceTDrinker (approved):

PR #1711, opened 2024-10-22
fix(core): fix decomposition over 1 level to be balanced

IceTDrinker:
It was brought to our attention that a decomposition over 1 level was not properly balanced and could have a mean value of 0.5 which can be an issue when being fed in an FFT (generating harmonics in the process) this fixes that.

A tests has been added to check on a selected 1 level decomposition that the average of the values returning during decomposition is ~0

cc @agnesLeroy and maybe @guillermo-oyarzun so that you can check how to update this on your end, I can explain the formula in more detail, but normally the comment in the code explains how it works

### DIFF
--- a/tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs
+++ b/tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs
@@ -46,6 +46,25 @@ pub fn native_closest_representable<Scalar: UnsignedInteger>(
     res << shift
 }
 
+/// With
+///
+/// B = 2^bit_count
+/// val < B
+/// random € [0, 1]
+///
+/// returns 1 if the following if condition is true otherwise 0
+///
+/// (val > B / 2) || ((val == B / 2) && (random == 1))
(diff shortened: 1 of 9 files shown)


IceTDrinker:
I launched the tests in advance and as you can see all green, so the fix is behaving properly

IceTDrinker:
tested my bit trick for input values, the bit trick agrees with the plain if, will check the one in the decomposition as well,

IceTDrinker:
Ok will push a test validating the bit trick, then once this hits the PR GPU team can look at updating their code

IceTDrinker:
updated the code, the bit tricks are now in a function marked inline(always) and have a dedicated test to verify their output is correct good to review, I'll launch the tests directly with approved

IceTDrinker:
something going wrong in the pbs 128, need to check what it is

IceTDrinker:
latest pbs 128 failure (mismatch between two implems we have) is now fixed

mayeul-zama (approved):
Thanks!

PR #1714, opened 2024-10-23
Mz/move strings

mayeul-zama:


tmontaigu:
are we sure that we want to move strings into the hlapi module ?

Makefile
```diff
@@ -788,11 +788,11 @@ test_user_doc_gpu: install_rs_build_toolchain
...
-.PHONY: test_fhe_strings # Run tests for fhe_strings example
-test_fhe_strings: install_rs_build_toolchain
-	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
-		--example fhe_strings \
-		--features=$(TARGET_ARCH_FEATURE),integer
+# .PHONY: test_fhe_strings # Run tests for fhe_strings example
+# test_fhe_strings: install_rs_build_toolchain
+# 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile $(CARGO_PROFILE) \
+# 		--example fhe_strings \
+# 		--features=$(TARGET_ARCH_FEATURE),integer
```
tmontaigu, line 795: needs to be removed or add a target to run string tests and make sure it runs in the CI

tmontaigu:
Hum, thing is that I would provably move the, is the create_trivial_ascii from the ClientKey to the ServerKey as its what exists for integers, imo makes a bit more sense as the user could need the ability to create trivial data, and might prevent user from mistaking trivial as a real encryption

mayeul-zama:
> Hum, thing is that I would provably move the, is the create_trivial_ascii from the ClientKey to the ServerKey as its what exists for integers, imo makes a bit more sense as the user could need the ability to create trivial data, and might prevent user from mistaking trivial as a real encryption

I agree it makes more sense to have it on the ServerKey.
But in this case, when switching from encryption to trivial encryption in tests, we must pass a `ServerKey` instead of a `ClientKey` which may need changes inmany places.
Maybe we could have this method on the `ServerKey`, and on the `ClientKey` behind `#[cfg(test)]` (applies to all APIs levels)

What do you think?

tmontaigu:
> > Hum, thing is that I would provably move the, is the create_trivial_ascii from the ClientKey to the ServerKey as its what exists for integers, imo makes a bit more sense as the user could need the ability to create trivial data, and might prevent user from mistaking trivial as a real encryption
> 
> 
> 
> I agree it makes more sense to have it on the ServerKey.
> 
> But in this case, when switching from encryption to trivial encryption in tests, we must pass a `ServerKey` instead of a `ClientKey` which may need changes inmany places.
> 
> Maybe we could have this method on the `ServerKey`, and on the `ClientKey` behind `#[cfg(test)]` (applies to all APIs levels)
> 
> 
> 
> What do you think?

That's ok i think

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -242,6 +244,81 @@ impl ClientKey {
...
+    pub fn unchecked_create_trivial(&self, value: u64) -> Ciphertext {
+        let lwe_size = match self
+            .parameters
+            .pbs_parameters()
+            .unwrap()
+            .encryption_key_choice()
+        {
+            EncryptionKeyChoice::Big => self.large_lwe_secret_key().lwe_dimension().to_lwe_size(),
+            EncryptionKeyChoice::Small => self.small_lwe_secret_key().lwe_dimension().to_lwe_size(),
+        };
```
tmontaigu, line 281: parameters should already have a `encryption_lwe_dimension` function

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -242,6 +244,81 @@ impl ClientKey {
...
+        &self,
+        value: u64,
+        lwe_size: LweSize,
+    ) -> Ciphertext {
+        let params = self.parameters.pbs_parameters().unwrap();
+
+        let pbs_order = match params.encryption_key_choice() {
+            EncryptionKeyChoice::Big => PBSOrder::KeyswitchBootstrap,
+            EncryptionKeyChoice::Small => PBSOrder::BootstrapKeyswitch,
+        };
```
tmontaigu, line 296: PBSOrder::from(params.encryption_key_choice)

tfhe/src/shortint/client_key/mod.rs
mayeul-zama: Done

tfhe/src/shortint/client_key/mod.rs
mayeul-zama: Done

tfhe/src/strings/char_iter.rs
```diff
@@ -0,0 +1,211 @@
...
+                index: 0,
+                back_index: self.len(),
+            }
+        }
+    }
+
+    pub struct OptionalEndSliceIterator<'a, T> {
+        iter: OptionalEndSliceIter<'a, T>,
+        index: usize,
+        back_index: usize,
```
tmontaigu, line 50: Would it be easier to resuse std::slice::Iter<'a> make it easier ?

```rust


struct OptionalEndSliceIter<'a, T> {
    slice_iter: std::slice::Iter<'a, T>,
    last: Option<&'a T>,
}

impl<'a, T> OptionalEndSliceIter<'a, T> {
    pub fn len(&self) -> usize {
        self.slice_iter.as_slice().len() + usize::from(self.last.is_some())
    }

    pub fn new(slice: &'a [T], last: Option<&'a T>) -> Self {
        Self { slice_iter: slice.iter(), last }
    }
}

impl<'a, T> Iterator for OptionalEndSliceIter<'a, T> {
    type Item = &'a T;
    
    fn next(&mut self) -> Option<Self::Item> {
        let thing = self.slice_iter.next();
        if thing.is_some() {
            return thing;
        }
        
        self.last.take()
    }
}


impl<'a, T> DoubleEndedIterator for OptionalEndSliceIter<'a, T> {
    fn next_back(&mut self) -> Option<Self::Item> {
         let thing = self.slice_iter.next_back();
        if thing.is_some() {
            return thing;
        }
        
        self.last.take()
    }
}
```

mayeul-zama: Sure, thanks for the idea

tmontaigu (commented):
Looks good, needs to fix CI

tmontaigu (approved):

PR #1720, opened 2024-10-28
chore(ci): update chrome version

IceTDrinker:
Fix benchmarks by updating chrome, feels like something like chrome auto update is fucking up the binary while the benchmark runs resulting in the driver losing connection to it, updating chrome seemingly fixes the issue as can be seen in this run https://github.com/zama-ai/tfhe-rs/actions/runs/11533886276

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,8 +163,8 @@ install_web_resource:
 	rm checksum && \
 	$(decompress_cmd) $(filename)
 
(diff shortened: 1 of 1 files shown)


soonum (approved):
Thanks. So this means we need to update Chrome as soon as a new version is released ?

PR #1722, opened 2024-10-28
chore(ci): add schedule event to core_crypto benchmarks

soonum:


IceTDrinker:
do we want to do that ? it was to lower the bench prices at some point

PR #1723, opened 2024-10-28
chore(ci): verify commit on release

soonum:
Enforce commit being associated to a tag.
The tag must be committed by a member of the release team. In addition, the tag needs to be verified. Finally, triggering actor must also be a member of the release team.

### DIFF
--- a/.github/workflows/make_release.yml
+++ b/.github/workflows/make_release.yml
@@ -30,8 +30,15 @@ env:
   NPM_TAG: ""
 
 jobs:
+  verify_tag:
+    uses: ./.github/workflows/verify_tagged_commit.yml
+    secrets:
+      RELEASE_TEAM: ${{ secrets.RELEASE_TEAM }}
+      READ_ORG_TOKEN: ${{ secrets.READ_ORG_TOKEN }}
+
   package:
     runs-on: ubuntu-latest
+    needs: verify_tag
     outputs:
       hash: ${{ steps.hash.outputs.hash }}
     steps:
--- a/.github/workflows/verify_tagged_commit.yml
+++ b/.github/workflows/verify_tagged_commit.yml
@@ -1,0 +1,32 @@
+# Verify a tagged commit
+name: Verify tagged commit
+
+on:
+  workflow_call:
+    secrets:
+      RELEASE_TEAM:
+        required: true
+      READ_ORG_TOKEN:
+        required: true
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/')
+    steps:
+      # Check triggering actor membership
+      - name: Actor verification
+        id: actor_check
+        uses: morfien101/actions-authorized-user@4a3cfbf0bcb3cafe4a71710a278920c5d94bb38b
+        with:
+          username: ${{ github.triggering_actor }}
+          org: ${{ github.repository_owner }}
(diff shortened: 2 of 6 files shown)


.github/workflows/make_release.yml
```diff
@@ -30,8 +30,59 @@ env:
...
+        run: |
+          if [ ${{ steps.author_check.outputs.authorized }} = false ]; then
+            echo "Author '${{ env.COMMITTER_LOGIN }}' is not part of authorized team"
+            exit 1
+          fi
+          
+          if [ ${{ steps.actor_check.outputs.authorized }} = false ]; then
+            echo "Actor '${{ env.COMMITTER_LOGIN }}' is not authorized to perform release"
+            exit 1
+          fi
```
IceTDrinker, line 76: wrong actor reported here

.github/workflows/make_release.yml
```diff
@@ -30,8 +30,59 @@ env:
...
+          org: ${{ github.repository_owner }}
+          team: ${{ secrets.RELEASE_TEAM }}
+          github_token: ${{ secrets.READ_ORG_TOKEN }}
+
+      - name: Commit verification
+        run: |
+          if [ ${{ steps.author_check.outputs.authorized }} = false ]; then
+            echo "Author '${{ env.COMMITTER_LOGIN }}' is not part of authorized team"
+            exit 1
+          fi
```
IceTDrinker, line 71: quote everything here, false is a built in in bash, the output on the left is a string, also the check should be I believe == ?

IceTDrinker (changes requested):
I think the shell scripting has issues, a mistake in a log

soonum:
Should we extend these checks to the rest of release workflows ?
If so I can do it in a another PR but first I'd like to factorize this step into its own workflow and use `workflow_call` event to use it in all release workflows.

IceTDrinker:
> Should we extend these checks to the rest of release workflows ? If so I can do it in a another PR but first I'd like to factorize this step into its own workflow and use `workflow_call` event to use it in all release workflows.

yes we should

IceTDrinker:
could be done via a callable workflow, we have example of those somewhere, I can show you

.github/workflows/verify_tagged_commit.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/')
+    steps:
+      - name: Get commit details
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          {
+            echo "COMMITTER_LOGIN=$(gh api repos/${{ github.repository }}/commits/${{ github.sha }} --jq '.committer.login')";
+            echo "VERIFICATION_STATUS=$(gh api repos/${{ github.repository }}/commits/${{ github.sha }} --jq '.commit.verification.verified')";
```
IceTDrinker, line 23: question here does the sha refer to the tag itself ?

soonum: Yes , but in the end I need to modify the URL to get the actual tag.
It needs some fixing, thanks for noticing.

soonum:
Regarding discussion on Slack to the impossibility to get the SHA and the committer login for the tag, we've decided to only check the triggering actor to ensure a certain amount of safety to make a release.

IceTDrinker (approved):

PR #1729, opened 2024-10-30
fix(zk): proof compatiblity between 32/64b platforms

nsarlin-zama:


IceTDrinker (approved):
Looks good, todo: non regression test, sync with David if needed, proposed workflow:

- generate a CRS and a CPK in rust native x86
- load that in a js environment (node or browser, but node is likely easier)
- generate a proof
- load in rust and verify

PR #1740, opened 2024-10-31
feat(capi): add mechanism to get panic message as const char *

tmontaigu:
Previously, when an error occurred in the rust side, the panic message
would get printed to stderr, then the c function would return 1 to
indicate error.
    
This commit adds the ability to disable the automatic prints of panic
messages and adds functions to get the panic message as a const char *
to allow user better control on how to display error messages.


tfhe/src/c_api/error.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// This should be used to directly print/display or copy the error.
+#[no_mangle]
+pub unsafe extern "C" fn tfhe_error_get_last() -> *const c_char {
+    LAST_LOCAL_ERROR.with_borrow(|last_error| last_error.as_ptr())
+}
+
+/// Returns the length of the current error message stored
+///
+/// The length **DOES NOT INCLUDE** the nul byte
```
IceTDrinker, line 98: shouldn't the C API get the nul byte in the len ?

IceTDrinker (commented):
Looks generally ok, should we have a primitive to clear the last error ?

tfhe/src/c_api/error.rs
tmontaigu: Hum, It's mostly a matter of convention and having it be clear in the docs
Here I chose it to be like `strlen` which returns the len of a nul-terminated string but the returned len does not account for the nul-byte

tmontaigu:
> Looks generally ok, should we have a primitive to clear the last error ?

Doesn't hurt to have one

tfhe/src/c_api/error.rs
IceTDrinker: if it matches C API primitives perfect

IceTDrinker (approved):

PR #1741, opened 2024-10-31
chore(bench): add throughput for integer benchmarks

soonum:


tfhe/benches/integer/bench.rs
```diff
@@ -4,15 +4,22 @@ mod oprf;
...
 use rand::prelude::*;
+use rayon::prelude::*;
 use std::env;
+use std::sync::OnceLock;
 use tfhe::integer::keycache::KEY_CACHE;
 use tfhe::integer::prelude::*;
 use tfhe::integer::{IntegerKeyKind, RadixCiphertext, RadixClientKey, ServerKey, U256};
 use tfhe::keycache::NamedParam;
 
+static BENCH_TYPE: OnceLock<BenchmarkType> = OnceLock::new();
```
IceTDrinker, line 21: Not a fan at all of that :|

tfhe/benches/utilities.rs
```diff
@@ -383,6 +383,29 @@ impl EnvConfig {
...
+}
+
+pub enum BenchmarkType {
+    Latency,
+    Throughput,
+}
+impl BenchmarkType {
+    pub fn from_env() -> Result<Self, String> {
+        let raw_value = env::var("__TFHE_RS_BENCH_TYPE")
+            .map_err(|_| "no benchmark type provided in environment".to_string())?;
```
IceTDrinker, line 400: let's default to latency ?

tfhe/benches/utilities.rs
```diff
@@ -383,6 +383,29 @@ impl EnvConfig {
     }
 }
 
+/// Generate a number of threads to use to saturate current machine for throughput measurements.
+pub fn throughput_num_threads() -> u64 {
+    let num_threads = rayon::current_num_threads() as f64;
+    // Add 20% more to maximum threads available.
+    (num_threads + (num_threads * 0.2)) as u64
+}
```
IceTDrinker, line 391: unclear this is enough, check with @tmontaigu probably

tfhe/benches/integer/bench.rs
```diff
@@ -111,28 +118,59 @@ fn bench_server_key_binary_function_clean_inputs<F>(
...
+                        },
+                        criterion::BatchSize::SmallInput,
+                    )
+                });
+            }
+            BenchmarkType::Throughput => {
+                bench_id = format!("{bench_name}::throughput::{param_name}::{bit_size}_bits");
+                let elements = throughput_num_threads();
+                bench_group.throughput(Throughput::Elements(elements));
+                bench_group.bench_function(&bench_id, |b| {
```
IceTDrinker, line 154: unclear to me if this is the right way to call the throughput, did you confirm it works as expected ?

tfhe/benches/utilities.rs
```diff
@@ -383,6 +383,29 @@ impl EnvConfig {
...
+impl BenchmarkType {
+    pub fn from_env() -> Result<Self, String> {
+        let raw_value = env::var("__TFHE_RS_BENCH_TYPE")
+            .map_err(|_| "no benchmark type provided in environment".to_string())?;
+        match raw_value.to_lowercase().as_str() {
+            "latency" => Ok(BenchmarkType::Latency),
+            "throughput" => Ok(BenchmarkType::Throughput),
+            _ => Err(format!("benchmark type '{raw_value}' is not supported")),
+        }
```
IceTDrinker, line 405: no way to do both ?

tfhe/benches/utilities.rs
tmontaigu: Yes its unclear to know if its enough, for some ops it might for others not, or it could be enough for all,

I think a manual check needs to be done by running and trying different values and see if the throughput drastically changes

agnesLeroy:
We'll need also to add throughput benches for the GPU :slightly_smiling_face:

soonum:
> We'll need also to add throughput benches for the GPU 🙂

Yes for sure, I just wanted to check for the design before deploying to all the functions.

tfhe/benches/utilities.rs
soonum: Agreed, it's less violent this way.

tfhe/benches/integer/bench.rs
soonum: Yes it works as expected, at least on my laptop. We'll need to double check on target machine to see if results are coherent.

tfhe/benches/utilities.rs
soonum: Nope, unless we want to somewhat duplicate calls to `write_to_json()` currently located at the end of the benchmark function.

soonum: Or it would implies a third `match` arm, with a duplicate of the current two `BenchmarkType` variants.

soonum: After second thought, it's not relevant to be able to do both latency and throughput.
Moreover it would duplicate a lot of code, hence increasing source of error for maintainers.

IceTDrinker:
@soonum I see commits not having the right format it seems and a failing build, should I still review ?

soonum:
> @soonum I see commits not having the right format it seems and a failing build, should I still review ?

Yes, please review now. I'll squash all the commits after this first review.

tfhe/benches/utilities.rs
IceTDrinker: did we do any manual checks ?

tfhe/benches/integer/bench.rs
```diff
@@ -465,6 +551,66 @@ fn if_then_else_parallelized(c: &mut Criterion) {
             )
         });
 
+        let bench_id;
+
+        match BENCH_TYPE.get().unwrap() {
+            BenchmarkType::Latency => {
```
IceTDrinker, line 557: this looks duplicated in this function, see above I see a bench loop already ?

tfhe/benches/integer/glwe_packing_compression.rs
```diff
@@ -41,23 +45,85 @@ fn cpu_glwe_packing(c: &mut Criterion) {
...
+                        _ = black_box(unpacked);
+                    })
+                });
+            }
+            BenchmarkType::Throughput => {
+                let elements = throughput_num_threads();
+                // FIXME thread usage seemed to be somewhat more "efficient".
+                //  For example, with bit_size = 2, my laptop is only using around 2/3 of the
+                // available threads  Thread usage increases with bit_size = 8 but
+                // still isn't fully loaded.
```
IceTDrinker, line 85: what's that ?

tfhe/benches/integer/glwe_packing_compression.rs
```diff
@@ -41,23 +45,85 @@ fn cpu_glwe_packing(c: &mut Criterion) {
...
+
+                let builders = (0..elements)
+                    .map(|_| {
+                        let ct = cks.encrypt_radix(0_u32, num_blocks);
+                        let mut builder = CompressedCiphertextListBuilder::new();
+                        builder.push(ct);
+
+                        builder
+                    })
+                    .collect::<Vec<_>>();
```
IceTDrinker, line 96: this won't fill a compressed cipehrtext list, we have 512 bits of capacity per builder before we are creating several lists

also begs the question are we interested in one builder with a lot of ciphertexts or several builders (like it's currently written) ?

tfhe/benches/integer/glwe_packing_compression.rs
```diff
@@ -127,44 +181,137 @@ mod cuda {
...
-            let cuda_decompression_key = compressed_decompression_key.decompress_to_cuda(
-                radix_cks.parameters().glwe_dimension(),
-                radix_cks.parameters().polynomial_size(),
-                radix_cks.parameters().message_modulus(),
-                radix_cks.parameters().carry_modulus(),
-                radix_cks.parameters().ciphertext_modulus(),
-                &stream,
-            );
+            let bench_id_1;
+            let bench_id_2;
```
IceTDrinker, line 185: what are those compared to pack/unpack ?

tfhe/benches/integer/zk_pke.rs
```diff
@@ -176,82 +215,142 @@ fn pke_zk_verify(c: &mut Criterion, results_file: &Path) {
...
+                                    )
+                                    .unwrap();
+                            });
+                        });
+                    }
+                    BenchmarkType::Throughput => {
+                        // In throughput mode object sizes are not recorded.
+                        let elements = throughput_num_threads();
+                        bench_group.throughput(Throughput::Elements(elements)); // FIXME CPU is not fully loaded in the current state (around 75% on my
+                        // laptop)
```
IceTDrinker, line 307: did you ever manage to get this to saturate the CPU ?

.github/workflows/benchmark_gpu_core_crypto.yml
```diff
@@ -126,8 +125,7 @@ jobs:
           --commit-date "${{ env.COMMIT_DATE }}" \
           --bench-date "${{ env.BENCH_DATE }}" \
           --name-suffix avx512 \
-          --walk-subdirs \
-          --throughput
+          --walk-subdirs
```
IceTDrinker, line 129: I don't see the bench-type flag meaning we never measure/parse throughput ?

same everywhere else

.github/workflows/benchmark_integer.yml
```diff
@@ -8,6 +8,10 @@ on:
         description: "Run all precisions"
         type: boolean
         default: false
+      run_throughput:
+        description: "Run throughput benchmarks"
+        type: boolean
+        default: false
```
IceTDrinker, line 14: as discussed we likely need a way to choose latency and/or run_throughput to have data that's up to date

IceTDrinker (changes requested):
Some things to update

ci/benchmark_parser.py
```diff
@@ -153,12 +168,15 @@ def recursive_parse(
                     filter(None, [test_name, stat_name, name_suffix])
                 )
 
+                if stat_name == "mean" and bench_type == BenchType.throughput:
+                    value = (elements * ONE_SECOND_IN_NANOSECONDS) / value
```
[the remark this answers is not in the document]
soonum: It's operations per second.

.github/workflows/benchmark_integer.yml
soonum: I've implemented a way to do so in `benchmark_signed_integer.yml`. Tell me what you think of it before I spread the logic to the others files.

IceTDrinker: will do

.github/workflows/benchmark_signed_integer.yml
```diff
@@ -53,6 +62,27 @@ jobs:
         run: |
           echo "op_flavor=${{ toJSON(env.OP_FLAVOR) }}" >> "${GITHUB_OUTPUT}"
 
+      - name: Default benchmark type
+        if: github.event_name != 'workflow_dispatch'
```
IceTDrinker, line 66: this means we don't get throughput on scheduled benchmarks right ?

is there a way in Grafana to not fetch benchmarks that don't match ? i.e. if I have old throughput data I don't want them to show next to the latest latency data, this way benchmarks are always consistent and we can tell when data is missing ?

.github/workflows/benchmark_signed_integer.yml
```diff
@@ -53,6 +62,27 @@ jobs:
...
+
+      - name: Set one benchmark type
+        if: inputs.bench_type != 'both'
+        run: |
+          echo "BENCH_TYPE=[\"${{ inputs.bench_type }}\"]" >> "${GITHUB_ENV}"
+
+      - name: Both benchmarks types
+        if: inputs.bench_type == 'both'
+        run: |
+          echo "BENCH_TYPE=[\"latency\", \"throughput\"]" >> "${GITHUB_ENV}"
```
IceTDrinker, line 78: can be a single small shell script

if [[ "${{ inputs.bench_type }}" == "both" ]]; then
    todo
else
    todo
fi

IceTDrinker (commented):
Looks generally ok for the pattern you want to use

tfhe/benches/integer/zk_pke.rs
soonum: Yes, after investigation, it's just the first bits size iteration (`64usize`) which never saturate the CPU.
Following elements in the iterator saturate the CPU just fine.

IceTDrinker: what do you mean by first bit size iteration ?

tfhe/benches/integer/glwe_packing_compression.rs
soonum: Do we care about filling them ? We just want to perform the same kind of operation in parallel. We don't want to fill-up the list but fully loading the CPU first.

IceTDrinker: actually that's fine as the throughput will be tested for different levels of fullness of the list

ci/benchmark_parser.py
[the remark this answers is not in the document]
IceTDrinker: the value after the / sign ?

.github/workflows/benchmark_signed_integer.yml
soonum: Regarding the data display, it's rather easy to show one or the other using SQL query since `::throughput::` is contained in the bench ID for any throughput benchmark.

About the scheduling, I can make latency + throughput the default measurements. Is that what we want ?
To be cost effective, we could do it only on the quarterly scheduled benchmarks.

IceTDrinker: thing is when we look at both data, or when the blockchain will query the DB, how do we make sure the data they get is consistent ?

soonum: We can make a top-down menu selector to switch between latency and throughput.

ci/benchmark_parser.py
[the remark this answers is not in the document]
soonum: It's the value in nanoseconds recorded by criterion.

ci/benchmark_parser.py
```diff
@@ -135,7 +145,12 @@ def recursive_parse(
             elif subdir.name != "new":
                 continue
 
-            full_name, test_name = parse_benchmark_file(subdir)
+            full_name, test_name, elements = parse_benchmark_file(subdir)
+
+            if bench_type == BenchType.throughput and elements is None:
+                # Current subdir contains only latency measurements
+                continue
```
IceTDrinker, line 152: could print a warning

ci/benchmark_parser.py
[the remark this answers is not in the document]
IceTDrinker: and elements ?

it's unclear by looking at the formula if this si correct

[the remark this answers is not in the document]
IceTDrinker: we are rescaling the nanosecond measurement to seconds ?

.github/workflows/benchmark_signed_integer.yml
```diff
@@ -48,11 +56,42 @@ jobs:
...
 
+      #      - name: Set one benchmark type
+      #        if: inputs.bench_type != 'both'
+      #        run: |
+      #          echo "BENCH_TYPE=[\"${{ inputs.bench_type }}\"]" >> "${GITHUB_ENV}"
+      #
+      #      - name: Both benchmarks types
+      #        if: inputs.bench_type == 'both'
+      #        run: |
+      #          echo "BENCH_TYPE=[\"latency\", \"throughput\"]" >> "${GITHUB_ENV}"
```
IceTDrinker, line 87: this is commented out, expected ?

.github/workflows/benchmark_signed_integer.yml
soonum: This is implemented in CPU workflows.

.github/workflows/benchmark_signed_integer.yml
soonum: Needed to be removed.

IceTDrinker (commented):
I see a conflict, I think you were ising the preprod slab at some point, please fix the conflict and remove the preprod slab if there was any usage left

IceTDrinker (approved):

PR #1743, opened 2024-10-31
Ns/zk curve conformance

nsarlin-zama:
Adds conformance for zk proof and crs.

Proof conformance checks that the points are on the curve and in the correct subgroup.
Crs conformance also check that the crs is valid for the parameters.


tfhe/src/zk.rs
```diff
@@ -16,6 +18,14 @@ impl Named for CompactPkeProof {
...
 }
 
+impl ParameterSetConformant for CompactPkeProof {
+    type ParameterSet = ();
+
+    fn is_conformant(&self, _parameter_set: &Self::ParameterSet) -> bool {
+        self.is_usable()
+    }
+}
+
```
mayeul-zama, line 28: Shouldn't `is_usable` also be checked on `proof` in the loop in `impl ParameterSetConformant for ProvenCompactCiphertextList`

tfhe-zk-pok/src/curve_api.rs
```diff
@@ -116,6 +116,10 @@ pub trait CurveGroupOps<Zp>:
     fn to_le_bytes(self) -> impl AsRef<[u8]>;
     fn double(self) -> Self;
     fn normalize(self) -> Self::Affine;
+    fn validate_projective(&self) -> bool {
+        Self::validate_affine(&self.normalize())
+    }
+    fn validate_affine(affine: &Self::Affine) -> bool;
```
mayeul-zama, line 122: In my mind it would be clearer to have a `trait ValidateAffine { fn validate_affine(&self) -> bool;}` and the contraint `Affine: ValidateAffine` here instead (names could be changed)
Having a trait method which does not use `self` is a bit confusing

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -197,6 +206,42 @@ pub struct Proof<G: Curve> {
...
+        {
+            if !c_hat_t.validate_projective()
+                || !c_h.validate_projective()
+                || !pi_kzg.validate_projective()
+            {
+                return false;
+            }
+        }
+
+        true
```
mayeul-zama, line 241: Could be written
```

c_hat.validate_projective()
&& c_y.validate_projective()
&& pi.validate_projective()
&& compute_load_proof_fields.as_ref().map_or(true,
     |&ComputeLoadProofFields { c_hat_t, c_h, pi_kzg }|
           c_hat_t.validate_projective() && c_h.validate_projective() && pi_kzg.validate_projective())
```

Comment also applies to v2 proof

tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -359,4 +373,48 @@ mod test {
             PkeTestCiphertext { c1, c2 }
         }
     }
+
+    /// Return a point with coordinates (x, y) that is randomly chosen and not on the curve
+    pub(super) fn point_not_on_curve<Config: short_weierstrass::SWCurveConfig>(
```
mayeul-zama, line 378: Could be in a `#[cfg(test)]` I guess
Applies to other functions

tfhe/src/zk.rs
```diff
@@ -24,6 +34,65 @@ impl Named for CompactPkePublicParams {
...
+            noise_bound,
+            ciphertext_modulus,
+            plaintext_modulus,
+            // CRS created from shortint params have 1 MSB 0bit
+            msbs_zero_padding_bit_count: ZkMSBZeroPaddingBitCount(1),
+        })
+    }
+}
+
+impl ParameterSetConformant for CompactPkePublicParams {
```
mayeul-zama, line 82: Not sure to understand the need to check conformance on public parameters

It looks to me that if we can trust a `CompactPkePublicParamsConformanceParams`, we can also trust a ` CompactPkePublicParams`

tfhe/src/zk.rs
```diff
@@ -24,6 +34,65 @@ impl Named for CompactPkePublicParams {
...
+
+    fn is_conformant(&self, parameter_set: &Self::ParameterSet) -> bool {
+        self.d == parameter_set.lwe_dim.0
+            && self.k == parameter_set.max_num_message
+            && self.b == parameter_set.noise_bound
+            && self.q == parameter_set.ciphertext_modulus
+            && self.t == parameter_set.plaintext_modulus
+            && self.msbs_zero_padding_bit_count == parameter_set.msbs_zero_padding_bit_count.0
+            && self.is_usable()
+    }
```
IceTDrinker, line 93: we would also have max_num_message <= d and k <= max_num_message those more closely reflect the constraints on the parametrs

tfhe-zk-pok/src/proofs/mod.rs
nsarlin-zama: it's already the case (it's inside `mod test`)

tfhe/src/zk.rs
nsarlin-zama: The `CompactPkePublicParamsConformanceParam` can be derived from the public key encryption parameters. 

Concretely if you receive a key and a crs, this checks that they are compatible.

tfhe-zk-pok/src/curve_api.rs
mayeul-zama: Makes sense

IceTDrinker (changes requested):
Needs some tests please :)

IceTDrinker (approved):

nsarlin-zama:
@IceTDrinker if you could just validate the workflow/Makefile part of the last commit (I rebased so looking at the commit will be easier)

IceTDrinker (approved):
If fast zk tests pass this si good for me

PR #1747, opened 2024-11-04
Am/refactor/use natural decomposition order

IceTDrinker:


IceTDrinker:
for @agnesLeroy focus on the GPU stuff

agnesLeroy:
Draft PR here, let's see if the tests pass, if they do we can port the changes to this branch: https://github.com/zama-ai/tfhe-rs/pull/1749. Would be good to run benchmarks as well when the tests have passed to check it does not introduce a regression  on performance.

agnesLeroy:
GPU tests are passing on the draft PR and benchmark results are not affected, we can go with it imo.

IceTDrinker:
ok @agnesLeroy should I include the patch in my PR ?

agnesLeroy:
Yes :+1:

tfhe/src/high_level_api/backward_compatibility/keys.rs
```diff
@@ -28,66 +27,29 @@ impl Upgrade<ClientKey> for ClientKeyV0 {
...
-
-    fn upgrade(self) -> Result<ServerKey, Self::Error> {
-        Ok(ServerKey {
-            key: self.key,
-            tag: Tag::default(),
-        })
-    }
+impl Deprecable for ServerKey {
+    const TYPE_NAME: &'static str = "ServerKey";
+    const MIN_SUPPORTED_APP_VERSION: &'static str = "TFHE-rs v0.10";
```
nsarlin-zama, line 32: Is this really needed ? Since the different key types are each independently versioned, I'm not sure if we need to deprecate the higher level ones.
The deprecation will be caught during the deserialization of the inner type.

nsarlin-zama: The same comment applies to all types that are not directly composed of a level container.

IceTDrinker: I thought about it but the error message would be less clear in that case if we don't do a mass deprecation, let me know what you think we should do, worse case I can revert some of the changes, but as they are done now does not matter much normally and I'd rather avoid unnecessary changes _now_ to get the thing ready for the blockchain

nsarlin-zama: Yes makes sense, I think we can keep it like this

PR #1751, opened 2024-11-04
fix(gpu): general fixes and improvements to PBS

pdroalves:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/743, https://github.com/zama-ai/tfhe-rs-internal/issues/811

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/pbs/pbs_multibit_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/pbs/pbs_multibit_utilities.h
@@ -106,7 +106,7 @@ template <typename Torus> struct pbs_buffer<Torus, PBS_TYPE::MULTI_BIT> {
   uint32_t lwe_chunk_size;
   double2 *keybundle_fft;
   Torus *global_accumulator;
-  double2 *global_accumulator_fft;
+  double2 *global_join_buffer;
 
   PBS_VARIANT pbs_variant;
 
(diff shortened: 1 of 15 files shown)


agnesLeroy (commented):
Thanks a lot for this @pdroalves! It would be good to make the same changes in this PR for the tbc variant as well I think, to keep things consistent between cg & tbc.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_multibit_utilities.h
```diff
@@ -225,10 +225,12 @@ template <typename Torus> struct pbs_buffer<Torus, PBS_TYPE::MULTI_BIT> {
           num_blocks_keybundle * (polynomial_size / 2) * sizeof(double2),
           stream, gpu_index);
       global_accumulator = (Torus *)cuda_malloc_async(
```
agnesLeroy, line 227: I think this global_accumulator needs to be kept as is for the "default" multibit PBS variant. Only for TBC and CG it should not be sized with levels, am I wrong?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -113,7 +113,7 @@ mul_ggsw_glwe(Torus *accumulator, double2 *fft, double2 *join_buffer,
```
agnesLeroy, line 113: In this file, this section:
```
  // accumulate rest of the products into fft buffer
  for (int l = 1; l < gridDim.x; l++) {
    auto cur_src_acc = get_join_buffer_element<G>(l, blockIdx.y, group,
                                                  join_buffer, polynomial_size,
                                                  glwe_dimension, support_dsm);
    tid = threadIdx.x;
    for (int i = 0; i < params::opt / 2; i++) {
      fft[tid] += cur_src_acc[tid];
      tid += params::degree / params::opt;
```
could be done on blocks whose blockIDx.x = 0 only as well no?

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_multibit_utilities.h
pdroalves: Ah yes, I just renamed it. It is used as the "join buffer" in `mul_ggsw_glwe`, so I'm calling it by the right name.

agnesLeroy (approved):
Thanks a lot for this @pdroalves! I believe we still have an issue with the cmux test though no?

pdroalves:
Performance holds for this branch.

PR #1750, opened 2024-11-04
chore(zk): add a proof compat test between x86_64 and wasm

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/810

### PR content/description
This PR adds a test that:
- generates a crs and public key in rust
- load them in js and generate a proven ct list
- verify the proof in rust

This is done to validate proof compatibility between wasm (32b) and x86_64

PR #1755, opened 2024-11-05
fix(zk): fix zk wasm x86_64 tests

nsarlin-zama:
Fixes from https://github.com/zama-ai/tfhe-rs/pull/1750


IceTDrinker (approved):
Thanks a lot ! and sorry for the late review

PR #1760, opened 2024-11-07
Ns/rename pp crs

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/800#issue-2603108396

### PR content/description
Use the CompactPkeCrs instead of the CompactPkePublicParams. It simplifies a bit the workflow for users, they generate a crs and are directly able to use it and serialize it without having to extract the params.

Having a release 0.11 in Cargo.toml was necessary to be able to update the data repo.

### Check-list:

### DIFF
--- a/tfhe/src/js_on_wasm_api/js_high_level_api/zk.rs
+++ b/tfhe/src/js_on_wasm_api/js_high_level_api/zk.rs
@@ -25,13 +25,10 @@ impl From<ZkComputeLoad> for crate::zk::ZkComputeLoad {
 #[wasm_bindgen]
 pub struct CompactPkeCrs(pub(crate) crate::core_crypto::entities::CompactPkeCrs);
 
-#[wasm_bindgen]
-pub struct CompactPkePublicParams(pub(crate) crate::zk::CompactPkePublicParams);
-
 // "wasm bindgen is fragile and prefers the actual type vs. Self"
 #[allow(clippy::use_self)]
 #[wasm_bindgen]
-impl CompactPkePublicParams {
+impl CompactPkeCrs {
     #[wasm_bindgen]
     pub fn serialize(&self, compress: bool) -> Result<Vec<u8>, JsError> {
         catch_panic_result(|| {
@@ -45,12 +42,12 @@ impl CompactPkePublicParams {
     }
 
     #[wasm_bindgen]
-    pub fn deserialize(buffer: &[u8]) -> Result<CompactPkePublicParams, JsError> {
+    pub fn deserialize(buffer: &[u8]) -> Result<CompactPkeCrs, JsError> {
         // If buffer is compressed it is automatically detected and uncompressed.
         // TODO: handle validation
         catch_panic_result(|| {
             bincode::deserialize(buffer)
-                .map(CompactPkePublicParams)
+                .map(CompactPkeCrs)
                 .map_err(into_js_error)
         })
     }
@@ -71,7 +68,7 @@ impl CompactPkePublicParams {
     pub fn safe_deserialize(
         buffer: &[u8],
         serialized_size_limit: u64,
-    ) -> Result<CompactPkePublicParams, JsError> {
+    ) -> Result<CompactPkeCrs, JsError> {
         catch_panic_result(|| {
             crate::safe_serialization::DeserializationConfig::new(serialized_size_limit)
                 .disable_conformance()
@@ -80,12 +77,7 @@ impl CompactPkePublicParams {
                 .map_err(into_js_error)
         })
     }
-}
 
-// "wasm bindgen is fragile and prefers the actual type vs. Self"
-#[allow(clippy::use_self)]
-#[wasm_bindgen]
-impl CompactPkeCrs {
     #[wasm_bindgen]
     pub fn from_parameters(
         parameters: &ShortintParameters,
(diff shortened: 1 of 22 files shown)


tfhe/src/zk.rs
```diff
@@ -123,12 +125,30 @@ impl ZkVerificationOutCome {
...
+#[repr(transparent)]
 pub struct CompactPkeCrs {
     public_params: CompactPkePublicParams,
 }
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct ZkMSBZeroPaddingBitCount(pub u64);
+impl Named for CompactPkeCrs {
+    const NAME: &'static str = "zk::CompactPkeCrs";
+}
```
IceTDrinker, line 143: isn't this an issue for versioning if the struct is transparent ?

tfhe/src/js_on_wasm_api/js_high_level_api/zk.rs
```diff
@@ -111,7 +103,70 @@ impl CompactPkeCrs {
...
+    }
+}
+
+#[wasm_bindgen]
+pub struct CompactPkePublicParams(pub(crate) crate::zk::CompactPkePublicParams);
+
+// "wasm bindgen is fragile and prefers the actual type vs. Self"
+#[allow(clippy::use_self)]
+#[wasm_bindgen]
+impl CompactPkePublicParams {
```
IceTDrinker, line 123: do we want to keep this on the js side ?

IceTDrinker (commented):
If we have functions that can create/generate the old struct we should likely mark them deprecated/remove them, wdyt ?

tfhe/src/zk.rs
nsarlin-zama: No, the name is included outside of the versioned data so it has no influence on the versioning. If one day we want to modify the struct to not be transparent we will have to do a CompactPkeCrsVersions enum that starts at v1:
```rust
enum CompactPkeCrsVersions {
  V0(CompactPkePublicParams),
  V1(CompactPkeCrs)
}
```

IceTDrinker: ok my bad I meant if we want to safe deserialize a CompactPkeCrs from the params, it won't work because of the named impl right ?

tfhe/src/js_on_wasm_api/js_high_level_api/zk.rs
nsarlin-zama: I think we at least need to keep the deserialize code for backward compatibility ?

IceTDrinker: ah is it only deserialization ?

IceTDrinker: could it be put on th Crs and a "deserialize_from_parameters" kind of thing ?

nsarlin-zama:
> If we have functions that can create/generate the old struct we should likely mark them deprecated/remove them, wdyt ?

`crs.public_params()` is needed inside TFHE-rs, but maybe I can remove the pub attribute ? 
However I don't know if blockchain won't need it somehow, I know that they have their own structs with Curve::G1/G2 but I don't know if they need to access points in the crs

tfhe/src/zk.rs
nsarlin-zama: yes, the idea is to deserialize it as a CompactPkePublicParams and then convert it to a CompactPkeCrs using the From impl

IceTDrinker:
> > If we have functions that can create/generate the old struct we should likely mark them deprecated/remove them, wdyt ?
> 
> `crs.public_params()` is needed inside TFHE-rs, but maybe I can remove the pub attribute ? However I don't know if blockchain won't need it somehow, I know that they have their own structs with Curve::G1/G2 but I don't know if they need to access points in the crs

or maybe make it borrow from the CRS ?

IceTDrinker:
maybe I'm wrong here, I don't have a precise enough view of the CRS and params

tfhe/src/zk.rs
IceTDrinker: could we somehow force people to always deserialize the CRS via some from/into tricks ?

tfhe/src/js_on_wasm_api/js_high_level_api/zk.rs
nsarlin-zama: you want to keep both ? Maybe it's better to keep it simple and have only the Crs::deserialize_from_parameters ?

IceTDrinker: sure, I'm just trying to think of users manipulating the old struct and not finding any function on it

tfhe/src/zk.rs
nsarlin-zama: The problem is that since they have been serialized with NAME = "CompactPkePublicParams" it means that the output of safe_deserialize needs to be something with NAME= "CompactPkePublicParams". 

We could have CompactPkeCrs::NAME = "CompactPkePublicParams", but since it's just a matter of calling `.into` I don't think it's worth having this discrepancy.

IceTDrinker: no yeah I agree, was just thinking out loud

IceTDrinker: or maybe a different way to write safe serialize/deserialize with a serialized type S that has Into<T> where T is the actual type we want to use later, but that would maybe defeat the purpose of the safe serialize

IceTDrinker (approved):
Looks good to me !

PR #1766, opened 2024-11-08
Mz/fix oom

mayeul-zama:
Use fewer CPU cores when linking C API tests (expect if there if more than 100GB of total RAM for the CI machines)


IceTDrinker:
does not work https://github.com/zama-ai/tfhe-rs/actions/runs/11744634363/job/32720202203?pr=1766#step:4:2413

which also shows we are missing -e in the C API cmopilation test

IceTDrinker (approved):

PR #1770, opened 2024-11-12
chore(ci): run bench profile as ubuntu

IceTDrinker:
will run benchmarks here to check it works fine


IceTDrinker:
do not approve, looks like ubuntu is not in docker group

IceTDrinker:
benchmarks launched and all seems well https://github.com/zama-ai/tfhe-rs/actions?query=branch%3Aam%2Fchore%2Frun-bench-as-ubuntu

good to review

soonum (approved):

PR #1771, opened 2024-11-12
Ns/bench tfhe zk pok

nsarlin-zama:
### PR content/description
Add benches to the tfhe-zk-pok crates to be able to compare timings between zk v1 and v2

### DIFF
--- a/.github/workflows/benchmark_tfhe_zk_pok.yml
+++ b/.github/workflows/benchmark_tfhe_zk_pok.yml
@@ -1,0 +1,173 @@
+# Run benchmarks of the tfhe-zk-pok crate on an instance and return parsed results to Slab CI bot.
+name: tfhe-zk-pok benchmarks
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  schedule:
+    # Weekly benchmarks will be triggered each Saturday at 3a.m.
+    - cron: '0 3 * * 6'
+env:
+  CARGO_TERM_COLOR: always
+  RESULTS_FILENAME: parsed_benchmark_results_${{ github.sha }}.json
+  PARSE_INTEGER_BENCH_CSV_FILE: tfhe_rs_integer_benches_${{ github.sha }}.csv
+  ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  RUST_BACKTRACE: "full"
+  RUST_MIN_STACK: "8388608"
+  SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
+  SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
+  SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
+  SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+jobs:
+  should-run:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'workflow_dispatch' ||
+      ((github.event_name == 'push' || github.event_name == 'schedule') && github.repository == 'zama-ai/tfhe-rs')
+    outputs:
+      zk_pok_changed: ${{ steps.changed-files.outputs.zk_pok_any_changed }}
+    steps:
+      - name: Checkout tfhe-rs
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          fetch-depth: 0
+
+      - name: Check for file changes
+        id: changed-files
+        uses: tj-actions/changed-files@c3a1bb2c992d77180ae65be6ae6c166cf40f857c
+        with:
+          since_last_remote_commit: true
+          files_yaml: |
+            zk_pok:
+              - tfhe-zk-pok/**
+              - .github/workflows/benchmark_tfhe_zk_pok.yml
+
+  setup-instance:
+    name: Setup instance (tfhe-zk-pok-benchmarks)
+    runs-on: ubuntu-latest
+    needs: should-run
+    if: github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'schedule' && github.repository == 'zama-ai/tfhe-rs') ||
+      (github.event_name == 'push' &&
+      github.repository == 'zama-ai/tfhe-rs' &&
+      needs.should-run.outputs.zk_pok_changed == 'true')
+    outputs:
+      runner-name: ${{ steps.start-instance.outputs.label }}
+    steps:
+      - name: Start instance
+        id: start-instance
+        uses: zama-ai/slab-github-runner@801df0b8db5ea2b06128b7476c652f5ed5f193a8
+        with:
+          mode: start
+          github-token: ${{ secrets.SLAB_ACTION_TOKEN }}
+          slab-url: ${{ secrets.SLAB_BASE_URL }}
+          job-secret: ${{ secrets.JOB_SECRET }}
+          backend: aws
+          profile: bench
+
+  tfhe-zk-pok-benchmarks:
+    name: Execute tfhe-zk-pok benchmarks
+    if: needs.setup-instance.result != 'skipped'
+    needs: setup-instance
+    concurrency:
+      group: ${{ github.workflow }}_${{github.event_name}}_${{ github.ref }}${{ github.ref == 'refs/heads/main' && github.sha || '' }}
+      cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+    runs-on: ${{ needs.setup-instance.outputs.runner-name }}
+    steps:
+      - name: Checkout tfhe-rs repo with tags
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.FHE_ACTIONS_TOKEN }}
+
+      - name: Get benchmark details
+        run: |
+          {
+            echo "BENCH_DATE=$(date --iso-8601=seconds)";
+            echo "COMMIT_DATE=$(git --no-pager show -s --format=%cd --date=iso8601-strict ${{ github.sha }})";
+            echo "COMMIT_HASH=$(git describe --tags --dirty)";
+          } >> "${GITHUB_ENV}"
+
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a
+        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/placeholder_workflow.yml
```diff
@@ -1,14 +1,178 @@
-# Placeholder workflow file allowing running it without having to merge to main first
-name: Placeholder Workflow
+# Run benchmarks of the tfhe-zk-pok crate on an instance and return parsed results to Slab CI bot.
+name: tfhe-zk-pok benchmarks
```
IceTDrinker, line 2: to be reverted once you are done testing

.gitignore
```diff
@@ -13,6 +13,7 @@ target/
 
 # Some of our bench outputs
 /tfhe/benchmarks_parameters
+/tfhe-zk-pok/benchmarks_parameters
```
IceTDrinker, line 16: what's that ?

ci/benchmark_parser.py
```diff
@@ -318,15 +326,16 @@ def parse_key_gen_time(result_file):
     return _parse_key_results(result_file, "latency")
 
 
-def get_parameters(bench_id):
+def get_parameters(bench_id, directory="tfhe"):
```
IceTDrinker, line 329: need for this default value ?

tfhe-zk-pok/benches/pke_prove.rs
```diff
@@ -0,0 +1,54 @@
...
+        .sample_size(15)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    let rng = &mut StdRng::seed_from_u64(0);
+    for (params, param_name) in [
+        (PKEV1_TEST_PARAMS, "PKEV1_TEST_PARAMS"),
+        (PKEV2_TEST_PARAMS, "PKEV2_TEST_PARAMS"),
+    ] {
+        let (public_param, public_commit, private_commit, metadata) = init_params_v1(params);
+        let bits = params.k * 64;
```
IceTDrinker, line 27: that could be way too big

tfhe-zk-pok/benches/pke_prove.rs
```diff
@@ -0,0 +1,54 @@
...
+
+fn bench_pke_prove(c: &mut Criterion) {
+    let bench_shortname = "pke_zk_proof_v1";
+    let bench_name = format!("tfhe_zk_pok::{bench_shortname}");
+    let mut bench_group = c.benchmark_group(&bench_name);
+    bench_group
+        .sample_size(15)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    let rng = &mut StdRng::seed_from_u64(0);
```
IceTDrinker, line 21: not sure we want seeding here

tfhe-zk-pok/benches/pke_v2_prove.rs
```diff
@@ -0,0 +1,52 @@
...
+
+fn bench_pke_v2_prove(c: &mut Criterion) {
+    let bench_shortname = "pke_zk_proof_v2";
+    let bench_name = format!("tfhe_zk_pok::{bench_shortname}");
+    let mut bench_group = c.benchmark_group(&bench_name);
+    bench_group
+        .sample_size(15)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    let rng = &mut StdRng::seed_from_u64(0);
```
IceTDrinker, line 19: seeding ?

tfhe-zk-pok/benches/pke_v2_prove.rs
```diff
@@ -0,0 +1,52 @@
...
+    for (params, param_name) in [
+        (PKEV1_TEST_PARAMS, "PKEV1_TEST_PARAMS"),
+        (PKEV2_TEST_PARAMS, "PKEV2_TEST_PARAMS"),
+    ] {
+        let (public_param, public_commit, private_commit, metadata) = init_params_v2(params);
+        let bits = params.k * 64;
```
IceTDrinker, line 25: potentially too big

tfhe-zk-pok/benches/pke_verify.rs
```diff
@@ -0,0 +1,57 @@
+use criterion::{criterion_group, criterion_main, Criterion};
```
IceTDrinker, line 1: can't those be in the same file ? see the integer bench for pke zk

tfhe-zk-pok/benches/pke_verify.rs
```diff
@@ -0,0 +1,57 @@
...
+
+fn bench_pke_verify(c: &mut Criterion) {
+    let bench_shortname = "pke_zk_verify_v1";
+    let bench_name = format!("tfhe_zk_pok::{bench_shortname}");
+    let mut bench_group = c.benchmark_group(&bench_name);
+    bench_group
+        .sample_size(15)
+        .measurement_time(std::time::Duration::from_secs(60));
+
+    let rng = &mut StdRng::seed_from_u64(0);
```
IceTDrinker, line 21: seed

tfhe-zk-pok/benches/pke_verify.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    for (params, param_name) in [
+        (PKEV1_TEST_PARAMS, "PKEV1_TEST_PARAMS"),
+        (PKEV2_TEST_PARAMS, "PKEV2_TEST_PARAMS"),
+    ] {
+        let (public_param, public_commit, private_commit, metadata) = init_params_v1(params);
+        let bits = params.k * 64;
```
IceTDrinker, line 28: size ?

tfhe-zk-pok/benches/utils.rs
```diff
@@ -0,0 +1,459 @@
+#![allow(non_snake_case)]
```
IceTDrinker, line 1: looks to be mostly a copy paste ?

tfhe-zk-pok/benches/pke_verify.rs
IceTDrinker: we should not go above the zk capacity which is

k * effective_t.ilog2()

here you chose

k * 64 bits ?

IceTDrinker (changes requested):
I think something is not right with the k * 64, seeding feels unnecessary as we have randomness in real world performance/rng

We must find a way to avoid the duplication for the pure in house benchmark utils stuff and the zk-sepcific parameters setup this won't be maintainable long term

nsarlin-zama:
manual workflow run: https://github.com/zama-ai/tfhe-rs/actions/runs/11800194113/job/32870653319

tfhe-zk-pok/benches/utils.rs
nsarlin-zama: I need the types in `tfhe/benches/utilities` to be able to create the json file with the params but I can't use it directly because I am not in the `tfhe` crate. This is mostly a copy paste with only the needed bits and `tfhe` internal types converted to builtin rust types.

I also integrated things from `tfhe-zk-pok/proof/mod.rs` `test` module because benches are not run with `#[cfg(test)]`

IceTDrinker: the utils can be marked with a cfg = "bench" I believe

I think we have something like that in TFHE-rs

as for the utils some work needs to be done with David to make the copied code into a small util crate, we can't copy things when we "just" need to bench something

.gitignore
nsarlin-zama: running the benchmarks will create this folder which should not be versioned.

tfhe-zk-pok/benches/pke_verify.rs
nsarlin-zama: But if this is not needed I can easily merge the files

IceTDrinker: you can filter with criterion normally

.github/workflows/benchmark_tfhe_zk_pok.yml
```diff
@@ -0,0 +1,178 @@
...
+          {
+            echo "BENCH_DATE=$(date --iso-8601=seconds)";
+            echo "COMMIT_DATE=$(git --no-pager show -s --format=%cd --date=iso8601-strict ${{ github.sha }})";
+            echo "COMMIT_HASH=$(git describe --tags --dirty)";
+          } >> "${GITHUB_ENV}"
+
+      - name: Set up home
+        # "Install rust" step require root user to have a HOME directory which is not set.
+        run: |
+          echo "HOME=/home/ubuntu" >> "${GITHUB_ENV}"
```
soonum, line 96: You can remove this step, now runner is installed as user `ubuntu` so the `HOME` is already set.

ci/benchmark_parser.py
soonum: Agreed, I think it's better to pass this argument as a positional one.

tfhe-zk-pok/benches/pke_verify.rs
nsarlin-zama: I think I misunderstood the code in `benches/integer/zk_pke.rs`, I'm fixing that !

tfhe-zk-pok/benches/utils.rs
soonum: You think about a crate like `tfhe-utils` that would contains `CryptoParametersRecord` and `write_to_json()` function ?

IceTDrinker: potentially, we can't duplicate those everywhere

nsarlin-zama: The problem is that the current bench utilities make heavy use of tfhe types so it cannot be used without it, it would need a rework that's out of the scope of this PR I think. Since tfhe itself depend on tfhe-zk-pok it should be written in a way that prevent circular dependencies.

I don't think there is a `cfg(bench)` and the code is already only compiled with benches, the problem is that I can't use code from `cfg(test)` from there.

IceTDrinker: see tfhe/src/core_crypto/commons/ciphertext_modulus.rs:23 fairly sure there is something for benchmarks

nsarlin-zama: Are you sure that this works ? I don't find any doc about it, it does not seem to have any effect on my code and cargo warns about it. Maybe there is some special trick in the tfhe crate to make it work ?

.github/workflows/placeholder_workflow.yml
```diff
@@ -1,14 +1,173 @@
...
```
IceTDrinker, line 2: to be reset once you have the confirmation data is uploaded

IceTDrinker (approved):
Thanks 💪

PR #1782, opened 2024-11-14
Tm/noise asserts

tmontaigu:


tfhe/src/shortint/ciphertext/standard.rs
```diff
@@ -138,7 +138,13 @@ impl Ciphertext {
         self.noise_level
     }
 
-    pub fn set_noise_level(&mut self, noise_level: NoiseLevel) {
+    #[cfg_attr(any(feature = "noise-asserts", test), track_caller)]
```
mayeul-zama, line 141: Could be a simple `#[track_caller]` to avoid some complexity with features

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -913,6 +922,14 @@ impl ServerKey {
         ct: &Ciphertext,
         acc: &ManyLookupTableOwned,
     ) -> Vec<Ciphertext> {
+        if cfg!(any(feature = "noise-asserts", test)) && ct.noise_level() != NoiseLevel::UNKNOWN {
```
mayeul-zama, line 925: `NoiseLevel::UNKNOWN` and `NoiseLevel::MAX` have the same value which means a `NoiseLevel::MAX` would not be checked here
It seems to me that it would be better to panic if we have `NoiseLevel::MAX` here

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
```diff
@@ -177,7 +177,10 @@ impl ServerKey {
             // And here, it's because shortint sub_assign adds a correcting term,
             // which we do not want here
             crate::core_crypto::algorithms::lwe_ciphertext_sub_assign(&mut lhs_b.ct, &borrow.ct);
-            lhs_b.set_noise_level(lhs_b.noise_level() + borrow.noise_level());
+            lhs_b.set_noise_level(
+                lhs_b.noise_level() + borrow.noise_level(),
+                self.key.max_noise_level,
```
mayeul-zama, line 182: `ServerKey.max_degree` is initialized with `MaxDegree::integer_radix_server_key` in radix integers
This keeps space for carry propagation in all operations (except in carry propagation I guess, not sure if there are degree checks there)

I think we should do something similar for noise level
Or are noise checks in carry propagation enough? If so, I guess it would mean some operations could panic in carry propagation even if we could have prevented it in the previous operation (adding a PBS if the `max_noise_level_keep_space_for_carry_propagation` is too big)  

In practice not sure this is a problem as noise and degree checks are often redundant.

mayeul-zama (commented):
Instead of asserting on max noise level, I seems to me that it would be better to do operations differently if the noise_level is going to be too big (as is done for the degree).

This would mean we could have more confidence in noise validity (not just in tests or when activating a feature) without the risk of panics.

I guess this would need more work 
That being said, this PR is an improvement over the current state

tmontaigu:
> Instead of asserting on max noise level, I seems to me that it would be better to do operations differently if the noise_level is going to be too big (as is done for the degree).

It's not the goal here, the goal is to have some checks that enable us in test to know when we, by mistake crossed a limit on the noise level. And since those are at shortint level, its already too late to do the things differently (like you add two shortint blocks, but the sum exceeds the noise level, its too late, you don't have another way to do the addition)

Later at integer level we will add test on noise level to pre-clean if needed

tfhe/src/shortint/server_key/mod.rs
tmontaigu: I think it would be better to make the values of unknown and max different, because I do need the distinction because of some things do in the casting keys

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
tmontaigu: While I do thing max_degree and max_noise_level should be aligned I don't know if its better to align max_degree to max_noise_level or the reverse

tfhe/src/shortint/server_key/mod.rs
tmontaigu: The idea is that if the noise level is unknown we may allow to do the PBS, but not if'ts at MAX

In the casting keys, at some point, there's a ciphertext that is set with noise_level unknown and we do a PBS on it, so either we allow to do PBS on NoiseLevel::UNKNOWN noise level, or We change the thing in the casting key to not be at NoiseLevel::UNKNOWN

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
tmontaigu: The only use of the ServerKey.max_degree having space for 1 carry addition is for smart_add.sub/neg to work if the carry_propagation algorithm is the sequential one, not parallel

All other algorithm don't rely on this

tfhe/src/shortint/server_key/mod.rs
IceTDrinker: so here to note that the MAX is usize::MAX i.e. an impossible value to reach, UNKNOWN was set to MAX as a safety

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
IceTDrinker: max_noise_level is separate from the max degree and depends on the optimization of parameters

IceTDrinker: I think it's the question about "what do the parameters allow ?" i.e. the max_max_degree and the max_max_noise_level

and

"what are my constraints for these integers" which are the max_degree and max_noise_level

dont know if this is the right PR for this change (having both notions)

tfhe/src/shortint/ciphertext/common.rs
```diff
@@ -26,6 +26,10 @@ impl std::error::Error for NotTrivialCiphertextError {}
 pub struct MaxNoiseLevel(usize);
 
 impl MaxNoiseLevel {
+    #[cfg(test)]
+    pub(crate) const MAX: Self = Self(usize::MAX);
```
IceTDrinker, line 30: why this config ?

tfhe/src/integer/server_key/radix_parallel/scalar_sub.rs
IceTDrinker: maybe NoiseLevel::MAX makes no sense to be honest

tfhe/src/shortint/list_compression/compression.rs
```diff
@@ -87,8 +87,9 @@ impl CompressionKey {
                     );
 
                     let mut ct = ct.clone();
-
-                    unchecked_scalar_mul_assign(&mut ct, message_modulus.0 as u8);
+                    let max_noise_level =
+                        MaxNoiseLevel::new((ct.noise_level() * message_modulus.0).get());
```
IceTDrinker, line 91: this noise comes from "nowhere" right, we have no key for a reference max noise ?

IceTDrinker:
> Is this something we should also port to the GPU backend you think?

Can be, depending on how you track the noise in your implem, this way during tests you can catch when noise gets too high, but I don't know how you manage the degree/noise data in CUDA :)

tfhe/src/shortint/server_key/mod.rs
IceTDrinker: maybe we need to think about using

MAX = usize::MAX - 1
UNKNOWN = usize::MAX

and see if that could make sense in the general case

given that a noise of usize::MAX - 1

would break anything

also maybe time to change the noise level stuff to use u32 or u64 instead of usize, given that those don't refer to memory 

not necessarily in this PR but it begs the question

tfhe/src/shortint/list_compression/compression.rs
IceTDrinker: @mayeul-zama the key has no information of the multiplication required to compress ?

tfhe/src/shortint/key_switching_key/mod.rs
```diff
@@ -575,7 +575,10 @@ impl<'keys> KeySwitchingKeyView<'keys> {
                         let wrong_key_ct = keyswitched;
                         let mut correct_key_ct = self.dest_server_key.create_trivial(0);
                         correct_key_ct.degree = wrong_key_ct.degree;
-                        correct_key_ct.set_noise_level(wrong_key_ct.noise_level());
+                        correct_key_ct.set_noise_level(
+                            wrong_key_ct.noise_level(),
+                            self.dest_server_key.max_noise_level,
```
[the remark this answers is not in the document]
tmontaigu: Yes, hence the fact that PBS allows noise level to be unknown, as we assume the noise level is not truly known but the caller trusts its within the correct range

tfhe/src/shortint/ciphertext/common.rs
tmontaigu: Its only used in tests and since its pub(crate) clippy gives an error

tfhe/src/shortint/key_switching_key/mod.rs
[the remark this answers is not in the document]
IceTDrinker: I see, ok yeah

tfhe/src/shortint/ciphertext/standard.rs
tmontaigu: Yep went a bit too far here 😁

tfhe/src/shortint/list_compression/compression.rs
mayeul-zama: The compression parameters take into account this shift of the message (which is clean, see assert `NoiseLevel::NOMINAL` above)  to the carry bits.
In my understanding, it means that, in this compression pattern contract,  the multiplication required to compress is by `message_modulus`.

@fakub could you confirm?

IceTDrinker: What I mean is, does it have the constant to multiply with in the key ?

mayeul-zama: It's `message_modulus` which is not stored in `CompressionKey`

IceTDrinker: well that's not good at all, means keys can be easily misused

mayeul-zama: When misused, it would most likely fail a few lines above on
```
 assert_eq!(
    lwe_size,
    ct.ct.lwe_size(),
    "All ciphertexts do not have the same lwe size as the packing keyswitch key"
);
```

But indeed it would be safer also check against the `CompressionKey` expected `message_modulus`

tfhe/src/shortint/ciphertext/common.rs
IceTDrinker: switch to u64 as well

tmontaigu:
note that switching from usize -> u64 is a API (small), and a potential serialization break depending on the serialization format (bincode should be fine)

IceTDrinker:
> note that switching from usize -> u64 is a API (small), and a potential serialization break depending on the serialization format (bincode should be fine)

if needed tfhe-versionable should help manage that

tfhe/src/shortint/list_compression/compression.rs
tmontaigu: Not sure I followed everything, is there something to fix in this PR ?

IceTDrinker: not for you but keys are missing an information as far as I can tell

tfhe/src/c_api/shortint/parameters.rs
```diff
@@ -62,7 +62,7 @@ impl TryFrom<ShortintPBSParameters> for crate::shortint::ClassicPBSParameters {
                 c_params.modulus_power_of_2_exponent,
             )?,
             max_noise_level: crate::shortint::parameters::MaxNoiseLevel::new(
-                c_params.max_noise_level,
+                c_params.max_noise_level as u64,
```
IceTDrinker, line 65: update C API ones ?

tfhe/src/integer/server_key/mod.rs
```diff
@@ -246,7 +246,7 @@ impl ServerKey {
             MaxDegree::from_msg_carry_modulus(self.message_modulus(), self.carry_modulus());
         let max_sum_to_full_carry = max_degree.get() / degree.get();
 
-        max_sum_to_full_carry.min(self.key.max_noise_level.get())
+        max_sum_to_full_carry.min(self.key.max_noise_level.get() as usize)
```
IceTDrinker, line 249: let's plan on updating the degrees as well ?

IceTDrinker: can be in a separate PR

tfhe/src/shortint/server_key/bivariate_pbs.rs
```diff
@@ -36,7 +36,7 @@ fn ciphertexts_can_be_packed_without_exceeding_space_or_noise(
 
     max_degree.validate(final_degree)?;
 
-    let final_noise_level = (lhs.noise_level * factor) + rhs.noise_level;
+    let final_noise_level = (lhs.noise_level * factor as u64) + rhs.noise_level;
```
IceTDrinker, line 39: this could be updated as well to avoid the conversion here

IceTDrinker (commented):
I see lots of small things that could be changed to make everything consistent, do we say it's for another PR ?

tmontaigu:
I'm already working on make the Degree,MaxDegree,MessageModulus,CarryModulus use u64, I can add the commit to this PR once its done

IceTDrinker:
> I'm already working on make the Degree,MaxDegree,MessageModulus,CarryModulus use u64, I can add the commit to this PR once its done

fine by me

IceTDrinker (approved):
Thanks a lot !

IceTDrinker:
Some integer test broke with the noise check

IceTDrinker:
Looking good

PR #1783, opened 2024-11-14
Ns/test zk bad noise

nsarlin-zama:
### PR content/description
Adds a test for zk proofs with a larger noise than the accepted one.
This pr also include small refactor of the zk proofs for easier testing, and a few fixes. Each one is done in a separate commit.

The tests for v1 and v2 are not exactly analogous to more effectively test the edge cases of the proven norms


tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -507,16 +507,17 @@ pub fn prove<G: Curve>(
 
     let effective_t_for_decomposition = t >> msbs_zero_padding_bit_count;
 
+    let effective_q = decode_q(q);
```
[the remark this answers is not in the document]
nsarlin-zama: In it's not really the type that I want to express but the fact that in the input q, 0 actually means u64::MAX, whereas in "effective_q", 0 means 0 and u64::MAX means u64::MAX.
Shadowing the definition might be a bit dangerous because when looking at code you might skip the part where it's redefined.
But I'm ok with that if you think that's clearer.

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -1843,10 +1864,19 @@ pub fn verify<G: Curve>(
...
+    let B_squared = inf_norm_bound_to_euclidean_squared(B_inf, d + k);
+    let (_, D, _, m_bound) = compute_crs_params(
+        d,
+        k,
+        B_squared,
+        t_input,
+        msbs_zero_padding_bit_count,
+        bound_type,
+    );
+    //let n = crs_n;
```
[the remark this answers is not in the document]
nsarlin-zama: Yes n is used a lot in the proof, but it's the one from the crs

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -761,7 +783,7 @@ fn prove_impl<G: Curve>(
         )
         .collect::<Box<[_]>>();
 
-    let v = four_squares(sqr(B as u128) - e_sqr_norm).map(|v| v as i64);
+    let v = four_squares(B_squared - e_sqr_norm).map(|v| v as i64);
```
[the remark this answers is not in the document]
nsarlin-zama: the v here should be smaller than (B_squared - e_sqr_norm), which itself is smaller than i64::MAX because we checked above that B_squared is greater than e_sqr_norm.

tfhe-zk-pok/src/proofs/pke.rs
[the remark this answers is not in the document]
IceTDrinker: is q used in its non encoded form ?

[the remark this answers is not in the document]
nsarlin-zama: or use q and cast in the hash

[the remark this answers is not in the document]
IceTDrinker: casting for the hash would result in a 0 so essentially the encoded q

for the hash it's better to have the 64 bits value I think as we don't do u128 proofs, but yeah maybe decoded_q then

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2861,6 +2861,36 @@ mod tests {
...
+        assert_eq!(D, 3328);
+        assert_eq!(B_bound_squared, 3867562496364372);
+        assert_eq!(m_bound, 27);
+
+        let (n, D, B_bound_squared, m_bound) =
+            compute_crs_params(d, k, B_squared, t, msbs_zero_padding_bit_count, Bound::CS);
+        assert_eq!(n, 7168);
+        assert_eq!(D, 3328);
+        assert_eq!(B_bound_squared, 192844141830554880);
+        assert_eq!(m_bound, 30);
```
IceTDrinker, line 2891: Have verified all those values by hand with numpy, all look good !

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1396,6 +1397,211 @@ mod tests {
...
+    fn assert_prove_and_verify<G: Curve>(
+        testcase: &PkeTestcase,
+        testcase_name: &str,
+        crs: &PublicParams<G>,
+        rng: &mut StdRng,
+        expected_result: VerificationResult,
+    ) {
+        for load in [ComputeLoad::Proof, ComputeLoad::Verify] {
+            if prove_and_verify(testcase, crs, load, rng) != expected_result {
+                panic!("Testcase {testcase_name} failed")
```
IceTDrinker, line 1446: maybe assert_eq! and put the message so that we can get the expected/actual value ?

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1396,6 +1397,211 @@ mod tests {
...
+        let bad_idx = rng.gen::<usize>() % d;
+        let bad_term = (rng.gen::<u64>() % u64::MAX - B) as i64 + B as i64;
+
+        bad_e1[bad_idx] = if rng.gen() { bad_term } else { -bad_term };
+        testcase_bad_e1.e1 = bad_e1;
+
+        type Curve = curve_api::Bls12_446;
+
+        // To check management of bigger k_max from CRS during test
+        let crs_k = k + 1 + (rng.gen::<usize>() % (d - k));
```
IceTDrinker, line 1480: just a note this always gives a crs_k > k, we could try to also have the case crs_k = actual k by slightly tweaking the formula (maybe in another PR, it's a minor change unrelated to this PR)

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1396,6 +1397,211 @@ mod tests {
...
+            q,
+            t,
+            msbs_zero_padding_bit_count,
+        } = PKEV1_TEST_PARAMS;
+
+        let rng = &mut StdRng::seed_from_u64(0);
+
+        let testcase = PkeTestcase::gen(rng, PKEV1_TEST_PARAMS);
+
+        let mut testcase_bad_e1 = testcase.clone();
```
IceTDrinker, line 1467: what is this testcase_bad_e1 doing here ? I see it being cloned again/recreated line 1485 ?

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2545,6 +2546,322 @@ mod tests {
...
+        testcase: &PkeTestcase,
+        testcase_name: &str,
+        crs: &PublicParams<G>,
+        rng: &mut StdRng,
+        expected_result: VerificationResult,
+    ) {
+        for load in [ComputeLoad::Proof, ComputeLoad::Verify] {
+            if prove_and_verify(testcase, crs, load, rng) != expected_result {
+                panic!("Testcase {testcase_name} failed")
+            }
```
IceTDrinker, line 2596: same comment could use an assert_eq!

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2545,6 +2546,322 @@ mod tests {
...
+                        .map(|_| if rng.gen() { good_term } else { -good_term })
+                        .collect();
+                    testcase.e2 = (0..k)
+                        .map(|_| if rng.gen() { good_term } else { -good_term })
+                        .collect();
+                }
+            };
+
+            let B_with_slack_squared = inf_norm_bound_to_euclidean_squared(B, d + k);
+            let B_with_slack = isqrt(B_with_slack_squared) as u64;
```
IceTDrinker, line 2727: no problem with floor vs round vs ceil here in this sqrt ?

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2545,6 +2546,322 @@ mod tests {
...
+                        .sum::<u128>();
+
+                    let orig_value = match coeff_type {
+                        TestedCoeffType::E1 => testcase.e1[tested_idx],
+                        TestedCoeffType::E2 => testcase.e2[tested_idx],
+                    };
+
+                    let bound_squared =
+                        B_with_slack_squared - (e_sqr_norm - sqr(orig_value as u128));
+                    isqrt(bound_squared) as i64
```
IceTDrinker, line 2748: same question no shenanigans with floor/round/ceil ?

tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -2545,6 +2546,322 @@ mod tests {
...
+
+        let B_with_slack_squared = inf_norm_bound_to_euclidean_squared(B, d + k);
+        let B_with_slack = isqrt(B_with_slack_squared) as u64;
+
+        // Generate test noise vectors with random coeffs and one completely out of bounds
+
+        let mut testcases = Vec::new();
+        let mut testcase_bad_e1 = testcase.clone();
+        let bad_idx = rng.gen::<usize>() % d;
+        let bad_term = (rng.gen::<i64>() % i64::MAX - B_with_slack as i64) + B_with_slack as i64;
```
IceTDrinker, line 2799: you could get a 0 for the RNG and get out B_with_slack which would be in bound right ?

IceTDrinker (changes requested):
I think there are a few small mistakes for some random bad term values, otherwise looks very solid, thanks a lot for the hard work !

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: yeah I will make it use the same pattern as pke_v2, with another specific crs for this case

tfhe-zk-pok/src/proofs/pke.rs
nsarlin-zama: copy paste error, thanks for catching it

tfhe-zk-pok/src/proofs/pke_v2.rs
nsarlin-zama: Here we want to always be within the bound so I think that floor is ok

tfhe-zk-pok/src/proofs/pke_v2.rs
nsarlin-zama: same answer. The goal is to have a value x that is always accepted and x+1 is always refused, so I think that floor is the right choice

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1442,6 +1397,207 @@ mod tests {
...
+
+        // A CRS where the number of slots is bigger than the number of messages to encrypt
+        let big_crs_k = k + 1 + (rng.gen::<usize>() % (d - k));
+        let crs_bigger_k =
+            crs_gen::<Curve>(d, big_crs_k, B, q, t, msbs_zero_padding_bit_count, rng);
+
+        // ==== Generate test noise vectors with random coeffs and one completely out of bounds ===
+        let mut testcase_bad_e1 = testcase.clone();
+        let bad_idx = rng.gen::<usize>() % d;
+        let bad_term = (rng.gen::<i64>() % i64::MAX - B as i64) + B as i64;
```
IceTDrinker, line 1482: does the bad term behave properly as well ? I'm having some doubts now

nsarlin-zama: It should be ok if we take it as a positive value and nok if we take the negative one, I'm fixing it for this case

tfhe-zk-pok/src/proofs/pke.rs
```diff
@@ -1442,6 +1397,208 @@ mod tests {
...
+        // A CRS where the number of slots is bigger than the number of messages to encrypt
+        let big_crs_k = k + 1 + (rng.gen::<usize>() % (d - k));
+        let crs_bigger_k =
+            crs_gen::<Curve>(d, big_crs_k, B, q, t, msbs_zero_padding_bit_count, rng);
+
+        // ==== Generate test noise vectors with random coeffs and one completely out of bounds ===
+        let mut testcase_bad_e1 = testcase.clone();
+        let bad_idx = rng.gen::<usize>() % d;
+        // Generate a value between B + 1 and i64::MAX to make sure that it is out of bounds
+        let bad_term = (rng.gen::<i64>() % i64::MAX - (B + 1) as i64) + (B + 1) as i64;
```
IceTDrinker, line 1483: does that work given this is signed ? I feel like the modulo operator can return signed values ?

nsarlin-zama: Oh yeah I missed the gen::<i64>, I fixed it to generate u64 then cast

tfhe-zk-pok/src/proofs/pke_v2.rs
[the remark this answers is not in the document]
IceTDrinker: we have a test case that covers this one ?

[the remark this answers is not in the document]
nsarlin-zama: what do you mean ?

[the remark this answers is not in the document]
nsarlin-zama: B_squared is greater than e_sqr_norm if we are inside the noise bounds and smaller if we are outside, so both cases are covered by noise tests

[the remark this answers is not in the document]
IceTDrinker: that was my question, then all good !

IceTDrinker (approved):

PR #1786, opened 2024-11-15
chore(hlapi): stabilize FheTypes

mayeul-zama:


tfhe/src/c_api/high_level_api/mod.rs
```diff
@@ -22,23 +22,23 @@ mod zk;
...
+    Type_FheBool = 0,
     Type_FheUint2,
-    Type_FheUint4,
+    Type_FheUint4 = 1,
     Type_FheUint6,
-    Type_FheUint8,
+    Type_FheUint8 = 2,
     Type_FheUint10,
     Type_FheUint12,
     Type_FheUint14,
```
IceTDrinker, line 32: We need all variants to have a value otherwise we'll have an issue later on, I'll check with BC for that and the fact we have more types than them

IceTDrinker (changes requested):
Wait why is this the C API enum being updated ?

tfhe/src/high_level_api/mod.rs
```diff
@@ -130,23 +130,23 @@ pub enum Device {
 
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum FheTypes {
```
IceTDrinker, line 132: this needs to be marked #[repr(i32)] and all the types need to have a value associated, then we'll check with BC it's fine with them

tfhe/src/c_api/high_level_api/mod.rs
IceTDrinker: we probably need a test (with a macro) to check that all values match between the C Enum and the HL one

tfhe/src/c_api/high_level_api/mod.rs
```diff
@@ -22,23 +22,23 @@ mod zk;
...
+    Type_FheUint64 = 5,
+    Type_FheUint128 = 6,
+    Type_FheUint160 = 7,
+    Type_FheUint256 = 8,
+    Type_FheUint512 = 9,
+    Type_FheUint1024 = 10,
+    Type_FheUint2048 = 11,
     Type_FheInt2,
     Type_FheInt4,
     Type_FheInt6,
```
IceTDrinker, line 44: still missing values here

IceTDrinker (approved):
To avoid blocking while I'm away I'm approving, does not mean this should be merged as is

tfhe/src/high_level_api/mod.rs
```diff
@@ -129,35 +130,37 @@ pub enum Device {
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(i32)]
+#[derive(EnumIter)]
```
IceTDrinker, line 134: derive only in #[cfg(test)] maybe ? avoids having another mandatory dependency

so #[cfg_attr(test, derive(EnumIter))] I believe

tfhe/src/c_api/high_level_api/mod.rs
```diff
@@ -90,3 +90,14 @@ impl From<crate::FheTypes> for FheTypes {
...
+
+#[test]
+fn fhe_types_enum_to_int_compatible2() {
+    use strum::IntoEnumIterator;
+
+    for rust_value in crate::FheTypes::iter() {
+        let c_value = FheTypes::from(rust_value);
+
+        assert_eq!(rust_value as i32, c_value as i32)
+    }
```
IceTDrinker, line 102: needs to be in a #[cfg(test)] module or have the annotation

IceTDrinker:
as I approved there was this crash that looks related to the new dependency you added https://github.com/zama-ai/tfhe-rs/actions/runs/12315454529/job/34373751603

tfhe/src/c_api/high_level_api/mod.rs
IceTDrinker: huh !

tfhe/src/high_level_api/mod.rs
mayeul-zama: Done

tfhe/src/c_api/high_level_api/mod.rs
IceTDrinker: did you check this test runs with `make test_c_api_rs` ?

mayeul-zama: Just did, it does

IceTDrinker (approved):
This time I'm good with this :)

as long as all is green !

PR #1788, opened 2024-11-15
Tm/fix sub signed overflow

tmontaigu:


nsarlin-zama (commented):
Small comment but aside from that it looks good to me

nsarlin-zama (approved):
thanks!

PR #1792, opened 2024-11-18
fix: bench key size parsing

nsarlin-zama:
For keygen and keysize benches, the `crate` argument was not set when calling `get_parameters` from `_parse_key_results`.
Since these benches only make sense from the main `tfhe` crate, the value of the argument is forced to be `tfhe`.


IceTDrinker (approved):

nsarlin-zama:
superseded by #1798

PR #1793, opened 2024-11-18
fix(hlapi): rework CompressedCiphertextListBuilder

tmontaigu:
The hlapi builder target device was selected depending on features (gpu enabled ? gpu : cpu), but if at `build` time the thread_local key did not match the expected device, an error would be returned.

This is a bit too limiting for users that might want to do some processing on GPU and compression on CPU.

So the Builder is changed to delay, the selection of device used to compress when `build` is called.
This new design is more flexible for end users, at the cost of a bit more memory copies

* There should be no API breaking change
* There is no serialization breaking change as only the builder
  (which is not serializable) has been changed

### DIFF
--- a/tfhe/src/high_level_api/compressed_ciphertext_list.rs
+++ b/tfhe/src/high_level_api/compressed_ciphertext_list.rs
@@ -3,43 +3,72 @@ use tfhe_versionable::{Unversionize, UnversionizeError, Versionize, VersionizeOw
 use super::keys::InternalServerKey;
 use crate::backward_compatibility::compressed_ciphertext_list::CompressedCiphertextListVersions;
 use crate::core_crypto::commons::math::random::{Deserialize, Serialize};
+use crate::high_level_api::booleans::InnerBoolean;
+use crate::high_level_api::errors::UninitializedServerKey;
 #[cfg(feature = "gpu")]
 use crate::high_level_api::global_state::with_thread_local_cuda_streams;
 use crate::high_level_api::integers::{FheIntId, FheUintId};
-use crate::integer::ciphertext::{Compressible, DataKind, Expandable};
+use crate::integer::ciphertext::{DataKind, Expandable};
 #[cfg(feature = "gpu")]
 use crate::integer::gpu::ciphertext::compressed_ciphertext_list::{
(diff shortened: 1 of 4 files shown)


tfhe/src/high_level_api/compressed_ciphertext_list.rs
```diff
@@ -515,67 +566,102 @@ mod tests {
...
+            ct2.move_to_device(crate::Device::CudaGpu);
+            ct3.move_to_device(crate::Device::CudaGpu);
+            ct4.move_to_device(crate::Device::CudaGpu);
+
+            let mut compressed_list_builder = CompressedCiphertextListBuilder::new();
+            compressed_list_builder
+                .push(ct1)
+                .push(ct2)
+                .push(ct3)
+                .push(ct4);
```
IceTDrinker, line 626: could we have a CPU/GPU mix as well ?

tmontaigu: Yes we can

tmontaigu:
> Only one question, otherwise this looks good, though it shows we need to upgrade primitives to use iterators over some types rather than force the WhateverList construct inherited from the previous iteration of the lib

Yeah I sort of considered to do it here, but for this kind of things, it generally we want to iterate twice: once to get the number of lwe to preallocate the glwe_list, and a second time so populate the glwes

tfhe/src/high_level_api/compressed_ciphertext_list.rs
IceTDrinker: Obama for president

IceTDrinker:
> > Only one question, otherwise this looks good, though it shows we need to upgrade primitives to use iterators over some types rather than force the WhateverList construct inherited from the previous iteration of the lib
> 
> Yeah I sort of considered to do it here, but for this kind of things, it generally we want to iterate twice: once to get the number of lwe to preallocate the glwe_list, and a second time so populate the glwes

otherwise we need primitives to extend a List instead of manipulating the container, you can enhance ContiguousEntityContainerMut if you want to be able to append

IceTDrinker (approved):

PR #1798, opened 2024-11-19
chore(ci): fix recipe for integer benchmarks on gpu

soonum:


IceTDrinker (approved):

soonum:
I've launched all the failed ones on `main`. I'll wait for any other failures to make a possible fix here.

IceTDrinker (commented):
this likely does not work for the parser fix, the functions don't seem to take an additional param

PR #1800, opened 2024-11-20
Fix: degree for min max

Tuditi:
### PR content/description

Contains a test to recreate an issue we have been having at Belfort. The min - max functions seem to fail at the assertion that [the carries are empty](https://github.com/zama-ai/tfhe-rs/blob/46cf46563770f8e8072ff97692bf18d7d8673158/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_comparison.rs#L611). IIUC, this is due to the fact that `unchecked_add` will add together the degrees, but we know that either the LHS or the RHS will be set to 0 after performing the LUT [here](https://github.com/zama-ai/tfhe-rs/blob/46cf46563770f8e8072ff97692bf18d7d8673158/tfhe/src/integer/server_key/comparator.rs#L405-L406).

 It can be solved in the following ways:

1. Set the degree to `Degree::new(msg_modulus - 1)`.
2. Add a remainder operation in the LUTs (`% message_modulus`).
3. Calculate the maximum degree before applying the additions and the LUTs.

- Which one is your preferred solution?

### Check-list:

### DIFF
--- a/tfhe/src/integer/server_key/radix/tests.rs
+++ b/tfhe/src/integer/server_key/radix/tests.rs
@@ -1,6 +1,7 @@
 use crate::integer::keycache::KEY_CACHE;
 use crate::integer::server_key::radix_parallel::tests_cases_unsigned::*;
 use crate::integer::server_key::radix_parallel::tests_unsigned::test_add::smart_add_test;
+use crate::integer::server_key::radix_parallel::tests_unsigned::test_comparison::test_unchecked_minmax;
 use crate::integer::server_key::radix_parallel::tests_unsigned::test_neg::smart_neg_test;
 use crate::integer::server_key::radix_parallel::tests_unsigned::test_slice::{
     default_scalar_bitslice_assign_test, default_scalar_bitslice_test,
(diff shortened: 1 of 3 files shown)


tfhe/src/integer/server_key/radix/tests.rs
```diff
@@ -117,6 +118,10 @@ create_parametrized_test_classical_params!(integer_default_scalar_slice);
 create_parametrized_test_classical_params!(integer_default_scalar_slice_assign);
 create_parametrized_test_classical_params!(integer_smart_scalar_slice);
 create_parametrized_test_classical_params!(integer_smart_scalar_slice_assign);
+create_parametrized_test!(integer_unchecked_min {
+    coverage => {COVERAGE_PARAM_MESSAGE_2_CARRY_2_KS_PBS},
+    no_coverage => {PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M64}
+});
```
tmontaigu, line 124:
```suggestion
create_parametrized_test_classical_params!(integer_unchecked_min);
```

tfhe/src/integer/server_key/radix/tests.rs
```diff
@@ -800,3 +805,8 @@ fn integer_smart_scalar_slice_assign(param: ClassicPBSParameters) {
     let executor = CpuFunctionExecutor::new(&ServerKey::smart_scalar_bitslice_assign);
     smart_scalar_bitslice_assign_test(param, executor);
 }
+
+fn integer_unchecked_min(param: ClassicPBSParameters) {
+    let executor = CpuFunctionExecutor::new(&ServerKey::unchecked_min);
+    test_default_minmax(param, 2, executor, std::cmp::min::<u64>);
```
tmontaigu, line 811: test_unchecked_minmax

tmontaigu:
While the fix to set the degree is correct and should be kept, you are not using the correct test

This makes you believe the function is incorrect but it is.
You are putting an `unchecked_ ` function to a function designed to test a "default" so test gives inputs that are not meant to be handled by an unchecked function and the test also does post condition checks that are not meant for unchecked function.

Also you should use the `_parallelized` version, its faster and more efficient

tmontaigu (changes requested):

tfhe/src/integer/server_key/radix/tests.rs
Tuditi: This panics for 1_MESSAGE_1_CARRY with the following message: `At least 4 bits of space (message + carry) are required to be able to do comparisons.` So I skipped that one.

tmontaigu:
Here also you have to do a `make fmt` and  fix any `make clippy_integer` errors

tmontaigu (approved):

PR #1810, opened 2024-11-22
chore: update dependencies

IceTDrinker:


IceTDrinker:
Waiting on a fix in wasm pack for the rayon bindgen stuff, maybe we can move ahead without it for now

soonum (approved):

Cargo.toml
```diff
@@ -19,13 +19,16 @@ exclude = [
...
 bytemuck = "1.14.3"
-dyn-stack = { version = "0.10", default-features = false }
+dyn-stack = { version = "0.11", default-features = false }
+itertools = "0.13"
 num-complex = "0.4"
-pulp = { version = "0.18.22", default-features = false }
+pulp = { version = "0.19.7", default-features = false }
+rand = "0.8"
+rayon = "1"
 serde = { version = "1.0", default-features = false }
```
mayeul-zama, line 30: Could also add here
- criterion (6 uses)
- clap (3 uses)
- bincode (4 uses)
- serde_json (5 uses)

Cargo.toml
```diff
@@ -19,13 +19,16 @@ exclude = [
...
 [workspace.dependencies]
-aligned-vec = { version = "0.5", default-features = false }
+aligned-vec = { version = "0.6", default-features = false }
 bytemuck = "1.14.3"
-dyn-stack = { version = "0.10", default-features = false }
+dyn-stack = { version = "0.11", default-features = false }
+itertools = "0.13"
 num-complex = "0.4"
-pulp = { version = "0.18.22", default-features = false }
+pulp = { version = "0.19.7", default-features = false }
```
mayeul-zama, line 27: Pulp 0.20 has been released for 2 hours

tfhe/Cargo.toml
```diff
@@ -33,7 +33,6 @@ serde_json = "1.0.94"
 clap = { version = "=4.4.4", features = ["derive"] }
 # Used in user documentation
 fs2 = { version = "0.4.3" }
-itertools = "0.11.0"
 statrs = "0.16"
```
mayeul-zama, line 36: statrs has a 0.17.1

Cargo.toml
IceTDrinker: ah

Cargo.toml
IceTDrinker: I focused on the deps, not dev deps for now

IceTDrinker:
I'll push what I have but the rustc-tools stuff for the tfhe-linter is not happy with modern toolchains @nsarlin-zama when you are done with ZK you could have a look to also update the toolchain in the linter, a few error pop up in rustc-tools, could be the occasion to propose an update over there

IceTDrinker:
also weird thing the tfhe lints fail but not the Makefile recipe

tfhe/src/lib.rs
```diff
@@ -71,7 +75,8 @@
 )]
 #![cfg_attr(all(doc, not(doctest)), feature(doc_auto_cfg))]
 #![cfg_attr(all(doc, not(doctest)), feature(doc_cfg))]
-#![warn(rustdoc::broken_intra_doc_links)]
+// Weird clippy lint triggering without any code location
+#![cfg_attr(test, allow(clippy::large_stack_arrays))]
```
mayeul-zama, line 79: Could have a common comment with large_stack_frames I think

tfhe/src/lib.rs
```diff
@@ -50,8 +56,6 @@
 #![allow(clippy::explicit_iter_loop)]
 // End allowed pedantic lints
 
-// Nursery lints
-#![warn(clippy::nursery)]
 // The following lints have been temporarily allowed
```
mayeul-zama, line 59: Why move?

I believe the following allowed lints are all nursery (or used to be)

mayeul-zama (commented):
Looks good to me!
Just a few comments

tfhe/src/lib.rs
IceTDrinker: I move because there were breaking changes in terms of which lint belong to which, so enable all warnings first, then re allow to make sure we don't have breaking stuff later

tfhe/src/lib.rs
IceTDrinker: could be,

mayeul-zama (approved):

PR #1818, opened 2024-11-25
function executor in string tests

mayeul-zama:


IceTDrinker:
cargo build is red and there is a conflict

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -542,83 +542,31 @@ impl<F> CpuFunctionExecutor<F> {
...
-        let sks = self.sks.as_ref().expect("setup was not properly called");
-        (self.func)(sks, input);
-    }
-}
+impl<T> NotTuple for &Vec<T> {}
 
-impl<'a, F> FunctionExecutor<&'a mut RadixCiphertext, RadixCiphertext> for CpuFunctionExecutor<F>
+/// For unary operations
+///
+/// Note, we need to `NotTuple` constraint to avoid conflicts with binary or ternary operations
```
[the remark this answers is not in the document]
tmontaigu:
```rust
impl<F, I1, O> FunctionExecutor<I1, O> for CpuFunctionExecutor<F>
where
    F: Fn(&ServerKey, I1) -> O,
    I1: NotTuple,
{
    fn execute(&mut self, input: I1) -> O {
        let sks = self.sks.as_ref().expect("setup was not properly called");
        (self.func)(sks, input)
    }
}
```
conflicts with
```rust
impl<F, I1, I2 O> FunctionExecutor<I1, O> for CpuFunctionExecutor<F>
where
    F: Fn(&ServerKey, I1) -> O,
    I1: NotTuple,
{
    fn execute(&mut self, input: (I1, I2)) -> O {
        let sks = self.sks.as_ref().expect("setup was not properly called");
        (self.func)(sks, input)
    }
}
```

Because the first case `input: I1` conflicts with `input: (I1, I2)`

tfhe/src/strings/test_functions/test_common.rs
```diff
@@ -1,350 +1,345 @@
...
-        PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64,
-        TestKind::Trivial,
-    );
+fn test_encrypt_decrypt_parametrized() {
+    test_encrypt_decrypt(PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64);
+}
+
+fn test_encrypt_decrypt<P>(param: P)
+where
+    P: Into<PBSParameters> + Clone,
```
tmontaigu, line 18: Why is clone needed ?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -542,83 +542,31 @@ impl<F> CpuFunctionExecutor<F> {
...
-/// would be possible tho.
-impl<'a, F> FunctionExecutor<&'a RadixCiphertext, (RadixCiphertext, BooleanBlock)>
-    for CpuFunctionExecutor<F>
-where
-    F: Fn(&ServerKey, &RadixCiphertext) -> (RadixCiphertext, BooleanBlock),
-{
-    fn setup(&mut self, _cks: &RadixClientKey, sks: Arc<ServerKey>) {
-        self.sks = Some(sks);
-    }
+pub(crate) trait NotTuple {}
```
tmontaigu, line 545: :ok_hand:

tmontaigu (commented):
I see some `P: Into<PBSParameters> + Clone` which surprisesme we need the clone


One minor thing that is not done on the integer tests, that slightly improves readability  is to give an explicit name to the executor, e.g:
```pub(crate) fn string_eq_ignore_case_test_impl<P, T>(
    param: P,
    mut eq_ingore_case_executor: T,
    f: for<'a> fn(&'a str, &'a str) -> bool,
) {
//...
}
```

tfhe/src/strings/test_functions/test_common.rs
mayeul-zama: Removed

tfhe/src/strings/test_functions/test_concat.rs
```diff
@@ -1,157 +1,146 @@
...
+    P: Into<PBSParameters>,
+{
+    let executor = CpuFunctionExecutor::new(&ServerKey::concat);
+    string_concat_test_impl(param, executor, |lhs, rhs| lhs.to_owned() + rhs);
+}
 
+pub(crate) fn string_concat_test_impl<P, T>(
+    param: P,
+    mut concat_executor: T,
+    clear_function: for<'a> fn(&'a str, &'a str) -> String,
```
tmontaigu, line 29: Actually, there is no need for the clear_function in any of the tests

The test_impl function is what does the test, and so it knows what the output is expected to be

For example,  If the caller of the test could give concat_executor and a clear_function that do the same things but are not concatenating

tmontaigu (changes requested):

tfhe/src/strings/test_functions/test_concat.rs
mayeul-zama: Many tests test more than one function

On the tests that only test one function, I could remove it.
But not sure it's better as there will be 2 kinds of test structures instead of one.
What do you think?

tmontaigu: Maybe you could do it like its done for some of the tests: the test functions receive multiple executor:

```rust
pub(crate) fn string_contains_test_impl<P, E1, E2, E3>(
    param: P,
    mut contains_executor: E1,
    mut starts_with_executor: E2,
    mut end_with_executor: E3,
) {
 //..
}
```
what do you think ?

tmontaigu (approved):

tmontaigu (approved):

PR #1822, opened 2024-11-25
string server key wrapper

mayeul-zama:


tfhe/src/strings/server_key/mod.rs
```diff
@@ -7,12 +7,39 @@ pub use trim::split_ascii_whitespace;
...
+use crate::integer::{BooleanBlock, RadixCiphertext, ServerKey as IntegerServerKey};
 use crate::strings::ciphertext::{num_ascii_blocks, FheAsciiChar, FheString};
 use crate::strings::N;
 use rayon::prelude::*;
+use std::borrow::Borrow;
 use std::cmp::Ordering;
 
+pub struct ServerKey<T>
+where
+    T: Borrow<IntegerServerKey> + Sync,
```
tmontaigu, line 19: Is the `Sync` bound really needed ?

as most if not all function would first call `let intsk = self.inner.borrow()` 
and the int sk is the one that needs to be Sync, and it is

mayeul-zama: `compare_shifted_index` needs it for example

We could borrow outside of the parallel iterator and rebuild a `strings::ServerKey<&ServerKey>` in the closure to be able to remove this bound.
Do you think it would be better this way?

tmontaigu: Hum I think I would remove the bound from the struct type, so that the Sync only appears where it is actually needed

mayeul-zama: Seems simpler to me to have a single global constraint even if this is a bit more restrictive

tmontaigu: Yeah we can do that

tmontaigu:
Since the string function are tied to a different key than integer key, should we remove the 'string' prefix from functions found in comp.rs (string_le, string_eq, etc) ?

IceTDrinker:
> Since the string function are tied to a different key than integer key, should we remove the 'string' prefix from functions found in comp.rs (string_le, string_eq, etc) ?

Most likely_yes

tmontaigu (approved):

PR #1826, opened 2024-11-26
feat: centralize encoding in shortint

tmontaigu:
The plaintext encoding in shortint was duplicated all over the code

This commit centralize the encoding used for shortint, so that if an encoding fix is needed there should be one place to do it.

### DIFF
--- a/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs
@@ -1,3 +1,4 @@
+use crate::core_crypto::entities::Cleartext;
 use crate::core_crypto::gpu::algorithms::{
     cuda_lwe_ciphertext_negate_assign, cuda_lwe_ciphertext_plaintext_add_assign,
 };
@@ -78,8 +79,8 @@ impl CudaServerKey {
         let ct_blocks = ct.as_ref().d_blocks.lwe_ciphertext_count().0;
 
         let scalar = self.message_modulus.0 as u8 - 1;
-        let delta = (1_u64 << 63) / (self.message_modulus.0 * self.carry_modulus.0);
-        let shift_plaintext = u64::from(scalar) * delta;
+
+        let shift_plaintext = self.encoding().encode(Cleartext(u64::from(scalar))).0;
 
         let scalar_vector = vec![shift_plaintext; ct_blocks];
         let mut d_decomposed_scalar =
--- a/tfhe/src/integer/gpu/server_key/radix/mod.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/mod.rs
@@ -1,4 +1,4 @@
-use crate::core_crypto::entities::{GlweCiphertext, LweCiphertextList};
+use crate::core_crypto::entities::{Cleartext, GlweCiphertext, LweCiphertextList};
 use crate::core_crypto::gpu::lwe_ciphertext_list::CudaLweCiphertextList;
 use crate::core_crypto::gpu::vec::CudaVec;
 use crate::core_crypto::gpu::{CudaLweList, CudaStreams};
@@ -26,7 +26,7 @@ use crate::shortint::engine::{
 use crate::shortint::server_key::{
     BivariateLookupTableOwned, LookupTableOwned, ManyLookupTableOwned,
 };
-use crate::shortint::PBSOrder;
+use crate::shortint::{PBSOrder, PaddingBit, ShortintEncoding};
 
 mod abs;
 mod add;
@@ -154,6 +154,15 @@ impl CudaServerKey {
         res
     }
 
+    pub(crate) fn encoding(&self) -> ShortintEncoding {
+        ShortintEncoding {
+            ciphertext_modulus: self.ciphertext_modulus,
+            message_modulus: self.message_modulus,
+            carry_modulus: self.carry_modulus,
+            padding_bit: PaddingBit::Yes,
+        }
+    }
+
     /// # Safety
     ///
     /// - `streams` __must__ be synchronized to guarantee computation has finished, and inputs must
@@ -173,8 +182,6 @@ impl CudaServerKey {
             PBSOrder::BootstrapKeyswitch => self.key_switching_key.output_key_lwe_size(),
         };
 
-        let delta = (1_u64 << 63) / (self.message_modulus.0 * self.carry_modulus.0);
-
         let decomposer = BlockDecomposer::new(scalar, self.message_modulus.0.ilog2())
             .iter_as::<u64>()
             .chain(std::iter::repeat(0))
@@ -187,7 +194,7 @@ impl CudaServerKey {
         );
         let mut info = Vec::with_capacity(num_blocks);
         for (block_value, mut lwe) in decomposer.zip(cpu_lwe_list.iter_mut()) {
-            *lwe.get_mut_body().data = block_value * delta;
+            *lwe.get_mut_body().data = self.encoding().encode(Cleartext(block_value)).0;
             info.push(CudaBlockInfo {
                 degree: Degree::new(block_value),
                 message_modulus: self.message_modulus,
--- a/tfhe/src/shortint/ciphertext/standard.rs
+++ b/tfhe/src/shortint/ciphertext/standard.rs
@@ -6,7 +6,7 @@ use crate::core_crypto::entities::*;
 use crate::core_crypto::prelude::{allocate_and_trivially_encrypt_new_lwe_ciphertext, LweSize};
 use crate::shortint::backward_compatibility::ciphertext::CiphertextVersions;
 use crate::shortint::parameters::{CarryModulus, MessageModulus};
-use crate::shortint::CiphertextModulus;
+use crate::shortint::{CiphertextModulus, PaddingBit, ShortintEncoding};
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
 use tfhe_versionable::Versionize;
@@ -199,6 +199,15 @@ impl Ciphertext {
             .map(|x| x % self.message_modulus.0)
     }
 
+    pub(crate) fn encoding(&self, padding_bit: PaddingBit) -> ShortintEncoding {
+        ShortintEncoding {
+            ciphertext_modulus: self.ct.ciphertext_modulus(),
+            message_modulus: self.message_modulus,
+            carry_modulus: self.carry_modulus,
+            padding_bit,
+        }
+    }
+
     /// See [Self::decrypt_trivial].
     /// # Example
     ///
@@ -225,8 +234,11 @@ impl Ciphertext {
     /// ```
     pub fn decrypt_trivial_message_and_carry(&self) -> Result<u64, NotTrivialCiphertextError> {
         if self.is_trivial() {
-            let delta = (1u64 << 63) / (self.message_modulus.0 * self.carry_modulus.0);
-            Ok(self.ct.get_body().data / delta)
+            let decoded = self
+                .encoding(PaddingBit::Yes)
+                .decode(Plaintext(*self.ct.get_body().data))
+                .0;
+            Ok(decoded)
         } else {
             Err(NotTrivialCiphertextError)
         }
@@ -234,23 +246,25 @@ impl Ciphertext {
 }
 
 pub(crate) fn unchecked_create_trivial_with_lwe_size(
-    value: u64,
+    value: Cleartext<u64>,
     lwe_size: LweSize,
     message_modulus: MessageModulus,
     carry_modulus: CarryModulus,
     pbs_order: PBSOrder,
     ciphertext_modulus: CiphertextModulus,
 ) -> Ciphertext {
-    let delta = (1_u64 << 63) / (message_modulus.0 * carry_modulus.0);
-
-    let shifted_value = value * delta;
-
-    let encoded = Plaintext(shifted_value);
+    let encoded = ShortintEncoding {
+        ciphertext_modulus,
+        message_modulus,
+        carry_modulus,
+        padding_bit: PaddingBit::Yes,
+    }
+    .encode(value);
 
     let ct =
         allocate_and_trivially_encrypt_new_lwe_ciphertext(lwe_size, encoded, ciphertext_modulus);
 
-    let degree = Degree::new(value);
+    let degree = Degree::new(value.0);
 
     Ciphertext::new(
         ct,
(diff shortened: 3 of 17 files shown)


tfhe/src/integer/gpu/server_key/radix/mod.rs
```diff
@@ -147,6 +147,15 @@ impl CudaServerKey {
         res
     }
 
+    pub(crate) fn encoding(&self, padding_bit: PaddingBit) -> ShortintEncoding {
+        ShortintEncoding {
+            ciphertext_modulus: self.ciphertext_modulus,
+            message_modulus: self.message_modulus,
+            carry_modulus: self.carry_modulus,
+            padding_bit,
```
mayeul-zama, line 155: Shouldn't we hardcode `PaddingBit::Yes` here
Or otherwise store it in `CudaServerKey`

tfhe/src/shortint/ciphertext/standard.rs
```diff
@@ -225,8 +225,16 @@ impl Ciphertext {
...
         if self.is_trivial() {
-            let delta = (1u64 << 63) / (self.message_modulus.0 * self.carry_modulus.0);
-            Ok(self.ct.get_body().data / delta)
+            // Value of the shift we multiply our messages by
+            let decoded = ShortintEncoding {
+                ciphertext_modulus: self.ct.ciphertext_modulus(),
+                message_modulus: self.message_modulus,
+                carry_modulus: self.carry_modulus,
+                padding_bit: PaddingBit::Yes,
+            }
```
mayeul-zama, line 234: We could add a method `encoding` to `Ciphertext`

tfhe/src/shortint/ciphertext/standard.rs
```diff
@@ -241,11 +249,13 @@ pub(crate) fn unchecked_create_trivial_with_lwe_size(
     pbs_order: PBSOrder,
     ciphertext_modulus: CiphertextModulus,
 ) -> Ciphertext {
```
mayeul-zama, line 251: This function could take an `ShortintEncoding` as input
I don't know this is a better though

tfhe/src/shortint/engine/mod.rs
```diff
@@ -92,6 +92,13 @@ where
     assert_eq!(accumulator.polynomial_size(), polynomial_size);
     assert_eq!(accumulator.glwe_size(), glwe_size);
```
mayeul-zama, line 94: This function could take an ShortintEncoding as input
I don't know this is a better though

tfhe/src/shortint/engine/mod.rs
```diff
@@ -175,6 +179,14 @@ where
...
 
+    // Value of the shift we multiply our messages by
```
mayeul-zama, line 182: Comment to remove

tfhe/src/shortint/mod.rs
```diff
@@ -135,3 +136,87 @@ where
...
+
+        // The bit before the message
+        let rounding_bit = delta >> 1;
+
+        // Compute the rounding bit
+        let rounding = (value.0 & rounding_bit) << 1;
+
+        Cleartext(value.0.wrapping_add(rounding) / delta)
+    }
+}
```
mayeul-zama, line 222: Could be in an `encoding` module

tfhe/src/shortint/public_key/compact.rs
```diff
@@ -148,15 +147,15 @@ fn to_plaintext_iterator(
         "Encryption modulus cannot exceed the plaintext modulus"
     );
 
+    let encoding = ShortintEncoding {
+        ciphertext_modulus: parameters.ciphertext_modulus,
+        message_modulus: parameters.message_modulus,
+        carry_modulus: parameters.carry_modulus,
+        padding_bit: PaddingBit::Yes,
+    };
```
mayeul-zama, line 155: Could be a method of `CompactPublicKeyEncryptionParameters`

tfhe/src/shortint/public_key/compact.rs
```diff
@@ -376,8 +371,14 @@ impl CompactPublicKey {
...
         let plaintext_modulus = self.parameters.message_modulus.0 * self.parameters.carry_modulus.0;
-        let delta = (1u64 << 63) / plaintext_modulus;
         assert!(encryption_modulus <= plaintext_modulus);
+        let delta = ShortintEncoding {
+            ciphertext_modulus: self.parameters.ciphertext_modulus,
+            message_modulus: self.parameters.message_modulus,
+            carry_modulus: self.parameters.carry_modulus,
+            padding_bit: PaddingBit::Yes,
+        }
+        .delta();
```
mayeul-zama, line 381: Could be a method of the type of `self.parameters`

tfhe/src/shortint/mod.rs
```diff
@@ -135,3 +136,87 @@ where
...
+            self.padding_bit,
+        )
+    }
+}
+
+impl ShortintEncoding {
+    pub(crate) fn from_parameters(
+        params: impl Into<ShortintParameterSet>,
+        padding_bit: PaddingBit,
+    ) -> Self {
```
mayeul-zama, line 188: When using the classical or multibit PBS, I think we should force the `PaddingBit` to be set.
But it looks like it would need bigger changes

Maybe we could think of removing no padding bit encoding and wop from shortint

mayeul-zama (commented):
Thanks for the refactor, it's much cleaner this way!

I left a few comments

tfhe/src/shortint/mod.rs
tmontaigu: I agree

tfhe/src/shortint/mod.rs
tmontaigu: Yeah I also think the no padding bit 'support' should go away  as we don't really test it, its not something we use either

tfhe/src/shortint/ciphertext/standard.rs
tmontaigu: I don't think it's better as to create the ciphertext, the metadata

tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs
```diff
@@ -9,6 +10,7 @@ use crate::integer::gpu::server_key::CudaBootstrappingKey;
 use crate::integer::gpu::{
     unchecked_bitop_integer_radix_kb_assign_async, BitOpType, CudaServerKey, PBSType,
 };
+use crate::shortint::PaddingBit;
```
IceTDrinker, line 13: what's the current thinking on the PaddingBit thing ?

tfhe/src/integer/gpu/server_key/radix/mod.rs
IceTDrinker: I would tend to agree

tfhe/src/shortint/ciphertext/standard.rs
IceTDrinker: worst case we can have a create_trivial_with_encoding and rewire some functions, but if it's not required currently not worth adding IMO

IceTDrinker (commented):
The current management is fine with me as it's pub(crate) and therefore does not leak out of the crate, agree with the fact some no padding stuff should be removed if we could... but currently we can't because of the freaking wopbs

Some places could have the encoding hardcoded (for the Cuda part notably I guess) and maybe for some integer cpu stuff as well

tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs
tmontaigu: Hum, the main reeason it exists is because some things allow to encrypt without padding bit, even in 99% of the code/ usage only work with a padding bit

IceTDrinker: yep to keep

tfhe/src/shortint/client_key/mod.rs
```diff
@@ -494,16 +494,9 @@ impl ClientKey {
     pub fn decrypt_message_and_carry(&self, ct: &Ciphertext) -> u64 {
         let decrypted_u64: u64 = self.decrypt_no_decode(ct);
```
mayeul-zama, line 495: `decrypt_no_decode` could return a `Plaintext`

tfhe/src/shortint/engine/client_side.rs
```diff
@@ -65,16 +66,11 @@ impl ShortintEngine {
         u64: RandomGenerable<NoiseDistribution, CustomModulus = u64>,
         KeyCont: crate::core_crypto::commons::traits::Container<Element = u64>,
     {
-        //The delta is the one defined by the parameters
-        let delta = (1_u64 << 63)
-            / (client_key_parameters.message_modulus().0 * client_key_parameters.carry_modulus().0);
-
-        //The input is reduced modulus the message_modulus
+        // The input is reduced modulus the message_modulus
```
mayeul-zama, line 69: This comment is not really useful and could be removed IMO
Applies to many places

tfhe/src/shortint/server_key/mod.rs
```diff
@@ -1164,11 +1171,9 @@ impl ServerKey {
...
         let modular_value = value % self.message_modulus.0;
 
-        let delta = (1_u64 << 63) / (self.message_modulus.0 * self.carry_modulus.0);
-
-        let shifted_value = modular_value * delta;
-
-        let encoded = Plaintext(shifted_value);
+        let encoded = self
+            .encoding(PaddingBit::Yes)
+            .encode(Cleartext(modular_value));
```
mayeul-zama, line 1176: Style suggestion: move `Cleartext` wrapping in modular_value definition
Applies to many places (some being with `Plaintext` wrapping before decoding instead)

tfhe/src/shortint/client_key/mod.rs
tmontaigu: Yep makes sense

tfhe/src/shortint/public_key/compact.rs
tmontaigu: done

tfhe/src/shortint/public_key/compact.rs
tmontaigu: done

tfhe/src/shortint/server_key/mod.rs
tmontaigu: It does apply for some cases but for others not that much

mayeul-zama (commented):
Thanks, I just have 2 questions remaining 
And at least the 2nd commit should be squashed (the 3rd could also, or be renamed)

tfhe/src/shortint/mod.rs
```diff
@@ -50,6 +50,7 @@
 pub mod backward_compatibility;
 pub mod ciphertext;
 pub mod client_key;
+pub(crate) mod encoding;
```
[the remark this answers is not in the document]
IceTDrinker: for now this is fine, avoids leaking too much outside and allows us to keep evolving the design if needed

tfhe/src/shortint/public_key/compact.rs
```diff
@@ -469,6 +461,11 @@ impl CompactPublicKey {
     pub fn parameters(&self) -> CompactPublicKeyEncryptionParameters {
         self.parameters
     }
+
+    #[cfg(feature = "zk-pok")]
```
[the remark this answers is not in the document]
tmontaigu: Its only used in encrypt_and_prove

mayeul-zama (approved):
Thanks!

tmontaigu:
Added an encrypt/decrypt test with a non native (but pow2) modulus, and fixed an assert un decode that was too strict

mayeul-zama (approved):
Thanks a lot!

IceTDrinker:
missing something to merge ?

tmontaigu:
Yes needs a small bug fix

tmontaigu:
Bug has been fixed and the test updated to catch it

IceTDrinker:
> Bug has been fixed and the test updated to catch it

Could you point out what it was and where we can see the fix/test, with the rebase the diff view of github is not helpful

tmontaigu:
The bug was an extra `* 2` in the delta for custom modulus, very likely due to copy paste from native

In the test  this was added
```
        let encoded = encoding.encode(Cleartext(m));
        assert_eq!(encoded.0 / CIPHERTEXT_MODULUS, 0, "padding bit is not null");
        assert!(
            encoded.0 < (CIPHERTEXT_MODULUS / 2),
            "encoded message goes beyond its allowed space"
        );
```

also these were added when decoding
```
       // Force the decoded value to be in the correct range
        Cleartext((value.0.wrapping_add(rounding) / delta) % (self.plaintext_space()))
  fn plaintext_space(&self) -> u64 {
        self.message_modulus.0
            * self.carry_modulus.0
            * if self.padding_bit == PaddingBit::No {
                1
            } else {
                2
            }
    }
```

tfhe/src/shortint/encoding.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    let mut params = V0_10_PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M64;
+    params.carry_modulus = CarryModulus(1);
+    params.ciphertext_modulus = CiphertextModulus::new(CIPHERTEXT_MODULUS as u128);
+
+    let encoding = ShortintEncoding::from_parameters(params, PaddingBit::Yes);
+    let (cks, _sks) = crate::shortint::gen_keys(params);
+    for m in 0..params.message_modulus.0 {
+        let encoded = encoding.encode(Cleartext(m));
+        assert_eq!(encoded.0 / CIPHERTEXT_MODULUS, 0, "padding bit is not null");
```
IceTDrinker, line 110: I think you want to divide by CIPHERTEXT_MODULUS /2 ?

IceTDrinker: that would be consistent with the < CIPHERTEXT_MODULUS / 2 below ?

tmontaigu: Yes that is what I wanted to do

IceTDrinker: re thinkg about this, isn't the check encoded < CIPHERTEXT_MODULUS /2 enough ? it guarantees the padding bit and that all the other bits above are correct ?

i.e. the check here is redundant ?

tmontaigu: yes, removed

IceTDrinker (approved):

PR #1828, opened 2024-11-27
chore(ci): add debug mode to shortint test script for easier debug later

IceTDrinker:
- only warn if no tests are run with the big parameters filter

### DIFF
--- a/scripts/shortint-tests.sh
+++ b/scripts/shortint-tests.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
 function usage() {
     echo "$0: shortint test runner"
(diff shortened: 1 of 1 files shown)


scripts/shortint-tests.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
```
agnesLeroy, line 3: What does this do?

scripts/shortint-tests.sh
```diff
@@ -111,15 +111,16 @@ and not test(~smart_add_and_mul)"""
         --cargo-profile "${cargo_profile}" \
         --package "${tfhe_package}" \
         --profile ci \
-        --features="${ARCH_FEATURE}",shortint,internal-keycache,experimental \
+        --features="${ARCH_FEATURE}",shortint,internal-keycache,zk-pok,experimental \
```
agnesLeroy, line 114: Why activate the zk-pok feature?

agnesLeroy (approved):
Looks good, two small questions :)

scripts/shortint-tests.sh
IceTDrinker: prints all the steps of the bash script

scripts/shortint-tests.sh
IceTDrinker: because it's enabled above and without it we are recompiling for no good reason

PR #1831, opened 2024-11-28
fix: clean noise of skipped block in full_propagate_parallelized

tmontaigu:
In full_propagate_parallelized we find the first block which has a degree >= msg_mod, meaning it has a carry and start propagating from there.

However, while the preceding blocks may have no carry, their noise level may not be nominal, and so to leave the radix in a consistent state and clean state, full_propagate now also clean the noise for blocks that are not propagated

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/mod.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/mod.rs
@@ -39,6 +39,8 @@ mod vector_find;
 
 use super::ServerKey;
 use crate::integer::ciphertext::IntegerRadixCiphertext;
+use crate::integer::RadixCiphertext;
+use crate::shortint::ciphertext::{Ciphertext, NoiseLevel};
 pub(crate) use add::OutputFlag;
 use rayon::prelude::*;
 pub use scalar_div_mod::{MiniUnsignedInteger, Reciprocable};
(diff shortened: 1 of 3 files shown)


tfhe/src/integer/server_key/radix_parallel/mod.rs
```diff
@@ -140,45 +142,51 @@ impl ServerKey {
...
+                let (carry, message) = rayon::join(
+                    || self.key.carry_extract(&blocks[i]),
+                    || self.key.message_extract(&blocks[i]),
+                );
+
+                blocks[i] = message;
+
+                //add the carry to the next block
+                if i < blocks.len() - 1 {
+                    self.key.unchecked_add_assign(&mut blocks[i + 1], &carry);
```
IceTDrinker, line 188: we can get out of propagation with non nominal noise ?

mayeul-zama (commented):
Looks good to me, thanks for the fix!

I think the commit could be split in 2: refactor and fix

mayeul-zama:
Does the GPU implementation have the same bug?
@agnesLeroy

tfhe/src/integer/server_key/radix_parallel/tests_cases_unsigned.rs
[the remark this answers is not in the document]
agnesLeroy: Actually `tests_cases_unsigned` is a leftover, it should be fully refactored to match the new organization of tests.

tfhe/src/integer/server_key/radix_parallel/mod.rs
tmontaigu: Yes there should be no problem:
For blocks in range 0..len we extract both the msg and carry, and if the block is not the last, we add the carry to the next block, which will then clean itself when its going to extract its msg and carry

IceTDrinker: indeed, misread the loop

IceTDrinker (commented):
Good for me I would say, though maybe a conversion to function executor is needed for GPU I believe (?)

IceTDrinker (approved):
Approving to be able to move forward, function executor can be a second PR

PR #1832, opened 2024-11-28
chore(gpu): replace inneficient log2 implementation and use it everyw…

pdroalves:
…here

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/polynomial/parameters.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/polynomial/parameters.cuh
@@ -1,7 +1,24 @@
 #ifndef CUDA_PARAMETERS_CUH
 #define CUDA_PARAMETERS_CUH
 
-constexpr int log2(int n) { return (n <= 2) ? 1 : 1 + log2(n / 2); }
+#include "device.h"
+#include <cstdint>
+
+// If decide to support something else than 32 and 64 bits, this method will
+// need to be adjusted
+template <typename T> constexpr unsigned log2_int(T n) {
+  if (n == 0) {
+    PANIC("Cuda error (log2): log2 is undefined for 0");
+  }
+
+  if constexpr (sizeof(T) == 4) { // uint32_t
+    return (unsigned)(8 * sizeof(uint32_t) - __builtin_clz(n) - 1);
+  } else if constexpr (sizeof(T) == 8) { // uint64_t
+    return (unsigned)(8 * sizeof(uint64_t) - __builtin_clzll(n) - 1);
+  } else {
+    return (n <= 2) ? 1 : 1 + log2_int(n / 2);
+  }
+}
 
 constexpr int choose_opt_amortized(int degree) {
   if (degree <= 1024)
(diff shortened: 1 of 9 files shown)


backends/tfhe-cuda-backend/cuda/src/polynomial/parameters.cuh
```diff
@@ -1,7 +1,7 @@
 #ifndef CUDA_PARAMETERS_CUH
 #define CUDA_PARAMETERS_CUH
 
-constexpr int log2(int n) { return (n <= 2) ? 1 : 1 + log2(n / 2); }
+#define log2_int(X) ((unsigned) (8*sizeof (unsigned long long) - __builtin_clzll((X)) - 1))
```
IceTDrinker, line 4: Only works for 8 bytes integers, could be a templated function

pdroalves: Probably 32 and 64 bits are enough for now.

backends/tfhe-cuda-backend/cuda/src/polynomial/parameters.cuh
```diff
@@ -1,7 +1,17 @@
...
 
-constexpr int log2(int n) { return (n <= 2) ? 1 : 1 + log2(n / 2); }
+#include <cstdint>
+
+// If decide to support something else than 32 and 64 bits, this method will
+// need to be adjusted
+template <typename T> constexpr unsigned log2_int(T x) {
+  if constexpr (sizeof(T) <= sizeof(uint32_t)) {
+    return (unsigned)(8 * sizeof(uint32_t) - __builtin_clz(x) - 1);
+  } else if constexpr (sizeof(T) <= sizeof(uint64_t)) {
```
IceTDrinker, line 11: Why not take strict equality (as a uint16_t would not get the right value I think) and the last else branch have a const assert with an error message ?

backends/tfhe-cuda-backend/cuda/src/polynomial/parameters.cuh
```diff
@@ -1,7 +1,17 @@
 #ifndef CUDA_PARAMETERS_CUH
 #define CUDA_PARAMETERS_CUH
 
-constexpr int log2(int n) { return (n <= 2) ? 1 : 1 + log2(n / 2); }
+#include <cstdint>
+
+// If decide to support something else than 32 and 64 bits, this method will
+// need to be adjusted
```
agnesLeroy, line 7: We'll need u128 for 128bit PBS actually

IceTDrinker: would likely need to look different if you don't have a native u128, e.g. having two u64 so requires a somwhat different logic

pdroalves: I don't think we have a clz function that nativelly supports uint8_t, uint16_t, or uint128_t. However, I couldn't find a single use currently for these sizes  in the codebase (although we will need uint128_t soon). One idea is to fallback to the original formula in these cases, and keep 32 and 64 bits as special cases (I just pushed a new commit doing that), or use this:

```
    if constexpr (sizeof(T) == 1) { // uint8_t
        return (unsigned)(8 * sizeof(uint8_t) - __builtin_clz(n) - 1 + 24);
    }else if constexpr (sizeof(T) == 2) { // uint16_t
        return (unsigned)(8 * sizeof(uint16_t) - __builtin_clz(n) - 1 + 16);
    }else if constexpr (sizeof(T) == 4) { // uint32_t
        return (unsigned)(8 * sizeof(uint32_t) - __builtin_clz(n) - 1);
    } else if constexpr (sizeof(T) == 8) { // uint64_t
        return (unsigned)(8 * sizeof(uint64_t) - __builtin_clzll(n) - 1);
    } else {
        return (n <= 2) ? 1 : 1 + log2_int(n / 2);
    }
```

By doing this we adjust the output  of  __builtin_clz to smaller sizes. The only issue is that I'm not confident this is future proof or plataform agnostic. Could be.

WDYT?

agnesLeroy (approved):

PR #1835, opened 2024-11-29
refactor(shortint): use view -> decompress to simplify key decompression

IceTDrinker:
- the decompress_into primitives consume the input entity, but we can use a view that won't consume the original owned key


mayeul-zama (approved):
Thanks!

PR #1837, opened 2024-11-29
chore(ci): run integer tests on approval only if file changed

soonum:


.github/workflows/aws_tfhe_integer_tests.yml
```diff
@@ -59,14 +59,15 @@ jobs:
...
+              - .github/workflows/aws_tfhe_integer_tests.yml
 
   setup-instance:
     name: Setup instance (unsigned-integer-tests)
     needs: should-run
     if:
       (github.event_name == 'push' && github.repository == 'zama-ai/tfhe-rs' && needs.should-run.outputs.integer_test == 'true') ||
       (github.event_name == 'schedule' && github.repository == 'zama-ai/tfhe-rs') ||
-      (github.event_name == 'pull_request' && contains(github.event.label.name, 'approved')) ||
+      (github.event_name == 'pull_request' && contains(github.event.label.name, 'approved') && needs.should-run.outputs.integer_test == 'true') ||
```
IceTDrinker, line 70: I see we check the should run on main, so we could keep it in the end, more test runs, plus we have new test checks, so maybe just update the should run thing ? and keep the old behavior for the pull request event (sorry I changed my mind in between)

soonum: Ok rolling back then.

IceTDrinker (approved):

PR #1846, opened 2024-12-04
Ns/plug zk v2

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/772
### PR content/description
Plugs ZK V2. The CompactPkeCrs type is now an enum with a variant for V1 and V2, each holding the right 'PublicParam' type from `tfhe_zk_pok`. The proof is similarly wrapped in an enum. A prove method has been added on the Crs object that will generate a proof compatible with this Crs. Similarly, a verify method will always reject proofs that are not of the right kind.

Backward compatibility is maintained and old crs/proofs are automatically wrapped in the V1 scheme. A fix was needed in versionable for this, to make it handle type that evolve from `#[repr(transparent)]` to their own type.

By default new Crs are created using v2, but methods have been added to create a Crs with v1.

Proof conformance checks the version of the scheme.

### Check-list:

* [x] Tests for the changes have been added -> Need to add a v2 proof in data repo
* [x] Param update


tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
```diff
@@ -6,13 +6,26 @@ use crate::shortint::parameters::{
     MessageModulus,
 };
 
+/// This parameter set should be used when doing zk proof of public key encryption
 pub const PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64: CompactPublicKeyEncryptionParameters =
     CompactPublicKeyEncryptionParameters {
-        encryption_lwe_dimension: LweDimension(1024),
-        encryption_noise_distribution: DynamicDistribution::new_t_uniform(42),
+        encryption_lwe_dimension: LweDimension(2048),
```
IceTDrinker, line 12: I'll check if the keyswitching parameters need an update as well

IceTDrinker: looks like the keyswitch to big key needs a small update, I'll give you the info to update and have legacy constants there as well

mayeul-zama (commented):
Just a few comments
FYI I only had a quick look at the "plug zk v2" commit

utils/tfhe-versionable/examples/transparent_then_not.rs
```diff
@@ -0,0 +1,162 @@
...
+
+// This type was transparent before, but it has now been transformed to a full type.
+#[derive(Versionize)]
+#[versionize(MyStructWrapperVersions)]
+enum MyStructWrapper<T> {
+    OldWrapper(MyStruct<T>),
+    NewWrapper(u64),
+}
+
+// We need to create a dispatch enum that starts at the last version that was transparent
```
tmontaigu, line 19: That sentence confuses me, to me, the last version that was transparent is V1, not V0, so if I follow the `that starts at the last version that was transparent` the enum should start at V1

utils/tfhe-versionable/examples/transparent_then_not.rs
```diff
@@ -0,0 +1,162 @@
+//! This example is similar to the "transparent" one, except that the wrapper type is transparent at
+//! a point in time, then converted into its own type that is not transparent.
+//!
+//! To make this work, we consider that the inner and the wrapper type share the same history up to
+//! the version where the transparent attribute has been removed.
+
+use std::convert::Infallible;
```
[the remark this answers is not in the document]
nsarlin-zama: Ok I will try to make the example more readable. I don't think this is just an ordering issue because I would like the order of the definitions the match the explanation in the comments

utils/tfhe-versionable/examples/transparent_then_not.rs
nsarlin-zama: Yeah I'm confused too, I will reword that

mayeul-zama (approved):
Thanks!

IceTDrinker:
please don't merge right away, I'd like to have a chance to discuss the keyswitching parameters before you do (and read the PR as well)

utils/tfhe-versionable/examples/transparent_then_not.rs
```diff
@@ -0,0 +1,173 @@
...
+// adding a new kind of metadata.
+#[derive(Versionize)]
+#[versionize(MyStructWrapperVersions)]
+struct MyStructWrapper<T> {
+    inner: MyStruct<T>,
+    count: u64,
+}
+
+// We need to create a dispatch enum that has the same history as the inner type until the point
+// where the wrapper is not transparent anymore.
```
tmontaigu, line 26: I think its better worded that way, but, wording now makes me understand that the enum should be
```rust
enum MyStructWrapperVersions<T> {
    V0(MyStructV0),
    V1(MyStrucV1<T>),
    V2(MyStructWrapper<T>),
}
```

as for V0 and V1 the Wrapper was transparent, and so the inner types were `MyStructV0` and `MyStrucV1`.

because as it stands,
```rust
enum MyStructWrapperVersions<T> {
    V0(MyStructWrapperV0),
    V1(MyStructWrapperV1<T>),
    V2(MyStructWrapper<T>),
}
```
Does not look different  from what is usally done for non transparent structs

nsarlin-zama: What I mean here is that in v0 and v1(versions of the app) the `MyStructWrapperVersions` enum does not exist, so we need to create it directly in v2 with at least the same versions as `MyStructVersions`.

Whereas with normal non transparent types you build your enum one version at a time, so you never create from scratch a dispatch enum with 3 variants

nsarlin-zama: the examples in this repo try to follow a pattern where "mod v0", "mod v1"... simulate previous versions of an application to show how the versioning evolve with time

nsarlin-zama: I reworded it a bit

tfhe/src/zk/mod.rs
```diff
@@ -0,0 +1,783 @@
...
+pub type SerializableCompactPkePublicParams =
+    tfhe_zk_pok::serialization::SerializablePKEv1PublicParams;
+
+impl Named for ZkCompactPkeV1PublicParams {
+    const NAME: &'static str = "zk::CompactPkePublicParams";
+}
+
+pub struct CompactPkeCrsConformanceParams {
+    lwe_dim: LweDimension,
+    max_num_message: usize,
```
IceTDrinker, line 72: new type ?

IceTDrinker (commented):
It seems we are missing a check on k_max <= d it seems, otherwise we are looking good, after the parameter stuff

tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
```diff
@@ -6,13 +6,26 @@ use crate::shortint::parameters::{
...
         message_modulus: MessageModulus(4),
         carry_modulus: CarryModulus(4),
         ciphertext_modulus: CiphertextModulus::new_native(),
         expansion_kind: CompactCiphertextListExpansionKind::RequiresCasting,
     }
     .validate();
+
+/// This legacy parameter set should be used with the v1 pke zk scheme
+pub const PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64_ZK_V1:
+    CompactPublicKeyEncryptionParameters = CompactPublicKeyEncryptionParameters {
```
IceTDrinker, line 23: did you do the keyswitch parameters update ?

tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
```diff
@@ -6,13 +6,26 @@ use crate::shortint::parameters::{
...
+        encryption_noise_distribution: DynamicDistribution::new_t_uniform(17),
         message_modulus: MessageModulus(4),
         carry_modulus: CarryModulus(4),
         ciphertext_modulus: CiphertextModulus::new_native(),
         expansion_kind: CompactCiphertextListExpansionKind::RequiresCasting,
     }
     .validate();
+
+/// This legacy parameter set should be used with the v1 pke zk scheme
+pub const PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64_ZK_V1:
```
IceTDrinker, line 22: also this does not respect the potential TFHE-rs version being in the name of consts

tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
nsarlin-zama: not yet !

tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
nsarlin-zama: I know, I did not reworked the params yet

PR #1849, opened 2024-12-05
fix(tfhe-lints): linter was not run, missing compile time env var

nsarlin-zama:
### PR content/description
Fix the custom linter that was not running anymore due to an error in the child compiler process that was not propagated. Failure in the child process will now make the ci fail.

This is linked to https://github.com/zama-ai/tfhe-rs/pull/1810 but we also need the PR in rustc-tools to be merged to fix the issue with the toolchain update

### DIFF
--- a/utils/cargo-tfhe-lints-inner/src/main.rs
+++ b/utils/cargo-tfhe-lints-inner/src/main.rs
@@ -27,6 +27,11 @@ fn main() {
         (tool_args.as_slice(), &[] as &[String])
     };
 
+    // The linter calls rustc without cargo, so these variables won't be set. Since we use them in
+    // our code, we need to set them to any value to avoid a compilation error.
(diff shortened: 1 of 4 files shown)


utils/cargo-tfhe-lints-inner/src/main.rs
```diff
@@ -27,6 +27,9 @@ fn main() {
         (tool_args.as_slice(), &[] as &[String])
     };
 
+    std::env::set_var("CARGO_PKG_VERSION_MAJOR", "X");
+    std::env::set_var("CARGO_PKG_VERSION_MINOR", "Y");
```
mayeul-zama, line 31: A comment would help understand why this is needed

nsarlin-zama: I added a small comment to explain

mayeul-zama (approved):
Thanks!

PR #1851, opened 2024-12-06
fix(shortint): add ciphertext_modulus_after_packing_ks to compression…

mayeul-zama:


tfhe/src/shortint/parameters/list_compression.rs
```diff
@@ -20,6 +21,7 @@ pub struct CompressionParameters {
     pub lwe_per_glwe: LweCiphertextCount,
     pub storage_log_modulus: CiphertextModulusLog,
     pub packing_ks_key_noise_distribution: DynamicDistribution<u64>,
+    pub ciphertext_modulus_after_packing_ks: CiphertextModulus,
```
IceTDrinker, line 24: need an upgrade function + backward data compatibility I would think ?

IceTDrinker:
I'm currently doubting whether we should merge this, or wait to have an actual use for it 🤔

IceTDrinker (commented):
As discussed let's at least unlock power of 2 moduli at the core level so that this change makes sense (sorry about that, wasn't clear at first to me)

PR #1866, opened 2024-12-13
chore: remove aliases for gaussian parameters for compact PK

IceTDrinker:
- we are TUniform by default so no more aliases for gaussian parameters


IceTDrinker:
@soonum and @nsarlin-zama as you are on the parameter stuff, those are old gaussian parameters whose aliases were not removed, this PR changes that

nsarlin-zama (approved):
looks good to me

PR #1869, opened 2024-12-13
chore(all): remove the dependency to lazy_static

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/247

### PR content/description
Use LazyLock instead of lazy_static. (Or nothing when the value is already const)


mayeul-zama (approved):
Thanks!

PR #1870, opened 2024-12-13
Am/refactor/factorize expansion code

IceTDrinker:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/787

Adds sanitization for all expansion code paths

Factorize in a single function to avoid issues later, managed to make it work without it being too weird @nsarlin-zama 

Maybe traits would be better, this works, I let you tweak if needed


nsarlin-zama:
Yes maybe this could use the Expandable trait ? 
And maybe have another parameter or encode in the ExpansionMode if the verification should be done or not ?

IceTDrinker:
This was not the goal of the PR, I let you do that on top/after if you want to

mayeul-zama (approved):
Looks good to me!
Not sure if some parts needs a closer look

IceTDrinker:
just rebased to avoid the branch getting out of sync

IceTDrinker:
I need some time to go over this change again and see if it's better this way (is certainly less error prone when a change needs to be applied for all lists)

PR #1882, opened 2024-12-16
feat(hlapi): add strings

tmontaigu:


tfhe/src/high_level_api/strings/ascii/no_pattern.rs
```diff
@@ -0,0 +1,331 @@
...
+use crate::high_level_api::strings::ascii::FheAsciiString;
+use crate::prelude::FheStringRepeat;
+use crate::strings::ciphertext::UIntArg;
+use crate::strings::client_key::EncU16;
+use crate::{FheBool, Tag};
+
+pub enum FheStringLen {
+    NoPadding(u16),
+    Padding(FheUint16),
+}
```
nsarlin-zama, line 13: Couldn't we use a trivial ciphertext here in the NoPadding case instead of having an enum ? Maybe it would make it easier to write generic code without having to match everywhere ?

tfhe/src/high_level_api/strings/ascii/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+            inner: inner.into(),
+            tag,
+        }
+    }
+
+    pub fn try_encrypt_with_padding(
+        str: impl AsRef<str>,
+        padding: u32,
+        client_key: &ClientKey,
+    ) -> crate::Result<Self> {
```
nsarlin-zama, line 74: Maybe it would be easier for the users to give the total size instead of the padding count here, so they don't have to compute it from the string size

tfhe/src/high_level_api/strings/ascii/strip.rs
```diff
@@ -0,0 +1,194 @@
+use crate::high_level_api::global_state::with_internal_keys;
+use crate::high_level_api::keys::InternalServerKey;
+use crate::high_level_api::strings::ascii::FheAsciiString;
+use crate::high_level_api::strings::traits::FheStringStrip;
+use crate::high_level_api::FheBool;
+use crate::strings::ciphertext::ClearString;
+
+impl FheStringStrip<&Self> for FheAsciiString {
+    /// If the pattern does match the start of the string, returns a new encrypted string
+    /// with the specified pattern from the start, and boolean set to `true`,
```
nsarlin-zama, line 10: missing a word ?

tfhe/src/high_level_api/strings/ascii/strip.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            #[cfg(feature = "gpu")]
+            InternalServerKey::Cuda(_) => {
+                panic!("gpu does not support strip_prefix");
+            }
+        })
+    }
+
+    /// If the pattern does match the end of the string, returns a new encrypted string
+    /// with the specified pattern from the end, and boolean set to `true`,
```
nsarlin-zama, line 56: missing "stripped" ?

tfhe/src/high_level_api/strings/ascii/strip.rs
```diff
@@ -0,0 +1,194 @@
...
+            InternalServerKey::Cuda(_) => {
+                panic!("gpu does not support strip_suffix");
+            }
+        })
+    }
+}
+
+impl FheStringStrip<&ClearString> for FheAsciiString {
+    /// If the pattern does match the start of the string, returns a new encrypted string
+    /// with the specified pattern from the start, and boolean set to `true`,
```
nsarlin-zama, line 104: idem

tfhe/src/high_level_api/strings/ascii/strip.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            #[cfg(feature = "gpu")]
+            InternalServerKey::Cuda(_) => {
+                panic!("gpu does not support strip_prefix");
+            }
+        })
+    }
+
+    /// If the pattern does match the end of the string, returns a new encrypted string
+    /// with the specified pattern from the end, and boolean set to `true`,
```
nsarlin-zama, line 150: idem

tfhe/src/high_level_api/strings/tests/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        assert_eq!(index as usize, clear_string.find("dog").unwrap());
+
+        let (index, found) = string1.rfind(&ClearString::new("dog".into()));
+        assert!(found.decrypt(client_key));
+        let index: u32 = index.decrypt(client_key);
+        assert_eq!(index as usize, clear_string.rfind("dog").unwrap());
+    }
+}
+
+fn test_string_len_is_empty(client_key: &ClientKey) {
```
nsarlin-zama, line 57: Maybe have a test where the string is actually empty ?

tfhe/src/high_level_api/strings/ascii/no_pattern.rs
tmontaigu: Its the design used in integers

I depends on how explicit we want the thing to be, if we always return a FheUint16 then we lose the fact that the len can be trivially known appears with an explicit type to the user.

But maybe it would indeed be better as I think most of the time the lengh is going to be non-trivial et even if you start with a string wihout padding, some ops will cause them to have padding

tfhe/src/high_level_api/strings/ascii/mod.rs
tmontaigu: we could add a `try_encrypt_with_fixed_sized` which encrypts and padds (or trims) so that the str has the given size

nsarlin-zama: yes good idea !

tfhe/src/high_level_api/strings/ascii/no_pattern.rs
nsarlin-zama: Yes indeed, it depends on how people plan to use it

tmontaigu: @IceTDrinker @mayeul-zama If you have a opinion on whether the len/is_empty functions should return an enum result (clear/encrypted) or always return a FheUint but when the len was known in clear the FheUint is trivial, (meaning user would need to use the `is_trivial` fn and `decrypt_trivial()` to get the clear value)

IceTDrinker: I would think an enum is maybe better for end users ? Avoids confusion about trivial encryptions

mayeul-zama: The user may not know that it can be trivial
Having an enum seems more explicit
The enum could have a function `to_ct` to support the case where we don't want to match

tfhe/src/high_level_api/strings/tests/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    let string = FheAsciiString::try_encrypt_with_padding("", 10, client_key).unwrap();
+    match string.len() {
+        FheStringLen::NoPadding(len) => assert_eq!(len, 0u16),
+        FheStringLen::Padding(_) => {
+            panic!("Unexpected result");
+        }
+    }
```
nsarlin-zama, line 117: In this case, shouldn't it fall in the "Padding(_)" branch ?

tfhe/src/high_level_api/strings/ascii/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+                Ok(unversioned)
+            }
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Versionize, Clone)]
+#[versionize(FheAsciiStringVersions)]
+pub struct FheAsciiString {
+    pub(crate) inner: AsciiDevice,
```
IceTDrinker, line 124: odd choice for the inner type name ?

tfhe/src/strings/ciphertext.rs
```diff
@@ -5,20 +5,25 @@ use crate::integer::{
...
+#[versionize(FheAsciiCharVersions)]
 pub struct FheAsciiChar {
     pub enc_char: RadixCiphertext,
 }
 
 /// Represents a encrypted string made up of [`FheAsciiChar`]s.
-#[derive(Clone)]
+#[derive(Clone, Serialize, Deserialize, Versionize)]
+#[versionize(FheStringVersions)]
 pub struct FheString {
```
IceTDrinker, line 27: naming mismatch between HL and this module "on purpose" ?

tfhe/src/high_level_api/strings/tests/mod.rs
IceTDrinker: agree likely a copy paste error

tmontaigu: yes

tfhe/src/high_level_api/strings/ascii/mod.rs
tmontaigu: Yes, I was not sure what name to give, maybe `InnerString` might be better

tfhe/src/strings/ciphertext.rs
tmontaigu: Yeah, I'm not 100% attached to this name, but due to how ascii is limited, I felt like making it clear in the name might avoid some disapointments

IceTDrinker: Fair enough

tfhe/src/high_level_api/strings/ascii/mod.rs
IceTDrinker: this is fully pub(crate) ?

IceTDrinker (approved):

tfhe/src/high_level_api/strings/ascii/mod.rs
tmontaigu: Yes

IceTDrinker: ok then let's get this ready for release and update naming later

IceTDrinker:
> Test should have failed, the make recipe for high level tests is missing the strings feature

then feel free to rebase and update the make target

nsarlin-zama (approved):

PR #1886, opened 2024-12-17
chore(shortint): new parameters

nsarlin-zama:


soonum:
Should we set aliases for Gaussian common parameters (like 2_2) just like the Tuniform flavour ?  @IceTDrinker

soonum (approved):
Incredible and tedious job, thanks a lot @nsarlin-zama  !

tfhe/src/shortint/parameters/compact_public_key_only/p_fail_2_minus_64/ks_pbs.rs
```diff
@@ -18,8 +18,35 @@ pub const V0_11_PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64:
 }
 .validate();
```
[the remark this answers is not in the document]
nsarlin-zama: I feel that "normal" users shouldn't have to worry about zkv1/v2, so by default it should be zkv2 ? Or maybe I can add another alias for explicit v2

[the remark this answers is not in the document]
nsarlin-zama: I added the alias

tfhe/src/shortint/parameters/key_switching/p_fail_2_minus_64/ks_pbs.rs
```diff
@@ -31,7 +31,25 @@ pub const V0_11_PARAM_KEYSWITCH_PKE_TO_SMALL_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2
 pub const V0_11_PARAM_KEYSWITCH_PKE_TO_BIG_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64:
```
IceTDrinker, line 31: this param and above don't have the ZKv2 naming, otherwise values look correct

nsarlin-zama: fixed

nsarlin-zama:
I added a new enum inside the `CompactPublicKeyEncryptionParameters`: ` SupportedCompactPkeZkScheme`. It can take 3 values: ZkNotSupported, V1 and V2.
ZkNotSupported is used when pke parameters are derived from compute paramters and we cannot guess the correct zk version to use.

V1 params will always generate a V1 crs and V2 params a V2 crs, which will make incompatibility issues less likely. However, since the zk version is not stored in the keyswitch paramters, it is still possible to mix pke parameters for v2 with keyswitch parameters for v1.

apps/trivium/benches/kreyvium_shortint.rs
```diff
@@ -12,12 +12,12 @@ pub fn kreyvium_shortint_warmup(c: &mut Criterion) {
     let underlying_sk: tfhe::shortint::ServerKey = (*hl_server_key.as_ref()).clone().into();
 
     let (client_key, server_key): (ClientKey, ServerKey) =
-        gen_keys(PARAM_MESSAGE_1_CARRY_1_KS_PBS_GAUSSIAN_2M64);
+        gen_keys(V0_11_PARAM_MESSAGE_1_CARRY_1_KS_PBS_GAUSSIAN_2M64);
```
IceTDrinker, line 15: huh I think this is not correct as the ConfigBuilder is using the default config, not required to change for the release tomorrow, this is purely at the GitHub repo level, please make a note to look for ConfigBuilder::default() usage

apps/trivium/benches/trivium_shortint.rs
```diff
@@ -12,12 +12,12 @@ pub fn trivium_shortint_warmup(c: &mut Criterion) {
     let underlying_sk: tfhe::shortint::ServerKey = (*hl_server_key.as_ref()).clone().into();
 
     let (client_key, server_key): (ClientKey, ServerKey) =
-        gen_keys(PARAM_MESSAGE_1_CARRY_1_KS_PBS_GAUSSIAN_2M64);
+        gen_keys(V0_11_PARAM_MESSAGE_1_CARRY_1_KS_PBS_GAUSSIAN_2M64);
```
IceTDrinker, line 15: same story here

apps/trivium/benches/kreyvium_shortint.rs
nsarlin-zama: I'm missing something, but this is just a rename ?

nsarlin-zama: I think the regression was introduced in a previous pr, so yeah we might as well fix it in another pr if that's not blocking for the release

IceTDrinker (approved):

apps/trivium/benches/kreyvium_shortint.rs
nsarlin-zama: I agree but the change to tuniform is not from this PR

IceTDrinker: yep 100%

PR #1892, opened 2024-12-18
chore(ci): allow external contribution in fast aws tests workflow

soonum:
This is a first commit to allow external contribution on only one file.
It seems that the event `pull_request_target` event need to be in `main` to be usable.


mayeul-zama (approved):
Thanks!

PR #1903, opened 2024-12-20
chore(gpu): port fix to compression encoding

pdroalves:
- Modifies the generation of the LUT used in decompression so that the delta is calculated with a different precision, as in the CPU implementation

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/845

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
@@ -627,26 +627,46 @@ void rotate_left(Torus *buffer, int mid, uint32_t array_length) {
   std::rotate(buffer, buffer + mid, buffer + array_length);
 }
 
+/// Caller needs to ensure that the operation applied is coherent from an
+/// encoding perspective.
+///
+/// For example:
+///
+/// Input encoding has 2 bits and output encoding has 4 bits, applying the
+/// identity lut would map the following:
+///
+/// 0|00|xx -> 0|00|00
+/// 0|01|xx -> 0|00|01
+/// 0|10|xx -> 0|00|10
+/// 0|11|xx -> 0|00|11
+///
+/// The reason is the identity function is computed in the input space but the
+/// scaling is done in the output space, as there are more bits in the output
+/// space, the delta is smaller hence the apparent "division" happening.
 template <typename Torus>
-void generate_lookup_table(Torus *acc, uint32_t glwe_dimension,
-                           uint32_t polynomial_size, uint32_t message_modulus,
-                           uint32_t carry_modulus,
-                           std::function<Torus(Torus)> f) {
+void generate_lookup_table_with_encoding(Torus *acc, uint32_t glwe_dimension,
+                                         uint32_t polynomial_size,
+                                         uint32_t input_message_modulus,
+                                         uint32_t input_carry_modulus,
+                                         uint32_t output_message_modulus,
+                                         uint32_t output_carry_modulus,
+                                         std::function<Torus(Torus)> f) {
 
-  uint32_t modulus_sup = message_modulus * carry_modulus;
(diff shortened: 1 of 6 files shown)


tmontaigu (commented):
looks ok to me

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -627,26 +627,46 @@ void rotate_left(Torus *buffer, int mid, uint32_t array_length) {
...
+                                         uint32_t input_message_modulus,
+                                         uint32_t input_carry_modulus,
+                                         uint32_t output_message_modulus,
+                                         uint32_t output_carry_modulus,
+                                         std::function<Torus(Torus)> f) {
+
+  uint32_t input_modulus_sup = input_message_modulus * input_carry_modulus;
+  uint32_t output_modulus_sup = output_message_modulus * output_carry_modulus;
+  uint32_t box_size = polynomial_size / input_modulus_sup;
+  Torus output_delta = (1ul << 63) / output_modulus_sup;
```
IceTDrinker, line 658: this does not work for Torus u32 (don't know if you still support that) also this can overflow as ul is not guaranteed to be 64 bits

https://en.cppreference.com/w/cpp/language/types

pdroalves: Indeed. Good catch.

We haven't dropped support, although we have no test or active development for u32. 

I will fix this.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -627,26 +627,46 @@ void rotate_left(Torus *buffer, int mid, uint32_t array_length) {
...
+                                         uint32_t input_message_modulus,
+                                         uint32_t input_carry_modulus,
+                                         uint32_t output_message_modulus,
+                                         uint32_t output_carry_modulus,
+                                         std::function<Torus(Torus)> f) {
+
+  uint32_t input_modulus_sup = input_message_modulus * input_carry_modulus;
+  uint32_t output_modulus_sup = output_message_modulus * output_carry_modulus;
+  uint32_t box_size = polynomial_size / input_modulus_sup;
+  Torus output_delta = std::numeric_limits<Torus>::max()  / (2 * output_modulus_sup);
```
pdroalves, line 658: @IceTDrinker this is not exactly `(1ul << 63) / modulus_sup` but it is close. `std::numeric_limits<Torus>::max() / 2` should be smaller than `1<<63` by one. Do you think that's ok?

IceTDrinker: another way to do it is 1 << (Torus::BITS - 1)

tmontaigu: `static_cast<Torus>(1) << (Torus::BITS - 1)` or `Torus(1) << (Torus::BITS - 1)`

as I think in `1 << (Torus::BITS - 1)` the left most constant `1` is going to be an int as all C constant that do not have postfix are int

IceTDrinker: yes of course, just meant the idea was the above, but you are right to precise it

IceTDrinker: to be fully precise it would be

```(static_cast<Torus>(1) << (Torus::BITS - 1)) / output_modulus_sup``` in this exact line you commented

IceTDrinker: Just to check you have Torus::BITS defined in CUDA ? I was using pseudo code

IceTDrinker (approved):

PR #1910, opened 2025-01-03
test(shortint): remove oprf test flakiness

mayeul-zama:


tfhe/src/shortint/oprf.rs
```diff
@@ -41,6 +42,21 @@ impl ServerKey {
...
+        log_modulus: CiphertextModulusLog,
+    ) -> LweCiphertext<Vec<u64>> {
+        let mut ct = self.create_random_from_seed(seed, lwe_size);
+
+        for i in ct.as_mut() {
+            *i = modulus_switch(*i, log_modulus) << (64 - log_modulus.0);
+        }
+
+        ct
+    }
```
IceTDrinker, line 58: forces the input to be exactly the one seen in the blind rotation ?

mayeul-zama: Exactly, this remove the MS noise which made the test flaky (previously fixed by using deterministic keys).

Plus this is more in line with the paper in which the output of the hashing function H is mod 2N.

guillermo-oyarzun (approved):

IceTDrinker (approved):

PR #1917, opened 2025-01-07
chore(zk): add bench zk v1 vs v2

nsarlin-zama:
### PR content/description
Adds benchmark of zkv1 vs zkv2 inside the tfhe crate


IceTDrinker:
is the bench parser fixed though ?

nsarlin-zama:
yes, we had 2 issues:
- one with versioned names which has been fixed here: https://github.com/zama-ai/tfhe-rs/pull/1929
- second issue in this pr because I prefixed compute parameters with ZKV1, but compute parameters are identical in V1 and V2. So I fixed it by appending ZKV1 or ZKV2 in the bench name and instead of the parameters name

nsarlin-zama:
manual run here: https://github.com/zama-ai/tfhe-rs/actions/runs/12672979600

IceTDrinker:
> manual run here: https://github.com/zama-ai/tfhe-rs/actions/runs/12672979600

perfect, did you check that the graphana display was good ?

nsarlin-zama:
I'm on it

IceTDrinker (approved):

PR #1936, opened 2025-01-09
chore(backward): move compat tests to avoid compiling them everytime

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/867

### PR content/description
Moved it to a workspace level "tests" folder, that way they are run with `cargo test -p tests` or when doing `cargo test` in the workspace root, but I don't think that we are doing this


Cargo.toml
```diff
@@ -11,10 +11,11 @@ members = [
     "backends/tfhe-cuda-backend",
     "utils/tfhe-versionable",
     "utils/tfhe-versionable-derive",
+		"tests"
```
IceTDrinker, line 14: this is a tab, please enable editor config

IceTDrinker, line 14: aren't the tests missing a Cargo.toml ?

nsarlin-zama: yes forgot to "git add" it

nsarlin-zama: I have it but editorconfig.yml only set space for .rs files

IceTDrinker: then let's update it for .c .h and .toml files as well

IceTDrinker (approved):
Thanks looks good !

PR #1939, opened 2025-01-09
chore: enable space in editorconfig for more file types

nsarlin-zama:


.editorconfig
```diff
@@ -9,7 +9,7 @@ end_of_line = lf
 insert_final_newline = true
 
 # 4 space indentation
-[*.rs]
+[*.{rs,js,c,h,toml}]
 charset = utf-8
 indent_style = space
 indent_size = 4
```
IceTDrinker, line 15: I don't think the js and c/h currently have indent_size = 4, meaning it will break with js formmating I believe ?

maybe create dedicated sections to avoid redoing all the formatting ?

tehre was a c file which had also gotten a tab in an update to test_high_level_api_zk.c

nsarlin-zama: in js some files are 2 spaces and some other are 4, so I unified it to 4. I can do the same for c/h

nsarlin-zama: looks like the js formatter respects editorconfig!

IceTDrinker: ok, just that this will lose the history of the files :|

IceTDrinker: also we don't currently have a C formatter because it's so hard to consistently have the exact version of a C formatter like clang... across sytems

nsarlin-zama: ok I can set 2 for c,h,js and 4 for rust,toml

However js fmt is currently not run everywhere so we have js files that are using 4 spaces

IceTDrinker: ok first:

toml with 4 spaces

second:

enable check_fmt_js always and update to either 2 or 4 tabs (but I do think it must run somewhere)

third:

c/h keep 2 tabs to avoid losing the history of most files, we don't have an auto formatter because C/C++ tooling sucks majorly so don't lose too much time on it unless there is a simple solution that does not involve recompiling it ourselves

IceTDrinker (approved):
Thanks a lot !

IceTDrinker (approved):

IceTDrinker:
good to merge

PR #1957, opened 2025-01-13
chore(ci): setup centralized workflow launcher

soonum:
To ease external contributions, this implements a workflow responsible for triggering workflows that are currently related to pull_request event.
To ensure security, changes on some files are allowed only to Zama members. Since workflows are included in this file, this protects against secret leaks.

For now only AWS fast CPU tests is targeted as a proof of concept.


soonum:
This implementation is discarded because of the solution used in #1962 and #1968

PR #1962, opened 2025-01-14
chore(ci): choose pull request event with path changes

soonum:
To be able to run CI for external contribution, through forked repository, and be able to test CI modification in a development branch, we need to discriminate pull request events. 
For the former 'pull_request_target' event is needed to have access to GitHub secrets. 'pull_request' event is required for the latter otherwise the workflow would be pulled from the HEAD of the base branch and thus wouldn't contain changes from the developer.


IceTDrinker (approved):
let's try this

PR #1968, opened 2025-01-14
chore(ci): skip aws fast tests if ci files changed

soonum:
This would skip 'check-user-permission' job if the event 'pull_request_target' is emitted and CI files have changed. It avoids overlapping of 'pull_request' and 'pull_request_target' events. CI changes would only be tested on 'pull_request' for Zama own pull requests.

### DIFF
--- a/.github/workflows/aws_tfhe_fast_tests.yml
+++ b/.github/workflows/aws_tfhe_fast_tests.yml
@@ -60,6 +60,7 @@ jobs:
       user_docs_test: ${{ env.IS_PULL_REQUEST == 'false' ||
         steps.changed-files.outputs.user_docs_any_changed ||
         steps.changed-files.outputs.dependencies_any_changed }}
+      ci_file_changed: ${{ env.IS_PULL_REQUEST == 'false' || steps.changed-files.outputs.ci_any_changed }}
       any_file_changed: ${{ env.IS_PULL_REQUEST == 'false' || steps.aggregated-changes.outputs.any_changed }}
     steps:
       - name: Checkout tfhe-rs
@@ -118,9 +119,13 @@ jobs:
               - '!tfhe/src/c_api/**'
               - 'tfhe/docs/**.md'
               - README.md
+            ci:
+              - .github/**
+              - ci/**
 
       - name: Aggregate file changes
         id: aggregated-changes
+        # CI files are not included in this aggregator.
         if: ( steps.changed-files.outputs.dependencies_any_changed == 'true' ||
           steps.changed-files.outputs.csprng_any_changed == 'true' ||
           steps.changed-files.outputs.zk_pok_any_changed == 'true' ||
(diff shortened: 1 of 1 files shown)


soonum:
I just tested the behavior manually. I confirm that it works as expected.

.github/workflows/aws_tfhe_fast_tests.yml
```diff
@@ -164,8 +172,8 @@ jobs:
 
   fast-tests:
     name: Fast CPU tests
-    if: github.event_name != 'pull_request_target' ||
-      (github.event_name == 'pull_request_target' && needs.setup-instance.result != 'skipped')
+    if: github.event_name == 'workflow_dispatch' ||
+      needs.setup-instance.result != 'skipped'
```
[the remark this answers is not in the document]
soonum: In fact the `if:` can be entirely removed I think, thanks to the `needs: [ should-run, setup-instance ]` and the `if:` in `setup-instance`.
Wydt ?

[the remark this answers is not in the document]
IceTDrinker: looks good, an additional question we did not put the workflow as part of the files which would trigger the tests to run, do you remember the logic ?

.github/workflows/aws_tfhe_fast_tests.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       user_docs_test: ${{ env.IS_PULL_REQUEST == 'false' ||
         steps.changed-files.outputs.user_docs_any_changed ||
         steps.changed-files.outputs.dependencies_any_changed }}
+      ci_file_changed: ${{ env.IS_PULL_REQUEST == 'false' || steps.changed-files.outputs.ci_any_changed }}
```
IceTDrinker, line 63: the token below in the checkout tfhe-rs step would not be available in a pull_request_target event from outside, correct ?

soonum: Yes it's available, but it's a special token that has only read permission on the repo.
The way it will work is: it'll pull this workflow version from base branch. Then the `should-run` will be executed in a safe way. Finally, if the triggering actor is not member of Zama, the workflow will fail with no possibility to leak this very secret, or any other in fact.

.github/workflows/aws_tfhe_fast_tests.yml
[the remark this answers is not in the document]
soonum: What do you mean ? Going back to the centralized workflow to trigger them all ?
If you think of that, then no, this new way of dealing with the files discard the need to have such centralized workflow.

.github/workflows/aws_tfhe_fast_tests.yml
IceTDrinker: isn't the github token enough for a checkout ?

.github/workflows/aws_tfhe_fast_tests.yml
```diff
@@ -11,6 +11,8 @@ env:
   SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
   SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
   SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  MSG_MINIMAL: event,action url,commit
```
IceTDrinker, line 14: what is the MSG_MINIMAL ?

.github/workflows/aws_tfhe_fast_tests.yml
soonum: Nope because GitHub token only works on public repositories.

.github/workflows/aws_tfhe_fast_tests.yml
soonum: The message format displayed on Slack. If you add `ref` to the list you'll find the current notification format.
Removing `ref` allows to get rid of the default behavior of the action.

From action's README:

> If set to true, removes: Ref, Event, Actions URL and Commit from the message. You can optionally whitelist any of these 4 removed values by passing it comma separated to the variable instead of true. (ex: MSG_MINIMAL: event or MSG_MINIMAL: ref,actions url, etc.)

IceTDrinker (approved):

PR #1974, opened 2025-01-15
fix(shortint): error message in packing keyswitch

nsarlin-zama:


tfhe/src/shortint/list_compression/compression.rs
```diff
@@ -46,7 +46,7 @@ impl CompressionKey {
 
         assert!(
             message_modulus.0 <= carry_modulus.0,
-            "GLWE packing is implemented with messages in carries, so carry_modulus (={}) can't be bigger than message_modulus(={})",
+            "GLWE packing is implemented with messages in carries, so carry_modulus (={}) must be bigger than message_modulus(={})",
```
mayeul-zama, line 49: I guess I meant `message_modulus` can't be bigger than `carry_modulus` (understood in the sense strictly bigger).

It think using the "must be bigger or equal" or inverting (as I meant) would be clearer

mayeul-zama (approved):
Thanks!

PR #1975, opened 2025-01-15
Am/test/hl api noise checks

IceTDrinker:
@mayeul-zama as I'm touching some of the stats primitives that you touched recently

### DIFF
--- a/tfhe/src/core_crypto/commons/math/random/t_uniform.rs
+++ b/tfhe/src/core_crypto/commons/math/random/t_uniform.rs
@@ -70,6 +70,22 @@ impl<T: UnsignedInteger> TUniform<T> {
     pub fn max_value_inclusive(&self) -> T::Signed {
         T::Signed::ONE << self.bound_log2 as usize
     }
+
+    // TODO: return the ModularVariance? Essentially the first part of the formula corresponds to
+    // what older iterations of the lib called ModularVariance, this struct does not currently have
+    // the DispersionParameter trait implemented, so here for convenience we compute the Variance
+    // directly by dividing the ModularVariance by the modulus squared to be in the Torus domain.
+    #[cfg(test)]
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub(crate) fn variance(
+        &self,
+        modulus: f64,
+    ) -> crate::core_crypto::commons::dispersion::Variance {
+        // (2^{2 * b + 1} + 1) / 6
+        crate::core_crypto::commons::dispersion::Variance(
+            ((2.0f64.powi(2 * (self.bound_log2 as i32) + 1) + 1.0f64) / 6.0f64) * modulus.powi(-2),
+        )
+    }
 }
 
 macro_rules! implement_t_uniform_uint {
--- a/tfhe/src/core_crypto/commons/mod.rs
+++ b/tfhe/src/core_crypto/commons/mod.rs
@@ -78,8 +78,20 @@ pub mod test_tools {
         }
     }
 
+    /// Can apply to any distribution as long as the samples are idependant and identically
+    /// distributed.
+    ///
+    /// As this construction relies on the Central Limit Theorem you must ensure that you have
+    /// enough samples for the Central Limit Theorem to apply. Some accepted values can be as low as
+    /// 30 or 50, though if you can afford it, to have tighter intervals and to have a better
+    /// convergence for the CLT you should try to gather more samples.
+    pub fn any_distribution_variance_confidence_interval() {
+        todo!()
+    }
+
     /// Note that the confidence_level is the probability to be in the interval to be computed.
-    pub fn variance_confidence_interval(
+    /// Applies only to samples known to come from a normal distribution
+    pub fn normal_distribution_variance_confidence_interval(
         sample_count: f64,
         measured_variance: Variance,
         confidence_level: f64,
@@ -423,8 +435,11 @@ pub mod test_tools {
 
         let confidence_level = 0.95;
 
-        let confidence_interval =
-            variance_confidence_interval(20., measured_variance, confidence_level);
+        let confidence_interval = normal_distribution_variance_confidence_interval(
+            20.,
+            measured_variance,
+            confidence_level,
+        );
 
         let lower_bound = confidence_interval.lower_bound();
         let upper_bound = confidence_interval.upper_bound();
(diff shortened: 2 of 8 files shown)


tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,227 @@
...
+        let noise = decrypt_lwe_ciphertext(&core_key, &lwe_ct).0;
+        let signed_noise: i64 = noise as i64;
+
+        // Rescale
+        noise_samples.push(signed_noise as f64 / modulus_as_f64);
+    }
+
+    let measured_variance = variance(&noise_samples);
+    let measured_confidence_interval =
+        variance_confidence_interval(noise_samples.len() as f64, measured_variance, 0.99);
```
mayeul-zama, line 212: `confidence_level` variable could be defined above and used here.

A significance level of 1% produces a quite flaky test.

IceTDrinker: haven't had issues when we tested those up until now

mayeul-zama: Sure, it's not that big
But 1% can be quite annoying, it means 1% of the CI runs will fail because of this test I believe

IceTDrinker: I'll run in a loop and try to get an estimate of failures

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -346,4 +397,45 @@ pub mod test_tools {
             assert_eq!(distance, -0.25);
         }
     }
+
+    #[test]
+    fn test_confidence_interval() {
+        // https://stats.libretexts.org/Bookshelves/Introductory_Statistics/
+        // Inferential_Statistics_and_Probability_-_A_Holistic_Approach_(Geraghty)/
+        // 09%3A_Point_Estimation_and_Confidence_Intervals/9.03%3A_Confidence_Intervals
```
nsarlin-zama, line 405: Maybe for links we could make a exception to the line width rule to make them clickable ?

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: 0 fails in a 1000 run

tfhe/src/core_crypto/commons/math/random/t_uniform.rs
```diff
@@ -70,6 +71,13 @@ impl<T: UnsignedInteger> TUniform<T> {
     pub fn max_value_inclusive(&self) -> T::Signed {
         T::Signed::ONE << self.bound_log2 as usize
     }
+
+    pub fn variance(&self, modulus: f64) -> Variance {
+        // (2^{2 * b + 1} + 1) / 6
+        Variance(
+            ((2.0f64.powi(2 * (self.bound_log2 as i32) + 1) + 1.0f64) / 6.0f64) * modulus.powi(-2),
+        )
+    }
```
nsarlin-zama, line 80: For me it feels a bit weird to divide by the modulus here, because the modulus is not a parameter of the distribution. I would rather return the variance without the division and do the rescaling outside of this function.

nsarlin-zama (commented):
small comments

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
mayeul-zama: Then either the confidence interval is not correct or I don't interpret the confidence level well
What's your understanding of what the confidence level is?

tfhe/src/core_crypto/commons/math/random/t_uniform.rs
IceTDrinker: that's fair, I did that quickly when I was doing that for the noise tests

IceTDrinker: though there is a notion of modular variance under the modulus of the ciphertext (those come from way back) so I think I tried to stick to the meaning of Variance and ModularVariance

IceTDrinker: I'll recheck

IceTDrinker: so the value computed without the modulus is a ModularVariance, this struct needs a modulus so the API would not change, the reason I did not return a ModularVariance it seems is that we are missing some trait implementation on it and I did not want to make mistakes implementing that when I did it for the noise check.

I can put this API pub(crate) and have a TODO on it

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: Long lines tend to create issues with rustfmt https://github.com/rust-lang/rustfmt/issues/3863 several comments in there about long lines causing the formatting to stop working

so no I won't change this to avoid rustfmt randomly stopping formatting stuff 🙂

tfhe/src/core_crypto/commons/math/random/t_uniform.rs
IceTDrinker: pub(crate) with a descriptive TODO, only used in tests today anyways I think this is fine

IceTDrinker:
I'll try the "bootstrapping" method found online for the confidence interval for non gaussian stuff, if it does not work we won't have much choice than to go with the current stuff even if the confidence interval seems to not behave as expected for the TUniform distribution here

IceTDrinker:
got something that looks more reasonable based on the central limit theorem the gist is

X ~ TUniform(b)
define the auxiliary variable Y ~ X^2 given E[X] = 0 then sum(Y_i) / (N - 1) is the unbiased variance estimate.

for a sufficient number of samples E[Y] = sum(Y_i) / N distribution follows a normal law https://en.wikipedia.org/wiki/Central_limit_theorem#Classical_CLT

using that relationship with the normal law we can get a link between the variance estimate we want and that normal law allowing to use the normal law inverse CDF to have a confidence interval, managed to confirm that the confidence interval looks ok

IceTDrinker:
superseded by https://github.com/zama-ai/tfhe-rs/pull/2377

PR #1981, opened 2025-01-16
Ns/doc strings

nsarlin-zama:
small fixes to string doc


IceTDrinker (approved):

nsarlin-zama:
TODO: rework tutorial instead of removing it

nsarlin-zama:
I reworded the tutorial to make it more explicit that there is now an included strings module, with an example at the end for comparison.

IceTDrinker (approved):

PR #1982, opened 2025-01-16
fix(safe_ser): aliases in named for renamed types deserialization

nsarlin-zama:

### PR content/description
Adds the possibility to have aliases in the Named trait.

Context:
In previous versions of TFHE-rs only the `CompactPkePublicParams` were serializable. Since 0.11, the `CompactPkeCrs` is the main exposed type and is serializable. For versioning, there is an upgrade path that allows to load and upgrade an old `CompactPkePublicParams` into a new `CompactPkeCrs`. However this could not work in practice for `CompactPkePublicParams` serialized with `safe_serialize` because of the `Named` impl was different and the data was rejected in the header check.

This PR allows to have aliases to a name, that will be accepted when loading this name. This is optional so it does not break previous code. That way, "zk::CompactPkePublicParams" is seen as an alias for "zk::CompactPkeCrs".

### DIFF
--- a/tfhe/src/named.rs
+++ b/tfhe/src/named.rs
@@ -1,3 +1,7 @@
 pub trait Named {
+    /// Default name for the type
     const NAME: &'static str;
+    /// Aliases that should also be accepted for backward compatibility when checking the name of
+    /// values of this type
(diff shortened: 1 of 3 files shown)


mayeul-zama (approved):
Thanks!

PR #1985, opened 2025-01-17
chore(ci): extend external contribution to all pr workflows

soonum:
User permission checking is done after the should-run, when there is such step, rather than before it.
This way, only workflows that should run would fail id triggering actor is not allowed to launch it.
Thus a repository maintainer would have to re-run only a handful of jobs that would effectively run afterward.

Resolves: zama-ai/tfhe-rs-internal#890


.github/workflows/check_ci_files_change.yml
```diff
@@ -0,0 +1,41 @@
...
+        value: ${{ jobs.check-changes.outputs.ci_file_changed }}
+    secrets:
+      REPO_CHECKOUT_TOKEN:
+        required: true
+
+jobs:
+  check-changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
nsarlin-zama, line 21: Why does this need write permission ?

soonum: It's an old mistake, `tj-actions/changed-files` only needs Read permission in our case. I'm changing it.

IceTDrinker:
@soonum lint and checks not happy ?

soonum:
> @soonum lint and checks not happy ?

Ok, I see what's going on here. I need to remove the dynamic referencing of workflow calls. `<my_workflow>@${{ github.sha }}` syntax induce a failure during checking because the version is not pinned.
The solution is: remove the reference, merge this PR, then create a new PR with a SHA from `main` for these workflows to correctly pin the action.
This would make any change to the two workflows a bit more subtle to test but this is the only solution I can think of to harden security.

IceTDrinker:
> > @soonum lint and checks not happy ?
> 
> Ok, I see what's going on here. I need to remove the dynamic referencing of workflow calls. `<my_workflow>@${{ github.sha }}` syntax induce a failure during checking because the version is not pinned. The solution is: remove the reference, merge this PR, then create a new PR with a SHA from `main` for these workflows to correctly pin the action. This would make any change to the two workflows a bit more subtle to test but this is the only solution I can think of to harden security.

there is a whitelist for the dep pinning

nsarlin-zama (approved):
it looks good to me, thanks!

nsarlin-zama (approved):

nsarlin-zama (approved):
Looks good to me, good job!

PR #1997, opened 2025-01-21
fix(gpu): general fixes on indexes used in multi-gpu context

pdroalves:
- fix a bug in which the wrong GPU may be queried for the max shared memory
- If multiple streams are running split through multiple GPUs, operations happening on a stream in GPU i should query GPU i about its max shared memory.
- also fixes wrong indexing at rust side

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/901

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/device.h
+++ b/backends/tfhe-cuda-backend/cuda/include/device.h
@@ -27,6 +27,8 @@ inline void cuda_error(cudaError_t code, const char *file, int line) {
     std::abort();                                                              \
   }
 
+void cuda_set_device(uint32_t gpu_index);
+
 cudaEvent_t cuda_create_event(uint32_t gpu_index);
 
 void cuda_event_record(cudaEvent_t event, cudaStream_t stream,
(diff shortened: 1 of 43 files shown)


pdroalves:
@agnesLertá maybe we could introduce a new test to cover this scenario. Wdyt?

agnesLeroy (commented):
Hey @pdroalves! Looks good to me thanks for the fix! Yes, we should definitely add a test for it. The thing is, for now only the long run tests are testing the use of a single GPU other than 0. What we could do is add long run tests with a set of classic PBS parameters maybe? Otherwise we could add signed & unsigned tests with a different FnExecutor, in the same fashion as I did for the long run tests. Wdyt? (see the executor here: tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs:34)

agnesLeroy (approved):

agnesLeroy:
@pdroalves I saw the multi-gpu tests passed: maybe you want to merge the cudaSetDevice encapsulation first and then rebase this one on main before merging?

pdroalves:
@agnesLeroy Actually that PR was based on this one.

agnesLeroy:
@pdroalves tests have passed, thanks a lot!

PR #2001, opened 2025-01-24
Al/refactor lut

agnesLeroy:

### PR content/description

This PR introduces degree / max degree data in the int_radix_lut structure, to prepare for noise tracking in apply_uni/bivariate_lookup_table.

I chose to track max_degree as well because for the many lut case it depends on the number of luts applied.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -10,6 +10,14 @@
 #include <cmath>
 #include <functional>
 
+class NoiseLevel {
+public:
+  // Constants equivalent to the Rust code
+  static const uint64_t NOMINAL = 1;
+  static const uint64_t ZERO = 0;
+  static const uint64_t UNKNOWN = std::numeric_limits<uint64_t>::max();
+};
+
 template <typename Torus>
 __global__ void radix_blocks_rotate_right(Torus *dst, Torus *src,
                                           uint32_t value, uint32_t blocks_count,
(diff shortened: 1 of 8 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -1604,10 +1670,13 @@ template <typename Torus> struct int_prop_simu_group_carries_memory {
         }
       };
 
-      auto lut = luts_array_second_step->get_lut(0, lut_id);
       generate_device_accumulator<Torus>(
-          streams[0], gpu_indexes[0], lut, glwe_dimension, polynomial_size,
-          message_modulus, carry_modulus, f_first_grouping_inner_propagation);
+          streams[0], gpu_indexes[0],
+          luts_array_second_step->get_lut(0, lut_id),
+          luts_array_second_step->get_degree(0),
```
guillermo-oyarzun, line 1676: this should be get_degree(lut_id) and get_max_degree(lut_id)

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -680,22 +680,23 @@ void generate_lookup_table_with_encoding(Torus *acc, uint32_t glwe_dimension,
   }
 
   rotate_left<Torus>(body, half_box_size, polynomial_size);
+  return degree;
```
guillermo-oyarzun, line 683: why do we need to return the degree now?

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -848,8 +861,9 @@ void generate_device_accumulator_with_encoding(
   Torus *h_lut =
       (Torus *)malloc((glwe_dimension + 1) * polynomial_size * sizeof(Torus));
 
+  *max_degree = input_message_modulus * input_carry_modulus - 1;
   // fill accumulator
-  generate_lookup_table_with_encoding<Torus>(
+  *degree = generate_lookup_table_with_encoding<Torus>(
```
guillermo-oyarzun, line 866: shouldn't degree and max_degree be an array instead of single_value?

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
agnesLeroy: I return it to store it in int_radix_lut. It's done when calling generate_device_lookup_table_...

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: Nice catch :pray:

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -1256,8 +1308,10 @@ template <typename Torus> struct int_seq_group_prop_memory {
       };
       auto seq_lut = lut_sequential_algorithm->get_lut(0, index);
       generate_device_accumulator<Torus>(
-          streams[0], gpu_indexes[0], seq_lut, glwe_dimension, polynomial_size,
-          message_modulus, carry_modulus, f_lut_sequential);
+          streams[0], gpu_indexes[0], seq_lut,
+          lut_sequential_algorithm->get_degree(0),
```
guillermo-oyarzun, line 1312: here it should be get_degree(index) same for max_degree

guillermo-oyarzun:
Maybe it would be good to launch the long tests on this PR before merging it, mainly to test the many combinations of the luts

agnesLeroy:
I'll do a quick run locally with less ops. Since only the C++ side is modified, this does not affect the computation of noise level/degree on the Rust side yet. Next PR should do that.

guillermo-oyarzun:
> I'll do a quick run locally with less ops. Since only the C++ side is modified, this does not affect the computation of noise level/degree on the Rust side yet. Next PR should do that.

ahh ok, then maybe it is not needed, if nothing is using this yet, probably we won't see anything with the tests anyway

PR #2010, opened 2025-01-27
[TEST] Test user permission check new triggering condition

soonum:
AWS Fast user permissions  check job shouldn't be run.


soonum:
It works as expected. :tada:

PR #2011, opened 2025-01-27
chore(ci): run user permissions check if job should run

soonum:
A small workflow optimization by running user permissions check only if a job is supposed to run.

### DIFF
--- a/.github/workflows/aws_tfhe_fast_tests.yml
+++ b/.github/workflows/aws_tfhe_fast_tests.yml
@@ -148,9 +148,11 @@ jobs:
   # Fail if the triggering actor is not part of Zama organization.
   # If pull_request_target is emitted and CI files have changed, skip this job. This would skip following jobs.
   check-user-permission:
-    needs: check-ci-files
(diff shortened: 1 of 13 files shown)


IceTDrinker:
@soonum I'm seeing cancelled workflows even in other PRs something to investigate

e.g. https://github.com/zama-ai/tfhe-rs/actions/runs/12994186340/job/36238356440?pr=2011

.github/workflows/aws_tfhe_fast_tests.yml
```diff
@@ -148,9 +148,11 @@ jobs:
   # Fail if the triggering actor is not part of Zama organization.
```
IceTDrinker, line 148: thinking about the aggregate file changes above, there is no outputs.any_changed that is global to avoid doing it by hand ?

IceTDrinker (approved):
Thanks ! this will allow to more easily identify which workflow requires a manual run

soonum:
Discarded since #2056

PR #2015, opened 2025-01-29
feat(integer): impl Named for compression keys

mayeul-zama:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2015)

### DIFF
--- a/tfhe/src/integer/compression_keys.rs
+++ b/tfhe/src/integer/compression_keys.rs
@@ -1,6 +1,7 @@
 use super::ClientKey;
 use crate::conformance::ParameterSetConformant;
 use crate::integer::backward_compatibility::list_compression::*;
+use crate::named::Named;
 use serde::{Deserialize, Serialize};
 use tfhe_versionable::Versionize;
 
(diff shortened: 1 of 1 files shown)


IceTDrinker:
conformance already done for all structs ?

mayeul-zama:
> conformance already done for all structs ?

Not for `CompressionPrivateKeys`, which is probably nor useful
It's done for the other types

mayeul-zama:
> > > conformance already done for all structs ?
> > 
> > 
> > Not for `CompressionPrivateKeys`, which is probably not useful It's done for the other types
> 
> Maybe for coherence we should still implement it or remove the serialize impl ?

`ClientKey` also impl `Named` but not `ParameterSetConformant`

It can still be useful for using `safe_deserialization` without conformance and avoid some deserialization errors  
IMO it's fine as is

IceTDrinker:
If it’s not too bad and we have it for client key could be worth adding the missing conformance

IceTDrinker (approved):
Thanks a lot !

PR #2019, opened 2025-01-30
fix(hlapi): ciphertext list decompress after safe_deser

tmontaigu:
After a safe_serialize/safe_deserialize, the CompressedCiphertextList was on Cpu. As the `get` method looked at the device of the data and not the device of the server_key to know where computation needs to happen, it meant that in this case decompressing using Gpu was impossible, only Cpu was usable (as data was always onlu on Cpu)

The fix is twofold:
* First, when deserializing, the data will use the current serverkey (if any) as a hint on where data should be placed
* the `get` method now uses the server_key to know where computations needs to be done, which may incur a temporary copy/transfer on every call to `get` if the device is not correct.

The API to move data has also been added

Note that this was not the case when using regular serialize/deserialize as this would store the device, so that deserialize was able to restore into the same device (hence why the test which use serialie/deserialize did not fail). In hindsight, the ser/de impl should not save which device the data originated from

Fixes zama-ai/tfhe-rs-internal/issues/905

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019)

### DIFF
--- a/tfhe/src/high_level_api/compressed_ciphertext_list.rs
+++ b/tfhe/src/high_level_api/compressed_ciphertext_list.rs
@@ -1,10 +1,18 @@
 use tfhe_versionable::{Unversionize, UnversionizeError, Versionize, VersionizeOwned};
 
+use super::details::MaybeCloned;
+#[cfg(feature = "gpu")]
+use super::global_state::with_thread_local_cuda_streams_for_gpu_indexes;
 use super::keys::InternalServerKey;
+#[cfg(feature = "gpu")]
+use super::GpuIndex;
 use crate::backward_compatibility::compressed_ciphertext_list::CompressedCiphertextListVersions;
 use crate::core_crypto::commons::math::random::{Deserialize, Serialize};
+#[cfg(feature = "gpu")]
+use crate::core_crypto::gpu::CudaStreams;
 use crate::high_level_api::booleans::InnerBoolean;
 use crate::high_level_api::errors::UninitializedServerKey;
+use crate::high_level_api::global_state::device_of_internal_keys;
 #[cfg(feature = "gpu")]
 use crate::high_level_api::global_state::with_thread_local_cuda_streams;
 use crate::high_level_api::integers::{FheIntId, FheUintId};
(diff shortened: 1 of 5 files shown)


tfhe/src/high_level_api/compressed_ciphertext_list.rs
```diff
@@ -206,13 +214,119 @@ impl CompressedCiphertextListBuilder {
...
+        #[derive(Deserialize)]
+        enum Fake {
+            Cpu(crate::integer::ciphertext::CompressedCiphertextList),
+            #[cfg(feature = "gpu")]
+            Cuda(crate::integer::gpu::ciphertext::compressed_ciphertext_list::CudaCompressedCiphertextList),
+        }
+        let mut new = match Fake::deserialize(deserializer)? {
+            Fake::Cpu(v) => Self::Cpu(v),
+            #[cfg(feature = "gpu")]
+            Fake::Cuda(v) => Self::Cuda(v),
```
IceTDrinker, line 238: does serialization of CUDA lists even work ? 🤔

tmontaigu: Yes it works, the impl in integer/gpu deos a tmp copy to Cpu then serializes that cpu ciphertext

IceTDrinker: So we can never encounter Cuda variants in practice during deserialize ?

tmontaigu: We can, when`Self::Cuda(v)`   where `v` is a cuda comrpessed list is serialized, serde will first store the fact that the variant is Cuda, then is asks v to serialize itself, and v is going to seriliaze itself using cpu format.

So when deserializing such a thin, serde will first see that the variant is Self::Cuda, then call deserialize, so now the CudaCompressedList, will take the bytes deserialize them as a CompressedList and then transfert that to a CudaCompressedList

IceTDrinker (commented):
Some comments/questions otherwise looks very good IMO

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-:-OHrmvK9-CYl6joc290o:bqgmg0j)* status: all files reviewed, 4 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 224 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrlG8zDEaWNezC9r_Z:-OHrlG8zDEaWNezC9r__:ba16sgi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L224)):*
> ```XML
> }
> 
> impl<'de> serde::Deserialize<'de> for InnerCompressedCiphertextList {
> ```

could this be simplified somewhat by having the Fake enum and a Deserialize Into bound ?

Not against manual serialization code, just that it may need a bit less manual wokr here

something like https://serde.rs/container-attrs.html#from perhaps and have a serializable intermediate struct and the into would do the conversion
___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 735 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrmNqY5x9iIDNymh0N:-OHrmNqY5x9iIDNymh0O:b-2etmum) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L735)):*
> ```XML
>                 let mut serialized = vec![];
>                 safe_serialize(&compressed_list, &mut serialized, 1024 * 1024 * 16)
>                     .expect("safe serialize succeeds");
> ```

here the message would be "compressed list safe serialize failed" as it would appear in the stdout right ?

others are just unwrap though so maybe just an unwrap here could be enough
___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 796 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrmgOf9YGdcUtUDQkL:-OHrmgOf9YGdcUtUDQkM:b-mum2sr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L796)):*
> ```XML
>                 let mut serialized = vec![];
>                 safe_serialize(&compressed_list, &mut serialized, 1024 * 1024 * 16)
>                     .expect("safe serialize succeeds");
> ```

same remark
___
*[`tfhe/src/core_crypto/gpu/mod.rs` line 739 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrkOLl7V0vZAllSAgt:-OHrkOLl7V0vZAllSAgu:b-5v6qbv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/core_crypto/gpu/mod.rs#L739)):*
> ```XML
>             d_vec
>         };
>         streams.synchronize();
> ```

is this enough to ensure copy went through ? I'm too rusty on cuda primitives currently to remember

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019)* status: all files reviewed, 4 unresolved discussions (waiting on @IceTDrinker)

___
*[`tfhe/src/core_crypto/gpu/mod.rs` line 739 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrkOLl7V0vZAllSAgt:-OHrnjxv-EIRZRBewMer:b-d2xh9v) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/core_crypto/gpu/mod.rs#L739)):*

is this enough to ensure copy went through ? I'm too rusty on cuda primitives currently to remember
</blockquote></details>

It should, that is what we do for all functions that call someting async
___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 224 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrlG8zDEaWNezC9r_Z:-OHroRtJA8pokLtp8gen:bur15ws) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L224)):*

could this be simplified somewhat by having the Fake enum and a Deserialize Into bound ?

Not against manual serialization code, just that it may need a bit less manual wokr here

something like https://serde.rs/container-attrs.html#from perhaps and have a serializable intermediate struct and the into would do the conversion
</blockquote></details>

Using `#[serde(from = "FromType")]` would probably work, it would just make us move the definition of fake, and the new+move_to_device into a impl of From, 

I can try it if you think that's better
___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 735 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrmNqY5x9iIDNymh0N:-OHrnu-c1bpXlNa1h0fL:b-srrpao) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L735)):*

here the message would be "compressed list safe serialize failed" as it would appear in the stdout right ?

others are just unwrap though so maybe just an unwrap here could be enough
</blockquote></details>

Yes, will change this one and the other, it was the original message from the bug report

IceTDrinker (approved):
Thanks looks good !

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-:-OHrpECmCFElp1BPVMpT:balxvrw)* status: all files reviewed, 2 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 224 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrlG8zDEaWNezC9r_Z:-OHrp1zh5h37KDbnuAVj:b-kkk61u) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L224)):*

Using `#[serde(from = "FromType")]` would probably work, it would just make us move the definition of fake, and the new+move_to_device into a impl of From, 

I can try it if you think that's better
</blockquote></details>

Not saying is better just wondering, if it's not worth it no need to do it

tmontaigu:
*[`tfhe/src/high_level_api/compressed_ciphertext_list.rs` line 224 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-OHrlG8zDEaWNezC9r_Z:-OHsEd7x8_q7KLTViGcg:bezuhxn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6b3a3200d0c4cea3fdee5445d47817cd646b7c37/tfhe/src/high_level_api/compressed_ciphertext_list.rs#L224)):*

Not saying is better just wondering, if it's not worth it no need to do it
</blockquote></details>

I would say that to me, since this fake enum is only used in one place, it's better to hide it in the custom impl of Deserialize

IceTDrinker:
pcc unhappy

IceTDrinker (approved):
Alright all good ! Thanks a lot :) 
:lgtm:

> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2019#-:-OHwvn3aAckSnh1K2iRN:bhxpst7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tmontaigu)

PR #2020, opened 2025-01-30
style(c_api): use destructuring to avoid missing fields

mayeul-zama:


tfhe/src/c_api/shortint/parameters.rs
```diff
@@ -398,6 +410,8 @@ expose_as_shortint_pbs_parameters!(
     PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M64,
 );
 
+// TODO: make API using this type
+#[allow(dead_code)]
 pub struct CompressionParameters(
     pub(crate) crate::shortint::parameters::list_compression::CompressionParameters,
 );
```
IceTDrinker, line 417: no configuration in the HL API to enable compression ?

mayeul-zama: Actually yes, it's used by `config_builder_enable_compression`
It appeared unused because I didn't have `high-level-c-api` feature enabled
Removed

PR #2043, opened 2025-02-10
chore: use tfhe-backward-compat-data version  from Cargo.toml

mayeul-zama:


scripts/backward_compat_data_version.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+awk 'match($0,/^tfhe-backward-compat-data = { git = "https:\/\/github\.com\/zama-ai\/tfhe-backward-compat-data\.git", branch = "([^"]*)"/, arr) { print arr[1]}' tests/Cargo.toml
```
IceTDrinker, line 3: missing newline at the end, make sure this file has chmod +x

IceTDrinker:
can you make sure it works on macOS as well ?

scripts/backward_compat_data_version.sh
nsarlin-zama: Maybe use a dedicated toml parser ? That way we are sure that it stays independent of the formatting

nsarlin-zama: I think there is one in the python stdlib if we don't want to install anything: https://docs.python.org/3/library/tomllib.html

scripts/backward_compat_data_version.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/python
```
IceTDrinker, line 1: python3

IceTDrinker (approved):

IceTDrinker:
@mayeul-zama go for merge as this slows/blocks other PRs

PR #2047, opened 2025-02-10
chore(core): remove log and modular standard_dev

mayeul-zama:


IceTDrinker:
going to be using the log std dev for my tests, so you can close this, as for the change we discussed, it can be done later if it means you can merge your PR for drift earlier without wasting time on the modular variance

mayeul-zama:
> going to be using the log std dev for my tests, so you can close this, as for the change we discussed, it can be done later if it means you can merge your PR for drift earlier without wasting time on the modular variance

I added `StandardDev::from_log_standard_dev` to replace `LogStandardDev::from_log_standard_dev` in some tests
Doesn't that cover your use case?

PR #2060, opened 2025-02-13
chore(core): use real modulus in test

mayeul-zama:


tfhe/src/core_crypto/algorithms/modulus_switch_noise_reduction.rs
```diff
@@ -128,6 +128,12 @@ where
         0,
         "Expected at least one encryption of zero"
     );
+    assert_eq!(
+        lwe.ciphertext_modulus(),
+        CiphertextModulus::new_native(),
+        "Non native modulus are not supported, got {}",
+        lwe.ciphertext_modulus(),
+    );
```
IceTDrinker, line 136: as an FYI you can do assert!(lwe.ciphertext_modulus().is_native_modulus(), msg)

IceTDrinker (approved):

PR #2065, opened 2025-02-14
refactor(hlapi): separate hlapi compressed/compact list from integer

tmontaigu:
This makes the HLAPI's compressed list use lower level struct (from shortint) rather than integer, this will allow the list to store non integer data.

BREAKING CHANGE: The trait bound on CompactCiphertextList changed from `Compactable` to `HlCompactable`.


IceTDrinker (commented):
Comment to make it available in reviewable

tmontaigu:
Closed in favor of a simpler thing

PR #2068, opened 2025-02-14
fix(gpu): enable large number of samples in pbs tbc

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
@@ -76,6 +76,55 @@ mul_ggsw_glwe_in_fourier_domain(double2 *fft, double2 *join_buffer,
 
   synchronize_threads_in_block();
 }
+// This is a temporary function to have a working version of the tbc-PBS
+// with large integers.
+// TO DO: we should erase this function when the other flavors of the PBS also
+// work with the number of samples in the first block dimension.
+template <typename G, class params>
+__device__ void mul_ggsw_glwe_in_fourier_domain_tbc(
+    double2 *fft, double2 *join_buffer,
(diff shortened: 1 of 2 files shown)


guillermo-oyarzun:
performance was not affected after this change, tested it with cuda benchmarks

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -77,6 +77,52 @@ mul_ggsw_glwe_in_fourier_domain(double2 *fft, double2 *join_buffer,
   synchronize_threads_in_block();
 }
 
+template <typename G, class params>
+__device__ void mul_ggsw_glwe_in_fourier_domain_tbc(
```
pdroalves, line 81: I don't think it's a good idea to duplicate this function. We should actually update the other flavours to swap the dimensions too.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_tbc_multibit.cuh
```diff
@@ -68,22 +68,22 @@ __global__ void __launch_bounds__(params::degree / params::opt)
   // The third dimension of the block is used to determine on which ciphertext
```
pdroalves, line 68: Update this comment

pdroalves (changes requested):

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
guillermo-oyarzun: totally agree, this is just an intermediate step to have an independent fix in the tbc pbs. In this way we can port the other pbs one by one, after that we remove this function. As agreed, i will add a comment to not forget to remove it.

pdroalves (approved):

PR #2087, opened 2025-02-19
docs: refactor and update benchmarks pages

soonum:
Benchmarks tables are rendered as descriptive SVG images.
Sort results by backend to have a clearer view in tree of content.
PBS benchmarks now display results for various p-fail and several precisions.


yuxizama (approved):
Looking very good! 
Thank you for this work @soonum @IceTDrinker

PR #2089, opened 2025-02-19
feat(gpu): refactor the sample extract entry point

pdroalves:
feat(gpu): refactor the sample extract entry point so the user can pass how many LWEs should be extracted per GLWE

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/925

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cuh
@@ -28,7 +28,7 @@ void cuda_convert_lwe_ciphertext_vector_to_cpu(cudaStream_t stream,
 
 template <typename Torus, class params>
 __global__ void sample_extract(Torus *lwe_array_out, Torus const *glwe_array_in,
-                               uint32_t const *nth_array,
+                               uint32_t const *nth_array, uint32_t lwe_per_glwe,
                                uint32_t glwe_dimension) {
 
   const int input_id = blockIdx.x;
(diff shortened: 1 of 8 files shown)


tfhe/src/core_crypto/gpu/algorithms/test/glwe_sample_extraction.rs
```diff
@@ -75,21 +75,22 @@ fn glwe_encrypt_sample_extract_decrypt_custom_mod<Scalar: UnsignedTorus + Send +
         let input_cuda_glwe_list =
             CudaGlweCiphertextList::from_glwe_ciphertext_list(&glwe_list, &streams);
 
+        let lwe_per_glwe = 2;
```
guillermo-oyarzun, line 78: why is this hardcoded to 2?

pdroalves: No particular reason. I just wanted to test for some value different than `lwe_per_glwe = polynomial_size`, since this case is already covered with compression.

Do you think it should be something else?

agnesLeroy (commented):
Hello! Thank you for this PR @pdroalves. I'm not sure what the use case is for it, @andrei-stoian-zama if it matches your use case then it's good for me :slightly_smiling_face: I just have some small comments otherwise it looks good.

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cuh
```diff
@@ -39,28 +39,29 @@ __global__ void sample_extract(Torus *lwe_array_out, Torus const *glwe_array_in,
...
 template <typename Torus, class params>
-__host__ void host_sample_extract(cudaStream_t stream, uint32_t gpu_index,
-                                  Torus *lwe_array_out,
-                                  Torus const *glwe_array_in,
-                                  uint32_t const *nth_array, uint32_t num_nths,
-                                  uint32_t glwe_dimension) {
+__host__ void
+host_sample_extract(cudaStream_t stream, uint32_t gpu_index,
+                    Torus *lwe_array_out, Torus const *glwe_array_in,
+                    uint32_t const *nth_array, uint32_t num_nths,
```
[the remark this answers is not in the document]
agnesLeroy: Maybe we should discuss again about the design of this function @pdroalves, I have the feeling this could be simplified but maybe I'm wrong.

tfhe/src/core_crypto/gpu/algorithms/test/glwe_sample_extraction.rs
pdroalves: That's an idea. It will increase considerably the time for this test but it's worthy. I will do the adjust.

pdroalves: Done.

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cuh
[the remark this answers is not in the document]
pdroalves: Yes, naming here is bad. The GPU implementation follows CPU's, so I didn't want to change that too much and get too far from the reference code.

The idea here is to extract the `nth`-LWE stored in a GLWE ciphertext. We can do it in batch, so we have an array with the indexes we want. That's `nth_array` and `num_nths`. 

What if I just add this explanation as a comment?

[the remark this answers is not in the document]
pdroalves: @agnesLeroy have a look on what I added

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cuh
```diff
@@ -39,28 +39,42 @@ __global__ void sample_extract(Torus *lwe_array_out, Torus const *glwe_array_in,
...
+ *
+ * This function extracts LWE ciphertexts from GLWE ciphertexts
+ *
+ * \param stream The CUDA stream to use for the operation.
+ * \param gpu_index The index of the GPU to use.
+ * \param lwe_array_out The output array for the extracted LWE ciphertexts.
+ * \param glwe_array_in The input array of GLWE ciphertexts.
+ * \param nth_array The array of indices specifying which LWE ciphertexts to
+ * extract from each GLWE ciphertext. \param num_nths The number of indices in
+ * the nth_array. \param lwe_per_glwe The number of LWE ciphertexts to extract
```
agnesLeroy, line 62: I don't understand why we need lwe_per_glwe, num_nths does not provide the same information?

pdroalves: `num_nths` will tell you how many LWEs will be extracted in total. `lwe_per_glwe` will tell you how many LWEs will be extracted from a single GLWE.

agnesLeroy: Ah ok makes sense :+1: :pray:

agnesLeroy (approved):
Good to go for me once the test is updated :slightly_smiling_face:

pdroalves:
Test was updated. Will merge as soon the CI runs it.

andrei-stoian-zama:
@agnesLeroy this helps to implement GLWE dot product, which requires to sample extract the Nth coefficient out of each GLWE in a list. 

I verified independently in the GLWE dot product backend that this PR helps with that use-case.

PR #2090, opened 2025-02-20
chore(ci): fix release workflow for tfhe-versionable

soonum:
tfhe-versionable crate depends on tfhe-versionable-derive. Workflow, now ensure that derive crate is published before attempting to package tfhe-versionable.

Dry-run option is removed since it cannot be use correctly due the reason aforementioned.


nsarlin-zama (approved):
Looks good to me, thanks!
Sadly we cannot test this before next release :\

IceTDrinker (approved):
Thanks, looks good indeed !

soonum:
closes: zama-ai/tfhe-rs-internal#921

PR #2091, opened 2025-02-20
chore(ci): add should-run to tfhe-fft and tfhe-ntt tests

soonum:
This is done to avoid testing tfhe-ftt/ntt crates if nothing changes in their source files.
However, these tests would be run unconditionally on each push on main branch.

### DIFF
--- a/.github/workflows/cargo_test_fft.yml
+++ b/.github/workflows/cargo_test_fft.yml
@@ -3,16 +3,46 @@ name: Cargo Test tfhe-fft
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
 
 env:
   CARGO_TERM_COLOR: always
+  IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
 
 jobs:
+  should-run:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      fft_test: ${{ env.IS_PULL_REQUEST == 'false' || steps.changed-files.outputs.fft_any_changed }}
+    steps:
+      - name: Checkout tfhe-rs
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          fetch-depth: 0
+          persist-credentials: 'false'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/cargo_test_fft.yml
```diff
@@ -61,11 +92,31 @@ jobs:
...
+        run: |
+          echo "tfhe-fft files haven't changed tests don't need to run"
+
+      - name: All tests success
+        if: needs.should-run.outputs.fft_test == 'true' &&
+          needs.cargo-tests-fft.result == 'success' &&
+          needs.cargo-tests-fft-nightly.result == 'success' &&
+          needs.cargo-tests-fft-node-js.result == 'success'
+        run: |
+          echo "All tfhe-fft test passed"
```
IceTDrinker, line 122: is this enough to fail the build if one of the tests fails ?

soonum: The push I just made should fix this. Thanks for noticing.

.github/workflows/cargo_test_fft.yml
```diff
@@ -3,16 +3,45 @@ name: Cargo Test tfhe-fft
...
+
+      - name: Check for file changes
+        id: changed-files
+        uses: tj-actions/changed-files@dcc7a0cba800f454d79fff4b993e8c3555bcc0a8
+        with:
+          files_yaml: |
+            fft:
+              - tfhe/Cargo.toml
+              - tfhe-fft/**
+              - '.github/workflows/cargo_test_fft.yml'
```
IceTDrinker, line 40: maybe the Makefile as well ?

.github/workflows/cargo_test_ntt.yml
```diff
@@ -3,16 +3,45 @@ name: Cargo Test tfhe-ntt
...
+
+      - name: Check for file changes
+        id: changed-files
+        uses: tj-actions/changed-files@dcc7a0cba800f454d79fff4b993e8c3555bcc0a8
+        with:
+          files_yaml: |
+            ntt:
+              - tfhe/Cargo.toml
+              - tfhe-ntt/**
+              - '.github/workflows/cargo_test_ntt.yml'
```
IceTDrinker, line 40: same remark here for Makefile

IceTDrinker (approved):
thanks !

PR #2102, opened 2025-02-24
test: make the bound on the base variance check a bit looser

IceTDrinker:
Temp solution to a check that is too strict (and in that case the obtained value was actually smaller than expected _which is good_)

@mayeul-zama do we have proper intervals for those checks ?


mayeul-zama:
There's a formula of the variance of the variance estimation: https://math.stackexchange.com/a/73080

I'm not sure it's worth spending time using it for computing this intervals

mayeul-zama (approved):

IceTDrinker:
a better fix could be to allow the result to be smaller, while keeping a tight bound when it's bigger

PR #2107, opened 2025-02-25
chore(ci): fix workflow concurrency condition

soonum:
Referencing current branch using `github.head_ref` is a leftover from handling pull_request_target event. 
This event being removed, there is no need to be specific and we can instead use `github.workflow_ref` which is more robust.

### DIFF
--- a/.github/workflows/aws_tfhe_backward_compat_tests.yml
+++ b/.github/workflows/aws_tfhe_backward_compat_tests.yml
@@ -51,7 +51,7 @@ jobs:
     name: Backward compatibility tests
     needs: [ setup-instance ]
     concurrency:
-      group: ${{ github.workflow }}_${{ github.head_ref || github.ref }}
+      group: ${{ github.workflow_ref }}
       cancel-in-progress: true
     runs-on: ${{ needs.setup-instance.outputs.runner-name }}
     steps:
@@ -73,7 +73,7 @@ jobs:
       - name: Use specific data branch
         if: ${{ contains(github.event.pull_request.labels.*.name, 'data_PR') }}
         env:
-          PR_BRANCH: ${{ github.head_ref || github.ref_name }}
+          PR_BRANCH: ${{ github.ref_name }}
         run: |
           echo "BACKWARD_COMPAT_DATA_BRANCH=${PR_BRANCH}" >> "${GITHUB_ENV}"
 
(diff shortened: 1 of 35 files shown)


.github/workflows/aws_tfhe_backward_compat_tests.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     name: Backward compatibility tests
     needs: [ setup-instance ]
     concurrency:
-      group: ${{ github.workflow }}_${{ github.head_ref || github.ref }}
+      group: ${{ github.workflow_ref }}
```
IceTDrinker, line 54: what is workflow ref ? vs github.ref we were using before the PR target thing ?

soonum: Before we were using two context variable:

* `github.workflow` uses what's in `name:` directive at the top of the workflow. If there is none it falls back to workflow path.
* `github.ref` will print the branch in the form `refs/heads/my_dev_branch`

Now we can do the same with only one:

* `github.workflow_ref` will display `path/to/workflow@refs/heads/my_dev_branch`

IceTDrinker: neat!

.github/workflows/aws_tfhe_tests.yml
```diff
@@ -169,7 +169,7 @@ jobs:
       (github.event_name == 'pull_request' && needs.setup-instance.result != 'skipped')
     needs: [ should-run, setup-instance ]
     concurrency:
-      group: ${{ github.workflow }}_${{github.event_name}}_${{ github.head_ref || github.ref }}
+      group: ${{ github.workflow }}_${{github.event_name}}_${{ github.ref }}
```
IceTDrinker, line 172: do you know why we have the event name here ?

IceTDrinker (commented):
Just one more question

.github/workflows/aws_tfhe_tests.yml
soonum: Oh i get it now, we have a `schedule` event occurring everyday @ 1a.m everyday.
For devs outside of Paris timezone, it might be troublesome if they have their CI in their PR being cancelled because of this `schedule` event.
I'll let the `event_name` but I'll refactor the couple `workflow` + `ref` using `workflow_ref` like elsewhere.

IceTDrinker (approved):
Thanks !

PR #2111, opened 2025-02-26
fix(c-api): add missing casts

tmontaigu:
cast_into FheUint{12, 512, 1024, 2048} were missing from the C API


IceTDrinker (approved):
Thanks !

PR #2113, opened 2025-02-26
docs(js): update JS example

tmontaigu:
The example was still using CompactFheUint32List
which as been removed in favor of the more generic CompactCiphertextList

closes https://github.com/zama-ai/tfhe-rs-internal/issues/798


IceTDrinker (approved):
Thanks !

IceTDrinker:
Will need a backport

PR #2117, opened 2025-02-27
fix: fix the atomic pattern used to cast in trivium and a test in shortint

IceTDrinker:
- parameters are optimized for a clean ciphertext, the ciphertext being keyswitched was noisy

Circuit is

```
        x = Input
        x = bootstrap(x, input_bootstrap_key)
        x = keyswitch(x, cast_keyswitch_key)
        x = bootstrap(x, output_bootstrap_key)
```

we were doing 

```
        x = Input
        x = bootstrap(x, input_bootstrap_key)
        x = x * 3
        x = keyswitch(x, cast_keyswitch_key)
        x = bootstrap(x, output_bootstrap_key)
```

for parameters with the drift mitigation this means that the noise after the multiplication by 3 is very large (as it gets mitigated right before the PBS)

### DIFF
--- a/apps/trivium/src/trans_ciphering/mod.rs
+++ b/apps/trivium/src/trans_ciphering/mod.rs
@@ -48,6 +48,8 @@ fn transcipher_from_1_1_stream(
 ) -> FheUint64 {
     assert_eq!(stream.len(), 64);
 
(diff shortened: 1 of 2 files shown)


agnesLeroy (approved):

mayeul-zama (approved):
Thanks!

PR #2128, opened 2025-02-28
fix: BlockDecomposer

tmontaigu:
The BlockDecomposer gave the possibility when the number of bits per block was not a multiple of the number of bits in the original integer to force the extra bits of the last block to a particular value.

However, the way this was done could only work when setting these bits to 1, when wanting to set them to 0 it would not work.

Good news is that we actually never wanted to set them to 0, but it should still be fixed for completeness, and allow other feature to be added without bugs

### DIFF
--- a/tfhe/src/integer/block_decomposition.rs
+++ b/tfhe/src/integer/block_decomposition.rs
@@ -2,7 +2,7 @@ use crate::core_crypto::prelude::{CastFrom, CastInto, Numeric};
 use crate::integer::bigint::static_signed::StaticSignedBigInt;
 use crate::integer::bigint::static_unsigned::StaticUnsignedBigInt;
 use core::ops::{AddAssign, BitAnd, ShlAssign, ShrAssign};
-use std::ops::{BitOrAssign, Shl, Sub};
+use std::ops::{BitOrAssign, Not, Shl, Shr, Sub};
 
 // These work for signed number as rust uses 2-Complements
 // And Arithmetic shift for signed number (logical for unsigned)
@@ -14,8 +14,10 @@ pub trait Decomposable:
     + ShrAssign<u32>
     + Eq
     + CastFrom<u32>
+    + Shr<u32, Output = Self>
     + Shl<u32, Output = Self>
     + BitOrAssign<Self>
+    + Not<Output = Self>
 {
 }
 pub trait Recomposable:
@@ -86,13 +88,20 @@ impl<const N: usize> RecomposableFrom<u8> for StaticUnsignedBigInt<N> {}
 impl<const N: usize> DecomposableInto<u64> for StaticUnsignedBigInt<N> {}
 impl<const N: usize> DecomposableInto<u8> for StaticUnsignedBigInt<N> {}
 
+#[derive(Copy, Clone)]
+#[repr(u32)]
+pub enum PaddingBitValue {
+    Zero = 0,
+    One = 1,
+}
+
 #[derive(Clone)]
 pub struct BlockDecomposer<T> {
     data: T,
     bit_mask: T,
     num_bits_in_mask: u32,
     num_bits_valid: u32,
-    padding_bit: T,
+    padding_bit: Option<PaddingBitValue>,
     limit: Option<T>,
 }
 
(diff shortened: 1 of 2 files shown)


tfhe/src/integer/block_decomposition.rs
```diff
@@ -92,27 +94,35 @@ pub struct BlockDecomposer<T> {
...
 
     pub fn new(value: T, bits_per_block: u32) -> Self {
-        Self::new_(value, bits_per_block, None, T::ZERO)
+        Self::new_(value, bits_per_block, None, None)
     }
 
-    fn new_(value: T, bits_per_block: u32, limit: Option<T>, padding_bit: T) -> Self {
+    fn new_(value: T, bits_per_block: u32, limit: Option<T>, padding_bit: Option<u32>) -> Self {
+        if let Some(padding_bit) = padding_bit {
+            assert!(padding_bit <= 1);
```
IceTDrinker, line 124: how about a PaddingBit struct which holds an Option<?> with repr(transparent) and a new_one(), new_zero() function ?

? can be bool or u32 as you like

will avoid the assert to enforce the invariant, wdyt ?

you could even have consts like PADDING_BIT_ZERO, PADDING_BIT_ONE if you like

tfhe/src/integer/block_decomposition.rs
```diff
@@ -159,11 +194,18 @@ where
...
             // This will be the case when self.num_bits_in_mask is not a multiple
-            // of T::BITS.  We replace bits that
-            // do not come from the actual T but from the padding
-            // intoduced by the shift, to a specific value.
-            for i in self.num_bits_valid..self.num_bits_in_mask {
-                masked |= self.padding_bit << i;
+            // of T::BITS.
+            //
+            // We replace bits that do not come from the actual T but from the padding
+            // intoduced by the shift, to a specific value. if one was provided.
```
IceTDrinker, line 200: value. -> value, 

?

tfhe/src/integer/block_decomposition.rs
```diff
@@ -159,11 +194,18 @@ where
...
+            if let Some(padding_bit) = self.padding_bit {
+                let padding_mask = (self.bit_mask >> self.num_bits_valid) << self.num_bits_valid;
+                masked = masked & !padding_mask;
+
+                let padding_bit = T::cast_from(padding_bit);
+                for i in self.num_bits_valid..self.num_bits_in_mask {
+                    masked |= padding_bit << i;
```
IceTDrinker, line 207: didn't you say this does not work if padding_bit is 0 ? Or am I misreading something ?

tfhe/src/integer/block_decomposition.rs
```diff
@@ -310,6 +322,35 @@ mod tests {
...
+    fn test_bit_block_decomposer_3() {
+        let bits_per_block = 3;
+
+        let value = -1i8;
+        let blocks = BlockDecomposer::new(value, bits_per_block)
+            .iter_as::<u64>()
+            .collect::<Vec<_>>();
+        // We expect the last block padded with 1s as a consequence of arithmetic shift
+        let expected_blocks = vec![7, 7, 7];
+        assert_eq!(expected_blocks, blocks);
```
IceTDrinker, line 335: why do we expect this ?

Block Decomposer has a 2's complement representation ? i.e. if there are no bits left we expect the negative values to have all 1s set in the equivalent bit representation, i.e. here for 3 bits per block, -1i8 gets sign extended to 9 bits ?

IceTDrinker (commented):
Question on what the block decomposer actually does

tfhe/src/integer/block_decomposition.rs
tmontaigu: I think bool would be enough

tfhe/src/integer/block_decomposition.rs
tmontaigu: We expect this because the block décomposer right shifts the value at each iteration to keep the next bits to be extracted at the least significant position

And here since the `T` is a i8 the right shift will be an arithmetic shift

tfhe/src/integer/block_decomposition.rs
IceTDrinker: ah, don't you need the Option thing to discrimnate some specific case ?

tmontaigu: Its only needed in the struct itself and in the private (new_) constructor as we use a different constructor if the user whishes to specify the padding

IceTDrinker: I just don't know if the bool is the best tool to express what the user wants, don't recall if this struct is pub or pub(crate)

tfhe/src/integer/block_decomposition.rs
IceTDrinker: Ok I guess my question is: is that a behavior we want ? it does not sound like what I would want from the decomposer ? unsure unless it's clear it is "sign aware"

tmontaigu: It's the behaviour we want, because when we decompose a scalar to a `n` amount of blocks that would need the scalar to have more bits (ie decompose an 8bit scalar to 16 blocks of 2_2, so 32bits) we want the 'sign expansion' to happen

so that `decrypt(encrypt(-128i8, 2_2, 16blocks)) = -128i32 and not 128u32`

If the block decomposer did not do this by default, then each creation of a block decomposer would need to make sure the padding_bit value given is proper (i.e BlockDecomposer::with_pading_bit(..., scalar < Scalar::ZERO))

Also, by exploiting this arithmetic shift thing, if we want to decompose a smaller type to a bigger number of blocks, we can just treat the scalar as having more bits than it truly has, and everything work, otherwise we would need to `.chain(...)` the block decomposer with `(msg_mod - 1) * (scalar < Scalar::ZERO) `

tfhe/src/integer/block_decomposition.rs
tmontaigu: The struct is pub

tfhe/src/integer/block_decomposition.rs
tmontaigu: All these could be abstracted with a special constructor (which is sort of planned in the next PR https://github.com/zama-ai/tfhe-rs/pull/2133 that is adding a `with_block_count` constructor) , but relying on sign expansion of arithmetic shift allows the constructor to simply return a BlockDecomposer and not an `impl Iterator<Item = T>` or  a `Chain<BlockDecomposer>`

IceTDrinker: alright fine by me

tfhe/src/integer/block_decomposition.rs
IceTDrinker: Then I don't feel like bool is the best to carry intent here

tmontaigu: So you'd prefer a enum `PaddingBitValue { Zero, One }` and possibly None ?

IceTDrinker: Don’t you think ?

IceTDrinker: integer is fine to break

IceTDrinker (approved):
Thanks!

PR #2132, opened 2025-02-28
chore(ci): fallback on permanent h100 instance on shortage

soonum:
When a shortage occurs on n3-H100x1 instances on Hyperstack, we'll fall back on the permanent one registered on GitHub. This can be done by using 'h100x1' as runner label to run a job on it.


agnesLeroy (commented):
Thank you! :pray: Could we fall back also for single H100 benchmarks you think?

soonum:
> Thank you! 🙏 Could we fall back also for single H100 ks/pbs, integer & erc20 benchmarks you think?

Done and the fallback has been successfully tested manually.

agnesLeroy (approved):

agnesLeroy (approved):

PR #2143, opened 2025-03-03
chore(ci): force installation of toolchain for tfhe-lints

IceTDrinker:


IceTDrinker (commented):
I can't self approve as this was a PR originated by me but it seems PCC worked so good to me

nsarlin-zama (approved):

PR #2147, opened 2025-03-03
feat(tfhe-ntt,gf64) Add custom root-of-unity for Goldilock Prime

Bapt-Roux:
Those root-of-unity enable friendly twiddle generation with low hamming-weigth. And thus, enable to replace some multiplication with simple shift.

This is hide behind the feature flag "gf64-ru64-8" and exposed at tfhe level through "ntt-gf64-ru64-8".


IceTDrinker:
done in https://github.com/zama-ai/tfhe-rs/pull/2213

PR #2153, opened 2025-03-05
feat(gpu): Implement 128 bit classic pbs

bbarbakadze:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
@@ -309,7 +309,7 @@ __global__ void
 batch_NSMFFT_128(double *in_re_hi, double *in_re_lo, double *in_im_hi,
                  double *in_im_lo, double *out_re_hi, double *out_re_lo,
                  double *out_im_hi, double *out_im_lo, double *buffer) {
-  extern __shared__ double sharedMemoryFFT[];
+  extern __shared__ double sharedMemoryFFT128[];
   double *re_hi, *re_lo, *im_hi, *im_lo;
 
   if (SMD == NOSM) {
@@ -322,10 +322,10 @@ batch_NSMFFT_128(double *in_re_hi, double *in_re_lo, double *in_im_hi,
     im_lo =
         &buffer[blockIdx.x * params::degree / 2 * 4 + params::degree / 2 * 3];
   } else {
-    re_hi = &sharedMemoryFFT[params::degree / 2 * 0];
-    re_lo = &sharedMemoryFFT[params::degree / 2 * 1];
-    im_hi = &sharedMemoryFFT[params::degree / 2 * 2];
-    im_lo = &sharedMemoryFFT[params::degree / 2 * 3];
+    re_hi = &sharedMemoryFFT128[params::degree / 2 * 0];
+    re_lo = &sharedMemoryFFT128[params::degree / 2 * 1];
+    im_hi = &sharedMemoryFFT128[params::degree / 2 * 2];
+    im_lo = &sharedMemoryFFT128[params::degree / 2 * 3];
   }
 
   Index tid = threadIdx.x;
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
@@ -3,6 +3,8 @@
 
 #include "device.h"
 #include "fft/bnsmfft.cuh"
+#include "fft128/fft128.cuh"
+
 #include "pbs/programmable_bootstrap.h"
 #include "pbs/programmable_bootstrap_multibit.h"
 #include "polynomial/parameters.cuh"
@@ -250,5 +252,116 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
   cuda_drop_async(buffer, stream, gpu_index);
   cudaFreeHost(h_bsk);
 }
+template <int N>
+__global__ void dprint_array(double *a) {
+  if (threadIdx.x == 0 && blockIdx.x == 0) {
+    for (int i = 0; i < N; i++)
+      printf("%.30f\n", a[i]);
+  }
+}
+template <class params>
+void convert_and_transform_128(cudaStream_t stream,
+                               uint32_t gpu_index, double *d_re0, double *d_re1, double *d_im0,
+                               double *d_im1,
+                               __uint128_t const *d_standard, uint32_t number_of_samples) {
+
+  printf("bsk transform\n");
+  size_t required_shared_memory_size = sizeof(double) * params::degree / 2 * 4;
+  int grid_size = number_of_samples;
+  int block_size = params::degree / params::opt;
+  bool full_sm =
+      (required_shared_memory_size <= cuda_get_max_shared_memory(gpu_index));
+  size_t buffer_size = full_sm ? 0 : (size_t)number_of_samples * params::degree / 2 * 4;
+  size_t shared_memory_size = full_sm ? required_shared_memory_size : 0;
+  double *buffer = (double *)cuda_malloc_async(buffer_size, stream, gpu_index);
+
+  // configure shared memory for batch fft kernel
+  if (full_sm) {
+    check_cuda_error(cudaFuncSetAttribute(
+        batch_NSMFFT_128<FFTDegree<params, ForwardFFT>, FULLSM>,
+        cudaFuncAttributeMaxDynamicSharedMemorySize, shared_memory_size));
+    check_cuda_error(cudaFuncSetCacheConfig(
+        batch_NSMFFT_128<FFTDegree<params, ForwardFFT>, FULLSM>,
+        cudaFuncCachePreferShared));
+  }
+
+  // convert u128 into 4 x double
+  batch_convert_u128_to_f128_as_torus<params>
+  <<<grid_size, block_size, 0, stream>>>(d_re0, d_re1, d_im0, d_im1,
+                                         d_standard);
+
+  // call negacyclic 128 bit forward fft.
+  if (full_sm) {
+    batch_NSMFFT_128<FFTDegree<params, ForwardFFT>, FULLSM>
+    <<<grid_size, block_size, shared_memory_size, stream>>>(
+        d_re0, d_re1, d_im0, d_im1, d_re0, d_re1, d_im0, d_im1, buffer);
+  } else {
+    batch_NSMFFT_128<FFTDegree<params, ForwardFFT>, NOSM>
+    <<<grid_size, block_size, shared_memory_size, stream>>>(
+        d_re0, d_re1, d_im0, d_im1, d_re0, d_re1, d_im0, d_im1, buffer);
+  }
+  cuda_drop_async(buffer, stream, gpu_index);
+
+  printf("#cuda\n");
+  cudaDeviceSynchronize();
+
+  dprint_array<params::degree / 2><<<1, 1>>>(d_re0);
+
+}
+
+
+
+inline void cuda_convert_lwe_programmable_bootstrap_key_u128(cudaStream_t stream,
+                                                 uint32_t gpu_index,
+                                                 double *dest, __uint128_t const *src,
+                                                 uint32_t polynomial_size,
+                                                 uint32_t total_polynomials) {
+  cuda_set_device(gpu_index);
+
+  // Here the buffer size is the size of double times the number of polynomials time 4
+  // each polynomial is represented with 4 double array with size polynomial_size / 2
+  // into the complex domain to perform the FFT
+  size_t buffer_size =
+      total_polynomials * polynomial_size / 2 * sizeof(double) * 4;
+
+  __uint128_t *d_standard = (__uint128_t *)cuda_malloc_async(buffer_size, stream, gpu_index);
+  double *d_bsk = (double *)cuda_malloc_async(buffer_size, stream, gpu_index);
+
+  double *d_re0 = d_bsk + 0ULL * total_polynomials * polynomial_size / 2;
+  double *d_re1 = d_bsk + 1ULL * total_polynomials * polynomial_size / 2;
+  double *d_im0 = d_bsk + 2ULL * total_polynomials * polynomial_size / 2;
+  double *d_im1 = d_bsk + 3ULL * total_polynomials * polynomial_size / 2;
+
+  cuda_memcpy_async_to_gpu(d_standard, src, buffer_size,
+                           stream, gpu_index);
+
+
+
+  switch (polynomial_size) {
+    case 256:
+      convert_and_transform_128<AmortizedDegree<256>>(stream, gpu_index, d_re0, d_re1, d_im0,
+          d_im1,
+          d_standard, total_polynomials);
+    case 512:
+      convert_and_transform_128<AmortizedDegree<512>>(stream, gpu_index, d_re0, d_re1, d_im0, d_im1, d_standard, total_polynomials);
+      break;
+    case 1024:
+      convert_and_transform_128<AmortizedDegree<1024>>(stream, gpu_index, d_re0, d_re1, d_im0, d_im1, d_standard, total_polynomials);
+      break;
+    case 2048:
+      convert_and_transform_128<AmortizedDegree<2048>>(stream, gpu_index, d_re0, d_re1, d_im0, d_im1, d_standard, total_polynomials);
+      break;
+    case 4096:
+      convert_and_transform_128<AmortizedDegree<4096>>(stream, gpu_index, d_re0, d_re1, d_im0, d_im1, d_standard, total_polynomials);
+      break;
+    default:
+    PANIC("Cuda error (convert BSK): unsupported polynomial size. Supported "
+          "N's are powers of two in the interval [256..4096].")
+  }
+
+  cuda_drop_async(d_standard, stream, gpu_index);
+  cuda_drop_async(d_bsk, stream, gpu_index);
+}
+
 
 #endif // CNCRT_BSK_H
--- a/tfhe/src/core_crypto/gpu/entities/lwe_bootstrap_key.rs
+++ b/tfhe/src/core_crypto/gpu/entities/lwe_bootstrap_key.rs
@@ -32,11 +32,22 @@ impl CudaLweBootstrapKey {
     where
         InputBskCont::Element: UnsignedInteger,
     {
+        println!("#1");
         let input_lwe_dimension = bsk.input_lwe_dimension();
         let polynomial_size = bsk.polynomial_size();
         let decomp_level_count = bsk.decomposition_level_count();
         let decomp_base_log = bsk.decomposition_base_log();
         let glwe_dimension = bsk.glwe_size().to_glwe_dimension();
+        let split_count = if size_of::<InputBskCont::Element>() == 16 {
+            2
+        } else {
+            1
+        };
+        println!("split_count: {:?}", split_count);
+        println!("input_lwe_dimension: {:?}", input_lwe_dimension);
+        println!("polynomial_size: {:?}", polynomial_size);
+        println!("decomp_level_count: {:?}", decomp_level_count);
+        println!("glwe_dimension: {:?}", glwe_dimension);
 
         // Allocate memory
         let mut d_vec = CudaVec::<f64>::new_multi_gpu(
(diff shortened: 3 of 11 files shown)


backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -9,20 +9,29 @@
 template <typename Torus>
 uint64_t get_buffer_size_full_sm_programmable_bootstrap_step_one(
     uint32_t polynomial_size) {
-  return sizeof(Torus) * polynomial_size +      // accumulator_rotated
-         sizeof(double2) * polynomial_size / 2; // accumulator fft
+  size_t scalar_size = sizeof(Torus);
+  size_t split_count = (scalar_size == 16) ? 2 : 1;
```
agnesLeroy, line 13: Maybe instead of split_count this could be named double_count?

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -9,20 +9,29 @@
...
+         sizeof(double) * 2 * split_count * polynomial_size /
+             2; // accumulator fft
 }
 template <typename Torus>
 uint64_t get_buffer_size_full_sm_programmable_bootstrap_step_two(
     uint32_t polynomial_size) {
-  return sizeof(Torus) * polynomial_size +      // accumulator
-         sizeof(double2) * polynomial_size / 2; // accumulator fft
+  size_t scalar_size = sizeof(Torus);
+  size_t split_count = (scalar_size == 16) ? 2 : 1;
```
agnesLeroy, line 22: Same here

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -9,20 +9,29 @@
...
+         sizeof(double) * 2 * split_count * polynomial_size /
+             2; // accumulator fft
 }
 
 template <typename Torus>
 uint64_t
 get_buffer_size_partial_sm_programmable_bootstrap(uint32_t polynomial_size) {
-  return sizeof(double2) * polynomial_size / 2; // accumulator fft
+  size_t scalar_size = sizeof(Torus);
+  size_t split_count = (scalar_size == 16) ? 2 : 1;
```
agnesLeroy, line 32: Same here

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -9,20 +9,29 @@
 template <typename Torus>
 uint64_t get_buffer_size_full_sm_programmable_bootstrap_step_one(
     uint32_t polynomial_size) {
-  return sizeof(Torus) * polynomial_size +      // accumulator_rotated
-         sizeof(double2) * polynomial_size / 2; // accumulator fft
+  size_t scalar_size = sizeof(Torus);
```
agnesLeroy, line 12: I think we could keep just `sizeof(Torus)` in the formula, it seems easy to read to me, wdyt?

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -215,6 +224,158 @@ template <typename Torus> struct pbs_buffer<Torus, PBS_TYPE::CLASSICAL> {
...
+                       (glwe_dimension + 1);
+        } else if (max_shared_memory < full_sm + minimum_sm_tbc) {
+          device_mem = partial_dm * input_lwe_ciphertext_count * level_count *
+                       (glwe_dimension + 1);
+        }
+
+        // Otherwise, both kernels run all in shared memory
+        d_mem = (int8_t *)cuda_malloc_async(device_mem, stream, gpu_index);
+
+        global_join_buffer = (double *)cuda_malloc_async(
```
agnesLeroy, line 358: This global_join_buffer size could maybe be returned by some function since we allocate it 3 times?

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -75,10 +83,58 @@ public:
...
+      res_re -= carry_re << base_log;
+      res_im -= carry_im << base_log;
+
+      auto out_re = u128_to_signed_to_f128(res_re);
+      auto out_im = u128_to_signed_to_f128(res_im);
+
+      auto out_re_hi = result + 0LL * params::degree / 2;
+      auto out_re_lo = result + 1LL * params::degree / 2;
+      auto out_im_hi = result + 2LL * params::degree / 2;
+      auto out_im_lo = result + 3LL * params::degree / 2;
```
agnesLeroy, line 117: Why do you need 0LL, 1LL, etc. for pointer arithmetics? You could just use 0, 1, 2, 3 no?

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -287,6 +287,25 @@ batch_convert_u128_to_f128_as_torus(double *out_re_hi, double *out_re_lo,
       &in[blockIdx.x * params::degree + params::degree / 2]);
 }
 
+// params is expected to be full degree not half degree
+template <class params>
+__global__ void
+batch_convert_u128_to_f128_strided_as_torus(double *d_out,
```
agnesLeroy, line 293: Why is there strided in the name?

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -356,6 +375,70 @@ batch_NSMFFT_128(double *in_re_hi, double *in_re_lo, double *in_im_hi,
   }
 }
 
+template <class params, sharedMemDegree SMD>
+__global__ void batch_NSMFFT_strided_128(double *d_in, double *d_out,
```
agnesLeroy, line 379: Why is there strided in the name?

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -356,6 +375,70 @@ batch_NSMFFT_128(double *in_re_hi, double *in_re_lo, double *in_im_hi,
...
+  } else {
+    re_hi = &sharedMemoryFFT128[params::degree / 2 * 0];
+    re_lo = &sharedMemoryFFT128[params::degree / 2 * 1];
+    im_hi = &sharedMemoryFFT128[params::degree / 2 * 2];
+    im_lo = &sharedMemoryFFT128[params::degree / 2 * 3];
+  }
+
+  constexpr size_t chunk_size = params::degree / 2 * 4;
+  double *chunk = &d_in[blockIdx.x * chunk_size];
+  double *tmp_re_hi = &chunk[0ULL * params::degree / 2];
```
agnesLeroy, line 402: Do you really need 0ULL, 1ULL, etc. here? You could maybe just use 0, 1, 2, 3?

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -251,4 +285,92 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
   cudaFreeHost(h_bsk);
 }
 
+template <class params>
+void convert_and_transform_128(cudaStream_t stream, uint32_t gpu_index,
```
agnesLeroy, line 289: Maybe the name could be `convert_u128_to_f128_and_forward_fft_128`?

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -251,4 +285,92 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
...
+    check_cuda_error(cudaFuncSetAttribute(
+        batch_NSMFFT_strided_128<FFTDegree<params, ForwardFFT>, FULLSM>,
+        cudaFuncAttributeMaxDynamicSharedMemorySize, shared_memory_size));
+    check_cuda_error(cudaFuncSetCacheConfig(
+        batch_NSMFFT_strided_128<FFTDegree<params, ForwardFFT>, FULLSM>,
+        cudaFuncCachePreferShared));
+  }
+
+  // convert u128 into 4 x double
+  batch_convert_u128_to_f128_strided_as_torus<params>
```
agnesLeroy, line 314: I'm not sure it's necessary to keep strided in the names, as anyway this FFT implementation can only work with the data layout that we use? Wdyt?

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -251,4 +285,92 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
   cudaFreeHost(h_bsk);
 }
 
+template <class params>
+void convert_and_transform_128(cudaStream_t stream, uint32_t gpu_index,
+                               double *d_bsk, __uint128_t const *d_standard,
+                               uint32_t number_of_samples) {
+
```
agnesLeroy, line 292: Here I think it's necessary to set the device

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cu
```diff
@@ -0,0 +1,205 @@
...
+    break;
+  default:
+    PANIC("Cuda error (classical PBS): unsupported polynomial size. "
+          "Supported N's are powers of two"
+          " in the interval [256..4096].")
+  }
+}
+
+template <typename Torus>
+void executor_cuda_programmable_bootstrap_lwe_ciphertext_vector_128(
```
agnesLeroy, line 53: Usually we don't have an `executor_` functions, only `cuda_programmable_bootstrap_lwe_ciphertext_vector_128` in this case. You could directly paste the content of this function there, no?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -172,6 +173,35 @@ __device__ void add_to_torus(double2 *m_values, Torus *result,
...
+
+    if (overwrite_result) {
+      result[tid] = torus_real;
+      result[tid + params::degree / 2] = torus_imag;
+    } else {
+      result[tid] += torus_real;
+      result[tid + params::degree / 2] += torus_imag;
+    }
+    tid = tid + params::degree / params::opt;
+  }
```
agnesLeroy, line 202: Is it expected not to syncthreads here?

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -56,6 +56,64 @@ __device__ void polynomial_product_accumulate_in_fourier_domain(
   }
 }
 
+// Computes result += first * second
+// If init_accumulator is set, assumes that result was not initialized and does
+// that with the outcome of first * second
+template <class params>
+__device__ void polynomial_product_accumulate_in_fourier_domain_128(
+    double *result, double *first, const double *second,
+    bool init_accumulator = false) {
```
agnesLeroy, line 65: Maybe not for this PR but init_accumulator could maybe be a template argument.

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -56,6 +56,64 @@ __device__ void polynomial_product_accumulate_in_fourier_domain(
...
+  int tid = threadIdx.x;
+  if (init_accumulator) {
+    for (int i = 0; i < params::opt / 2; i++) {
+      f128 a_re(first[tid + 0ULL * params::degree / 2],
```
agnesLeroy, line 69: Here maybe you could use simply 0,  1, 2, 3 for indexing?

tfhe/src/core_crypto/gpu/entities/lwe_bootstrap_key.rs
```diff
@@ -37,6 +37,11 @@ impl CudaLweBootstrapKey {
         let decomp_level_count = bsk.decomposition_level_count();
         let decomp_base_log = bsk.decomposition_base_log();
         let glwe_dimension = bsk.glwe_size().to_glwe_dimension();
+        let split_count = if size_of::<InputBskCont::Element>() == 16 {
```
agnesLeroy, line 40: Maybe split_count -> double_count?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -0,0 +1,487 @@
...
+  // decomposition, for the mask and the body (so block 0 will have the
+  // accumulator decomposed at level 0, 1 at 1, etc.)
+  GadgetMatrix<Torus, params> gadget_acc(base_log, level_count, accumulator);
+  gadget_acc.decompose_and_compress_level_128(accumulator_fft, blockIdx.z);
+
+  // We are using the same memory space for accumulator_fft and
+  // accumulator_rotated, so we need to synchronize here to make sure they
+  // don't modify the same memory space at the same time
+  // Switch to the FFT space
+  auto acc_fft_re_hi = accumulator_fft + 0LL * params::degree / 2;
```
agnesLeroy, line 125: Same question about indexing

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -0,0 +1,487 @@
...
+    }
+  }
+}
+
+template <typename params>
+__host__ void scratch_programmable_bootstrap_128(
+    cudaStream_t stream, uint32_t gpu_index, pbs_buffer_128<CLASSICAL> **buffer,
+    uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t level_count,
+    uint32_t input_lwe_ciphertext_count, bool allocate_gpu_memory) {
+
```
agnesLeroy, line 290: I think here we need to set device to gpu_index, otherwise the shared mem configuration on multiple devices might be broken.

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -341,19 +384,30 @@ pub unsafe fn convert_lwe_programmable_bootstrap_key_async<T: UnsignedInteger>(
...
-            stream_ptr,
-            streams.gpu_indexes[i].get(),
-            dest.as_mut_c_ptr(i as u32),
-            src.as_ptr().cast(),
-            input_lwe_dim.0 as u32,
-            glwe_dim.0 as u32,
-            l_gadget.0 as u32,
-            polynomial_size.0 as u32,
-        );
+        if size_of::<T>() == 16 {
```
agnesLeroy, line 388: I'm not sure that this is the best way to do this in Rust. @IceTDrinker maybe you could give us some advice here? We have convert_lwe_programmable_bootstrap_key_async that should call the u128 entry point or the u64 one depending on the type T. Is using an if in this way alright?

agnesLeroy (commented):
Hey @bbarbakadze! Thanks a lot for this PR :pray: This is a very significant contribution and I (as well as the whole team I think) really appreciate the effort you put there. Here comes my review, overall it's only minor comments you'll see. @pdroalves don't hesitate to tell us as well what you think about naming, always good to have several pairs of eyes on that. :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
bbarbakadze: Yes I agree.

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
bbarbakadze: since we have 64 bit pointers I used `LL`, maybe compiler would cast it without explicitely writing it need to double check, I am not sure we need those `LL` s

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
bbarbakadze: there are two versions of batch convert:
1) `[re_hi_0, re_hi_1, ... re_hi_n, re_lo_0, re_lo_1, ... re_lo_n, im_hi_0, im_hi_1, ..., im_hi_n,  im_lo_0, im_lo_1, ..., im_lo_n]`
2) strided one: `[re_hi_0, re_lo_0, im_hi_0, im_lo_0, re_hi_1, re_lo_1, im_hi_1, im_lo_1, ...,re_hi_n, re_lo_n, im_hi_n, im_lo_n, ]`

I need to add comments there.

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
bbarbakadze: depends on how what pattern was used to store `re_hi, re_lo, im_hi and im_lo` parts we either call
`batch_NSMFFT_strided_128` or `batch_NSMFFT_128`

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
agnesLeroy: Aaah ok! Hmm do you observe that one is more efficient than the other? Do we need to keep both?

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
bbarbakadze: I will double check it if we need it and remove if it is not necessary

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
bbarbakadze: we have two options  for it strided and non strided. lets discuss it.

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
bbarbakadze: we don't need sync here, but will double check

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
agnesLeroy: From our discussion the idea is to drop the coalesced version, we would add it back if needed if we do the 128 bit multi-bit PBS at some point. Not sure we'll do it because key sizes will become an issue.

tfhe/src/core_crypto/gpu/mod.rs
agnesLeroy: Actually it is pretty convenient this way, we only have 1 entry point for the GPU bootstrap in Rust, regardless whether it's u64 or u128, I quite like it.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
bbarbakadze: will calculate once in the beginning of the constructor  and will use the variable in those three places, once this will be used only in that constructor I don't think it is worth to create another function that returns the size

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cu
bbarbakadze: I would keep it like this,  when we add new variants of 128 bit PBS like CG, TBC or even multi_bit variants there will be another switch for them and it will be a big amount of code with nested switches without executor.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
bbarbakadze: we are not setting device in scratch for 64 bit PBS do we really need it here

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cu
agnesLeroy: Ok :+1:

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
agnesLeroy: We should have it in both I think. It's not so much of an issue for 64bit PBS because the device is set when creating int_radix_lut before we call the scratch_pbs on each device, but still would be better to set it in the scratch.

agnesLeroy (approved):
Awesome! Thank you very much @bbarbakadze!

PR #2156, opened 2025-03-05
chore(ci): fix permanent instance selection condition

soonum:
Due to 'continue-on-error' directive 'use-permanent-instance' step could not rely on failure() function.

I've tested the behavior, and now it falls back on the permanent instance.


IceTDrinker (approved):
Thanks !

PR #2161, opened 2025-03-07
docs: change svg benchmark tables appearance for pbs

soonum:
Fresh SVGs for PBS benchmarks results.
I'm wondering if we should keep these new tables with the same overall size in width than integer ones, or if we should make it more fitted ? @yuxizama


yuxizama (approved):
It looks very polished! 
Thank you @soonum

IceTDrinker (commented):
I see at least two arrays where the content changed ?

IceTDrinker:
you can see the diff in files changed

soonum:
Oh yes ! You're right, it misses the multi-bit PBS in the results for CPU benchmarks.

IceTDrinker (changes requested):
It's displaying FheUint in some cases which does not really apply to PBS

IceTDrinker (approved):
Thanks, good to backport !

IceTDrinker:
don't forget to backport please :)

PR #2162, opened 2025-03-07
chore: pre-generate keyswitching keys for shortint tests

IceTDrinker:
- we run in a cross process race condition which fucks up the key file
- no rust crate seems to help and linux locks are just a fucking mess
- also avoid truncating file when we are going to write to it, get a lock first

Added the key gen for the test as well to avoid issues

### DIFF
--- a/tfhe/examples/utilities/generates_test_keys.rs
+++ b/tfhe/examples/utilities/generates_test_keys.rs
@@ -27,6 +27,16 @@ use tfhe::shortint::parameters::{
 };
 use tfhe::shortint::MultiBitPBSParameters;
 
+const KSK_PARAMS: [(
+    ClassicPBSParameters,
+    ClassicPBSParameters,
+    ShortintKeySwitchingParameters,
+); 1] = [(
+    V1_0_PARAM_MESSAGE_1_CARRY_1_KS_PBS_GAUSSIAN_2M128,
+    V1_0_PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128,
+    V1_0_PARAM_KEYSWITCH_1_1_KS_PBS_TO_2_2_KS_PBS_GAUSSIAN_2M128,
+)];
+
 fn client_server_keys() {
     let matches = Command::new("test key gen")
         .arg(
@@ -44,6 +54,9 @@ fn client_server_keys() {
                 .exclusive(true),
         )
         .get_matches();
+
+    // Always generate those as they may be used in the different cases
+    generate_ksk_keys(&KSK_PARAMS);
 
     // If set using the command line flag "--ladner-fischer" this algorithm will be used in
     // additions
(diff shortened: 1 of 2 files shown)


nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2162)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`tfhe/src/keycache/mod.rs` line 114 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2162#-OKkDnok8i8MmvLXCjOw:-OKkDnol6v9y9RUfSWF9:b9gdwg4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c269fbc6eaf4e2361d0c2f1047814c1da3f0d433/tfhe/src/keycache/mod.rs#L114)):*
> ```XML
>                 let file = File::open(&path_buf).unwrap();
>                 // TODO Manage file locking for inter process stuff, unfortunately linux locks are a
>                 // mess and nothing seems to work
> ```

looks like there is a File::lock method in std that will be stabilized with 1.87: https://github.com/rust-lang/rust/pull/136794
Maybe it will be worth retrying when this is released 
___
*[`tfhe/examples/utilities/generates_test_keys.rs` line 59 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2162#-OKkBbGS19eZAENezjcK:-OKkBbGS19eZAENezjcL:b70i3gi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c269fbc6eaf4e2361d0c2f1047814c1da3f0d433/tfhe/examples/utilities/generates_test_keys.rs#L59)):*
> ```XML
> 
>     // Always generate those as they may be used in the different cases
>     generate_ksk_keys(&KSK_PARAMS);
> ```

This is not directly linked to this pr but I think it should be "generate_ks_keys"

IceTDrinker:
@nsarlin-zama need changes or approved ?

nsarlin-zama (approved):
Depends on if you want to fix the function name.

But this is not really important so approved anyways, thanks!

IceTDrinker:
> Depends on if you want to fix the function name.
> 
> But this is not really important so approved anyways, thanks!

In the current case, even if the failure is rare I'd rather have it merged sooner than later :)

nsarlin-zama:
Oh I was thinking maybe instead of doing the truncate after lock thing we could use [create_new](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new) and skip in case of error ?

IceTDrinker:
> Oh I was thinking maybe instead of doing the truncate after lock thing we could use [create_new](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new) and skip in case of error ?

Can be done later if you want to, I don't think it mattered much anyways as we pre generated the keys, not sure in terms of cache consistency what our various tests expect

IceTDrinker:
I mean create_new probably would have avoided the issue we had most likely, but then what are we supposed to do if the write fails ? Meaning two test processes have generated keys

nsarlin-zama:
I think if the second process just use its own key but does not store it this is equivalent as what's happening here where the first key is overwritten by the second ?

Or worst case the second process could drop its generated key and reload the one from the file

IceTDrinker:
> I think if the second process just uses its own key but does not store it this is equivalent as what's happening here where the first key is overwritten by the second ?
> 
> Or worst case the second process could drop its generated key and reload the one from the file

I don't think the second option would work with how the keycache is organized

PR #2173, opened 2025-03-13
fix(gpu): fix logic of host_compare_with_zero_equality

agnesLeroy:

### PR content/description

host_compare_with_zero_equality was not following the same logic as the CPU, because it included the call to are_all_comparison_blocks_true. This PR fixes it, and thus avoids doing redundant computation for scalar comparisons (and division).

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
@@ -70,7 +70,6 @@ __host__ void host_unsigned_integer_div_rem_kb(
   auto did_not_overflow = mem_ptr->did_not_overflow;
   auto overflow_sum = mem_ptr->overflow_sum;
   auto overflow_sum_radix = mem_ptr->overflow_sum_radix;
-  auto tmp_1 = mem_ptr->tmp_1;
   auto at_least_one_upper_block_is_non_zero =
       mem_ptr->at_least_one_upper_block_is_non_zero;
   auto cleaned_merged_interesting_remainder =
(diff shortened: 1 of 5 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
```diff
@@ -343,7 +343,7 @@ __host__ void host_unsigned_integer_div_rem_kb(
         is_at_least_one_comparisons_block_true<Torus>(
             streams, gpu_indexes, gpu_count,
             at_least_one_upper_block_is_non_zero, tmp_1,
-            mem_ptr->comparison_buffer, bsks, ksks, tmp_1->num_radix_blocks);
+            mem_ptr->comparison_buffer, bsks, ksks, 1);
```
guillermo-oyarzun, line 346: since we are just checking one block, isn't better to use a cmux?

agnesLeroy: Hmm I don't know, it's done in this way on the CPU. @tmontaigu is there a specific reason?

agnesLeroy: Ah wait I have the logic wrong still.

agnesLeroy: Ok I need to fix the logic of `compare_blocks_with_zero` to match the CPU

agnesLeroy:
@guillermo-oyarzun the PR changed completely :sweat_smile:

backends/tfhe-cuda-backend/cuda/src/integer/scalar_comparison.cuh
```diff
@@ -813,9 +828,9 @@ __host__ void host_integer_radix_scalar_equality_check_kb(
       PANIC("Cuda error: integer operation not supported")
     }
 
-    host_compare_with_zero_equality<Torus>(msb_streams, gpu_indexes, gpu_count,
```
pdroalves, line 816: If I read this right, we are removing `are_all_comparisons_block_true()` from inside `host_compare_blocks_with_zero()` to save ops when the latter is not need, as here and inside division. Is that right?

pdroalves (approved):
Good catch. `compare_blocks_with_zero()` indeed doesn't call `are_all_comparisons_block_true()`. I don't remember why we merged both in the CUDA backend but clearly this PR is a better way.

IMO code is good to merge as soon as CI is green.

backends/tfhe-cuda-backend/cuda/src/integer/scalar_comparison.cuh
agnesLeroy: Exactly

agnesLeroy:
> Good catch. `compare_blocks_with_zero()` indeed doesn't call `are_all_comparisons_block_true()`. I don't remember why we merged both in the CUDA backend but clearly this PR is a better way.
> 
> IMO code is good to merge as soon as CI is green.

Ah the failing test was due to a wrong assert in host_compare_with_zero, all good now

PR #2175, opened 2025-03-14
chore(ci): install dependencies as standalone job

soonum:
Installing dependencies several times, due to matrix strategy, lead to job failure. Now, if the workflow uses the remote instance, the dependencies will be installed only once.


agnesLeroy (commented):
Thanks a lot @soonum! May I ask you to do the same for benchmark_gpu_erc20_common.yml please? :slightly_smiling_face:

soonum:
> Thanks a lot @soonum! May I ask you to do the same for benchmark_gpu_erc20_common.yml please? 🙂

No need to,  since it doesn't use a matrix to run multiple benchmarks on the same instance.

agnesLeroy (approved):

PR #2186, opened 2025-03-20
chore(ci): cache backward compatibility data

soonum:
Git LFS transfers use a lot of bandwidth. Since data used to test backward compatibility won't change every day, we can leverage GitHub cache action.


nsarlin-zama (approved):
Nice, thanks! 

Maybe this should be backported to the v1.0 branch to use the new AMI there too ? Otherwise if we upgrade the data repo version it will break in v1.0

IceTDrinker:
thanks ! very nicely done :)

IceTDrinker (approved):

PR #2190, opened 2025-03-21
chore(ci): fix size benchmark

IceTDrinker:
- don't expand, we have tests for that, the server key would be required with the "new" parameters we are interested in


agnesLeroy:
What are the new parameters you're referring to?

IceTDrinker:
> What are the new parameters you're referring to?

Blockchain vs. what we did a very long time ago with "compact public key parameters"

agnesLeroy (approved):
Oooh ok, all good then.

PR #2195, opened 2025-03-21
feat(integer): boolean_scalar_dot_prod

tmontaigu:
Add function do compute a dot product between a vector of encrypted boolean values and a vector of clear values

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)

### DIFF
--- a/tfhe/src/core_crypto/commons/numeric/mod.rs
+++ b/tfhe/src/core_crypto/commons/numeric/mod.rs
@@ -130,3 +130,9 @@ where
         }
     }
 }
+
+pub trait OverflowingAdd<Rhs> {
+    type Output;
+
+    fn overflowing_add(self, other: Rhs) -> (Self::Output, bool);
+}
--- a/tfhe/src/core_crypto/commons/numeric/signed.rs
+++ b/tfhe/src/core_crypto/commons/numeric/signed.rs
@@ -1,4 +1,5 @@
 use super::{CastFrom, CastInto, Numeric, SignedNumeric, UnsignedInteger};
+use crate::core_crypto::prelude::OverflowingAdd;
 use std::ops::{
     Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign,
     Mul, MulAssign, Neg, Not, Rem, RemAssign, Shl, ShlAssign, Shr, ShrAssign, Sub, SubAssign,
@@ -29,6 +30,7 @@ pub trait SignedInteger:
     + ShlAssign<usize>
     + Shr<usize, Output = Self>
     + ShrAssign<usize>
+    + OverflowingAdd<Self, Output = Self>
     + CastFrom<f64>
     + CastInto<f64>
 {
@@ -59,6 +61,15 @@ macro_rules! implement {
 
         impl SignedNumeric for $Type {
             type NumericUnsignedType = $UnsignedType;
+        }
+
+        impl OverflowingAdd<Self> for $Type {
+            type Output = Self;
+
+            #[inline]
+            fn overflowing_add(self, rhs: Self) -> (Self, bool) {
+                self.overflowing_add(rhs)
+            }
         }
 
         impl SignedInteger for $Type {
--- a/tfhe/src/core_crypto/commons/numeric/unsigned.rs
+++ b/tfhe/src/core_crypto/commons/numeric/unsigned.rs
@@ -1,4 +1,5 @@
 use super::{CastFrom, CastInto, Numeric, SignedInteger, UnsignedNumeric};
+use crate::core_crypto::prelude::OverflowingAdd;
 use std::ops::{
     Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign,
     Mul, MulAssign, Not, Rem, RemAssign, Shl, ShlAssign, Shr, ShrAssign, Sub, SubAssign,
@@ -30,6 +31,7 @@ pub trait UnsignedInteger:
     + ShlAssign<usize>
     + Shr<usize, Output = Self>
     + ShrAssign<usize>
+    + OverflowingAdd<Self, Output = Self>
     + CastFrom<Self::Signed>
     + CastFrom<f64>
     + CastInto<f64>
@@ -83,8 +85,6 @@ pub trait UnsignedInteger:
     #[must_use]
     fn wrapping_shr(self, rhs: u32) -> Self;
     #[must_use]
-    fn overflowing_add(self, rhs: Self) -> (Self, bool);
-    #[must_use]
     fn overflowing_sub(self, rhs: Self) -> (Self, bool);
     #[must_use]
     fn is_power_of_two(self) -> bool;
@@ -119,6 +119,15 @@ macro_rules! implement {
 
         impl UnsignedNumeric for $Type {
             type NumericSignedType = $SignedType;
+        }
+
+        impl OverflowingAdd<Self> for $Type {
+            type Output = Self;
+
+            #[inline]
+            fn overflowing_add(self, rhs: Self) -> (Self, bool) {
+                self.overflowing_add(rhs)
+            }
         }
 
         impl UnsignedInteger for $Type {
@@ -217,10 +226,6 @@ macro_rules! implement {
             #[inline]
             fn wrapping_pow(self, exp: u32) -> Self {
                 self.wrapping_pow(exp)
-            }
-            #[inline]
-            fn overflowing_add(self, rhs: Self) -> (Self, bool) {
-                self.overflowing_add(rhs)
             }
             #[inline]
             fn overflowing_sub(self, rhs: Self) -> (Self, bool) {
--- a/tfhe/src/high_level_api/array/mod.rs
+++ b/tfhe/src/high_level_api/array/mod.rs
@@ -15,14 +15,18 @@ use crate::high_level_api::array::traits::HasClear;
 use crate::high_level_api::global_state;
 #[cfg(feature = "gpu")]
 use crate::high_level_api::global_state::with_thread_local_cuda_streams;
-use crate::high_level_api::integers::FheUintId;
+use crate::high_level_api::integers::{FheIntId, FheUintId};
 use crate::high_level_api::keys::InternalServerKey;
-use crate::{FheBool, FheId, FheUint};
-use std::ops::RangeBounds;
+use crate::{FheBool, FheId, FheInt, FheUint, Tag};
+use std::ops::{AddAssign, Mul, RangeBounds};
 use traits::{ArrayBackend, BackendDataContainer, BackendDataContainerMut};
 pub use traits::{IOwnedArray, Slicing, SlicingMut};
 
 use crate::array::stride::DynDimensions;
+use crate::core_crypto::prelude::{Numeric, OverflowingAdd, SignedNumeric, UnsignedNumeric};
+use crate::integer::block_decomposition::DecomposableInto;
+use crate::integer::RadixCiphertext;
+use crate::prelude::{CastFrom, CastInto};
 pub use cpu::{
     CpuFheBoolArray, CpuFheBoolSlice, CpuFheBoolSliceMut, CpuFheIntArray, CpuFheIntSlice,
     CpuFheIntSliceMut, CpuFheUintArray, CpuFheUintSlice, CpuFheUintSliceMut, FheBoolId,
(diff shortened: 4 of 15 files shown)


nsarlin-zama (commented):
Some small comments but the algorithm looks good

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-:-OMC4UJzC1bAorclTy7w:br5r1hw)* status: all files reviewed, 4 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 16 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBuChf71Ao38h4ioIM:-OMBuChf71Ao38h4ioIN:bsay7y9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L16)):*
> ```XML
>         boolean_blocks: &[BooleanBlock],
>         clears: &[Clear],
>         n_blocks: u32,
> ```

If I understand correctly, this is the number of blocks in the returned Ciphertext, maybe this could be made explicit in the comment above
___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBgKNq1X_GlFS00T9n:-OMBgKNq1X_GlFS00T9o:bfey5ss) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L33)):*
> ```XML
>         );
> 
>         assert!(!boolean_blocks.is_empty(), "operands must not be empty");
> ```

Maybe this could be an error since this is not an overloaded operator ? (same for the other assert above)
If we want to keep the panic I think this should be documented here and in the calling functions.

Also we could check that all the clear in the list fit in `n_blocks * self.message_modulus().ilog2()`  ?
___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 80 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMC1_X24Gn8Pyul_YhS:-OMC1_X24Gn8Pyul_YhT:b-ub7bpj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L80)):*
> ```XML
> 
>                             let shift = block_index as u32 * self.message_modulus().0.ilog2();
>                             ((summed_clear >> shift) & block_mask).cast_into()
> ```

I don't know if it's worth the effort, but there are cases where we know that the resulting block will be 0 whatever the boolean are, so we can skip the lut applications. 
For example if the scalar are small and n_block is big, we know that the high blocks will be 0 at this step. One thing to do could be to iterate not until n_block but until something like `clear_chunk.sum().ilog2() / packing_size` (or n_block in case of overflow) and fill the remaining blocks with zeros.
___
*[`tfhe/src/high_level_api/array/mod.rs` line 436 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBc_wv4WgkE0-dJgsH:-OMBc_wv4WgkE0-dJgsI:br6tk8z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/high_level_api/array/mod.rs#L436)):*
> ```XML
>     clears: &[Clear],
>     num_bits: u32,
>     create_fhe_value: F,
> ```

Instead of using a callback this could return a RadixCiphertext or a Vec of blocks and have the constructor in the calling function ?
Or maybe this is too cumbersome because of the tag

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 4 unresolved discussions (waiting on @nsarlin-zama)

___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBgKNq1X_GlFS00T9n:-OMCgA81EDhI0jp9D0W1:b-8zk46t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L33)):*

Maybe this could be an error since this is not an overloaded operator ? (same for the other assert above)
If we want to keep the panic I think this should be documented here and in the calling functions.

Also we could check that all the clear in the list fit in `n_blocks * self.message_modulus().ilog2()`  ?
</blockquote></details>

Hum I think we should discuss if this something we want to do, as none of the integer ops return a Result, despite multiple of them having assert, so before changing that pattern we should see if that's something we want to do
___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 80 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMC1_X24Gn8Pyul_YhS:-OMCfnZN8o5Lmwk0Emlw:b-jmhhtv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L80)):*

I don't know if it's worth the effort, but there are cases where we know that the resulting block will be 0 whatever the boolean are, so we can skip the lut applications. 
For example if the scalar are small and n_block is big, we know that the high blocks will be 0 at this step. One thing to do could be to iterate not until n_block but until something like `clear_chunk.sum().ilog2() / packing_size` (or n_block in case of overflow) and fill the remaining blocks with zeros.
</blockquote></details>

Yeah we can do that, it's not going to impact computations alot but it's still less CPU usage

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 4 unresolved discussions (waiting on @nsarlin-zama)

___
*[`tfhe/src/high_level_api/array/mod.rs` line 436 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBc_wv4WgkE0-dJgsH:-OMCvT9I9Qh7vtw0QjNG:b-s9wnzm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/high_level_api/array/mod.rs#L436)):*

Instead of using a callback this could return a RadixCiphertext or a Vec of blocks and have the constructor in the calling function ?
Or maybe this is too cumbersome because of the tag
</blockquote></details>

We could make the inner function return `(Vec<Ciphertext>, Tag)` so that the caller can then construct the final radix
___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBgKNq1X_GlFS00T9n:-OMCutqW3pwbxFg7kALo:b-zff5m1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L33)):*

Ok but this one feels like something that can easily be triggered with default parameters without doing anything strange, this is not some "user played with fire" panic. I see that `sum_ciphertexts` returns an option when called on an empty list, so it's a bit analogous.
</blockquote></details>

We could return an Option, but I'm not super convinced yet

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 4 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/high_level_api/array/mod.rs` line 436 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBc_wv4WgkE0-dJgsH:-OMCxk_N6I70T0e-MedF:bbacldk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/high_level_api/array/mod.rs#L436)):*

We could make the inner function return `(Vec<Ciphertext>, Tag)` so that the caller can then construct the final radix
</blockquote></details>

yes I think it might be a bit more readable than with a callback
___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMBgKNq1X_GlFS00T9n:-OMCxFhW02fvfnYl-ZvT:b9f7cyy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/03dd11d9cca7c67da1d113eb096492d5a0540c62/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L33)):*

We could return an Option, but I'm not super convinced yet
</blockquote></details>

If it panics it should be documented, because it's not super obvious from the function signature that the input list should not be empty (it could just return 0 in that case)

nsarlin-zama (changes requested):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 2 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 78 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMH1Rsk09TCyGDx007o:-OMH1Rsk09TCyGDx007p:bo3dm50) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9339aa6b87bdf1a127af11347af070a96639b816/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L78)):*
> ```XML
>                 let mut summed_clear = Clear::ZERO;
>                 for c in clear_chunk.iter() {
>                     summed_clear += *c;
> ```

I think this could overflow and return a smaller number, in that case we can't guess the number of blocks in the dot_product so we should return n_blocks

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 78 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMH1Rsk09TCyGDx007o:-OMHJg8nDgUvDik1dJzT:b-fuugg7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9339aa6b87bdf1a127af11347af070a96639b816/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L78)):*

I think this could overflow and return a smaller number, in that case we can't guess the number of blocks in the dot_product so we should return n_blocks
</blockquote></details>

Yes indeed,

The problem is that detecting the correct number of block to use is hard to get right in every case, e.g If I say n_blocks = 4,and do`63 + 1` -> 63 needs 3 blocks, 1 needs 1 blocks, so we could think we need only 3 blocks, but as 63+1 = 64 => 4 blocks we need 4 blocks4

So getting the correct real_n_blocks why imply tring all combinations

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 78 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMH1Rsk09TCyGDx007o:-OMHSHkd1uIZgnQPO89_:bmf3m9n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9339aa6b87bdf1a127af11347af070a96639b816/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L78)):*
We could use overflowing_add, but that's kinda annoying as it means having to add a new trait and impl it for many types, 

> The expected behavior in the cases where the scalar types are different from the fhe type size are a bit unclear to me.
I think if we require Clear::BITS >= n_blocks*msg_log2 then the behavior will be correct and predictable is it will be the dot prod modulo 2**(n_blocks*msg_log2)

nsarlin-zama (changes requested):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 2 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 78 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMH1Rsk09TCyGDx007o:-OMHTsnN65faXYGKqOyH:bef3twk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9339aa6b87bdf1a127af11347af070a96639b816/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L78)):*

We could use overflowing_add, but that's kinda annoying as it means having to add a new trait and impl it for many types, 

> The expected behavior in the cases where the scalar types are different from the fhe type size are a bit unclear to me.
I think if we require Clear::BITS >= n_blocks*msg_log2 then the behavior will be correct and predictable is it will be the dot prod modulo 2**(n_blocks*msg_log2)
</blockquote></details>

Sorry I thought that Clear was UnsignedInteger. Maybe it can be done using `CastInto<usize>`since we only care about the bit values ? But maybe it's not worth adding this complexity for a gain that only occurs with small numbers.

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 2 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs` line 78 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195#-OMH1Rsk09TCyGDx007o:-OMH_ei5412WAK7aGUgu:bv8jzyp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9339aa6b87bdf1a127af11347af070a96639b816/tfhe/src/integer/server_key/radix_parallel/scalar_dot_prod.rs#L78)):*

I don't see how we could use `CastInto<usize>`, if Clear is u128,i128,I256,etc then it won't work

I'll take some time to think about wether we should add this OverflowingAdd trait
</blockquote></details>

Yes indeed this wouldn't work for larger bitwidth :\

nsarlin-zama (approved):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2195)* status: all files reviewed, 2 unresolved discussions (waiting on @tmontaigu)

PR #2208, opened 2025-03-25
feat(hlapi): add scalar cmux

tmontaigu:
The current IfThenElse trait was too restrictive
to be able to use it to add these methods so a new trait was added.
This trait uses method name with the scalar prefix to avoid conflicts and avoid introducing breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2208)

### DIFF
--- a/tfhe/src/high_level_api/booleans/base.rs
+++ b/tfhe/src/high_level_api/booleans/base.rs
@@ -1,12 +1,14 @@
 use super::inner::InnerBoolean;
 use crate::backward_compatibility::booleans::FheBoolVersions;
 use crate::conformance::ParameterSetConformant;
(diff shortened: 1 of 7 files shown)


IceTDrinker (approved):
Thanks a lot !



> Reviewed 4 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2208#-:-OMSPKUsA02tXDSk1dxU:bm7683k)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IceTDrinker)

PR #2212, opened 2025-03-27
Ns/chore/conformance unwrap

nsarlin-zama:
### PR content/description
Removes some unwraps in conformance check. It's not supposed to be reachable by a client but I think it would be good to be panic free in conformance.
This comes from https://github.com/zama-ai/tfhe-rs/pull/2024 but I made it a separate PR because it is not linked to the AP.
Also includes some typo fixes


IceTDrinker (approved):
Thanks !

PR #2220, opened 2025-03-28
refactor(hlapi): dedup impl code in scalar_ops

tmontaigu:
Closes https://github.com/zama-ai/tfhe-rs-internal/issues/941


nsarlin-zama (approved):
Feels good

PR #2226, opened 2025-03-31
feat(hl): create FheTypes from i32

nsarlin-zama:
This PR adds a conversion method from i32 into FheTypes.
This is done using the `strum` crate. It could also be used later to remove some big matches on FheTypes in the C and JS APIs.


IceTDrinker (approved):
Thanks !

PR #2227, opened 2025-03-31
fix(gpu): update panic condition to apply only to tbc

guillermo-oyarzun:

### Check-list:


pdroalves (commented):
Good catch @guillermo-oyarzun. This assert is only valid  to the TBC PBS since in that case we have a upper bound for the cluster size. 

What if we move this assert to `has_support_to_cuda_programmable_bootstrap_tbc()` so the user will still  be able to run the computation using some other flavour?

guillermo-oyarzun:
> Good catch @guillermo-oyarzun. This assert is only valid to the TBC PBS since in that case we have a upper bound for the cluster size.
> 
> What if we move this assert to `has_support_to_cuda_programmable_bootstrap_tbc()` so the user will still be able to run the computation using some other flavour?

good idea, i like that solution better!

PR #2228, opened 2025-03-31
chore(apis): replace the C/JS FheTypes conversions with strum

nsarlin-zama:
### PR content/description
Since #2226 added strum as a dependency, this PR also uses it in C/JS api to remove big manual matches for FheTypes conversions.

I added a test in the rust part of the JS API. I could not run this test in wasm because some of the dev-dependencies are not wasm compatible, so this would have required a bigger refactor.


IceTDrinker (commented):
Wondering if the tests really test the proper matching of variants

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2228#-:-OMqDSgeC2vP01v2fhkk:b-hpb95j)* status: all files reviewed, 1 unresolved discussion (waiting on @nsarlin-zama)

___
*[`tfhe/src/c_api/high_level_api/mod.rs` line 140 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2228#-OMqD9R35kzGKYkBZ-Vm:-OMqD9R35kzGKYkBZ-Vn:b9jqv8e) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/80a52bed1172bcfd9acc5da53d1b39fd0ea9da15/tfhe/src/c_api/high_level_api/mod.rs#L140)):*
> ```XML
>         let c_value = FheTypes::from(rust_value);
> 
>         assert_eq!(rust_value as i32, c_value as i32)
> ```

I'm tryin to think about this, but are we really testing anything here ?

What I mean is, before we had the manual conversion, but here we are essentially just checking that an i32 value is conserved without verifying that the variants match :thinking:

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2228)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker)

___
*[`tfhe/src/c_api/high_level_api/mod.rs` line 140 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2228#-OMqD9R35kzGKYkBZ-Vm:-OMqH8mo5GPaP_83EWIx:bjnt4sd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/80a52bed1172bcfd9acc5da53d1b39fd0ea9da15/tfhe/src/c_api/high_level_api/mod.rs#L140)):*

I'm tryin to think about this, but are we really testing anything here ?

What I mean is, before we had the manual conversion, but here we are essentially just checking that an i32 value is conserved without verifying that the variants match :thinking:
</blockquote></details>

yes you're right, this is only testing that the values are in the same range.
Maybe in the end it is safer to keep the manual impl with the test

PR #2238, opened 2025-04-03
overflowing neg

tmontaigu:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238)

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/neg.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/neg.rs
@@ -78,19 +78,19 @@ impl ServerKey {
     where
         T: IntegerRadixCiphertext,
     {
-        let mut tmp_ctxt;
-
-        let ct = if ctxt.block_carries_are_empty() {
-            ctxt
+        if ctxt.block_carries_are_empty() {
+            let mut result = self.bitnot(ctxt);
+            self.scalar_add_assign_parallelized(&mut result, 1);
+            result
+        } else if self.is_neg_possible(ctxt).is_ok() {
+            let mut result = self.unchecked_neg(ctxt);
+            self.full_propagate_parallelized(&mut result);
+            result
         } else {
-            tmp_ctxt = ctxt.clone();
-            self.full_propagate_parallelized(&mut tmp_ctxt);
-            &tmp_ctxt
-        };
-
-        let mut ct = self.unchecked_neg(ct);
-        self.full_propagate_parallelized(&mut ct);
-        ct
+            let mut cleaned_ctxt = ctxt.clone();
+            self.full_propagate_parallelized(&mut cleaned_ctxt);
+            self.neg_parallelized(&cleaned_ctxt)
+        }
     }
 
     pub fn overflowing_neg_parallelized<T>(&self, ctxt: &T) -> (T, BooleanBlock)
@@ -99,6 +99,8 @@ impl ServerKey {
     {
         let mut tmp_ctxt;
 
+        // As we want to compute the overflow we need a truly clean state
+        // And so we cannot avoid the full_propagate like we may in non overflowing_block
         let ct = if ctxt.block_carries_are_empty() {
             ctxt
         } else {
--- a/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_neg.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_neg.rs
@@ -4,8 +4,8 @@ use crate::integer::server_key::radix_parallel::tests_cases_unsigned::{FunctionE
 use crate::integer::server_key::radix_parallel::tests_unsigned::{
     nb_tests_for_params, nb_tests_smaller_for_params,
     panic_if_any_block_info_exceeds_max_degree_or_noise, panic_if_any_block_is_not_clean,
-    panic_if_any_block_values_exceeds_its_degree, unsigned_modulus, CpuFunctionExecutor,
-    ExpectedDegrees, ExpectedNoiseLevels, MAX_NB_CTXT,
+    panic_if_any_block_values_exceeds_its_degree, random_non_zero_value, unsigned_modulus,
+    CpuFunctionExecutor, ExpectedDegrees, ExpectedNoiseLevels, MAX_NB_CTXT,
 };
 use crate::integer::tests::create_parameterized_test;
 use crate::integer::{BooleanBlock, IntegerKeyKind, RadixCiphertext, RadixClientKey, ServerKey};
@@ -174,7 +174,7 @@ where
 // Default Tests
 //=============================================================================
 
-pub(crate) fn default_neg_test<P, T>(param: P, mut executor: T)
+pub(crate) fn default_neg_test<P, T>(param: P, mut neg: T)
 where
     P: Into<TestParameters>,
     T: for<'a> FunctionExecutor<&'a RadixCiphertext, RadixCiphertext>,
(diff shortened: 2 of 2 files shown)


nsarlin-zama (changes requested):
> Reviewed 9 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238)* status: 9 of 13 files reviewed, 2 unresolved discussions (waiting on @agnesLeroy)

___
*[`tfhe/src/integer/server_key/radix_parallel/neg.rs` line 122 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONFArbtDooNmlCMS9uk:-ONFArbtDooNmlCMS9ul:bx9xoi4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/server_key/radix_parallel/neg.rs#L122)):*
> ```XML
>             // And in unsigned integers, the only case that is not an overflow is -0,
>             // so we can just invert the result
>             self.boolean_bitnot_assign(&mut overflowed);
> ```

I am not sure I understand this correctly, but I don't think this is how overflows for neg should be handled. If you take the rust doc for neg on [unsigned](https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_neg) and [signed](https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_neg):
- signed: overflows when input is i32::MIN (0x80000000)
- unsigned: overflows when input is not 0

I think the case where `!input + 1` overflows is not considered an overflow of the neg operation because it still yields the correct result. The fact that the addition overflows is an implementation detail but does not really matter as the result is correct for the user. On intel CPU the neg operation behaves as "0 - input" so the OF is not set when input is 0.

neg overflows for i32::MIN because the 2's complement representation is not balanced so there is no correct value for "-i32::MIN". I think for unsigned it overflows everywhere except for 0 because you can't represent negative numbers.
___
*[`tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs` line 104 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONF2FSBDqMycGFePL6H:-ONF2FSBDqMycGFePL6I:b-di04wu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs#L104)):*
> ```XML
>         streams: &CudaStreams,
>     ) {
>         // We do (-ciphertext) + (msg_mod -1) as it allows to avoid an allocation
> ```

I'm not sure because of the cuda impl, but since this is only a boolean maybe there is an easier way to do the bitnot ?

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/neg.rs` line 122 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONFArbtDooNmlCMS9uk:-ONJHn7F5hIqRFr92pPC:bdmfcyv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/server_key/radix_parallel/neg.rs#L122)):*

I am not sure I understand this correctly, but I don't think this is how overflows for neg should be handled. If you take the rust doc for neg on [unsigned](https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_neg) and [signed](https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_neg):
- signed: overflows when input is i32::MIN (0x80000000)
- unsigned: overflows when input is not 0

I think the case where `!input + 1` overflows is not considered an overflow of the neg operation because it still yields the correct result. The fact that the addition overflows is an implementation detail but does not really matter as the result is correct for the user. On intel CPU the neg operation behaves as "0 - input" so the OF is not set when input is 0.

neg overflows for i32::MIN because the 2's complement representation is not balanced so there is no correct value for "-i32::MIN". I think for unsigned it overflows everywhere except for 0 because you can't represent negative numbers.
</blockquote></details>

The end result we have is correct, for unsigned, 0 will not be considered an overflow for unsigned (but all others will),and for signed only the min will overflow

And to me this is the correct way to handle it:

doing `0 -input` is like doing `0 + -input` <=> `0 + (!input + 1) <=> `!input + 1`

However, when doing `0 + (!input + 1)` since that is like doing an subtraction in twos complement, it means the that for unsigned integer neither the carry flag nor oveflowflag are valid indicators of overflow. You can look at unsigned oveflowing sub which uses the subtraction with borrow

tmontaigu:
*[`tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs` line 104 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONF2FSBDqMycGFePL6H:-ONJL3mQDpd5GUBh_oyn:bcmfe49) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs#L104)):*

I'm not sure because of the cuda impl, but since this is only a boolean maybe there is an easier way to do the bitnot ?
</blockquote></details>

What do you have in mind ? I just took the code used for `CudaUnsignedRadixCiphertext`, and boolean block is defined as `pub struct CudaBooleanBlock(pub CudaUnsignedRadixCiphertext);` so apart from hardcoding a Vec of len 1 , I dont seen any possible changes

nsarlin-zama (approved):
Looks good to me!

> Reviewed 4 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-:-ONJUXf-EmkWORVeSJeL:b9j9n3q)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @agnesLeroy)

___
*[`tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs` line 104 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONF2FSBDqMycGFePL6H:-ONJSyNT5wTX4yRiBo9y:b-rc5nx1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs#L104)):*

What do you have in mind ? I just took the code used for `CudaUnsignedRadixCiphertext`, and boolean block is defined as `pub struct CudaBooleanBlock(pub CudaUnsignedRadixCiphertext);` so apart from hardcoding a Vec of len 1 , I dont seen any possible changes
</blockquote></details>

Ok it just felt weird to have to build the vec but it looks like gpu operations only work on ciphertext lists
___
*[`tfhe/src/integer/server_key/radix_parallel/neg.rs` line 122 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONFArbtDooNmlCMS9uk:-ONJM8qa7_7TvC39y0cV:b-6l3jrf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/server_key/radix_parallel/neg.rs#L122)):*

So the comment is just trying to explain that we cant use the overflow flag since we are in unsigned  mode, but 'luckily' flipping it yields correct result, otherwise we would have to do a `sub_with_borrow(0, input` which would need a full implementation, so its just faster and less code to use this trick
</blockquote></details>

Ok sorry I misunderstood the comment. Indeed this looks correct!

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/neg.rs` line 122 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONFArbtDooNmlCMS9uk:-ONJWSHT7jXMfiqUmKpg:bdtd1qs) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/server_key/radix_parallel/neg.rs#L122)):*

Ok sorry I misunderstood the comment. Indeed this looks correct!
</blockquote></details>

I can update the comment in the code so that it may be clearer

nsarlin-zama (approved):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @agnesLeroy)

___
*[`tfhe/src/integer/server_key/radix_parallel/neg.rs` line 122 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2238#-ONFArbtDooNmlCMS9uk:-ONJX2DR8TVd4ZBQddGs:b-iupq40) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/210bf7686eba61b31101975ac020a4747f808aad/tfhe/src/integer/server_key/radix_parallel/neg.rs#L122)):*

I can update the comment in the code so that it may be clearer
</blockquote></details>

Maybe just a function doc comment that specifies the cases where it overflows for signed and unsigned will do it. 

What confused me is that I thought that the overflow in the unsigned was the opposite of the signed case, but I forgot that `overflowing_scalar_add_assign_parallelized` was generic too so it was already handling signed and unsigned overflows differently.

IceTDrinker:
to be merged maybe ?

PR #2269, opened 2025-04-15
feat(hl): add AsRef implementation on ServerKey to access NoiseSquashingKey

IceTDrinker:
cc @kc1212 this one is going into main but the same patch will make it to release/1.1.x

to note that you likely will need to call

```rust
let sns_key: Option<&tfhe::integer::NoiseSquashingKey> = sk.as_ref().as_ref();
```

 the reason being that we can only return a reference to the internal option from the `AsRef` Implem, and then you call `Option::as_ref` to convert the `&Option<T>` to `Option<&T>`

### DIFF
--- a/tfhe/src/high_level_api/keys/server.rs
+++ b/tfhe/src/high_level_api/keys/server.rs
@@ -112,6 +112,12 @@ impl ServerKey {
         self.key.cpk_casting_key()
     }
 
+    pub fn noise_squashing_key(
+        &self,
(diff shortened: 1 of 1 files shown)


nsarlin-zama:
Looks like having multiple asref impl makes it ambiguous so it's breaking for the integer::ServerKey. Another solution would be to have a dedicated method like `fn noise_squashing_key(&self) -> Option<&NoiseSquashingKey>`, would remove the needs for the outer as_ref as well.

nsarlin-zama (approved):
thanks!

PR #2271, opened 2025-04-16
chore: update toolchain to 2025-04-16

nsarlin-zama:
update latest toolchain, should fix ci


utils/tfhe-versionable-derive/src/versionize_impl.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 pub(crate) enum VersionizeImplementor {
     Classic(ClassicVersionizeAttribute),
     Convert(ConvertVersionizeAttribute),
-    Transparent(TransparentStruct),
+    Transparent(Box<TransparentStruct>),
```
IceTDrinker, line 24: what's this for ?

nsarlin-zama: That's because of the large enum variant lint. I think it only triggers now because some type inside the field grew in size between compiler versions

IceTDrinker: ok

IceTDrinker (approved):
Thanks !

PR #2275, opened 2025-04-16
feat(core): enable custom modulus generation for TUniform

IceTDrinker:
Only real tricky part here is making sure that the custom modulus can properly generate the TUniform value and encode that value properly.

Had to put an assert in the generate_one_custom_modulus because the distribution system (which is ancient and was just patched up to work with custom moduli) is not very well suited to the custom modulus stuff to enforce invariants.

Changes in the test file are mainly updating the helper trait for TUniform to make sure that we are able to properly encode/decode the values when the modulus is not native.

One day it might get reworked 😅

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275)

### DIFF
--- a/tfhe/src/core_crypto/commons/math/random/mod.rs
+++ b/tfhe/src/core_crypto/commons/math/random/mod.rs
@@ -47,11 +47,19 @@ where
     // pair of elements as custom modulus
     type CustomModulus: Copy;
 
+    /// Generate a value from `distribution`.
     fn generate_one<G: ByteRandomGenerator>(
         generator: &mut RandomGenerator<G>,
         distribution: D,
     ) -> Self;
 
+    /// Generate a value from `distribution` modulo the given `custom_modulus`.
+    ///
+    /// This `custom_modulus` must be able to represent all possible values from `distribution`.
+    ///
+    /// Implementations are allowed to panic if the given `custom_modulus` cannot represent all
+    /// possible values from `distribution` as the effective distribution would differ from the
+    /// desired distribution.
     fn generate_one_custom_modulus<G: ByteRandomGenerator>(
         generator: &mut RandomGenerator<G>,
         distribution: D,
--- a/tfhe/src/core_crypto/commons/math/random/t_uniform.rs
+++ b/tfhe/src/core_crypto/commons/math/random/t_uniform.rs
@@ -59,8 +59,8 @@ impl<T: UnsignedInteger> TUniform<T> {
         self.bound_log2
     }
 
-    pub const fn distinct_value_count(&self) -> usize {
-        (1 << (self.bound_log2 + 1)) + 1
+    pub(crate) fn distinct_value_count(self) -> T {
+        (T::ONE << ((self.bound_log2 + 1) as usize)) + T::ONE
     }
 
     pub fn min_value_inclusive(&self) -> T::Signed {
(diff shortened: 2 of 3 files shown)


nsarlin-zama (changes requested):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`tfhe/src/core_crypto/commons/math/random/t_uniform.rs` line 114 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275#-OO1qvru1wkWk5ie64kx:-OO1qvru1wkWk5ie64ky:b-p308i0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e55fa00cab48411c5cc88d4cdc101e7ba81c0111/tfhe/src/core_crypto/commons/math/random/t_uniform.rs#L114)):*
> ```XML
>                 // If q = 5, we have 5 representable values, which means all TUniform values can
>                 // be represented, 0 == 0, 1 == 1, 2 == 2, 3 == -2, 4 == -1
>                 // So we need tuniform_unique_values <= q
> ```

I am not sure I understand this well, for me intuitively `generate_one_custom_modulus` for a distribution means that we have a distribution that is larger than the modulus and we want to make the returned value fit in a smaller set, like we do in C with 
```c
int val = rand() % 20;
```
So here the modulus is actually smaller than the distribution size.

It seems the behavior in this PR specific to Tuniform. TUniform has a generic UnsignedInteger type so here the "custom modulus" is related to this generic. But this generic is not part of the RandomGenerable trait, so in my mind what this method is doing should not be part of the trait impl.

Maybe have a dedicated method on `Tuniform<T>` would be less confusing ? Because in the end, the user should use this method differently if this is a TUniform so we rely on the fact that they know that they have a TUniform distribution at hand.

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @nsarlin-zama)

___
*[`tfhe/src/core_crypto/commons/math/random/t_uniform.rs` line 114 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275#-OO1qvru1wkWk5ie64kx:-OO238GU1EtD2tv9VU5N:b-3by982) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e55fa00cab48411c5cc88d4cdc101e7ba81c0111/tfhe/src/core_crypto/commons/math/random/t_uniform.rs#L114)):*

I am not sure I understand this well, for me intuitively `generate_one_custom_modulus` for a distribution means that we have a distribution that is larger than the modulus and we want to make the returned value fit in a smaller set, like we do in C with 
```c
int val = rand() % 20;
```
So here the modulus is actually smaller than the distribution size.

It seems the behavior in this PR specific to Tuniform. TUniform has a generic UnsignedInteger type so here the "custom modulus" is related to this generic. But this generic is not part of the RandomGenerable trait, so in my mind what this method is doing should not be part of the trait impl.

Maybe have a dedicated method on `Tuniform<T>` would be less confusing ? Because in the end, the user should use this method differently if this is a TUniform so we rely on the fact that they know that they have a TUniform distribution at hand.
</blockquote></details>

No actually the custom modulus indicates in what space the data lives, it's the space used to represent the torus.

So if we have a modulus 2^64 then -1 == u64::MAX == 2^64 - 1

if we have a modulus 2^21 then -1 == 2^21 - 1

The other generation functions using the custom modulus follow the same pattern

Gaussian works slightly differently as the gaussian gets scaled by Q, uniform works with rejection sampling to make sure the value is in the right range

Here the important thing, it seems, is to make sure all values of TUniform we can generate will fit in the destination modulus, otherwise the actual distribution will not be conserved I think ?

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker)

___
*[`tfhe/src/core_crypto/commons/math/random/t_uniform.rs` line 114 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275#-OO1qvru1wkWk5ie64kx:-OO24heA33gZ6_E_RSwn:b-opsp8a) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e55fa00cab48411c5cc88d4cdc101e7ba81c0111/tfhe/src/core_crypto/commons/math/random/t_uniform.rs#L114)):*

No actually the custom modulus indicates in what space the data lives, it's the space used to represent the torus.

So if we have a modulus 2^64 then -1 == u64::MAX == 2^64 - 1

if we have a modulus 2^21 then -1 == 2^21 - 1

The other generation functions using the custom modulus follow the same pattern

Gaussian works slightly differently as the gaussian gets scaled by Q, uniform works with rejection sampling to make sure the value is in the right range

Here the important thing, it seems, is to make sure all values of TUniform we can generate will fit in the destination modulus, otherwise the actual distribution will not be conserved I think ?
</blockquote></details>

Ok sorry it makes sense, I understand it better now. So you use it with modulus that are smaller than the native modulus of the type but greater than the number of elements that can be drawn from the distribution.

Maybe it might be worth having a doc comment for the `generate_one_custom_modulus` method in the trait definition ?
 
And adding a doc comment here to say that it can panic if the modulus is smaller than the number of distinct values ?

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275)* status: all files reviewed, 1 unresolved discussion (waiting on @nsarlin-zama)

___
*[`tfhe/src/core_crypto/commons/math/random/t_uniform.rs` line 114 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2275#-OO1qvru1wkWk5ie64kx:-OO2UIEl6rdqSEi3nOqF:bijefq5) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e55fa00cab48411c5cc88d4cdc101e7ba81c0111/tfhe/src/core_crypto/commons/math/random/t_uniform.rs#L114)):*

Ok sorry it makes sense, I understand it better now. So you use it with modulus that are smaller than the native modulus of the type but greater than the number of elements that can be drawn from the distribution.

Maybe it might be worth having a doc comment for the `generate_one_custom_modulus` method in the trait definition ?
 
And adding a doc comment here to say that it can panic if the modulus is smaller than the number of distinct values ?
</blockquote></details>

Don't be sorry, it's a mix of very legacy systems and additions slapped on top of it with non trivial implications

As for the documentation proposal that's a very good point

PR #2278, opened 2025-04-17
Al/cuda malloc mem tracking

agnesLeroy:

### PR content/description

With this PR it is possible to write:

```
let config = ConfigBuilder::default();
let client_key = ClientKey::generate(config);
let csks = CompressedServerKey::new(&client_key);
let server_key = csks.decompress_to_gpu();
set_server_key(server_key);
let clear_a = rng.gen_range(1..=i32::MAX);
let clear_b = rng.gen_range(1..=i32::MAX);
let mut a = FheInt32::try_encrypt(clear_a, &client_key).unwrap();
let mut b = FheInt32::try_encrypt(clear_b, &client_key).unwrap();
let ciphertexts_size = a.get_size_on_gpu() + b.get_size_on_gpu();
assert!(check_valid_cuda_malloc(ciphertexts_size, GpuIndex::new(0)));

a.move_to_current_device();
b.move_to_current_device();

let tmp_buffer_size = a.get_add_assign_size_on_gpu(&b);
assert!(check_valid_cuda_malloc(tmp_buffer_size, GpuIndex::new(0)));
a += &b;
```

This then would have to be extended to other functions.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/device.h
+++ b/backends/tfhe-cuda-backend/cuda/include/device.h
@@ -47,12 +47,27 @@ uint32_t cuda_is_available();
 
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
 
+void *cuda_malloc_with_size_tracking_async(uint64_t size, cudaStream_t stream,
+                                           uint32_t gpu_index,
+                                           uint64_t *size_tracker,
+                                           bool allocate_gpu_memory);
+
 void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index);
 
-void cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
+bool cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
+
+void cuda_memcpy_with_size_tracking_async_to_gpu(void *dest, const void *src,
+                                                 uint64_t size,
+                                                 cudaStream_t stream,
+                                                 uint32_t gpu_index,
+                                                 bool gpu_memory_allocated);
 
 void cuda_memcpy_async_to_gpu(void *dest, const void *src, uint64_t size,
                               cudaStream_t stream, uint32_t gpu_index);
+
+void cuda_memcpy_with_size_tracking_async_gpu_to_gpu(
+    void *dest, void const *src, uint64_t size, cudaStream_t stream,
+    uint32_t gpu_index, bool gpu_memory_allocated);
 
 void cuda_memcpy_async_gpu_to_gpu(void *dest, void const *src, uint64_t size,
                                   cudaStream_t stream, uint32_t gpu_index);
@@ -63,6 +78,11 @@ void cuda_memcpy_gpu_to_gpu(void *dest, void const *src, uint64_t size,
 void cuda_memcpy_async_to_cpu(void *dest, const void *src, uint64_t size,
                               cudaStream_t stream, uint32_t gpu_index);
 
+void cuda_memset_with_size_tracking_async(void *dest, uint64_t val,
+                                          uint64_t size, cudaStream_t stream,
+                                          uint32_t gpu_index,
+                                          bool gpu_memory_allocated);
+
 void cuda_memset_async(void *dest, uint64_t val, uint64_t size,
                        cudaStream_t stream, uint32_t gpu_index);
 
@@ -71,6 +91,10 @@ int cuda_get_number_of_gpus();
 void cuda_synchronize_device(uint32_t gpu_index);
 
 void cuda_drop(void *ptr, uint32_t gpu_index);
+
+void cuda_drop_with_size_tracking_async(void *ptr, cudaStream_t stream,
+                                        uint32_t gpu_index,
+                                        bool gpu_memory_allocated);
 
 void cuda_drop_async(void *ptr, cudaStream_t stream, uint32_t gpu_index);
 }
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
@@ -48,7 +48,7 @@ typedef struct {
   uint32_t lwe_dimension;
 } CudaRadixCiphertextFFI;
 
-void scratch_cuda_apply_univariate_lut_kb_64(
+uint64_t scratch_cuda_apply_univariate_lut_kb_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, void const *input_lut, uint32_t lwe_dimension,
     uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t ks_level,
@@ -56,7 +56,7 @@ void scratch_cuda_apply_univariate_lut_kb_64(
     uint32_t grouping_factor, uint32_t input_lwe_ciphertext_count,
     uint32_t message_modulus, uint32_t carry_modulus, PBS_TYPE pbs_type,
     uint64_t lut_degree, bool allocate_gpu_memory, bool allocate_ms_array);
-void scratch_cuda_apply_many_univariate_lut_kb_64(
+uint64_t scratch_cuda_apply_many_univariate_lut_kb_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, void const *input_lut, uint32_t lwe_dimension,
     uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t ks_level,
@@ -78,7 +78,7 @@ void cleanup_cuda_apply_univariate_lut_kb_64(void *const *streams,
                                              uint32_t gpu_count,
                                              int8_t **mem_ptr_void);
 
-void scratch_cuda_apply_bivariate_lut_kb_64(
+uint64_t scratch_cuda_apply_bivariate_lut_kb_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, void const *input_lut, uint32_t lwe_dimension,
     uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t ks_level,
@@ -109,7 +109,7 @@ void cuda_apply_many_univariate_lut_kb_64(
     CudaModulusSwitchNoiseReductionKeyFFI const *ms_noise_reduction_key,
     void *const *bsks, uint32_t num_luts, uint32_t lut_stride);
 
-void scratch_cuda_full_propagation_64(
+uint64_t scratch_cuda_full_propagation_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, uint32_t lwe_dimension, uint32_t glwe_dimension,
     uint32_t polynomial_size, uint32_t ks_level, uint32_t ks_base_log,
@@ -127,7 +127,7 @@ void cleanup_cuda_full_propagation(void *const *streams,
                                    uint32_t const *gpu_indexes,
                                    uint32_t gpu_count, int8_t **mem_ptr_void);
 
-void scratch_cuda_integer_mult_radix_ciphertext_kb_64(
+uint64_t scratch_cuda_integer_mult_radix_ciphertext_kb_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, bool const is_boolean_left, bool const is_boolean_right,
     uint32_t message_modulus, uint32_t carry_modulus, uint32_t glwe_dimension,
@@ -161,7 +161,7 @@ void cuda_scalar_addition_integer_radix_ciphertext_64_inplace(
     void const *h_scalar_input, uint32_t num_scalars, uint32_t message_modulus,
     uint32_t carry_modulus);
 
-void scratch_cuda_integer_radix_logical_scalar_shift_kb_64(
+uint64_t scratch_cuda_integer_radix_logical_scalar_shift_kb_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr, uint32_t glwe_dimension, uint32_t polynomial_size,
     uint32_t big_lwe_dimension, uint32_t small_lwe_dimension, uint32_t ks_level,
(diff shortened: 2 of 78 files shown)


tfhe/src/high_level_api/gpu_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+/// let mut rng = rand::thread_rng();
+/// let clear_a = rng.gen_range(1..=i32::MAX);
+/// let clear_b = rng.gen_range(1..=i32::MAX);
+/// let mut a = FheInt32::try_encrypt(clear_a, &client_key).unwrap();
+/// let mut b = FheInt32::try_encrypt(clear_b, &client_key).unwrap();
+/// let ciphertexts_size = a.get_size_on_gpu() + b.get_size_on_gpu();
+/// assert!(check_valid_cuda_malloc(ciphertexts_size, GpuIndex::new(0)));
+///
+/// a.move_to_current_device();
+/// b.move_to_current_device();
```
antoniupop, line 28: Is the explicit move to device call needed if it targets the device selected on current thread?

antoniupop (commented):
Looks great - sorry I only had time to skim over most of it, I'm technically on holidays.

Something that might make easier the synchronisation between concurrent streams offloading to the same GPU would be to allow reserving memory for an upcoming operation - not sure if it's possible. Something like:

```
let tmp_buffer_size = a.get_add_assign_size_on_gpu(&b);
assert!(try_reserve_cuda_memory(tmp_buffer_size, GpuIndex::new(0)));
a += &b;
release_reserved_cuda_memory(tmp_buffer_size, GpuIndex::new(0)));
```
This assuming we can't straight up try to allocate, as too complicated to track all buffers, so instead just keep an atomic counter of how much memory is reserved and use that to ensure we don't need to synchronise between threads that could check at the same time, both get OK return, then they both try to execute and run out of memory.
I think this would only need an atomic value to be used along with the cuda free memory function to implement.

tfhe/src/high_level_api/gpu_utils.rs
agnesLeroy: a and b after encryption are located on the CPU, that's why I'm moving them to the current device in the example.

agnesLeroy:
> Looks great - sorry I only had time to skim over most of it, I'm technically on holidays.
> 
> Something that might make easier the synchronisation between concurrent streams offloading to the same GPU would be to allow reserving memory for an upcoming operation - not sure if it's possible. Something like:
> 
> ```
> let tmp_buffer_size = a.get_add_assign_size_on_gpu(&b);
> assert!(try_reserve_cuda_memory(tmp_buffer_size, GpuIndex::new(0)));
> a += &b;
> release_reserved_cuda_memory(tmp_buffer_size, GpuIndex::new(0)));
> ```
> 
> This assuming we can't straight up try to allocate, as too complicated to track all buffers, so instead just keep an atomic counter of how much memory is reserved and use that to ensure we don't need to synchronise between threads that could check at the same time, both get OK return, then they both try to execute and run out of memory. I think this would only need an atomic value to be used along with the cuda free memory function to implement.

Very good idea, we had implemented this scratch/cleanup logic at the backend level to do something like this eventually. With the memory tracking capability we're introducing, now it would be possible for the  user to compute the max amount of memory needed for all their operations, reserve memory for it and then in the scratch/cleanup we would skip cuda malloc/free. This will require a lot more changes though so wouldn't be for this quarter. First I need to expose the function to get required memory size for all operations...

tfhe/src/high_level_api/gpu_utils.rs
antoniupop: Sure, but isn't that done implicitly? I guess it's to get the memory usage of the two taken into account before checking if the operation can execute.

agnesLeroy: Yes exactly :slightly_smiling_face:

antoniupop (approved):

backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -48,15 +48,18 @@ uint32_t cuda_is_available();
...
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
 
-void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index);
+void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index,
+                        uint64_t *size_tracker, bool allocate_gpu_memory);
 
-void cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
+bool cuda_check_valid_malloc(uint64_t size, uint32_t gpu_index);
 
 void cuda_memcpy_async_to_gpu(void *dest, const void *src, uint64_t size,
```
pdroalves, line 56: The same suggestion about having two versions of each `cuda_memcpy_` method applies here. However, I'm not convinced we can't simply reorganize the scratch functions so that all operations that know they don't allocate memory remain within a conditional (checking the boolean). This would simplify a lot this PR and include the memcpy methods. Are you sure we need to modify so many functions to include that parameter?

backends/tfhe-cuda-backend/cuda/include/integer/integer.h
```diff
@@ -517,7 +517,7 @@ void cleanup_cuda_integer_are_all_comparisons_block_true(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr_void);
 
-void scratch_cuda_integer_is_at_least_one_comparisons_block_true_kb_64(
+uint64_t scratch_cuda_integer_is_at_least_one_comparisons_block_true_kb_64(
```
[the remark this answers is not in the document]
agnesLeroy: No!

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -198,87 +200,98 @@ template <typename Torus> struct int_radix_lut {
...
+        num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0],
+        size_tracker, allocate_gpu_memory);
+    lwe_trivial_indexes = (Torus *)cuda_malloc_async(
+        num_radix_blocks * sizeof(Torus), streams[0], gpu_indexes[0],
+        size_tracker, allocate_gpu_memory);
+
+    h_lwe_indexes_in = (Torus *)malloc(num_radix_blocks * sizeof(Torus));
+    h_lwe_indexes_out = (Torus *)malloc(num_radix_blocks * sizeof(Torus));
+
+    for (int i = 0; i < num_radix_blocks; i++)
```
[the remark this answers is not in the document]
agnesLeroy: Totally, I will make the change!

backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -48,15 +48,18 @@ uint32_t cuda_is_available();
 
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
 
-void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index);
+void *cuda_malloc_async(uint64_t size, cudaStream_t stream, uint32_t gpu_index,
```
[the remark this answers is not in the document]
agnesLeroy: Good idea! I will make the change!

backends/tfhe-cuda-backend/cuda/src/utils/helper.cu
```diff
@@ -0,0 +1,6 @@
+#include "helper.cuh"
+template <> __global__ void print_debug_kernel(const double2 *src, int N) {
```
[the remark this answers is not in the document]
agnesLeroy: This is not necessary I will move it back.

backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -48,15 +48,18 @@ uint32_t cuda_is_available();
 
 void *cuda_malloc(uint64_t size, uint32_t gpu_index);
```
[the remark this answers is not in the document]
agnesLeroy: We actually never use cuda_malloc

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -531,61 +569,67 @@ template <typename Torus> struct int_radix_lut {
...
   void release(cudaStream_t const *streams, uint32_t const *gpu_indexes,
                uint32_t gpu_count) {
     free(this->gpu_indexes);
-    for (uint i = 0; i < active_gpu_count; i++) {
-      cuda_drop_async(lut_vec[i], streams[i], gpu_indexes[i]);
-      cuda_drop_async(lut_indexes_vec[i], streams[i], gpu_indexes[i]);
-    }
+    if (gpu_memory_allocated) {
+      for (uint i = 0; i < active_gpu_count; i++) {
+        cuda_drop_async(lut_vec[i], streams[i], gpu_indexes[i]);
```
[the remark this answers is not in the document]
agnesLeroy: I actually didn't want to do that, even looking at it now I would rather keep `cuda_drop_async` as it is. Maybe we could have a `cuda_drop_async_with_size_track` to have the option there? :thinking: But I'm not sure it's going to make our lives easier.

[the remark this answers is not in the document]
agnesLeroy: Looking at it again it seems better to do that, so I implemented it

tfhe/src/high_level_api/integers/signed/ops.rs
```diff
@@ -1940,3 +1942,50 @@ where
...
+            InternalServerKey::Cuda(cuda_key) => with_thread_local_cuda_streams(|streams| {
+                tmp_buffer_size = cuda_key.key.key.get_add_assign_size_on_gpu(
+                    &*self.ciphertext.on_gpu(streams),
+                    &rhs.ciphertext.on_gpu(streams),
+                    streams,
+                );
+            }),
+        });
+        tmp_buffer_size
+    }
```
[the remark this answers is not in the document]
agnesLeroy: I had this error:
```
error: wildcard matches only a single variant and will also match any future added variants
    --> tfhe/src/high_level_api/integers/signed/ops.rs:1984:13
     |
1984 |             _ => 0,
     |             ^ help: try: `InternalServerKey::Cpu(_)`
     |
```
so I kept `InternalServerKey::Cpu(_) => 0` as a second option.

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -282,9 +322,14 @@ void cuda_drop(void *ptr, uint32_t gpu_index) {
   check_cuda_error(cudaFree(ptr));
 }
 
-/// Drop a cuda array asynchronously, if supported on the device
-void cuda_drop_async(void *ptr, cudaStream_t stream, uint32_t gpu_index) {
+/// Drop a cuda array asynchronously, if the data was allocated & it's supported
+/// on the device
+void cuda_drop_with_size_tracking_async(void *ptr, cudaStream_t stream,
```
pdroalves, line 327: Probably not necessary, but have you considered adding the pointer to `size_track` here so we can assert it finishes computation as zero? This would point out memory leaks.

Of course, we can check them using compute-sanitizer as well.

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -135,7 +135,7 @@ __host__ void are_all_comparisons_block_true(
             is_max_value_lut->get_degree(1),
             is_max_value_lut->get_max_degree(1), glwe_dimension,
             polynomial_size, message_modulus, carry_modulus,
-            is_equal_to_num_blocks_lut_f);
+            is_equal_to_num_blocks_lut_f, true);
```
pdroalves, line 138: Have you considered having two versions of `generate_device_accumulator` as you did in the device functions?>

backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cu
```diff
@@ -2,8 +2,11 @@
 
 void release_radix_ciphertext_async(cudaStream_t const stream,
                                     uint32_t const gpu_index,
-                                    CudaRadixCiphertextFFI *data) {
-  cuda_drop_async(data->ptr, stream, gpu_index);
+                                    CudaRadixCiphertextFFI *data,
+                                    const bool gpu_memory_allocated) {
+  if (gpu_memory_allocated)
```
pdroalves, line 7: I think you forgot the brackets to include the frees above this line.

Anyway, we don't need the cuda drop in the conditional.

pdroalves (approved):
I just found a single mistake on a release function. Other than that, I think we may be good to go. I wrote a few comments if you want to consider.

backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cu
agnesLeroy: Indeed we don't need this `if`. However we need to free memory on the CPU side as it gets allocated even if `allocate_gpu_memory` is set to false.

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
agnesLeroy: No, I didn't think it was too bad to add the extra argument for those.

backends/tfhe-cuda-backend/cuda/src/device.cu
agnesLeroy: That's a good idea, I'll open an issue for that because it would require extensive modifications to this PR. I'm not sure it's easily doable because it would require to keep track for every pointer of its own size, which we don't do at the moment: we only track the memory allocated for the whole function.

backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cu
agnesLeroy: We'll probably want to come back on this, to have the scratch/cleanup functions stop allocating & freeing memory on the CPU when called with `gpu_memory_allocated=false`, but as mentioned before that would require a lot of extra work and could be done in a second stage.

tfhe/src/high_level_api/integers/signed/ops.rs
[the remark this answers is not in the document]
tmontaigu: I'm responding late, but I think that is a valid case where we could have silenced this clippy thing

But its not too bad the way it it now, I'm not sure we want to support compilation with mulitple backend enabled as there is no support for mixing backend execution in terms of  parameters and bitwise reproducible results

PR #2279, opened 2025-04-18
chore(all): make clippy_rustdoc output less noisy

mayeul-zama:


IceTDrinker (commented):
What does it output in case of failure ?

mayeul-zama:
> What does it output in case of failure ?

It shows the error and prints the location of the failing doctest

IceTDrinker (approved):
Checked the formatting looks good enough 👍

PR #2285, opened 2025-04-22
chore(ci): fix slack notify in case of cancelled step

soonum:
If a step is cancelled, it is not considered as failure by GitHub. 
So if a user cancelled a task or if a job timed out, then no Slack notification was sent and devs weren't able to track down these events.

### DIFF
--- a/.github/workflows/aws_tfhe_backward_compat_tests.yml
+++ b/.github/workflows/aws_tfhe_backward_compat_tests.yml
@@ -124,7 +124,7 @@ jobs:
           echo "PULL_REQUEST_MD_LINK=[pull-request](${{ vars.PR_BASE_URL }}${{ github.event.pull_request.number }}), "  >> "${GITHUB_ENV}"
 
       - name: Slack Notification
-        if: ${{ failure() }}
+        if: ${{ failure() || (cancelled() && github.event_name != 'pull_request') }}
         continue-on-error: true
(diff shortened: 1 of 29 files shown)


.github/workflows/aws_tfhe_backward_compat_tests.yml
```diff
@@ -124,7 +124,7 @@ jobs:
           echo "PULL_REQUEST_MD_LINK=[pull-request](${{ vars.PR_BASE_URL }}${{ github.event.pull_request.number }}), "  >> "${GITHUB_ENV}"
 
       - name: Slack Notification
-        if: ${{ failure() }}
+        if: ${{ failure() || cancelled() }}
```
IceTDrinker, line 127: I'm going to be annoying here but I think we removed cancelled on tests to avoid spamming the channel for notifications

IceTDrinker: could be good to keep on the long run tests however

IceTDrinker: you could make it "generic" in the sense that it checks for the github event and this way you can copy/paste, up to you, but yeah cancelled on PRs not interesting

soonum: Done.

nsarlin-zama (approved):
Looks good to me, thanks!

PR #2286, opened 2025-04-22
chore: add CODEOWNERS file to avoid unwanted changes to core_crypto source

IceTDrinker:
cc @agnesLeroy 

For now the reviews are not mandatory from code owners to avoid blocking people, but it likely will be enabled with a backup owner to avoid blocking people when both are on leave.


soonum (approved):
I guess this file will be extended. Like for CI files, for example ?

IceTDrinker:
> I guess this file will be extended. Like for CI files, for example ?

Totally, just that I've had two strikes on changes I would never have approved, so code owners file it is

PR #2291, opened 2025-04-23
fix(core): remove additional body coeff in multi bit ms compression

nsarlin-zama:
An additional body term was included with the packed mask in the multibit ms compressed ct.
Since this term was simply ignored during extraction I don't think this has any impact on the backward compatibility. From what I understand the Conformance already uses the lwe dimension to check the size of the packed list.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)

### DIFF
--- a/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
+++ b/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
@@ -1,8 +1,9 @@
 use self::packed_integers::PackedIntegers;
 use crate::conformance::ParameterSetConformant;
-use crate::core_crypto::backward_compatibility::entities::compressed_modulus_switched_multi_bit_lwe_ciphertext::CompressedModulusSwitchedMultiBitLweCiphertextVersions;
+use crate::core_crypto::backward_compatibility::entities;
 use crate::core_crypto::fft_impl::common::modulus_switch;
 use crate::core_crypto::prelude::*;
+use entities::compressed_modulus_switched_multi_bit_lwe_ciphertext::*;
 use itertools::Itertools;
 use tfhe_versionable::Versionize;
 
@@ -180,6 +181,86 @@ pub struct CompressedModulusSwitchedMultiBitLweCiphertext<
 impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
     CompressedModulusSwitchedMultiBitLweCiphertext<Scalar>
 {
+    pub(crate) fn from_raw_parts(
+        body: usize,
+        packed_mask: PackedIntegers<usize>,
+        packed_diffs: Option<PackedIntegers<usize>>,
+        lwe_dimension: LweDimension,
+        uncompressed_ciphertext_modulus: CiphertextModulus<Scalar>,
+        grouping_factor: LweBskGroupingFactor,
+    ) -> Self {
+        assert_eq!(
+            packed_mask.initial_len() % grouping_factor.0,
+            0,
+            "Packed mask len (={}) should be a multiple of grouping factor (={})",
+            packed_mask.initial_len(),
+            grouping_factor.0
+        );
+        assert_eq!(packed_mask.initial_len(), lwe_dimension.0,
+            "Packed mask list is not of the correct size for the uncompressed LWE: expected {}, got {}",
+            lwe_dimension.0,
+            packed_mask.initial_len());
+
+        if let Some(diffs) = packed_diffs.as_ref() {
+            // Get the expected number of diffs from the compression code
+            let multi_bit_elements = packed_mask.initial_len() / grouping_factor.0;
+            let ggsw_per_multi_bit_element = grouping_factor.ggsw_per_multi_bit_element().0;
+
+            // In the diff list creation, we skip every power of two, so we have to remove them from
+            // the total count
+            let num_powers_of_2 = ggsw_per_multi_bit_element.ceil_ilog2() as usize;
+            let expected_diffs_len =
+                (ggsw_per_multi_bit_element - 1 - num_powers_of_2) * multi_bit_elements;
+            assert_eq!(
+                diffs.initial_len(),
+                expected_diffs_len,
+                "Packed diff list is not of the correct size for the uncompressed LWE: expected {}, got {}",
+                diffs.initial_len(),
+                expected_diffs_len
+            );
+        }
+
+        Self {
+            body,
+            packed_mask,
+            packed_diffs,
+            lwe_dimension,
+            uncompressed_ciphertext_modulus,
+            grouping_factor,
+        }
+    }
+
+    #[cfg(test)]
+    #[allow(clippy::type_complexity)]
+    pub(crate) fn into_raw_parts(
+        self,
+    ) -> (
+        usize,
+        PackedIntegers<usize>,
+        Option<PackedIntegers<usize>>,
+        LweDimension,
+        CiphertextModulus<Scalar>,
+        LweBskGroupingFactor,
+    ) {
+        let Self {
+            body,
+            packed_mask,
+            packed_diffs,
+            lwe_dimension,
+            uncompressed_ciphertext_modulus,
+            grouping_factor,
+        } = self;
+
+        (
+            body,
+            packed_mask,
+            packed_diffs,
+            lwe_dimension,
+            uncompressed_ciphertext_modulus,
+            grouping_factor,
+        )
+    }
+
     /// Compresses a ciphertext by reducing its modulus
     /// This operation adds a lot of noise
     pub fn compress<Cont: Container<Element = Scalar>>(
@@ -212,7 +293,7 @@ impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
 
         let body = modulus_switch(*input_lwe_body.data, log_modulus).cast_into();
 
-        let modulus_switched: Vec<usize> = ct
+        let modulus_switched: Vec<usize> = input_lwe_mask
             .as_ref()
             .iter()
             .map(|a| modulus_switch(*a, log_modulus).cast_into())
@@ -298,14 +379,14 @@ impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
             ))
         };
 
-        Self {
+        Self::from_raw_parts(
             body,
             packed_mask,
             packed_diffs,
-            lwe_dimension: ct.lwe_size().to_lwe_dimension(),
+            ct.lwe_size().to_lwe_dimension(),
             uncompressed_ciphertext_modulus,
             grouping_factor,
-        }
+        )
     }
 
     /// Converts back a compressed ciphertext to its initial modulus
@@ -313,6 +394,14 @@ impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
     /// The output must got through a PBS to reduce the noise
     pub fn extract(&self) -> FromCompressionMultiBitModulusSwitchedCt {
         let masks: Vec<usize> = self.packed_mask.unpack().collect();
+
+        assert_eq!(
+            masks.len() % self.grouping_factor.0,
+            0,
+            "Mask len (={}) should be a multiple of grouping factor (={})",
+            masks.len(),
+            self.grouping_factor.0
+        );
 
         let mut diffs_two_complement: Vec<usize> = vec![];
 
@@ -430,5 +519,53 @@ impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>> ParameterSetCo
                 }
             }
             && *uncompressed_ciphertext_modulus == lwe_ct_parameters.ct_modulus
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use rand::Rng;
+
+    use super::*;
+    #[test]
+    fn test_from_raw_parts() {
+        type Scalar = u64;
+
+        let len = 694;
+        let log_modulus = 12;
+        let grouping_factor = LweBskGroupingFactor(3);
+
+        let ciphertext_modulus = CiphertextModulus::new_native();
+
+        let mut lwe = LweCiphertext::new(Scalar::ZERO, LweSize(len), ciphertext_modulus);
+
+        // We don't care about the exact content here
+        rand::thread_rng().fill(lwe.as_mut());
+
+        let compressed = CompressedModulusSwitchedMultiBitLweCiphertext::compress(
+            &lwe,
+            CiphertextModulusLog(log_modulus),
+            grouping_factor,
+        );
+
+        let (
+            body,
+            packed_mask,
+            packed_diffs,
+            lwe_dimension,
+            uncompressed_ciphertext_modulus,
+            grouping_factor,
+        ) = compressed.into_raw_parts();
+
+        let rebuilt = CompressedModulusSwitchedMultiBitLweCiphertext::from_raw_parts(
+            body,
+            packed_mask,
+            packed_diffs,
+            lwe_dimension,
+            uncompressed_ciphertext_modulus,
+            grouping_factor,
+        );
+
+        let _ = rebuilt.extract();
     }
 }
--- a/tfhe/src/core_crypto/backward_compatibility/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
+++ b/tfhe/src/core_crypto/backward_compatibility/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
@@ -1,12 +1,56 @@
-use tfhe_versionable::VersionsDispatch;
+use std::convert::Infallible;
 
-use crate::core_crypto::prelude::{
-    CastFrom, CastInto, CompressedModulusSwitchedMultiBitLweCiphertext, UnsignedInteger,
-};
+use packed_integers::PackedIntegers;
+use tfhe_versionable::{Upgrade, Version, VersionsDispatch};
+
+use crate::core_crypto::prelude::*;
+
+#[derive(Version)]
+pub struct CompressedModulusSwitchedMultiBitLweCiphertextV0<
+    Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>,
+> {
+    body: usize,
+    packed_mask: PackedIntegers<usize>,
+    packed_diffs: Option<PackedIntegers<usize>>,
+    lwe_dimension: LweDimension,
+    uncompressed_ciphertext_modulus: CiphertextModulus<Scalar>,
+    grouping_factor: LweBskGroupingFactor,
+}
+
+impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
+    Upgrade<CompressedModulusSwitchedMultiBitLweCiphertext<Scalar>>
+    for CompressedModulusSwitchedMultiBitLweCiphertextV0<Scalar>
+{
+    type Error = Infallible;
+
+    fn upgrade(
+        self,
+    ) -> Result<CompressedModulusSwitchedMultiBitLweCiphertext<Scalar>, Self::Error> {
+        // In the previous version a last coefficient for the body was stored along the mask
+        // elements. This was a duplicate, so we remove it.
+
+        let mask_without_body = PackedIntegers::from_raw_parts(
+            self.packed_mask.packed_coeffs()[..self.lwe_dimension.0].to_vec(),
+            self.packed_mask.log_modulus(),
+            self.packed_mask.initial_len(),
+        );
+        Ok(
+            CompressedModulusSwitchedMultiBitLweCiphertext::from_raw_parts(
+                self.body,
+                mask_without_body,
+                self.packed_diffs,
+                self.lwe_dimension,
+                self.uncompressed_ciphertext_modulus,
+                self.grouping_factor,
+            ),
+        )
+    }
+}
 
 #[derive(VersionsDispatch)]
 pub enum CompressedModulusSwitchedMultiBitLweCiphertextVersions<
     Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>,
 > {
-    V0(CompressedModulusSwitchedMultiBitLweCiphertext<Scalar>),
+    V0(CompressedModulusSwitchedMultiBitLweCiphertextV0<Scalar>),
+    V1(CompressedModulusSwitchedMultiBitLweCiphertext<Scalar>),
 }


tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs
```diff
@@ -212,7 +212,7 @@ impl<Scalar: UnsignedInteger + CastInto<usize> + CastFrom<usize>>
 
         let body = modulus_switch(*input_lwe_body.data, log_modulus).cast_into();
 
-        let modulus_switched: Vec<usize> = ct
+        let modulus_switched: Vec<usize> = input_lwe_mask
```
mayeul-zama, line 215: A `assert_eq!(masks.len() % grouping_factor, 0)` could be added in the `extract` function

mayeul-zama: We could upgrade the type with a new version
Not sure it's worth it, but that would be safer

nsarlin-zama: Yes you're right!

nsarlin-zama: I added the type upgrade, with a from_raw_parts impl and a test for it

IceTDrinker (commented):
A few questions, don't recall everything on this

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-:-OOcOsJP4DOb6LhrP0pX:b-jkk3gk)* status: all files reviewed, 5 unresolved discussions (waiting on @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 204 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOSaT4Cm4barse6Nb:-OOcOSaT4Cm4barse6Nc:b1awebu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L204)):*
> ```XML
>             packed_mask.initial_len());
> 
>         if let Some(diffs) = packed_diffs.as_ref() {
> ```

don't recall but why is this an option ?
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOcOZgN9E91O_egSL17:bvptsyk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*
> ```XML
>             let multi_bit_elements = packed_mask.initial_len() / grouping_factor.0;
>             let ggsw_per_multi_bit_element = grouping_factor.ggsw_per_multi_bit_element().0;
>             let num_powers_of_2 =
> ```

what is this ?
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 208 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOeEZ4hgTmsC497TD:-OOcOeEZ4hgTmsC497TE:b-drchrp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L208)):*
> ```XML
>             let ggsw_per_multi_bit_element = grouping_factor.ggsw_per_multi_bit_element().0;
>             let num_powers_of_2 =
>                 (usize::BITS - ggsw_per_multi_bit_element.leading_zeros()) as usize;
> ```

ilog2 ? unsure on that one
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcO5FsBuuzYeWRjOnY:-OOcO5FsBuuzYeWRjOnZ:b-2wc0s8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L396)):*
> ```XML
> 
>         assert_eq!(
>             masks.len() % self.grouping_factor.0,
> ```

isn't that guaranteed by the construction of the entity ? :thinking: 

looks like the from_raw_parts checks it ?

the compress function should call it if you want the invariant ?

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: all files reviewed, 5 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 204 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOSaT4Cm4barse6Nb:-OOfnR0836xfU4qUr-tZ:bfbnp7t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L204)):*

don't recall but why is this an option ?
</blockquote></details>

I did not create the type, but from the compress code below if the list is full of zeros then it is replaced by "None".
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOfnUbN1pEF3_ptH9Bj:bh77y2f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*

what is this ?
</blockquote></details>

The goal is to re-compute the expected len of "diffs".

In the compression code below, the diff list is built like this
```rust
       for lwe_mask_elements in input_lwe_mask.as_ref().chunks_exact(grouping_factor.0) {
            for ggsw_idx in 1..grouping_factor.ggsw_per_multi_bit_element().0 {
                // We need to store the diff sums of more than one element as we store the
                // individual modulus_switched elements
                if ggsw_idx.is_power_of_two() {
                    continue;
                }
               
                 // ...
                 diffs.push(diff);
        }
}
```
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 208 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOeEZ4hgTmsC497TD:-OOfr9SX2WwZCFrKQ2ym:bdtr2he) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L208)):*

ilog2 ? unsure on that one
</blockquote></details>

yes ceil_ilog2 works too !
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcO5FsBuuzYeWRjOnY:-OOfrcDM2AunPQfueyGK:b-90kxad) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L396)):*

isn't that guaranteed by the construction of the entity ? :thinking: 

looks like the from_raw_parts checks it ?

the compress function should call it if you want the invariant ?
</blockquote></details>

I was applying the suggestion from Mayeul, but yes I can add it in the compression code too

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @mayeul-zama and @nsarlin-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 204 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOSaT4Cm4barse6Nb:-OOg3Hs_98pmuxE6VbjK:b-9ldshb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L204)):*

I did not create the type, but from the compress code below if the list is full of zeros then it is replaced by "None".
</blockquote></details>

erf... such a silly optimization given we'll almost always have non zero diffs, thanks
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOg3QEC8fbKXkZ0eVmj:b-walait) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*

The goal is to re-compute the expected len of "diffs".

In the compression code below, the diff list is built like this
```rust
       for lwe_mask_elements in input_lwe_mask.as_ref().chunks_exact(grouping_factor.0) {
            for ggsw_idx in 1..grouping_factor.ggsw_per_multi_bit_element().0 {
                // We need to store the diff sums of more than one element as we store the
                // individual modulus_switched elements
                if ggsw_idx.is_power_of_two() {
                    continue;
                }
               
                 // ...
                 diffs.push(diff);
        }
}
```
</blockquote></details>

ok but the name does not have meaning to me :sweat_smile: 
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 208 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOeEZ4hgTmsC497TD:-OOg3WK54Ep8Ma0qgOOa:b-3e6gsn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L208)):*

yes ceil_ilog2 works too !
</blockquote></details>

ok could be easier to read
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcO5FsBuuzYeWRjOnY:-OOg3_bvA_1egRy6HQST:b-pc0byg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L396)):*

I was applying the suggestion from Mayeul, but yes I can add it in the compression code too
</blockquote></details>

what I man is, you can call the from_raw_parts after compression so that the invariant is enforced

IceTDrinker (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @mayeul-zama and @nsarlin-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 209 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4LwR0kzIlSd5wxxG:-OOg4LwR0kzIlSd5wxxH:beojh0k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L209)):*
> ```XML
>             let ggsw_per_multi_bit_element = grouping_factor.ggsw_per_multi_bit_element().0;
> 
>             // No diff is stored when ggsw id is a power of 2, so we remove them from the number of
> ```

that comment is so odd, if you want we can talk about it in a call, I don't understand what's happening nor the new comment
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 347 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4Yv91i_jjL2vX5_Z:-OOg4YvA-aXFZz9oK9N0:b-j59j3e) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L347)):*
> ```XML
> 
>         // Sanity check
>         assert_eq!(
> ```

instead of asserting here could use the from_raw_parts at the bottom of the function, it will enforce invariants

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOgG0eA87PgFCOgvNC_:b5skyva) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*

I think we should have `num_powers_of_2 = ggsw_per_multi_bit_element.ceil_ilog2()` and remove `1` in `expected_diffs_len` because the loop starts at `1`

The result is the same, but it would be clearer IMO
</blockquote></details>

for me the loop starts at 1 only because 0 is itself a power of 2, so I find it easier to reason about if I don't make a special case of it. 
(I would be even in favor of starting the loop at 0 and let it be handled by the `if ggsw_idx.is_power_of_two()`, but I think it's starting to become a bit heavy for what was supposed to be a quick fix ! We could also start the loop at 3 since 1 and 2 will be skipped too)

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOgHtWc-5coQI0pibWn:b-qshlif) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*

for me the loop starts at 1 only because 0 is itself a power of 2, so I find it easier to reason about if I don't make a special case of it. 
(I would be even in favor of starting the loop at 0 and let it be handled by the `if ggsw_idx.is_power_of_two()`, but I think it's starting to become a bit heavy for what was supposed to be a quick fix ! We could also start the loop at 3 since 1 and 2 will be skipped too)
</blockquote></details>

Ah no I forgot that 0 is not a power of 2 😅

mayeul-zama:
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 207 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcOZgN9E91O_egSL16:-OOgIKgs47XUbUYaKvg_:b-3w15gw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L207)):*

Ah no forgot that 0 is not a power of 2 😅
</blockquote></details>

Nitpick in advance, `((ggsw_per_multi_bit_element - 1) - num_powers_of_2)`, would be clearer to me than `((ggsw_per_multi_bit_element - num_powers_of_2) - 1)`

mayeul-zama (approved):
:lgtm:
Thanks!

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-:-OOgKiS48hEtW6p4rj1l:b-aw02zl)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker and @nsarlin-zama)

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @mayeul-zama and @nsarlin-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcO5FsBuuzYeWRjOnY:-OOgkUJS52BMKqkC6W7A:b6upzhn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L396)):*

done
</blockquote></details>

then do we still need the assert ?
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 209 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4LwR0kzIlSd5wxxG:-OOgkH-l554hmPiaRD-Y:b-y379xk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L209)):*

I updated it, tell me if it's better. But remember that it is supposed to be read with the compression code just below in mind, not in reviewable.
</blockquote></details>

@mayeul-zama I don't understand the skips, we can skip the diffs for individual elements but not the groups ?

e.g. for a grouping factor == 4

we have

mod_switch(a0), and modswitch(a1), mod_switch(a2), and modswitch(a3) which can be stored without a diff (has nothing to do with a power of two as a3 is not a power of 2), but the sums need a diff right ?

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @mayeul-zama and @nsarlin-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 209 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4LwR0kzIlSd5wxxG:-OOgmYQKDEjPFVZsv_Oy:btwfe84) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L209)):*

Ok finally understood what was going on, the is_power_of_two indicates that only one key bit is encrypted/only one mask is selected (as the index is used as a selector), I would like a more precise comment, explaining that a power of 2 index corresponds to a mask or key bit selector of a single value and the mod switch in that case does not require a correction as it's not a sum of terms
</blockquote></details>

same for the part in compress

                // We need to store the diff sums of more than one element as we store the

                // individual modulus\_switched elements

                if ggsw\_idx.is\_power\_of\_two() {

                    continue;

                }

indicate it corresponds to a selector of a single value as tehre is a single bit set in the index in that case

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOcO5FsBuuzYeWRjOnY:-OOgn4fYB3TBPvVHkrJ9:b-jl0mp4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6cb39868939d8f114b8c15561137f1595300b441/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L396)):*

then do we still need the assert ?
</blockquote></details>

I don't really have an opinion about having it or not, but Mayeul suggested to add it. Maybe it still adds a bit of explicitness to show that the invariant is still valid here too.
___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 209 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4LwR0kzIlSd5wxxG:-OOgmopgBeZia72uOPTC:bksr4be) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L209)):*

same for the part in compress

                // We need to store the diff sums of more than one element as we store the

                // individual modulus\_switched elements

                if ggsw\_idx.is\_power\_of\_two() {

                    continue;

                }

indicate it corresponds to a selector of a single value as tehre is a single bit set in the index in that case
</blockquote></details>

I don't know, here I just try to be coherent with the compression, I don't think this comment should explain how the compression itself works...

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @mayeul-zama and @nsarlin-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 209 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOg4LwR0kzIlSd5wxxG:-OOgoJtH-qvk8KMZcCZa:b-oemrrx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/25470739fcee79a1f58d3314c626494d6fe34da3/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L209)):*

I don't know, here I just try to be coherent with the compression, I don't think this comment should explain how the compression itself works...
</blockquote></details>

No of course, I meant to add the precision about the bit selector being a power of two in the compression as well, both go hand in hand

mayeul-zama:
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 211 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOgs2YPDyWTZVVtIDc8:-OOgs2YPDyWTZVVtIDc9:b-rk2dc1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d5a6133b03897a1785747c1d7a369563d99af26a/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L211)):*
> ```XML
>             // In the diff list creation, we skip every power of two, so we have to remove them from
>             // the total count
>             let num_powers_of_2 = (ggsw_per_multi_bit_element.ceil_ilog2()) as usize;
> ```

You could use  `grouping_factor.0` here

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs` line 211 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291#-OOgs2YPDyWTZVVtIDc8:-OOgt9jP82WDeNevDJ33:b5vuwro) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d5a6133b03897a1785747c1d7a369563d99af26a/tfhe/src/core_crypto/entities/compressed_modulus_switched_multi_bit_lwe_ciphertext.rs#L211)):*

You could use  `grouping_factor.0` here
</blockquote></details>

This is correct but it is less explicit I think, I try to stay as close as possible to how the len is computed in the `compress` method. 
You don't have to know how `grouping_factor.ggsw_per_multi_bit_element()` is implemented to understand what is happening.

IceTDrinker (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2291)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @nsarlin-zama)

PR #2298, opened 2025-04-28
chore(shortint): add asserts

mayeul-zama:


tfhe/src/conformance.rs
```diff
@@ -44,6 +44,11 @@ impl ListSizeConstraint {
         if max_inclusive_group_count < min_inclusive_group_count {
             return Err("max_inclusive < min_inclusive".to_owned());
         }
+
+        if group_size == 0 {
+            return Err("group_size == 0".to_owned());
```
IceTDrinker, line 49: more descriptive error please

tfhe/src/conformance.rs
```diff
@@ -52,6 +57,8 @@ impl ListSizeConstraint {
     }
 
     pub fn multiply_group_size(&self, group_size_multiplier: usize) -> Self {
+        assert_ne!(group_size_multiplier, 0);
```
IceTDrinker, line 60: add a descriptive error message

tfhe/src/conformance.rs
```diff
@@ -60,6 +67,8 @@ impl ListSizeConstraint {
     }
 
     pub fn is_valid(&self, size: usize) -> bool {
+        assert_ne!(self.group_size, 0);
```
IceTDrinker, line 70: add a descriptive error message

tfhe/src/conformance.rs
```diff
@@ -60,8 +60,12 @@ impl ListSizeConstraint {
...
-        size % self.group_size == 0
-            && size >= self.min_inclusive_group_count * self.group_size
-            && size <= self.max_inclusive_group_count * self.group_size
+        if self.group_size != 0 {
+            size % self.group_size == 0
+                && size >= self.min_inclusive_group_count * self.group_size
+                && size <= self.max_inclusive_group_count * self.group_size
+        } else {
+            size == 0
+        }
```
IceTDrinker, line 69: self.group_size == 0 means only a size of 0 is accepted ? correct ?

as a general rule (and I think clippy might be saying the same) prefer positive predicates so

if self.group_size == 0 {
    size == 0
} else {
...
}

IceTDrinker (approved):

PR #2300, opened 2025-04-30
chore: fix clippy lint for aarch64 targets, use variables in format string

IceTDrinker:


soonum (approved):

IceTDrinker:
pcc checks on M1 are green, merging

PR #2312, opened 2025-05-06
chore(ci): upgrade actionlint to v1.7.7

soonum:
Usage of bash script to download and extract the final binary has been dropped.
Instead, the tarball is directly fetched according to the ACTIONLINT_VERSION value and the integrity of the tarball is checked with an hardcoded SHA256 sum.


IceTDrinker:
looks like it would need the toolchain update

soonum:
Toolchain update is too cumbersome for just being able to pin actionlint binary.
Closing.

PR #2319, opened 2025-05-09
fix(shortint): store correct ap in ciphertext during encryption

nsarlin-zama:
During encryption the AP inside the ciphertext was always hardcoded to standard

### DIFF
--- a/tfhe/src/core_crypto/commons/parameters.rs
+++ b/tfhe/src/core_crypto/commons/parameters.rs
@@ -289,12 +289,18 @@ pub enum EncryptionKeyChoice {
     Small,
 }
 
+impl EncryptionKeyChoice {
+    pub const fn into_pbs_order(self) -> PBSOrder {
+        match self {
+            Self::Big => PBSOrder::KeyswitchBootstrap,
+            Self::Small => PBSOrder::BootstrapKeyswitch,
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


tfhe/src/core_crypto/commons/parameters.rs
```diff
@@ -289,12 +289,18 @@ pub enum EncryptionKeyChoice {
     Small,
 }
 
+impl EncryptionKeyChoice {
+    pub const fn into_pbs_order(self) -> PBSOrder {
+        match self {
+            Self::Big => PBSOrder::KeyswitchBootstrap,
+            Self::Small => PBSOrder::BootstrapKeyswitch,
+        }
+    }
```
IceTDrinker, line 298: self.into()

IceTDrinker: ah nevermind it needs to be const, the From for PBSOrder could be updated though

nsarlin-zama: no this is const context I cannot call into, that's why I made this function that is now itself called by into

IceTDrinker: and I missed the into update, all good then sorry

IceTDrinker (approved):

PR #2328, opened 2025-05-13
chore(ci): minimum workflows run on draft pull-request

soonum:


IceTDrinker:
do we agree the no ci should be done with the github [no ci] tags in commit ?

IceTDrinker:
as discussed we will use [no ci] for now

PR #2329, opened 2025-05-13
chore(zk): check that crs group element at index n is 0

nsarlin-zama:
Adds a check in the ZK CRS that the group element at index n+1 (one based) is equal to zero.
This is checked during decompression and in `PublicParam::is_usable`, which is used by crs conformance

### DIFF
--- a/tfhe-zk-pok/src/proofs/pke_v2.rs
+++ b/tfhe-zk-pok/src/proofs/pke_v2.rs
@@ -139,8 +139,16 @@ where
             hash_z,
             hash_chi,
         } = compressed;
+
+        let uncompressed_g_lists = GroupElements::uncompress(g_lists)?;
+        if G::G1::projective(uncompressed_g_lists.g_list[n + 1]) != G::G1::ZERO {
+            return Err(InvalidSerializedPublicParamsError::InvalidGroupElements(
+                InvalidSerializedGroupElementsError::MissingPuncteredElement,
+            ));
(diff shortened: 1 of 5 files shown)


tfhe-zk-pok/src/proofs/pke_v2.rs
```diff
@@ -3290,6 +3302,37 @@ mod tests {
...
+
+        assert!(public_param.is_usable());
+
+        let public_param_that_was_compressed =
+            serialize_then_deserialize(&public_param, Compress::Yes).unwrap();
+
+        assert!(public_param_that_was_compressed.is_usable());
+
+        let mut bad_crs = public_param.clone();
+        bad_crs.g_lists.g_list[public_param.n + 1] = bad_crs.g_lists.g_list[public_param.n];
```
kc1212, line 3331: this is because of the zero/one-based indexing?

kc1212: ah I misunderstood, ok this is fine.

nsarlin-zama:
> LGTM!
> 
> The only thing I would add is check that calling `PublicParams::uncompress` actually returns the right error `InvalidSerializedGroupElementsError::MissingPuncteredElement` with `matches!` if it doesn't implement `PartialEq`.

Thanks! I don't think I can use matches here because the constant ZERO comes from a generic parameter.

PR #2331, opened 2025-05-13
fix(gpu): add indexes to modulus switch noise reduction

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
@@ -248,6 +248,7 @@ template <> struct pbs_buffer_128<PBS_TYPE::CLASSICAL> {
   __uint128_t *global_accumulator;
   double *global_join_buffer;
   __uint128_t *temp_lwe_array_in;
+  uint64_t *trivial_indexes;
 
   PBS_VARIANT pbs_variant;
   bool uses_noise_reduction;
@@ -263,11 +264,27 @@ template <> struct pbs_buffer_128<PBS_TYPE::CLASSICAL> {
     cuda_set_device(gpu_index);
     this->pbs_variant = pbs_variant;
     this->uses_noise_reduction = allocate_ms_array;
-    this->temp_lwe_array_in =
-        (__uint128_t *)cuda_malloc_with_size_tracking_async(
-            (lwe_dimension + 1) * input_lwe_ciphertext_count *
-                sizeof(__uint128_t),
-            stream, gpu_index, size_tracker, allocate_ms_array);
+    if (allocate_ms_array) {
+      this->temp_lwe_array_in =
+          (__uint128_t *)cuda_malloc_with_size_tracking_async(
+              (lwe_dimension + 1) * input_lwe_ciphertext_count *
+                  sizeof(__uint128_t),
+              stream, gpu_index, size_tracker, allocate_ms_array);
+      this->trivial_indexes = (uint64_t *)cuda_malloc_with_size_tracking_async(
+          input_lwe_ciphertext_count * sizeof(uint64_t), stream, gpu_index,
+          size_tracker, allocate_ms_array);
+      uint64_t *h_trivial_indexes = new uint64_t[input_lwe_ciphertext_count];
+      for (uint32_t i = 0; i < input_lwe_ciphertext_count; i++)
+        h_trivial_indexes[i] = i;
+
+      cuda_memcpy_with_size_tracking_async_to_gpu(
+          trivial_indexes, h_trivial_indexes,
+          input_lwe_ciphertext_count * sizeof(uint64_t), stream, gpu_index,
+          allocate_gpu_memory);
+
+      cuda_synchronize_stream(stream, gpu_index);
+      delete[] h_trivial_indexes;
+    }
     auto max_shared_memory = cuda_get_max_shared_memory(gpu_index);
     size_t global_join_buffer_size = (glwe_dimension + 1) * level_count *
                                      input_lwe_ciphertext_count *
@@ -404,9 +421,12 @@ template <> struct pbs_buffer_128<PBS_TYPE::CLASSICAL> {
       cuda_drop_with_size_tracking_async(global_accumulator, stream, gpu_index,
                                          gpu_memory_allocated);
 
-    if (uses_noise_reduction)
+    if (uses_noise_reduction) {
       cuda_drop_with_size_tracking_async(temp_lwe_array_in, stream, gpu_index,
                                          gpu_memory_allocated);
+      cuda_drop_with_size_tracking_async(trivial_indexes, stream, gpu_index,
+                                         gpu_memory_allocated);
+    }
   }
 };
 
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cu
@@ -86,13 +86,15 @@ void cuda_modulus_switch_inplace_64(void *stream, uint32_t gpu_index,
 
 void cuda_improve_noise_modulus_switch_64(
     void *stream, uint32_t gpu_index, void *lwe_array_out,
-    void const *lwe_array_in, void const *encrypted_zeros, uint32_t lwe_size,
-    uint32_t num_lwes, uint32_t num_zeros, double input_variance,
-    double r_sigma, double bound, uint32_t log_modulus) {
+    void const *lwe_array_in, void const *lwe_array_indexes,
+    void const *encrypted_zeros, uint32_t lwe_size, uint32_t num_lwes,
+    uint32_t num_zeros, double input_variance, double r_sigma, double bound,
+    uint32_t log_modulus) {
   host_improve_noise_modulus_switch<uint64_t>(
       static_cast<cudaStream_t>(stream), gpu_index,
       static_cast<uint64_t *>(lwe_array_out),
       static_cast<uint64_t const *>(lwe_array_in),
+      static_cast<uint64_t const *>(lwe_array_indexes),
       static_cast<const uint64_t *>(encrypted_zeros), lwe_size, num_lwes,
       num_zeros, input_variance, r_sigma, bound, log_modulus);
 }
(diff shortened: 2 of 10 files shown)


backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -248,6 +248,7 @@ template <> struct pbs_buffer_128<PBS_TYPE::CLASSICAL> {
   __uint128_t *global_accumulator;
   double *global_join_buffer;
   __uint128_t *temp_lwe_array_in;
+  __uint128_t *trivial_indexes;
```
agnesLeroy, line 251: For the indexes we could use uint32_t instead, no?

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cu
```diff
@@ -86,13 +86,15 @@ void cuda_modulus_switch_inplace_64(void *stream, uint32_t gpu_index,
...
-    double r_sigma, double bound, uint32_t log_modulus) {
+    void const *lwe_array_in, void const *lwe_array_indexes,
+    void const *encrypted_zeros, uint32_t lwe_size, uint32_t num_lwes,
+    uint32_t num_zeros, double input_variance, double r_sigma, double bound,
+    uint32_t log_modulus) {
   host_improve_noise_modulus_switch<uint64_t>(
       static_cast<cudaStream_t>(stream), gpu_index,
       static_cast<uint64_t *>(lwe_array_out),
       static_cast<uint64_t const *>(lwe_array_in),
+      static_cast<uint64_t const *>(lwe_array_indexes),
```
agnesLeroy, line 97: Same here the indexes type could be always u32 maybe

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
```diff
@@ -198,13 +197,13 @@ improve_noise_modulus_switch(Torus *array_out, const Torus *array_in,
   // This probably are not needed cause we are setting the values
   sum_mask_errors[threadIdx.x] = 0.f;
   sum_squared_mask_errors[threadIdx.x] = 0.f;
+  auto array_in_index = array_in + indexes[blockIdx.x] * lwe_size;
```
agnesLeroy, line 200: For the name, maybe it could be `this_block_lwe_in`, `this_block_lwe_out`? Wdyt?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -194,7 +194,8 @@ void execute_pbs_async(
             lut_indexes_vec[i] + (ptrdiff_t)(gpu_offset);
 
         void *zeros = nullptr;
-        if (ms_noise_reduction_key != nullptr)
+        if (ms_noise_reduction_key != nullptr &&
+            ms_noise_reduction_key->ptr != nullptr)
```
agnesLeroy, line 198: Could the pointer to the noise reduction key be nullptr if ms_noise_reduction_key isn't a nullptr? :thinking:

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -660,13 +660,15 @@ void cuda_programmable_bootstrap_lwe_ciphertext_vector_64(
       (pbs_buffer<uint64_t, CLASSICAL> *)mem_ptr;
 
   // If the parameters contain noise reduction key, then apply it
-  if (ms_noise_reduction_key != nullptr) {
+  if (ms_noise_reduction_key != nullptr &&
+      ms_noise_reduction_key->ptr != nullptr) {
```
agnesLeroy, line 664: Same question here

tfhe/src/core_crypto/gpu/algorithms/test/modulus_switch_noise_reduction.rs
```diff
@@ -147,6 +148,14 @@ fn check_noise_improve_modulus_switch_noise(
 
     let gpu_index = 0;
     let streams = CudaStreams::new_single_gpu(GpuIndex::new(gpu_index));
+    let num_blocks = 1;
+    let lwe_indexes_usize = (0..num_blocks).collect_vec();
+    let lwe_indexes = lwe_indexes_usize
+        .iter()
+        .map(|&x| <usize as CastInto<u64>>::cast_into(x))
+        .collect_vec();
```
agnesLeroy, line 156: Here you could define lwe_indexes directly as
```
let lwe_indexes: Vec<u64> = (0..num_blocks)
    .map(|x| x as u64)
    .collect();
```

agnesLeroy (commented):
Hey @guillermo-oyarzun! Thanks a lot for this PR, here are a few minor comments

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cu
guillermo-oyarzun: I agree and thought so too, the thing is that i am using the indexes that come from the pbs and those for now have the pbs type, i could leave it fixed in uint64_t for the moment so i can reduce the size of the indexes in the pbs128 at least. Maybe other PR could address to change all those indexes to uint32_t which seems to be more than sufficient.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
guillermo-oyarzun: i answer in the comment below

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
guillermo-oyarzun: yup i like it, i was trying to do a quick patch and didn't give much thought in the names

backends/tfhe-cuda-backend/cuda/src/crypto/ciphertext.cu
bbarbakadze: I don't think `uint32_t` is a good option now, since in `int_radix_lut` we use `Torus` as a type for indexes, maybe for now we use uint64_t and later we do a refactor for all the indexes in different pr.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
guillermo-oyarzun: strangely as it is, it is what we were observing, not sure if it is a matter of the way we interact with rust

guillermo-oyarzun: For some reason the pointer of noise_reduction_key was not nullptr but the noise_reduction_key->ptr was nullptr, not sure if it is related on how we handle the structure when we pass it to rust.

agnesLeroy: Aaah ok, we always pass a valid pointer to the ms_noise_reduction_key from Rust because of how the prepare_cuda_ms_noise_reduction_key_ffi is written. So maybe only checking if `ms_noise_reduction_key->ptr` is not a `nullptr` is enough here.

agnesLeroy: In Rust we do:
```
pub fn prepare_cuda_ms_noise_reduction_key_ffi(
    input_ms_key: Option<&CudaModulusSwitchNoiseReductionKey>,
    modulus: f64,
) -> CudaModulusSwitchNoiseReductionKeyFFI {
    input_ms_key.map_or(
        CudaModulusSwitchNoiseReductionKeyFFI {
            ptr: std::ptr::null_mut(),
            num_zeros: 0,
            ms_bound: 0.0,
            ms_r_sigma: 0.0,
            ms_input_variance: 0.0,
        },
        |ms_key| CudaModulusSwitchNoiseReductionKeyFFI {
            ptr: ms_key.modulus_switch_zeros.ptr.as_ptr(),
            num_zeros: ms_key.num_zeros,
            ms_bound: ms_key.ms_bound.0,
            ms_r_sigma: ms_key.ms_r_sigma_factor.0,
            ms_input_variance: ms_key.ms_input_variance.get_modular_variance(modulus).value,
        },
    )
}
```
so `ms_noise_reduction_key` is always a valid pointer

guillermo-oyarzun: the problem comes when we call it from inside the backend, and we pass a nullptr in the ms_noise_reduction_key, in that case if we only check the ->ptr it crashes

agnesLeroy: Aaah ok in that case both checks are necessary indeed!

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
```diff
@@ -198,13 +197,13 @@ improve_noise_modulus_switch(Torus *array_out, const Torus *array_in,
   // This probably are not needed cause we are setting the values
   sum_mask_errors[threadIdx.x] = 0.f;
   sum_squared_mask_errors[threadIdx.x] = 0.f;
+  auto this_block_lwe_in = array_in + indexes[blockIdx.x] * lwe_size;
+  auto this_block_lwe_out = array_out + blockIdx.x * lwe_size;
```
bbarbakadze, line 201: should be: `auto this_block_lwe_out = array_out + indexes[blockIdx.x] * lwe_size;`

bbarbakadze (changes requested):

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
agnesLeroy: Hmm no, I think it should not otherwise pbs indexes will be messed up no?

guillermo-oyarzun: beka is right, we need to save them considering the same indexing order, cause this is the input of the pbs

agnesLeroy (approved):

bbarbakadze (approved):

PR #2335, opened 2025-05-14
chore(deps): bump rand 0.8 -> 0.9.1, rand_core 0.6.4 -> 0.9.3

mkmks:


IceTDrinker:
let's close this if we are not moving ahead for now ? @mkmks

PR #2337, opened 2025-05-15
fix(bench): compilation error without the internal-keycache feature

nsarlin-zama:
fix build in tfhe-benchmark: shortint without internal-keycache and gpu without integer


soonum (approved):
Thanks for the fix :goat:

soonum (approved):

PR #2342, opened 2025-05-16
chore(bench): add suitable heuristic for zk throughput

soonum:
Heuristic based on PBS count was flawed since a ZK verification operation will eat up to 32 threads on the machine. 
The previous heuristic could generate an input data vector way bigger than the total of threads divided by 32. 
This in turn lead to long execution time for benchmark and generate bad results.


nsarlin-zama (approved):
thanks!!

PR #2351, opened 2025-05-19
chore: use wrapping div in long_run

tmontaigu:


tmontaigu:
We could, but technically, unsigned division cannot overflow, only the signed division can overflow (and only when doing T::MIN / -1, like -128i8 / -1 = 128, but 128 cannot be represented on i8, so it overflows to -128

agnesLeroy (approved):

PR #2355, opened 2025-05-19
Pg/hpu bench fix

pgardratzama:
Issue is appearing when running bench with iter > 1, for example:
`cargo run --profile devo --features=integer,internal-keycache,hpu-v80 --example hpu_bench -- --iter 10 --integer-w 64`
because of usage of .next_back() measured latency is only the latency of the 1st operation.

No emergency to merge.

### DIFF
--- a/tfhe/examples/hpu/bench.rs
+++ b/tfhe/examples/hpu/bench.rs
@@ -265,7 +265,7 @@ pub fn main() {
                     std::hint::black_box(&res);
                     res
                 })
-                .next_back()
+                .last()
                 .expect("Iteration must be greater than 0");
 
             // let res_fhe = $fhe_type::from(res_hpu);
--- a/tfhe/examples/hpu/bench.rs
+++ b/tfhe/examples/hpu/bench.rs
@@ -259,6 +259,7 @@ pub fn main() {
             );
             let roi_start = Instant::now();
 
+            #[allow(clippy::double_ended_iterator_last)]
             let res_hpu = (0..args.iter)
                 .map(|_i| {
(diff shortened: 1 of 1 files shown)


tfhe/examples/hpu/bench.rs
```diff
@@ -259,13 +259,14 @@ pub fn main() {
...
+            #[allow(clippy::double_ended_iterator_last)]
             let res_hpu = (0..args.iter)
                 .map(|_i| {
                     let res = HpuRadixCiphertext::exec(&proto, iop.opcode(), &srcs_enc, &imms);
                     std::hint::black_box(&res);
                     res
                 })
-                .next_back()
+                .last()
                 .expect("Iteration must be greater than 0");
```
IceTDrinker, line 270: better to loop or collect first, iterators being greedy, without comment here it's easy to re introduce the bug

Bapt-Roux:
Commits were squashed and rebased on main.
@IceTDrinker could you have a look 🙏

tfhe/examples/hpu/bench.rs
```diff
@@ -260,16 +260,19 @@ pub fn main() {
...
-                .map(|_i| {
+                .filter_map(|i| {
                     let res = HpuRadixCiphertext::exec(&proto, iop.opcode(), &srcs_enc, &imms);
-                    std::hint::black_box(&res);
-                    res
+                    if i == (args.iter - 1) {
+                        Some(res)
+                    } else {
+                        None
+                    }
```
IceTDrinker, line 269: Just for the record, there could be a world where a very smart optimizer might trivialize this and not run some perf stuff, but thankfully today it works so all good :)

IceTDrinker (approved):
Thanks !

PR #2368, opened 2025-05-22
chore(shortint): prepare the v1.3 params folder

nsarlin-zama:
prepare the V1.3 parameters.
This is currently just a duplicate of the 1.2 folder, but it will allow to add 1.3 specific parameters (for example for the noise squashing compression)


IceTDrinker (approved):
Thanks !

PR #2377, opened 2025-05-26
Am/test/future noise checks

IceTDrinker:
Replacement of https://github.com/zama-ai/tfhe-rs/pull/1975


IceTDrinker:
confirmed just now, sigh

tfhe/src/core_crypto/algorithms/test/modulus_switch_noise_reduction.rs
```diff
@@ -436,19 +436,23 @@ fn check_noise_improve_modulus_switch_noise(
         .unzip();
 
     println!(
-        "mean(&ms_errors)                     {}2^{:.2}",
-        if mean(&ms_errors) > 0_f64 { "+" } else { "-" },
-        mean(&ms_errors).abs().log2()
+        "arithmetic_mean(&ms_errors)                     {}2^{:.2}",
+        if arithmetic_mean(&ms_errors) > 0_f64 {
```
mayeul-zama: `arithmetic_mean(&ms_errors)` could be stored in a variable to avoid recomputing

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -50,21 +51,164 @@ pub mod test_tools {
...
+        pub fn lower_bound(&self) -> f64 {
+            self.lower_bound
+        }
+
+        pub fn upper_bound(&self) -> f64 {
+            self.upper_bound
+        }
+    }
+
+    pub fn mean_confidence_interval(
```
mayeul-zama: Maybe add a comment to clarify that
- the input variable must be a normal or
- `sample_count` must be big enough so that we can make the assumption that Central limit theorem applies

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -50,21 +51,164 @@ pub mod test_tools {
...
+        // observed_failures = round(n_tests * p_fail)
+        //
+        // # Clopper-Pearson Exact Confidence Interval
+        // lower_bound = stats.beta.ppf(alpha / 2, observed_failures, n_tests - observed_failures +
+        // 1) upper_bound = stats.beta.ppf(1 - alpha / 2, observed_failures + 1, n_tests -
+        // observed_failures)
+        //
+        // print("Observed number of failures (k):", observed_failures)
+        // print(f"Confidence Interval ({(1-alpha)*100}%): [2^{round(math.log2(lower_bound),3)},
+        // 2^{round(math.log2(upper_bound),3)}]")
```
mayeul-zama: To remove

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -50,21 +51,164 @@ pub mod test_tools {
...
+
+        MeanConfidenceInterval {
+            lower_bound,
+            upper_bound,
+        }
+    }
+
+    /// Return a MeanConfidenceInterval when you cannot evaluate the standard deviation of a random
+    /// variable
+    pub fn clopper_pearson_exact_confidence_interval(
```
mayeul-zama: If I understand correctly, this function returns a confidence interval on the estimated probability of success

So returning a `MeanConfidenceInterval` is confusing to me

even if we have a confidence interval around `estimated_p_success = mean(success)` with `success in {0, 1}`

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -50,21 +51,166 @@ pub mod test_tools {
...
+        pub fn lower_bound(&self) -> Variance {
+            self.lower_bound
+        }
+
+        pub fn upper_bound(&self) -> Variance {
+            self.upper_bound
+        }
+    }
+
+    pub fn variance_confidence_interval(
```
mayeul-zama: Maybe add a comment to clarify that
- the input variable must be a normal or
- `sample_count` must be big enough so that we can make the assumption that Central limit theorem applies

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -122,36 +268,45 @@ pub mod test_tools {
         pub null_hypothesis_is_valid: bool,
     }
 
+    pub fn equivalent_pfail_gaussian_noise(
+        original_precision_with_padding: u32,
+        original_pfail: f64,
+        new_precision_with_padding: u32,
+    ) -> f64 {
+        let original_precision_with_padding_and_noise_gap =
+            original_precision_with_padding as f64 + 1.0;
```
mayeul-zama: Not sure what this noise gap represents?

tfhe/src/core_crypto/commons/mod.rs
```diff
@@ -122,36 +268,45 @@ pub mod test_tools {
...
+        let new_precision_with_padding_and_noise_gap = new_precision_with_padding as f64 + 1.0;
+
+        let z_original_pfail =
```
mayeul-zama, line 304: I'm not familiar with this notation, it it standard?

Otherwise I think this represents the std_dev

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,232 @@
...
+
+#[test]
+fn test_noise_check_secret_key_encryption_noise_tuniform() {
+    let params: AtomicPatternParameters = TEST_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128.into();
+
+    let modulus_as_f64 = if params.ciphertext_modulus().is_native_modulus() {
+        2.0f64.powi(u64::BITS as i32)
+    } else {
+        params.ciphertext_modulus().get_custom_modulus() as f64
+    };
```
mayeul-zama: There's `raw_modulus_float`

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,232 @@
...
+                "{signed_noise} is not <= {inclusive_max_val}"
+            );
+
+            // Rescale
+            signed_noise as f64 / modulus_as_f64
+        })
+        .collect();
+
+    let measured_variance = variance(&noise_samples);
+    // With our test settings this gives a range of around 1/100 of a bit, this should be accurate
```
mayeul-zama: Not sure what you mean by "1/100 of a bit"

Is it that `max_bound.sqrt() - min_bound.sqrt() ≈ encoding_step / 100`?

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,232 @@
...
+    let chunk_size = NB_TEST.div_ceil(pool_count);
+
+    let noise_samples: Vec<_> = thread_pools
+        .into_par_iter()
+        .zip((0..NB_TEST).into_par_iter().chunks(chunk_size))
+        .flat_map(|(pool, work_idx)| {
+            pool.install(|| {
+                work_idx
+                    .into_iter()
+                    .map(|_| {
```
mayeul-zama, line 167: Does this improve performance compared to a simple `into_par_iter`?

tfhe/src/core_crypto/algorithms/test/modulus_switch_noise_reduction.rs
IceTDrinker: look at you making me clean your old tests 😂

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: good remark

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: not sure what you mean but the MeanConfidenceInterval returns a confidence interval given a measured value which is on average a value we are estimating without having access to its standard dev.

Essentially we compute an mean/average pfail ONCE, so we don't have the way to estimate it's std dev, this is a formula provided by Sam to estimate the confidence interval around this single measurement

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: fair

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: so we have 3 different notions of precision

- the cleartext space: usually 4 bits
- cleartext with padding: 5 bits
- we have the bit just before the message that needs to be 0 for a correct decryption, so that's what this is, and here it would be 6 bits, essentially it counts the bit at $$\Delta/2$$

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: z_... is the z/standard score it counts how many std dev a value has for a given gaussian https://en.wikipedia.org/wiki/Standard_score

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: indeed, it's a recent addition and it will be useful here

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: the expect std dev value has a log2 value of e.g. 10.00 and we have a lower bound in log2 that is ~9.99 and an upper bound of 10.01

meaning the measurement is pretty precise, as I'm using 0.999 one could fear the interval gets very large/not very reprensentative

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: actually no idea, this weird code (and I should have added a comment) is to avoid the infamous rayon borrow mut crash we get in that context

IceTDrinker: no way around it at the HL level I'm afraid

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: for now I'm going to limit it to gaussian distributions as we have a mean measurement, and not a sum of iid random variables

tfhe/src/core_crypto/commons/mod.rs
mayeul-zama: Ok
The bit before the message may be 1 if the noise is negative.
I think the term "precision" for this last notion is a bit confusing.
Maybe `log_noise_bound` ?

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
mayeul-zama: Ok
Maybe the comment could say "this gives log2_upper_bound - log2_lower_bound < 0.02 * expected_value"

tfhe/src/core_crypto/commons/mod.rs
mayeul-zama: In my understanding, the clopper pearson only applies to a binomial distribution which is only characterized by its `p_success` (and number of samples). It returns a confidence interval on this `p_success`.

A set of samples from a continuous distribution must first be converted to a set of yes/no (with a loss of information) to use this estimation.
IMO this use is too specific to appear in this function documentation.
Maybe a wrapper function for this use case could make sense.

Anyway, this function isn't used anywhere, maybe we could remove it

IceTDrinker: will be used in further commits for shortint

IceTDrinker: can rename to indicate it's a confidence interval for our probability of failure

tfhe/src/core_crypto/commons/mod.rs
IceTDrinker: precision is the naming form the R&D optimization stuff, I'll take a look for naming

IceTDrinker: it's not a log noise bound, and the bit needs to be 0 for correctness IIRC

IceTDrinker: renamed noise free bits, added a long docstring to explain the logic

tfhe/src/core_crypto/gpu/algorithms/test/modulus_switch_noise_reduction.rs
```diff
@@ -209,19 +209,23 @@ fn check_noise_improve_modulus_switch_noise(
...
-        "mean(&ms_errors_improved)            {}2^{:.2}",
-        if mean(&ms_errors_improved) > 0_f64 {
+        "arithmetic_mean(&ms_errors_improved)            {}2^{:.2}",
+        if arithmetic_mean(&ms_errors_improved) > 0_f64 {
             "+"
         } else {
             "-"
         },
-        mean(&ms_errors_improved).abs().log2()
+        arithmetic_mean(&ms_errors_improved).abs().log2()
```
mayeul-zama: The result can also be precomputed and stored

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,230 @@
...
+    let (inclusive_min_val, inclusive_max_val, expected_variance) = match encryption_noise {
+        DynamicDistribution::Gaussian(_gaussian) => {
+            panic!("This test is written for TUniform, wrong parameter set used")
+        }
+        DynamicDistribution::TUniform(tuniform) => (
+            tuniform.min_value_inclusive(),
+            tuniform.max_value_inclusive(),
+            tuniform.variance(modulus_as_f64),
+        ),
+    };
```
mayeul-zama, line 39: Nitpick
I think this could be written
```
let DynamicDistribution::TUniform(tuniform)= encryption_noise else {
            panic!("This test is written for TUniform, wrong parameter set used")
};

let inclusive_min_val = tuniform.min_value_inclusive();
let inclusive_max_val = tuniform.max_value_inclusive();
let expected_variance = tuniform.variance(modulus_as_f64);
```

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
```diff
@@ -0,0 +1,230 @@
...
+    // Hack to avoid server key needs and get the ciphertext directly
+    cpke_params.expansion_kind =
+        CompactCiphertextListExpansionKind::NoCasting(block_params.encryption_key_choice.into());
+
+    let modulus_as_f64 = if cpke_params.ciphertext_modulus.is_native_modulus() {
+        2.0f64.powi(u64::BITS as i32)
+    } else {
+        cpke_params.ciphertext_modulus.get_custom_modulus() as f64
+    };
+
```
mayeul-zama: raw_modulus_float

IceTDrinker: missed that one indeed

tfhe/src/core_crypto/gpu/algorithms/test/modulus_switch_noise_reduction.rs
IceTDrinker: 🙄

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: the alternative is the following :

    let (inclusive_min_val, inclusive_max_val, expected_variance) =
        if let DynamicDistribution::TUniform(tuniform) = encryption_noise {
            (
                tuniform.min_value_inclusive(),
                tuniform.max_value_inclusive(),
                tuniform.variance(modulus_as_f64),
            )
        } else {
            panic!()
        };

not worth it with the if let IMO

mayeul-zama: Not sure why you considered this alternative which is IMO worse

mayeul-zama (approved):
:lgtm:
Thanks!

> Reviewed 1 of 15 files at r1, 8 of 8 files at r4, 5 of 5 files at r5, 1 of 1 files at r6, 2 of 3 files at r7, 7 of 14 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2377#-:-OS4mP5v6oR0BXY1TuY1:b-aw02zl)* status: 8 of 15 files reviewed, 8 unresolved discussions (waiting on @IceTDrinker, @nsarlin-zama, and @tmontaigu)

tfhe/src/high_level_api/tests/noise_distribution/encryption_noise.rs
IceTDrinker: the way you wrote your code, I don't have access to tuniform

IceTDrinker: Ah no ok, don't know what happened when I tried it 🤔

PR #2381, opened 2025-05-27
feat(csprng): add Xof random generation

tmontaigu:
This adds a new kind of seed to the csprng

When created which such seed, the AES-CTR random generator initialization changes:
- The AES-KEY used is initialized differently
- The AES-CTR starts with a CTR that may not be 0

The changes make it so that the counter still goes from 0..MAX, but now the AES-CTR will encrypt the counter + some offset allowing to keep the regular behavior and the new one

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381)


IceTDrinker (commented):
Looks generally ok but want to make sure we get this right, some questions/comments

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-:-ORMZse_-_O22x42LbNK:b-iasyur)* status: all files reviewed, 10 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe-csprng/src/generators/implem/soft/block_cipher.rs` line 41 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMOpKwDgfSoBoiYB3e:-ORMOpKwDgfSoBoiYB3f:b-q3e9h7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/soft/block_cipher.rs#L41)):*
> ```XML
> }
> 
> #[allow(clippy::too_many_arguments)]
> ```

too many arguments ?
___
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 239 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQTnY7Cvjuv-usNCS:-ORMQTnY7Cvjuv-usNCT:b-f88id0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/aes_ctr/mod.rs#L239)):*
> ```XML
>     for mi in blocks {
>         prev_c = c;
>         c = u128::from_le_bytes(aes.generate_next(AesIndex(prev_c ^ mi)));
> ```

endianness wise is that ok ? I don't seem to see it in the XOF object definition in the NIST doc

also asking because I have seen some ne_bytes elsewhere in the code base and it's unclear to me if that would work properly on a big endian target (though as we know, those are very niche)
___
*[`tfhe-csprng/src/seeders/mod.rs` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMOyHe4Y3uiTcTrOp1:-ORMOyHe4Y3uiTcTrOp2:b9luutv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L12)):*
> ```XML
> pub struct Seed(pub u128);
> 
> /// A Seed as described in the NIST document
> ```

think there is an eprint ?
___
*[`tfhe-csprng/src/seeders/mod.rs` line 25 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMP5LhAhEiqaFlbQIW:-ORMP5LhAhEiqaFlbQIX:b-ve7ft) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L25)):*
> ```XML
> 
> impl XofSeed {
>     const DOMAIN_SEP_LEN: usize = 8;
> ```

just to make sure (better than being sorry) the 8 bytes for domain sep was confirmed ?
___
*[`tfhe-csprng/src/seeders/mod.rs` line 31 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYN0a9GdOPkLTapta:-ORMYN0a9GdOPkLTaptb:b5dx4gn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L31)):*
> ```XML
>         let mut data = vec![0u8; size_of::<u128>() + domain_separator.len()];
>         data[..Self::DOMAIN_SEP_LEN].copy_from_slice(domain_separator.as_slice());
>         data[Self::DOMAIN_SEP_LEN..].copy_from_slice(seed.to_le_bytes().as_slice());
> ```

layout ok with respect to description in the doc ?

doc seems to say seed || str and here I kind of read it as str || seed, but I'm guessing it's a matter of endianness
___
*[`tfhe-csprng/src/seeders/mod.rs` line 63 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYkeT2I_1K2aNAlVu:-ORMYkeT2I_1K2aNAlVv:bxf39jy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L63)):*
> ```XML
>     /// as blocks of u128 bits
>     pub(crate) fn iter_u128_blocks(&self) -> impl Iterator<Item = u128> + '_ {
>         self.data.chunks(size_of::<u128>()).map(move |chunk| {
> ```

first chunk is part domain sep part seed ?
___
*[`tfhe-csprng/src/seeders/mod.rs` line 73 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMJ0B2CL2Fw3jYRIHO:-ORMJ0B2CL2Fw3jYRIHP:b-s9wb9h) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L73)):*
> ```XML
> pub enum SeedKind {
>     /// Legacy kind of seed
>     Legacy(Seed),
> ```

wondering about this, the Xof stuff is for the nist doc, the "old" seed could still be used though ?

Not sure what the name should be though, could ask the guys who worked on the doc
___
*[`tfhe-csprng/src/generators/implem/aesni/block_cipher.rs` line 58 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMZEdR7qSxkv-tOHLg:-ORMZEdR7qSxkv-tOHLh:b-n1zt6n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aesni/block_cipher.rs#L58)):*
> ```XML
> 
>     fn generate_next(&mut self, index: AesIndex) -> [u8; BYTES_PER_AES_CALL] {
>         unsafe { transmute(aes_encrypt_one(u128_to_si128(index.0), &self.round_keys)) }
> ```

something feels off with &mut self not being modified here, we are going to output the same values for each call ? generate_next seems to indicate something whose state evolves
___
*[`tfhe-csprng/src/generators/implem/aesni/block_cipher.rs` line 70 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMOedACmjxLhBnkgtn:-ORMOedACmjxLhBnkgto:b-q3e9h7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aesni/block_cipher.rs#L70)):*
> ```XML
> }
> 
> #[allow(clippy::too_many_arguments)]
> ```

too many arguments ?
___
*[`tfhe-csprng/src/generators/implem/aarch64/block_cipher.rs` line 60 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQxI27VRSkuXoymnV:-ORMQxI27VRSkuXoymnW:bt57zxo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aarch64/block_cipher.rs#L60)):*
> ```XML
> 
>     fn generate_next(&mut self, index: AesIndex) -> [u8; BYTES_PER_AES_CALL] {
>         unsafe { encrypt(index.0, &self.round_keys) }.to_ne_bytes()
> ```

I have a weird feeling here, we call encrypt on a mutable self, but in fact self is not modified so it means we would always generate strictly the same bytes ?

IceTDrinker (commented):
forgot to add, we likely want to export the XofSeed in TFHE-rs, but I'm guessing it will be done in the PR introducing the KeySet ?

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-:-ORMZyAEBr0Qh88rJ-aO:bcbimfw)* status: all files reviewed, 10 unresolved discussions (waiting on @tmontaigu)

tmontaigu (commented):
yes, but i would make sense to already expose it

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-:-ORVXkj_C1rxTz_Y0REk:b-9raue1)* status: all files reviewed, 10 unresolved discussions (waiting on @IceTDrinker)

___
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 239 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQTnY7Cvjuv-usNCS:-ORVXWG1AnLr3tTwsltH:b106f5t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/aes_ctr/mod.rs#L239)):*

endianness wise is that ok ? I don't seem to see it in the XOF object definition in the NIST doc

also asking because I have seen some ne_bytes elsewhere in the code base and it's unclear to me if that would work properly on a big endian target (though as we know, those are very niche)
</blockquote></details>

Endianness has to be checked with nist autor, especially since the seed in an array of bytes, and we use it as iter of u128 values
___
*[`tfhe-csprng/src/generators/implem/aarch64/block_cipher.rs` line 60 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQxI27VRSkuXoymnV:-ORVXGe46xEReHuGM8KF:b-5jhp1l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aarch64/block_cipher.rs#L60)):*

I have a weird feeling here, we call encrypt on a mutable self, but in fact self is not modified so it means we would always generate strictly the same bytes ?
</blockquote></details>

no, we encrypt an index with the aes key stored, so as long as the index changes, the encryption changes
___
*[`tfhe-csprng/src/generators/implem/aesni/block_cipher.rs` line 58 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMZEdR7qSxkv-tOHLg:-ORVWca9C8V1UTnZJO7J:bqjsa4g) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aesni/block_cipher.rs#L58)):*

something feels off with &mut self not being modified here, we are going to output the same values for each call ? generate_next seems to indicate something whose state evolves
</blockquote></details>

That's because I copied the generate_batch signature which also has `&mut self`, although none of the impl seems to use that, so we could change both
___
*[`tfhe-csprng/src/seeders/mod.rs` line 25 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMP5LhAhEiqaFlbQIW:-ORVWZXW-gfzDfqChQwl:b2nign6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L25)):*

just to make sure (better than being sorry) the 8 bytes for domain sep was confirmed ?
</blockquote></details>

yes
___
*[`tfhe-csprng/src/seeders/mod.rs` line 31 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYN0a9GdOPkLTapta:-ORVWJWSEWz5nAHwVprw:b4of486) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L31)):*

layout ok with respect to description in the doc ?

doc seems to say seed || str and here I kind of read it as str || seed, but I'm guessing it's a matter of endianness
</blockquote></details>

The doc of `data` says str||seed and thats what NIST says, the doc of `XofSeed` could be swapped to also match the order
___
*[`tfhe-csprng/src/seeders/mod.rs` line 63 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYkeT2I_1K2aNAlVu:-ORVW1ay8IcX17Gkb6zD:b2nhswy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L63)):*

first chunk is part domain sep part seed ?
</blockquote></details>

Yes
___
*[`tfhe-csprng/src/seeders/mod.rs` line 73 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMJ0B2CL2Fw3jYRIHO:-ORVW32d4z7fLVQMd3P3:ba35tyu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L73)):*

wondering about this, the Xof stuff is for the nist doc, the "old" seed could still be used though ?

Not sure what the name should be though, could ask the guys who worked on the doc
</blockquote></details>

The 'old' seed is whats going to be used 99% of the time actually maybe a name like `Default`, `Normal` or`Vanilla` would be better

tmontaigu:
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 239 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQTnY7Cvjuv-usNCS:-ORWNpzJ0rGZK3Q-Or1G:b-hb98tz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/aes_ctr/mod.rs#L239)):*

Endianness has to be checked with nist autor, especially since the seed in an array of bytes, and we use it as iter of u128 values
</blockquote></details>

NIST says that convertion from/to bytes for integers are little-endian

We have that:
  * AES takes 128bits and output 128bits, the output of AES does not depend on the endian of the executing machine.
  * So that should mean we can use to_ne_bytes when converting the output u128 to [16; u8]
  * So if we give an array of 16 bytes (128bits) to 2 different impl, one than runs on little endian, the other that runs on big endian, they both have to load the bytes using from_ne_bytes, since the order of the input byte array is already fixed

So, here I believe  we should use from_ne_bytes to create `c`

Also, this means our current implementation of AES-CTR is not endian-agnostic, as the input u128 is converted to bytes using to_ne_bytes,
however, for example, 1u128 does not have the same byte order between little and big endian, meaning the output is not going to be the same.
The AesIndex passed to the BlockCipher should go through a round of `to_le()` if we want the CTR to be endian-agnostic

IceTDrinker (commented):
Comments here and there, for endianness I stopped asking everywhere and tests will be the deciders

> Reviewed 9 of 14 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-:-ORp9T_-C4BY0dtr0MF2:b2miblz)* status: all files reviewed, 8 unresolved discussions (waiting on @IceTDrinker, @soonum, and @tmontaigu)

___
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 239 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMQTnY7Cvjuv-usNCS:-ORpCJQP6d9-cSGSkvSN:bal5hxk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/aes_ctr/mod.rs#L239)):*

NIST says that convertion from/to bytes for integers are little-endian

We have that:
  * AES takes 128bits and output 128bits, the output of AES does not depend on the endian of the executing machine.
  * So that should mean we can use to_ne_bytes when converting the output u128 to [16; u8]
  * So if we give an array of 16 bytes (128bits) to 2 different impl, one than runs on little endian, the other that runs on big endian, they both have to load the bytes using from_ne_bytes, since the order of the input byte array is already fixed

So, here I believe  we should use from_ne_bytes to create `c`

Also, this means our current implementation of AES-CTR is not endian-agnostic, as the input u128 is converted to bytes using to_ne_bytes,
however, for example, 1u128 does not have the same byte order between little and big endian, meaning the output is not going to be the same.
The AesIndex passed to the BlockCipher should go through a round of `to_le()` if we want the CTR to be endian-agnostic
</blockquote></details>

Thanks, the big endian tests will be our verifiers in that case !
___
*[`tfhe-csprng/src/generators/implem/aesni/block_cipher.rs` line 58 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMZEdR7qSxkv-tOHLg:-ORp9ztyAzGhTm6uUOx9:b-z6g7vv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/generators/implem/aesni/block_cipher.rs#L58)):*

That's because I copied the generate_batch signature which also has `&mut self`, although none of the impl seems to use that, so we could change both
</blockquote></details>

I'm guessing it was a future proofing measure for the trait, not critical
___
*[`tfhe-csprng/src/seeders/mod.rs` line 31 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYN0a9GdOPkLTapta:-ORpAfzq9TElj6Ip7G39:bwpv5md) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L31)):*

The doc of `data` says str||seed and thats what NIST says, the doc of `XofSeed` could be swapped to also match the order
</blockquote></details>

yes let's properly map to NIST everywhere possible

for the endianness questions I'll stop asking and we'll rely on the rust cross thing with a big endian target
___
*[`tfhe-csprng/src/seeders/mod.rs` line 73 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMJ0B2CL2Fw3jYRIHO:-ORpB4GiCQtlBCxZqCkI:b-rs2nnd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L73)):*

The 'old' seed is whats going to be used 99% of the time actually maybe a name like `Default`, `Normal` or`Vanilla` would be better
</blockquote></details>

Maybe just Ctr ? and a docstring saying it uses the seed to initialize an AES block cipher in Counter mode without additional stuff ?
___
*[`Makefile` line 177 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORpBeijAf4D2tanRTHn:-ORpBeijAf4D2tanRTHo:b-7m2jqm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f26f879a63b7ca7ae037ef3cd2d6ac38cd7469b3/Makefile#L177)):*
> ```Makefile
> 
> .PHONY: install_cargo_cross # Install custom tfhe-rs lints
> install_cargo_cross:
> ```

safer to install stable toolchan and specify the toolchain as well ?
___
*[`tfhe-csprng/src/generators/aes_ctr/block_cipher.rs` line 18 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORpC1tM-IwIJqZ5FtXc:-ORpC1tM-IwIJqZ5FtXd:b-lf5lmg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f26f879a63b7ca7ae037ef3cd2d6ac38cd7469b3/tfhe-csprng/src/generators/aes_ctr/block_cipher.rs#L18)):*
> ```XML
>     fn new(key: AesKey) -> Self;
>     /// Generates the batch corresponding to the given index.
>     fn generate_batch(&mut self, data: [u128; AES_CALLS_PER_BATCH]) -> [u8; BYTES_PER_BATCH];
> ```

Have a doubt on the effect of that for the stack, we used to have stack issues as said elsewhere, I did not realize when you first asked me
___
*[`tfhe-csprng/src/generators/implem/aesni/block_cipher.rs` line 35 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORpAFwe5CfBqUqG4QA1:-ORpAFwe5CfBqUqG4QA2:bddbfx2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f26f879a63b7ca7ae037ef3cd2d6ac38cd7469b3/tfhe-csprng/src/generators/implem/aesni/block_cipher.rs#L35)):*
> ```XML
>     }
> 
>     fn generate_batch(&mut self, data: [u128; AES_CALLS_PER_BATCH]) -> [u8; BYTES_PER_BATCH] {
> ```

directly on stack might cause issues ? we have had stack overflows in the past, try running the M1 tests with the label

tmontaigu:
*[`tfhe-csprng/src/seeders/mod.rs` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMOyHe4Y3uiTcTrOp1:-ORpMDq94hRacZoqjRxM:b-u64yrd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L12)):*

think there is an eprint ?
</blockquote></details>

I added it

tmontaigu:
*[`tfhe-csprng/src/seeders/mod.rs` line 31 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMYN0a9GdOPkLTapta:-ORW_qEw4K071OVELPvd:bgw3w2c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L31)):*

The doc of `data` says str||seed and thats what NIST says, the doc of `XofSeed` could be swapped to also match the order
</blockquote></details>

(note that the eprint is not up to date on this)

tmontaigu:
*[`tfhe-csprng/src/seeders/mod.rs` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2381#-ORMOyHe4Y3uiTcTrOp1:-ORpjAUQEjH3z9VIoibn:b-oxbw61) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6f63f0ef7dc24f13e39226d2818cf73807661b1/tfhe-csprng/src/seeders/mod.rs#L12)):*

oh my god, this confuses me every single time, btw why do you prefer not having it inline ?
</blockquote></details>

Personnal preference, It keeps the line short and readable

PR #2394, opened 2025-06-02
chore(ci): lock zizmor version to avoid breaking ci pipelines

soonum:
Newer version of Zizmor can trigger errors due to new findings in workflows. To avoid breaking any ongoing pull-request, due to this unhandled update, zizmor version is locked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ install_typos_checker: install_rs_build_toolchain
 .PHONY: install_zizmor # Install zizmor workflow security checker
 install_zizmor: install_rs_build_toolchain
 	@zizmor --version > /dev/null 2>&1 || \
-	cargo $(CARGO_RS_BUILD_TOOLCHAIN) install zizmor || \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -170,7 +170,7 @@ install_typos_checker: install_rs_build_toolchain
 .PHONY: install_zizmor # Install zizmor workflow security checker
 install_zizmor: install_rs_build_toolchain
 	@zizmor --version > /dev/null 2>&1 || \
-	cargo $(CARGO_RS_BUILD_TOOLCHAIN) install zizmor || \
+	cargo $(CARGO_RS_BUILD_TOOLCHAIN) install zizmor --version ~1.9 || \
```
IceTDrinker, line 173: we need a string ? "~1.9" ?

IceTDrinker: or is this ok ?

IceTDrinker: ok workflow ran, all good

PR #2397, opened 2025-06-02
refactor(shortint): improve handling of empty compressed ct list

nsarlin-zama:
Improve handling of empty CompressedCiphertextList, following https://github.com/zama-ai/tfhe-rs/pull/2193. 
The metadata are now stored in an optional field that is not filled if the list is empty. The same pattern has been used for the Cuda and NoiseSquashed compressed lists.

For the NoiseSquashed list I did not add the backward data, hoping that this will be merged before the next release


IceTDrinker (commented):
Remarks, I don't know if everything GPU related is correct, but this looks generally Ok, apart from the count thing which looks a bit weird in hindsight

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-:-ORp4nOhBoRurxu7Bea2:b-hbye7c)* status: all files reviewed, 4 unresolved discussions (waiting on @agnesLeroy, @mayeul-zama, and @tmontaigu)

___
*[`tfhe/src/integer/gpu/list_compression/server_keys.rs` line 41 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-ORp3phI9KrO4lBmhDRM:-ORp3phI9KrO4lBmhDRN:bhkdh4g) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43564507eb0e074fc2b8dac61c4e3fe1e4d72c5d/tfhe/src/integer/gpu/list_compression/server_keys.rs#L41)):*
> ```XML
> }
> 
> #[derive(Clone)]
> ```

derive Copy even ?
___
*[`tfhe/src/shortint/ciphertext/compressed_ciphertext_list.rs` line 34 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-ORp4HyU8qENUUTAiUtO:-ORp4HyU8qENUUTAiUtP:ba6xqlz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43564507eb0e074fc2b8dac61c4e3fe1e4d72c5d/tfhe/src/shortint/ciphertext/compressed_ciphertext_list.rs#L34)):*
> ```XML
>         Vec<CompressedModulusSwitchedGlweCiphertext<u64>>,
>     pub(crate) meta: Option<CompressedCiphertextListMeta>,
>     pub(crate) count: CiphertextCount,
> ```

wonder if count should be a function instead to avoid desyncs with the actual list ? (all comments here apply to the noise squashing stuff as well)
___
*[`tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs` line 384 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-ORp3ZXQ-EeHugYhLB5a:-ORp3ZXR3L5H3ubb1Xqd:b-rqed72) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43564507eb0e074fc2b8dac61c4e3fe1e4d72c5d/tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs#L384)):*
> ```XML
>             .sum();
> 
>         let meta = self.packed_list.meta.as_ref().and_then(|cpu_meta| {
> ```

might need a check that initial_len == 0 if meta is None perhaps ?

Might need to be added in conformance ?

Can be done as a second step as this already solves some issues
___
*[`tfhe/src/shortint/backward_compatibility/ciphertext/mod.rs` line 293 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-ORp4AU-4ZHcL0NdfhtA:-ORp4AU-4ZHcL0NdfhtB:b-n218gc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43564507eb0e074fc2b8dac61c4e3fe1e4d72c5d/tfhe/src/shortint/backward_compatibility/ciphertext/mod.rs#L293)):*
> ```XML
> 
> #[derive(VersionsDispatch)]
> #[allow(dead_code)]
> ```

surprised by the dead code here ?

tfhe/src/integer/gpu/list_compression/server_keys.rs
```diff
@@ -207,16 +220,20 @@ impl CudaCompressionKey {
...
             ciphertext_modulus,
-            bodies_count: num_lwes,
             storage_log_modulus: self.storage_log_modulus,
             lwe_per_glwe: LweCiphertextCount(compressed_polynomial_size.0),
+        });
+
+        CudaPackedGlweCiphertextList {
+            data: packed_glwe_list,
+            meta,
+            bodies_count: num_lwes,
```
agnesLeroy: Maybe the bodies_count could be in the meta as well?

tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs
```diff
@@ -368,17 +373,34 @@ impl CompressedCiphertextList {
...
-                glwe_dimension: first_ct.glwe_dimension(),
-                polynomial_size: first_ct.polynomial_size(),
-                message_modulus,
-                carry_modulus,
-                ciphertext_modulus: self.packed_list.ciphertext_modulus,
-                bodies_count: self.packed_list.count.0,
-                storage_log_modulus,
-                lwe_per_glwe,
+                meta,
+                bodies_count: self.packed_list.len(),
```
agnesLeroy: Is self.packed_list.len()` equivalent to `self.packed_list.count.0`?

agnesLeroy (commented):
Thank you @nsarlin-zama! The GPU part looks good to me, I only have two small questions.

tfhe/src/integer/gpu/list_compression/server_keys.rs
nsarlin-zama: Initially I only put in meta the things that are undefined if the list is empty, and this one can just be 0 for the empty list. But I can move it to meta if you think it makes more sense like this! Maybe with a `len` method that returns this or 0 if meta is None

tfhe/src/integer/gpu/ciphertext/compressed_ciphertext_list.rs
nsarlin-zama: Yes this is equivalent. For the cpu I removed the `count` attribute, it is recomputed in the `len` method as the sum of each individual glwe `bodies_count`

tfhe/src/integer/gpu/list_compression/server_keys.rs
agnesLeroy: It would look better if this was in the meta as well I think, I don't know if we want a len() method as this is the total number of LWEs in the list, I think keeping it like this would be enough.

nsarlin-zama: done

IceTDrinker (approved):
Looks good to me!

> Reviewed 1 of 16 files at r2, 2 of 5 files at r3, 17 of 18 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-:-OSwnP_M5ofo3RVL33WG:b9j9n3q)* status: all files reviewed, 2 unresolved discussions (waiting on @agnesLeroy and @mayeul-zama)

___
*[`tfhe/src/integer/gpu/list_compression/server_keys.rs` line 41 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2397#-ORp3phI9KrO4lBmhDRM:-OSwlZ064SNG2Gs4Q0o0:b-g2ti08) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43564507eb0e074fc2b8dac61c4e3fe1e4d72c5d/tfhe/src/integer/gpu/list_compression/server_keys.rs#L41)):*

done
On a side note, the is a manual Clone impl for `CudaPackedGlweCiphertextList`, which I don't think is necessary. Maybe I should replace it with derive(Clone) ?
</blockquote></details>

unsure for the gpu what's preferred

IceTDrinker:
@nsarlin-zama go for merge

PR #2398, opened 2025-06-03
feat: add upgrade key chain

tmontaigu:
This adds an UpgradeKeyChain struct
that can be used to easily upgrade parameters of ciphertexts if some some upgrade keys are provided

### Check-list:


nsarlin-zama (commented):
A few comments, the general algorithm, data types and test look good!

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-:-OS-VIZ7E0-CXD8r5q-p:b-34zjgt)* status: all files reviewed, 14 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-VxLb7f387Si3Jm4x:-OS-VxLb7f387Si3Jm4y:b-voomf0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L12)):*
> ```XML
> 
> /// Decompression key used to upgrade to some parameters while decompressing
> pub struct DecompressionUpgradeKey {
> ```

I am wondering if there is value to have the decompression part inside the keychain mechanism:
- Decompression is only possible as the first step in the path, so it can be done before entering the graph?
- It adds a bit of complexity for us and the user (there is a dedicated method for compressed inputs)

Maybe it would be simpler if the keychain only handles decompressed ciphertexts and it is up to the user to decompress it before?

Or maybe there is some plans to have compression/decompression happening in the middle of an upgrade path (I remember reading something like this for inter device upgrades?)
___
*[`tfhe/src/high_level_api/upgrade.rs` line 14 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-2mWl1m8V17xY_chJ:-OS-2mWl1m8V17xY_chK:bkarae0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L14)):*
> ```XML
> pub struct DecompressionUpgradeKey {
>     inner: DecompressionKey,
>     tag_in: Tag,
> ```

The Tag documentation specifies that the Tag are not interpreted by TFHE-rs, I think that this should be updated to explain that tags are used by the keychain
___
*[`tfhe/src/high_level_api/upgrade.rs` line 34 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-JYhs5_k5IYLmPhHR:-OS-JYhs5_k5IYLmPhHS:bmddy5c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L34)):*
> ```XML
>         let glwe_secret_key = match &cks_out.key.key.key.atomic_pattern {
>             AtomicPatternClientKey::Standard(ap) => &ap.glwe_secret_key,
>             AtomicPatternClientKey::KeySwitch32(ap) => &ap.glwe_secret_key,
> ```

Compression is currently not supported by the KS32 AP
___
*[`tfhe/src/high_level_api/upgrade.rs` line 56 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-HSnY205VKy2HVeNR:-OS-HSnY205VKy2HVeNS:b-fp4fgl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L56)):*
> ```XML
>             blind_rotate_key,
>             lwe_per_glwe: params.lwe_per_glwe,
>         };
> ```

Maybe this could be an API inside the shortint layer ? For example ClientKey::new_decompression_key, which could also be called by `ClientKey::new_compression_decompression_key` to avoid code duplication



_Code quote:_
```XML
        let glwe_secret_key = match &cks_out.key.key.key.atomic_pattern {
            AtomicPatternClientKey::Standard(ap) => &ap.glwe_secret_key,
            AtomicPatternClientKey::KeySwitch32(ap) => &ap.glwe_secret_key,
        };

        let blind_rotate_key =
            ShortintEngine::with_thread_local_mut(|engine| ShortintBootstrappingKey::Classic {
                bsk: engine.new_classic_bootstrapping_key(
                    &private_compression_key
                        .key
                        .post_packing_ks_key
                        .as_lwe_secret_key(),
                    glwe_secret_key,
                    cks_out.key.key.key.parameters().glwe_noise_distribution(),
                    params.br_base_log,
                    params.br_level,
                    cks_out.key.key.key.parameters().ciphertext_modulus(),
                ),
                modulus_switch_noise_reduction_key: None,
            });

        let glwe_decompression_key = crate::shortint::list_compression::DecompressionKey {
            blind_rotate_key,
            lwe_per_glwe: params.lwe_per_glwe,
        };
```
___
*[`tfhe/src/high_level_api/upgrade.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-1Il-2wZRMx11Ma7C:-OS-1Il00p75u3TDZ6tw:b6uexbx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L78)):*
> ```XML
> 
> #[derive(Debug, PartialEq, Eq, Clone, Copy)]
> pub enum CipherKind {
> ```

CiphertextKind ? For me a Cipher is an algorithm
___
*[`tfhe/src/high_level_api/upgrade.rs` line 85 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-2qLbAup89svAsxza:-OS-2qLbAup89svAsxzb:bwqlz9i) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L85)):*
> ```XML
> /// Describes some parameters
> #[derive(Debug, PartialEq, Eq)]
> pub struct ParamDescriptor {
> ```

Since there is also the Tag inside, maybe it is more a "KeyDescriptor" or "KeyId"? Because you can have different tags for the same parameters. 
This already supports upgrades between different keys with the same parameters, and I feel like there could be real life use cases for this too.
___
*[`tfhe/src/high_level_api/upgrade.rs` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-3Pev0rFsbaLyZftH:-OS-3Pev0rFsbaLyZftI:b-ty6moh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L152)):*
> ```XML
>     }
> 
>     fn add_path(&mut self, node_in: NodeId, node_out: NodeId, key_index: usize) {
> ```

this is more `add_vertex` ?
___
*[`tfhe/src/high_level_api/upgrade.rs` line 294 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-8iWaAEQ_Gk4vV6bM:-OS-8iWaAEQ_Gk4vV6bN:b-991c2t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L294)):*
> ```XML
> ///
> /// Then upgrade keys that allow to go from one parameter set to another should
> /// be added with [Self::add_upgrade_key]
> ```

There could be an example here ?
___
*[`tfhe/src/high_level_api/upgrade.rs` line 295 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-XLtT1zmSPQiQRDwF:-OS-XLtT1zmSPQiQRDwG:b-k271ro) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L295)):*
> ```XML
> /// Then upgrade keys that allow to go from one parameter set to another should
> /// be added with [Self::add_upgrade_key]
> pub struct UpgradeKeyChain {
> ```

Won't this need serialization too? Or maybe this can be done in a subsequent PR
___
*[`tfhe/src/high_level_api/upgrade.rs` line 397 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-8OHz0VixWz0N8iLt:-OS-8OHz0VixWz0N8iLu:b2ngd26) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L397)):*
> ```XML
>     /// Upgrades the input ciphertext to the compute params of the selected tag and device
>     ///
>     /// Returns an error if no upgrade path could be found
> ```

Doc comment could also explain that this returns the intermediate ciphertexts
___
*[`tfhe/src/high_level_api/upgrade.rs` line 488 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-8FEl1DRW7twjcAql:-OS-8FEl1DRW7twjcAqm:b-uj15vw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L488)):*
> ```XML
>     //
>     // NOTE: only keyswitch are allowed in the upgrade path
>     fn follow_upgrade_path<T>(&self, ct: &T, upgrade_path: &[usize]) -> (Vec<T>, T)
> ```

maybe more `apply_upgrade_path` ?
___
*[`tfhe/src/high_level_api/upgrade.rs` line 527 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-E7v464iFxoR3cLo8:-OS-E7v464iFxoR3cLo9:bon0fga) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L527)):*
> ```XML
>     use crate::upgrade::{DecompressionUpgradeKey, UpgradeKeyChain};
>     use crate::*;
> ```

You could also have a test for the graph part here, since it does not need real keys it can be easier/faster to test the path finding in various configurations this way
___
*[`tfhe/src/high_level_api/upgrade.rs` line 564 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-JrnZBx0wCPuLjyhz:-OS-JrnZBx0wCPuLjyi-:bnzxhh3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L564)):*
> ```XML
>         }
> 
>         // We add an upgrade path, we add an upgrade path to form a chain
> ```

sentence is duplicated
___
*[`tfhe/src/high_level_api/upgrade.rs` line 602 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-T3ZqBdS2P3hi8qIN:-OS-T3ZqBdS2P3hi8qIO:undefined) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L602)):*
> ```XML
>         // we find a way to upgrade to the latest one
>         //
>         // This also tests that upgrading to from latest to latest is ok
> ```





_Suggestion:_
```XML
 from latest to latest is ok
```

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398)* status: all files reviewed, 14 unresolved discussions (waiting on @nsarlin-zama)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-VxLb7f387Si3Jm4x:-OSOmVvMCA5q38q3ARR0:b-10ex2x) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L12)):*

I am wondering if there is value to have the decompression part inside the keychain mechanism:
- Decompression is only possible as the first step in the path, so it can be done before entering the graph?
- It adds a bit of complexity for us and the user (there is a dedicated method for compressed inputs)

Maybe it would be simpler if the keychain only handles decompressed ciphertexts and it is up to the user to decompress it before?

Or maybe there is some plans to have compression/decompression happening in the middle of an upgrade path (I remember reading something like this for inter device upgrades?)
</blockquote></details>

The main intended usage for BC is to use decompression as the keyswitch mechanism

I believe all their inputs to be switched are compressed, and ideally there is a decompression key that exists and allows to decompress to latest compute params directly

But as I don't know if it will always be the case, I made the graph support decompressing to some intermediate compute params and then try to find some keyswiches
___
*[`tfhe/src/high_level_api/upgrade.rs` line 56 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-HSnY205VKy2HVeNR:-OSOlLfu9wy3iebUhFhk:b-cak6bk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L56)):*

Maybe this could be an API inside the shortint layer ? For example ClientKey::new_decompression_key, which could also be called by `ClientKey::new_compression_decompression_key` to avoid code duplication
</blockquote></details>

I had it in the ClientKey but wasn't sure it was best to have it there
___
*[`tfhe/src/high_level_api/upgrade.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-1Il-2wZRMx11Ma7C:-OSOlRuj-q6_My2vlKVn:bb77u1k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L78)):*

CiphertextKind ? For me a Cipher is an algorithm
</blockquote></details>

Yeah
___
*[`tfhe/src/high_level_api/upgrade.rs` line 295 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-XLtT1zmSPQiQRDwF:-OSOlXwj02EjWgi58dny:bvmu102) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L295)):*

Won't this need serialization too? Or maybe this can be done in a subsequent PR
</blockquote></details>

This probably will at some point, but I think it's for later

And the serialization should maybe only serialize info that can be used to re-construct the graph so that the graph struct is not serialized and we have more freedom to change we represent the graph
___
*[`tfhe/src/high_level_api/upgrade.rs` line 397 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-8OHz0VixWz0N8iLt:-OSOm2_07qaHCKLaukHx:b-ln5gyo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L397)):*

Doc comment could also explain that this returns the intermediate ciphertexts
</blockquote></details>

Yes, I need to ask for confirmation that returning intermediate upg5rades is needed

IceTDrinker (commented):
Reviewed until upgrade_from_compressed, looks generally good maybe some details here and there to confirm, will finish after my sync

> Reviewed 2 of 5 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-:-OSsaxyJD906mIh2ftL6:bkusyj3)* status: 5 of 6 files reviewed, 19 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 397 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OS-8OHz0VixWz0N8iLt:-OSsaOQ81u56Cqp0WUe7:b53mbka) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7d52555177aa6ecf74aa843df411086aebef8032/tfhe/src/high_level_api/upgrade.rs#L397)):*

Yes, I need to ask for confirmation that returning intermediate upg5rades is needed
</blockquote></details>

probably not required contrary to what was first believed
___
*[`tfhe/src/shortint/list_compression/server_keys.rs` line 120 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSs_vPqBkTbkBoaRYjU:-OSs_vPr0QLmWPqWB4LB:bmz4uyf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/shortint/list_compression/server_keys.rs#L120)):*
> ```XML
>         };
> 
>         assert!(
> ```

assert looks to be in a weird place if we abort, could be done before the key is actually generated
___
*[`tfhe/src/high_level_api/upgrade.rs` line 132 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsX0zLCI9HN6P2uvHx:-OSsX0zLCI9HN6P2uvHy:bp8xath) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L132)):*
> ```XML
>         self.adjacency[node_in.0].push(Vertex {
>             out_index: node_out,
>             data: key_index,
> ```

just wondering why the Vertex needs a key_index ? Given a Node contains a KeyDescriptor ?

just worried about potentially graph consistency
___
*[`tfhe/src/high_level_api/upgrade.rs` line 136 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsYn2Y3dU1qZe5BVIO:-OSsYn2Y3dU1qZe5BVIP:b-dvesqd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L136)):*
> ```XML
>     }
> 
>     fn find_upgrade_path(&self, source: NodeId, destination: NodeId) -> Option<Vec<usize>> {
> ```

looks like a Depth first search, correct ?

Feels like it means we may not get the shortest path ?

Maybe it does not matter for our upgrade strategies, but I have a doubt if we later have the intermediate compute keyswitches
___
*[`tfhe/src/high_level_api/upgrade.rs` line 391 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSs_3yf6kj1fpQY0t86:-OSs_3yf6kj1fpQY0t87:bbpaext) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L391)):*
> ```XML
>         if sks.key.compression_key.is_some() {
>             let node = Node::new(sks.tag(), CiphertextKind::Compressed, Device::CudaGpu);
>             let _compressed_node_id = self.graph.get_or_insert_node(node);
> ```

no updates on the graph and upgrade keys, does it mean the graph does nothing for GPU whend a decompression key is available ?
___
*[`tfhe/src/high_level_api/upgrade.rs` line 423 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsaEGP4KwU47soiFEt:-OSsaEGP4KwU47soiFEu:blz79ys) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L423)):*
> ```XML
>             ))
>             .ok_or_else(|| {
>                 crate::error!("The input of this key does not match anything in the upgrade graph")
> ```

the output ?
___
*[`tfhe/src/high_level_api/upgrade.rs` line 443 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsafdw49ilO7x_44Vc:-OSsafdw49ilO7x_44Vd:b-oiy7di) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L443)):*
> ```XML
>     ) -> crate::Result<(Vec<T>, T)>
>     where
>         T: Tagged + Clone,
> ```

this is not restricted to FheUint, as CompressedCiphertextList also implements tagged, this could be a source of mistake from users I would think, or perhaps the error would catch it, but it feels a little brittle

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398)* status: 5 of 6 files reviewed, 21 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 559 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OStWRi-2cft5rvc_kGC:-OStWRi-2cft5rvc_kGD:bma3skw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L559)):*
> ```XML
>     #[cfg(feature = "gpu")]
>     use crate::shortint::parameters::current_params::V1_2_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
>     use crate::shortint::parameters::{
> ```

consider using test_params perhaps ? I see comp without a suffix which looks a bit odd
___
*[`tfhe/src/high_level_api/upgrade.rs` line 845 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OStY77G8y9GzKeDEAzb:-OStY77G8y9GzKeDEAzc:b-a009o9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L845)):*
> ```XML
>             // Create Compressed ServerKey that is special for GPU/CPU compression inter ops
>             let gpu_compression_params =
>                 V1_1_COMP_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
> ```

V1_1 ? I think you import V1_2 earlier in the file and we will soon be in V1_3

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398)* status: 5 of 6 files reviewed, 13 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 132 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsX0zLCI9HN6P2uvHx:-OT28f4o2Pr4UgmZpeDU:b32wtd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L132)):*

Its not the same thing, the KeyDescriptor decribe a key on a ServerKey i.e compute key or compession key

here the key index is used to indentify the upgrade key to use, so that when the upgrader applies to path, it does not have to search which upgrade key it has to use to follow the vertex
</blockquote></details>

ok
___
*[`tfhe/src/high_level_api/upgrade.rs` line 136 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsYn2Y3dU1qZe5BVIO:-OT28grr85DJURvBUsbh:bpur3gn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L136)):*

Yes this is a graph traversal algorithm, not a shortest path finding

I think that for now this is sufficient, and changing the path finding algorithm can be done later when needed
</blockquote></details>

if we end up using a keyswitch path instead of a direct decompression it would be a bit silly, we should look at updating this quickly next quarter, if you can open an issue about it

Dijkstra Algorithm I guess is what we would want since our graph should not be cyclic
___
*[`tfhe/src/high_level_api/upgrade.rs` line 391 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSs_3yf6kj1fpQY0t86:-OT29--REuHgzWVhgNHv:b-kvtx0e) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L391)):*

The problem here is that the decompression key is a GPU decompression key (the data is on GPU, and the computation has to be done on GPU), while here we need it to be on CPU, so I would need a convertion from GPU key to CPU key
</blockquote></details>

hmm, don't we expect to do decompression on GPU at some point ?

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398)* status: 5 of 6 files reviewed, 12 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 136 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsYn2Y3dU1qZe5BVIO:-OT6CJwQAxZFR2MhhVcg:b-fbw3bb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L136)):*

If there is direct path is will always be taken, this thing is already tested, so i don't think its that necessary to change it quickly
</blockquote></details>

Ok, not sure what the algorithm is equivalent to, the equivalent algorithm will give the properties

tmontaigu:
*[`tfhe/src/high_level_api/upgrade.rs` line 391 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSs_3yf6kj1fpQY0t86:-OT6C5IEA9lMqtPhNnKn:b-3hodcy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L391)):*

hmm, don't we expect to do decompression on GPU at some point ?
</blockquote></details>

What do you mean by decompression on GPU ? doing the actual decompression on GPU or decompressing on whatever device to params that are meant for GPU ?

Here the key is not added because it is a key that is located on GPU, thus meant to be used on GPU whereas for now everything is made for CPU

GPU support should be done later because there are 2 choices,:
* The upgrader either have all its upgrade keys on one device
* Or it can have keys on multiple different devices

IceTDrinker (commented):
generally looks ok, main thing would be to use TEST PARAMS aliases to avoid having to update these tests

> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-:-OT6atZP5Gg_XJCMHm4P:bwq4xs)* status: all files reviewed, 12 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398)* status: all files reviewed, 12 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @tmontaigu)

___
*[`tfhe/src/high_level_api/upgrade.rs` line 443 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-OSsafdw49ilO7x_44Vc:-OTmMaFP-ZM2goMTOHta:bkk7y9c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c5a5068bc55c33537589b924339acfb09c4d51f5/tfhe/src/high_level_api/upgrade.rs#L443)):*

Actually the second bound  `KeySwitchingKey: FheKeyswitch<T>,`  restricts to FheUint/Int/Bool as the keyswitch trait is only implemented for these

We could add a marker trait that is only implemented for FheBool/Int/Uint if you think its better
Another solution could be to have one upgrade method, by using traits a bit like this

```
pub trait ParamRotation<Output> {
    type ExtraParams;
    
    fn upgrade(&self, params: &Self::ExtraParams, upgrader: &Upgrader) -> Output;
}

struct Upgrader;

impl Upgrader {
    
    pub fn upgrade<In, Out>(&self, value: &In, params: &In::ExtraParams) -> Out
    where In: ParamRotation<Out> {
        value.upgrade(params, self)
    }
}
```
</blockquote></details>

if there is no confusion possible, it's fine by me, did not realize the second bound would be enough, do we expect FheKeyswitch to be implemented for other types ?

nsarlin-zama (approved):
:lgtm:, thanks!

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2398#-:-OU-hmis2rTxL2-wsYpn:bpqxd5f)* status: 3 of 7 files reviewed, 6 unresolved discussions (waiting on @IceTDrinker, @mayeul-zama, and @tmontaigu)

PR #2399, opened 2025-06-03
fix(gpu): fix race condition on expand when on multi-gpu

pdroalves:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1024 and https://github.com/zama-ai/tfhe-rs-internal/issues/1006

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
@@ -112,15 +112,15 @@ template <typename Torus> struct zk_expand_mem {
 
     // Hint for future readers: if message_modulus == 4 then
     // packed_messages_per_lwe becomes 2
-    auto packed_messages_per_lwe = log2_int(params.message_modulus);
+    auto num_packed_msgs = log2_int(params.message_modulus);
 
     // Adjust indexes to permute the output and access the correct LUT
     auto h_indexes_in = static_cast<Torus *>(
-        malloc(packed_messages_per_lwe * num_lwes * sizeof(Torus)));
+        malloc(num_packed_msgs * num_lwes * sizeof(Torus)));
     auto h_indexes_out = static_cast<Torus *>(
-        malloc(packed_messages_per_lwe * num_lwes * sizeof(Torus)));
+        malloc(num_packed_msgs * num_lwes * sizeof(Torus)));
     auto h_lut_indexes = static_cast<Torus *>(
-        malloc(packed_messages_per_lwe * num_lwes * sizeof(Torus)));
+        malloc(num_packed_msgs * num_lwes * sizeof(Torus)));
     auto h_body_id_per_compact_list =
         static_cast<uint32_t *>(malloc(num_lwes * sizeof(uint32_t)));
     auto h_lwe_compact_input_indexes =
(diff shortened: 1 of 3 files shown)


tfhe/src/integer/gpu/ciphertext/compact_list.rs
```diff
@@ -83,9 +83,9 @@ impl CudaCompactCiphertextListExpander {
         index: usize,
         streams: &CudaStreams,
     ) -> Option<(CudaRadixCiphertext, DataKind)> {
-        let preceding_infos = self.blocks_info.get(..index).unwrap();
-        let current_info = self.blocks_info.get(index).unwrap();
-        let message_modulus = self.blocks_info.get(index).unwrap().info.message_modulus;
+        let preceding_infos = self.blocks_info.get(..index)?;
```
agnesLeroy, line 86: Why not unwrapping anymore?

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cu
```diff
@@ -11,31 +11,28 @@ int32_t cuda_setup_multi_gpu(int device_0_id) {
   int num_gpus = cuda_get_number_of_gpus();
   if (num_gpus == 0)
     PANIC("GPU error: the number of GPUs should be > 0.")
-  int num_used_gpus = 1;
```
agnesLeroy: Maybe this is the cause of the failing test in the HL API? We should maybe change this in another PR?

tfhe/src/integer/gpu/ciphertext/compact_list.rs
pdroalves: If an exception happens we want to pass it upstream. There is an assert on a test expecting for it.

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cu
pdroalves: Not impossible, but I don't see how. Very weird. Yesterday I ran a intensive test with all possible combinations between 1 and 8 H100s. I will repeat it also on L40s.

pdroalves: And rollback this, just in case.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -559,7 +559,7 @@ template <typename Torus> struct int_radix_lut {
     using_trivial_lwe_indexes = false;
   }
 
-  // Broadcast luts from gpu src_gpu_idx to all active gpus
+  // Broadcast luts from gpu gpu_indexes[src_gpu_idx] to all active gpus
```
agnesLeroy: I think this comment can be put back to what is was before?

agnesLeroy (commented):
@pdroalves  maybe it would be good to have a call to double check the indexing logic, it is quite hard to read to me :sweat_smile:

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
pdroalves: Indeed. Done.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -159,46 +165,75 @@ template <typename Torus> struct zk_expand_mem {
...
+     *
+     * // First 4 use message LUT, last 4 use carry LUT
+     * h_lut_indexes  = {0, 0, 0, 0, 1, 1, 1, 1}
+     *
+     * // Reorders output so message and carry for each LWE appear together
+     * h_indexes_out  = {0, 2, 4, 6, 1, 3, 5, 7}
+     *
+     * If an LWE contains a boolean value, its LUT index is shifted by
+     * num_packed_msgs to use the sanitization LUT (which ensures output is
+     * exactly 0 or 1).
```
agnesLeroy, line 199: Looks great, thanks a lot!

PR #2400, opened 2025-06-03
chore(zk): use 8 bytes dsep and 128bits SID in hash functions

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1004

### PR content/description
- Use 8 bytes (instead of 256) for the domain separators used by hash functions
- Adds a random 128bits sid in all the hashes. The SID is defined by the CRS

These changes are made backward compatible, which means:
- the SID is an option, set to None for old CRS
- the 256 bytes domain separators are still supported

This PR also fixes a bug in the `bad_noise` test where the generated noise for the test ended up within the bounds if the tested noise coeff was negative (this is bad because we want to test that proofs are correctly rejected with "out of bound" noise). This was not detected because the test vectors are seeded and the seed always generated a positive coeff.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2400)

### DIFF
--- a/.github/workflows/aws_tfhe_fast_tests.yml
+++ b/.github/workflows/aws_tfhe_fast_tests.yml
@@ -185,7 +185,8 @@ jobs:
           make test_tfhe_csprng_big_endian
 
       - name: Run tfhe-zk-pok tests
-        if: needs.should-run.outputs.zk_pok_test == 'true'
+        # Always run it to catch non deterministic bugs earlier
+        # if: needs.should-run.outputs.zk_pok_test == 'true'
         run: |
           make test_zk_pok
 
--- a/tests/backward_compatibility/high_level_api.rs
+++ b/tests/backward_compatibility/high_level_api.rs
@@ -1,10 +1,18 @@
 use super::shortint::load_params;
 use crate::{load_and_unversionize, TestedModule};
 use std::path::Path;
-use tfhe::prelude::{CiphertextList, FheDecrypt, FheEncrypt};
-use tfhe::shortint::{AtomicPatternParameters, PBSParameters};
+use tfhe::integer::parameters::DynamicDistribution;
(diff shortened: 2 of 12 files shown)


soonum (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2400)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker and @mayeul-zama)

___
*[`.github/workflows/aws_tfhe_fast_tests.yml` line 189 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2400#-OS-0_ys9Gymoswd3KG7:-OS-0_ys9Gymoswd3KG8:bd33u9h) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4526d35196688b063d956fa67f302090698cd7a9/.github/workflows/aws_tfhe_fast_tests.yml#L189)):*
> ```YAML
>       - name: Run tfhe-zk-pok tests
>         # Always run it to catch non deterministic bugs earlier
>         # if: needs.should-run.outputs.zk_pok_test == 'true'
> ```

Maybe we should remove this condition then ?
I'm wondering why running it every time would catch non-deterministic bugs ? Just by relying on the sheer number of runs we do across PRs ?

IceTDrinker:
an issue won't hurt to not forget, or maybe we enable the tests to run on merge to main or nightly, don't recall when our other tests run

soonum (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2400)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IceTDrinker and @mayeul-zama)

PR #2401, opened 2025-06-04
chore(gpu): prevent nvToolsExt inclusion when not profiling

andrei-stoian-zama:
When the `profile` feature is not activated, the `USE_NVTOOLS` is not defined. This prevents compilation of some profiling helper functions but doesn't disable inclusion of `nvToolsExt.h` which doesn't exist for some cuda installations. This PR removes the include when the flag is off.

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/utils/helper_profile.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/utils/helper_profile.cuh
@@ -1,6 +1,9 @@
 #ifndef HELPER_PROFILE
 #define HELPER_PROFILE
+
+#ifdef USE_NVTOOLS
 #include <nvToolsExt.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/utils/helper_profile.cuh
```diff
@@ -9,5 +11,9 @@ void cuda_nvtx_pop();
   { cuda_nvtx_label_with_color(name); }
 #define POP_RANGE()                                                            \
   { cuda_nvtx_pop(); }
+#else
```
guillermo-oyarzun: this part is not needed cause we disabled it inside the functions.

andrei-stoian-zama: thanks, I changed it

PR #2430, opened 2025-06-11
chore(ci): use gh cli instead of api calls done with curl

soonum:
This refactor simplifies reading of data_pr_close workflow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2430)


nsarlin-zama (approved):
way simpler and robust that way, thanks !

PR #2431, opened 2025-06-11
fix(zk): test failed with trivial ct equal to 0

nsarlin-zama:
Fix a failed zk test when the seed generated a trivial 0 which didn't modified the input ct. This made the proof being accepted, while the test expected it to be rejected


IceTDrinker (approved):
Thanks !

PR #2434, opened 2025-06-12
chore(gpu): skip packing ks test if it needs more ram than available

andrei-stoian-zama:
GPU Packing KS benches allocate a large scratch buffer. Depending on parameter sets, this buffer could exceed GPU RAM. This PR skips tests that may fail because of OOM.


tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,157 +591,199 @@ mod cuda {
...
+                                        gpu_keys.pksk.as_ref().unwrap(),
+                                        &d_input_lwe_list,
+                                        &mut d_output_glwe,
+                                        &streams,
+                                    );
+                                    black_box(&mut d_input_lwe_list);
+                                })
+                            });
+                        }
                     }
```
agnesLeroy: If skip_bench = true maybe we could print a warning message indicating that the bench was skipped due to lack of memory?

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,157 +591,199 @@ mod cuda {
...
+                                                    local_stream,
+                                                );
+                                            },
+                                        )
+                                },
+                                criterion::BatchSize::SmallInput,
+                            )
+                        });
+                    }
                 }
```
agnesLeroy: Same here

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,157 +591,199 @@ mod cuda {
...
+                    );
 
-                                    CudaLweCiphertextList::from_lwe_ciphertext_list(
-                                        &input_ct_list,
-                                        &local_streams[i],
-                                    )
-                                })
-                                .collect::<Vec<_>>();
+                    skip_bench = false;
+                    for gpu_index in 0..gpu_count {
```
agnesLeroy, line 685: Here you only need to check on gpu_indexes[0], which is where the function is executed.

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,157 +591,199 @@ mod cuda {
...
                         glwe_sk.polynomial_size(),
-                        GlweCiphertextCount(1),
-                        ciphertext_modulus,
-                        &streams,
+                        LweCiphertextCount(glwe_sk.polynomial_size().0),
                     );
 
-                    streams.synchronize();
+                    skip_bench = false;
+                    streams.gpu_indexes.iter().for_each(|gpu_index| {
```
agnesLeroy: Here you only need to check on gpu_indexes[0]

agnesLeroy (commented):
Thank you @andrei-stoian-zama! Looks good to me: there's nothing to do for multi-GPU configurations in this case because the core crypto entry points are only executed on 1 GPU regardless of the number of GPUs on the machine.

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -1062,6 +1062,38 @@ pub fn is_cuda_available() -> bool {
...
+            num_lwes.0 as u32,
+            false,
+        )
+    };
+    unsafe {
+        cleanup_packing_keyswitch_lwe_list_to_glwe(
+            streams.ptr[0],
+            streams.gpu_indexes[0].get(),
+            std::ptr::addr_of_mut!(fp_ks_buffer),
+            true,
```
agnesLeroy: false

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,157 +591,199 @@ mod cuda {
...
+                    *params,
+                    name,
+                    "packing_ks",
+                    &OperatorType::Atomic,
+                    bit_size,
+                    vec![bit_size],
+                );
+            }
         }
     }
```
agnesLeroy, line 781: For cuda_ks_group, cuda_multi_bit_ks_group you can also set

```
 let mut criterion: Criterion<_> = (Criterion::default()
        .sample_size(15)
        .measurement_time(std::time::Duration::from_secs(60)))
```
as on CPU

tfhe-benchmark/benches/core_crypto/ks_bench.rs
andrei-stoian-zama: will the warning show ? I had a println but it only shows when I use `--nocapture` in the cargo command

agnesLeroy: Hmm you're right, it would only show with nocapture. Well that's ok let's leave this as is then.

IceTDrinker (commented):
Seems to me the logic around skipping is maybe not exactly correct for saving files, also there is probably a way to make it simple

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2434#-:-OSYueGG8MROmy059c_5:bdva7ub)* status: all files reviewed, 9 unresolved discussions (waiting on @andrei-stoian-zama and @soonum)

___
*[`tfhe-benchmark/benches/core_crypto/ks_bench.rs` line 611 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2434#-OSYo3Mm53JVBU26ZrGk:-OSYo3Mm53JVBU26ZrGl:bds95by) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5ec985f90daf1c6e815271a105f084f3d6c707a/tfhe-benchmark/benches/core_crypto/ks_bench.rs#L611)):*
> ```XML
>                     skip_bench = !check_valid_cuda_malloc(mem_size, GpuIndex::new(0));
> 
>                     if !skip_bench {
> ```

let's use positive statements to make things easier to read (brains have trouble with negatives and even more double negatives), I'm guessing that if you are skipping the bench you could "just" continue the loop ?

if skip_bench {continue;} ?

also means you likely don't need the skip_bench at all if the cuda malloc failed ?
___
*[`tfhe-benchmark/benches/core_crypto/ks_bench.rs` line 682 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2434#-OSYuXBp50pNoNR5O1pO:-OSYuXBp50pNoNR5O1pP:b-m8cruh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5ec985f90daf1c6e815271a105f084f3d6c707a/tfhe-benchmark/benches/core_crypto/ks_bench.rs#L682)):*
> ```XML
>                     for gpu_index in 0..gpu_count {
>                         if !check_valid_cuda_malloc(mem_size, GpuIndex::new(gpu_index as u32)) {
>                             skip_bench = true;
> ```

to skip here you can add a label on the outer loop to indicate where to break 

https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html
___
*[`tfhe/src/core_crypto/gpu/mod.rs` line 1065 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2434#-OSYnW123YwxNCDOrGaW:-OSYnW123YwxNCDOrGaX:b-3vg8g3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5ec985f90daf1c6e815271a105f084f3d6c707a/tfhe/src/core_crypto/gpu/mod.rs#L1065)):*
> ```XML
> }
> 
> #[allow(clippy::too_many_arguments)]
> ```

I would be surprised if too many arguments triggered on that function, given the default threshold should be 7
___
*[`tfhe/src/core_crypto/gpu/mod.rs` line 1073 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2434#-OSYnifn4S6STCimYR3i:-OSYnifo939Xlvnm2t07:br43ujv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5ec985f90daf1c6e815271a105f084f3d6c707a/tfhe/src/core_crypto/gpu/mod.rs#L1073)):*
> ```XML
>     num_lwes: LweCiphertextCount,
> ) -> u64 {
>     let mut fp_ks_buffer: *mut i8 = std::ptr::null_mut();
> ```

i8 for GPU buffers ? :thinking:

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -588,163 +591,200 @@ mod cuda {
...
-                                        GlweCiphertextCount(1),
-                                        ciphertext_modulus,
-                                        &local_streams[i],
-                                    )
-                                })
-                                .collect::<Vec<_>>();
+                    if !skip_bench {
+                        let blocks: usize = 1;
+                        let elements = throughput_num_threads(blocks, 1);
+                        let elements_per_stream = std::cmp::min(elements as usize / gpu_count, glwe_sk.polynomial_size().0);
```
andrei-stoian-zama: @agnesLeroy : I changed the lwe list size here. all streams process a list of this size. 

 for pure throughput measurement (no induced latency due to oversubscription), the list size should be equal to the `throughput_num_threads` (this was already the case) - only  when possible. the list size can be no bigger than the GLWE poly size.

do you think this is ok?

agnesLeroy: Yes I think that should fix it

PR #2435, opened 2025-06-12
feat(hpu): Add support for tandem pdi

Bapt-Roux:
Add support for pdi archive and Fpga reloading.
Rely on Tandem implementation for hot-reloading of FPGA. Add reload procedure inside ffi/v80 backend.

Now when starting application on HpuV80, a first check of version is done. If version mismatch, a pdi reload is triggere


backends/tfhe-hpu-backend/config_store/v80/hpu_config.toml
```diff
@@ -6,7 +6,8 @@
           "${HPU_BACKEND_DIR}/config_store/${HPU_CONFIG}/hpu_regif_core_prc_3in3.toml"]
   polling_us=10
 [fpga.ffi.V80]
-  ami_id=1 # First ami device in the list
+  id= 0
+  pdi_path="${HPU_BACKEND_DIR}/config_store/v80_pdi/psi64.hpu"
```
hcampos-zama: Since we're changing nomenclature to have everything called **hpu**, maybe we could change this variable to **hpu_path**.

backends/tfhe-hpu-backend/src/utils/pdi_mgmt.rs
```diff
@@ -0,0 +1,36 @@
...
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let cli = Cli::parse();
+
+    match cli.command {
+        Commands::Pack { from_path, to_file } => {
+            let hpu_pdi = HpuV80Pdi::from_folder(&from_path)?;
+            hpu_pdi.to_bincode(&to_file)?;
+
+            println!("Successfully Pack folder {from_path} into {to_file}.");
```
hcampos-zama: **packed**

backends/tfhe-hpu-backend/src/utils/pdi_mgmt.rs
```diff
@@ -0,0 +1,36 @@
...
+        }
+        Commands::Unpack { from_file, to_path } => {
+            let hpu_pdi = HpuV80Pdi::from_bincode(&from_file)?;
+            hpu_pdi.to_folder(&to_path)?;
+            println!("Successfully Unpack file {from_file} into {to_path} folder.");
```
hcampos-zama: **unpacked**

hcampos-zama (approved):
The only downside of this approach is that if we ever have to add/change fields of the .hpu file format, we'll have no way to distinguish old formats from new ones.. Can we at least add a format version specifier or something that would allow us to do that?

Everything else looks great!

backends/tfhe-hpu-backend/config_store/v80/hpu_config.toml
Bapt-Roux: done

pgardratzama (approved):

PR #2441, opened 2025-06-13
feat(hpu): normality & expected value checks for HPU parameters

pgardratzama:
also adds noise measurement tests for 2M128 HPU parameter set

### DIFF
--- a/tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
+++ b/tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
@@ -1,14 +1,13 @@
 use super::*;
 use crate::core_crypto::commons::math::ntt::ntt64::Ntt64;
-use crate::core_crypto::commons::test_tools::{torus_modular_diff, variance};
+use crate::core_crypto::commons::test_tools::{
+    arithmetic_mean, normality_test_f64, torus_modular_diff, variance,
+};
 use std::io;
 
 // This is 1 / 16 which is exactly representable in an f64 (even an f32)
 // 1 / 32 is too strict and fails the tests
 const RELATIVE_TOLERANCE: f64 = 0.0625;
-
-const NB_HPU_TESTS: usize = 5;
-const NB_PBS: usize = 200;
 
 #[derive(Clone, Copy)]
 pub struct HpuTestParams {
@@ -107,6 +106,26 @@ pub const HPU_TEST_PARAMS_4_BITS_HPU_64_KS_21_132_TUNIFORM: HpuTestParams = HpuT
     message_modulus_log: CiphertextModulusLog(4),
     ct_width: 64,
     ksk_width: 21,
+    //norm2: 8,
+    norm2: 5,
+    ntt_modulus: 18446744069414584321,
+};
+
+#[allow(unused)]
+pub const HPU_TEST_PARAMS_4_BITS_HPU_64_KS_21_132_TUNIFORM_2M128: HpuTestParams = HpuTestParams {
+    lwe_dimension: LweDimension(879),
+    glwe_dimension: GlweDimension(1),
+    polynomial_size: PolynomialSize(2048),
+    lwe_noise_distribution: DynamicDistribution::new_t_uniform(3),
+    glwe_noise_distribution: DynamicDistribution::new_t_uniform(17),
+    pbs_base_log: DecompositionBaseLog(23),
+    pbs_level: DecompositionLevelCount(1),
+    ks_level: DecompositionLevelCount(8),
+    ks_base_log: DecompositionBaseLog(2),
+    message_modulus_log: CiphertextModulusLog(4),
+    ct_width: 64,
+    ksk_width: 21,
+    //norm2: 8,
     norm2: 5,
     ntt_modulus: 18446744069414584321,
 };
@@ -165,7 +184,13 @@ pub fn get_modulo_value<T: UnsignedInteger>(modulus: &CiphertextModulus<T>) -> u
 }
 
 //fn lwe_noise_distribution_hpu<Scalar: UnsignedTorus + CastInto<usize> + CastFrom<u64>>(
-fn hpu_noise_distribution(params: HpuTestParams) {
+fn hpu_noise_distribution(
+    params: HpuTestParams,
+    max_msg_val: u64,
+    nb_tests: usize,
+    nb_pbs_per_test: usize,
+    check_variance: bool,
+) {
     let lwe_dimension = params.lwe_dimension;
     let glwe_dimension = params.glwe_dimension;
     let polynomial_size = params.polynomial_size;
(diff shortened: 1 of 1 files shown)


tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
```diff
@@ -107,6 +106,26 @@ pub const HPU_TEST_PARAMS_4_BITS_HPU_64_KS_21_132_TUNIFORM: HpuTestParams = HpuT
...
+    norm2: 5,
+    ntt_modulus: 18446744069414584321,
+};
+
+#[allow(unused)]
+pub const HPU_TEST_PARAMS_4_BITS_HPU_64_KS_21_132_TUNIFORM_2M128: HpuTestParams = HpuTestParams {
+    lwe_dimension: LweDimension(879),
+    glwe_dimension: GlweDimension(1),
+    polynomial_size: PolynomialSize(2048),
+    lwe_noise_distribution: DynamicDistribution::new_t_uniform(3),
```
IceTDrinker, line 118: this looks incorrect  for lwe noise distribution

tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
```diff
@@ -165,7 +184,13 @@ pub fn get_modulo_value<T: UnsignedInteger>(modulus: &CiphertextModulus<T>) -> u
 }
 
 //fn lwe_noise_distribution_hpu<Scalar: UnsignedTorus + CastInto<usize> + CastFrom<u64>>(
-fn hpu_noise_distribution(params: HpuTestParams) {
+fn hpu_noise_distribution(
+    params: HpuTestParams,
+    max_msg_val: u64,
+    nb_tests: usize,
+    nb_pbs_per_test: usize,
+    check_variance: bool,
```
Bapt-Roux: Instead of a boolean, I will have used an enum like: 
```
enum HpuNoiseMode {
 Variance,
 Normality,
}
```
And replace if stmt by match

tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
pgardratzama: I was also a bit surprised but that is the output of minimal_bound from the optimizer and the lattice estimator looks ok with that :
> standard_deviations: [(879, 2.2215530137414073e-06, 3)]
> log2_modulus : 21.0
> lambda= 134.84988138366035

IceTDrinker: ~~it cannot be correct~~

ah nevermind it's the 21 bit thing

IceTDrinker: time to go in the weekend I guess 😵‍💫

pgardratzama: Have a great WE

tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs
pgardratzama: It has not been easy :sweat: but I think it is done

pgardratzama:
To be clear, I had occurences of failure with expected value score check with ~7% probability. But most of the time it works well and is below 6.25%.

IceTDrinker:
> To be clear, I had occurences of failure with expected value score check with ~7% probability. But most of the time it works well and is below 6.25%.

when do those tests run ? if they fail with such high probability we can't have them like this in CI, I know they are statistical tests but we can't have very flaky tests

pgardratzama:
noise measurement are ok and stable
normality tests are checking 2 things: normality (stable under 6.25% failure rate), expected value near enough 0 with a score (failure rate between 3 and 7%)
I will just raise expected failure rate of expected value score and we will continue like that

IceTDrinker:
Well there at least something to discuss with R&D and it is that : the normality is degraded after the 21 bits keyswitch and we know quantization on low bit width is a topic, if we expect 5% and can get +50% it is something we need to discuss theoretically

IceTDrinker:
The normality being unachievable after mod switch because of normalization requires us to have normality before the mod switch apparently, so if we are seeing trouble there as well it’s something worth verifying.

if it’s after the PBS it’s even more concerning

pgardratzama:
both check are done after KS (and *nu before):
normality_test_f64() is fine with failure probability < 5%
what is not always < 5% is the expected value score: mean(samples) / std_dev(samples) * sqrt(samples.len()) € [-1.96..1.96]
@mballandras did not seem surprised by this failure probability

mballandras:
Just to add that the test will still be powerful. In this same setting (100 tests with 100 samples each), if instead of zero the mean of the Gaussian is std_dev/10 (which would affect the p_fail by less than a bit), the failure rate of the test will be 17%, which will be catch almost always having set the threshold to 13%

IceTDrinker (commented):
Thanks a few nits, otherwise looks to be in line with what Mathieu proposed for the tests, to check if the 8% bound should be changed to 13% given his recommendations, I'm guessing @mballandras that the test for the average value is the one we will want to add for other primitives as well, is there a write-up about the specs of the test ? will be of interest when we update older core noise tests for the CPU part

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2441#-:-OT6QITiBCZLj5_qWmJj:bgokx0r)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @Bapt-Roux)

___
*[`tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs` line 128 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2441#-OT6P_nr8Cjjsn0eEew5:-OT6P_nr8Cjjsn0eEew6:bl5eds2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88ab60bacb280f5099a6af95eab67093a7c59fac/tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs#L128)):*
> ```XML
>     ct_width: 64,
>     ksk_width: 21,
>     //norm2: 8,
> ```

comment to remove ?
___
*[`tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs` line 735 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2441#-OT6Q91Z80id-CMgKVxb:-OT6Q91Z80id-CMgKVxc:b-va6l56) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88ab60bacb280f5099a6af95eab67093a7c59fac/tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs#L735)):*
> ```XML
> create_parameterized_test_hpu!(
>     hpu_noise_distribution {
>         //HPU_TEST_PARAMS_4_BITS_NATIVE_U64,
> ```

commented param sets could be removed ?
___
*[`tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs` line 752 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2441#-OT6QAjh3mA1Pry44QUk:-OT6QAjh3mA1Pry44QUl:b-2guby3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88ab60bacb280f5099a6af95eab67093a7c59fac/tfhe/src/core_crypto/algorithms/test/noise_distribution/lwe_hpu_noise.rs#L752)):*
> ```XML
> create_parameterized_test_hpu!(
>     hpu_noise_distribution {
>         //HPU_TEST_PARAMS_4_BITS_HPU_64_KS_21_132_GAUSSIAN,
> ```

same here

pgardratzama:
removed useless commented lines, rebased & reduced to 2 commits

IceTDrinker (changes requested):
Normality check failed, this is too flaky too merge as is

mballandras:
The normality test is similar to the zero-mean test, it is a null hypothesis test at 5% level of significance. If it is performed 100 times and we set the limit at 13% failure we should see only 0.15% of false positive

pgardratzama:
Agreed, what is really strange is that I get failure rate on normality check below 5% on my laptop and at the same time 7% on the CI. I have the impression failure rate is always worse on CI...
Is this because of what does the test:
```
iter A (100) {
  - generate keys
  - encrypt msg
  - iter B (160) {
    - * nu
    - KS
    - sample noise in vec
    - PBS
  }
  check normality & expected value
}
```
Could it be because I do not re-encrypt the msg at each run of iter B?

IceTDrinker:
@pgardratzama do you run strictly the same tests between CI and laptop ?

you could try to seed the test to compare both CI and laptop runs, given it is the ntt PBS the polynomial mul algo should be the same between both

PR #2451, opened 2025-06-17
chore(ci): add throughput option for erc20 and dex benchmarks

soonum:
Weekly benchmarks would run on both latency and throughput configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2451)


soonum:
This PR is useless since, throughput are already running  alongside latency each time ERC20 or DEX benchmarks targets are run.

PR #2456, opened 2025-06-17
HPU integer bench need to do LFS pull

pgardratzama:
to get the bistreams
(also comment ERC20 flavor that is not working yet with HPU)


Bapt-Roux (approved):
Look good to me.
We will add a proper sequential(i.e. without rayon) version of transfer_safe later and enable back the associated benches.

soonum (approved):
Great, thanks !

PR #2463, opened 2025-06-19
chore(ci): set pull-requests permission to write in commit checks

soonum:
This is mandatory according to the action documentation, notably to be able to write issue comment within the pull-request.


IceTDrinker:
What issue is this adressing ?

IceTDrinker:
zizmor seems to say you can use write at a more granular scope, perhaps like steps

PR #2466, opened 2025-06-19
chore(gpu): move sum ctxt lut allocation to host to save memory

agnesLeroy:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -639,6 +536,14 @@ __host__ void host_integer_partial_sum_ciphertexts_vec_kb(
                          current_blocks, &current_blocks_slice,
                          num_radix_blocks);
   }
+  // Since setup_lookup_tables is called in the host the release of memory for
+  // the lut has to be done here as well
+  if (!mem_ptr->mem_reuse) {
```
andrei-stoian-zama: I suggest this logic should be moved into the `int_sum_ciphertexts_vec_memory` class. In general I suggest avoiding new/delete pairs that are located in different files.

andrei-stoian-zama (commented):
I can't judge the whole PR, just a suggestion to improve encapsulation in `int_sum_ciphertexts_vec_memory`

backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
agnesLeroy: The purpose of this PR is to make an exception for sum ciphertexts to avoid allocating too much memory that we don't use actually. In the scratch we don't know yet exactly how many blocks are going to be needed for the lut. But maybe I could pass the input degrees to the scratch so we have the information there and move back the release in int_sum_ciphertexts_buffer? The way things are done now, we avoid allocating a huge lut when it's not needed (only 1 or 2 inputs for example), but I guess I could also check for that in the scratch.

agnesLeroy: I can't actually pass degrees in the scratch directly, sometimes degrees come from intermediate buffers.

agnesLeroy:
@andrei-stoian-zama @bbarbakadze without this change we can't execute the 4090 tests, swap benchmarks and integer throughput benchmarks, they get out of memory. This PR is thus making an exception for sum ciphertexts to allocate the lut in the host function instead of the scratch. The logic is slightly simplified compared to the previous version you read @andrei-stoian-zama and honestly I don't see how to make that simpler. But we have to merge this change to fix the CI and be able to release something tomorrow.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -82,6 +83,159 @@ void generate_many_lut_device_accumulator(
...
+        const uint32_t prev_c_id = c_id - 1;
+        const uint32_t prev_column_len = columns_counter[prev_c_id];
+        for (uint32_t i = 0; i + chunk_size <= prev_column_len;
+             i += chunk_size) {
+          const Torus in_index = columns[prev_c_id][i];
+          const Torus out_index = columns[prev_c_id][i + 1];
+          new_columns[c_id][ct_count] = out_index;
+          h_indexes_in[pbs_count] = in_index;
+          h_indexes_out[pbs_count] = out_index;
+          h_lut_indexes[pbs_count] = 1;
```
bbarbakadze: Maybe check if those three arrays are not a `nullptr`

bbarbakadze (approved):
Thank you @agnesLeroy.

agnesLeroy:
@bbarbakadze I'm also reverting your commit from Monday here to give you more time to stabilize it.

PR #2469, opened 2025-06-19
fix(hpu): Remove some hardcoded filename in tandem

Bapt-Roux:
Also enhance error handling related to user misconfiguration. And remove a bug with ami devn reading

### DIFF
--- a/backends/tfhe-hpu-backend/src/ffi/v80/mod.rs
+++ b/backends/tfhe-hpu-backend/src/ffi/v80/mod.rs
@@ -112,10 +112,12 @@ impl HpuHw {
         // NB: Separate the call to match sudoers rules
         let _ = Command::new("sudo")
             .arg("/usr/sbin/rmmod")
+            .arg("--syslog") // Output to syslog instead of stderr
             .arg("ami")
             .status();
         let _ = Command::new("sudo")
             .arg("/usr/sbin/rmmod")
+            .arg("--syslog") // Output to syslog instead of stderr
             .arg("qdma_pf")
             .status();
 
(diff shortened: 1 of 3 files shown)


backends/tfhe-hpu-backend/src/ffi/v80/mod.rs
```diff
@@ -149,7 +151,8 @@ impl HpuHw {
                 )
                 .arg("-tclargs")
                 .arg(format!("{}/{}", tmp_dir_str, &pdi_stg1_tmp))
-                .output();
+                .output()
+                .expect("Stage1 loading encounter error");
```
JJ-hw:
```suggestion
                .expect("Stage1 loading encounters error");
```

setup_hpu.sh
```diff
@@ -25,10 +25,10 @@ else
 fi
 
 # V80 bitstream refresh rely on XilinxVivado tools
-XILINX_VIVADO=${XILINX_VIVADO:-"/opt/xilinx/Vivado/2024.2"}
+XILINX_VIVADO=${XILINX_VIVADO:-"/opt/amd/Vivado/2024.2"}
```
JJ-hw, line 28: 1. Unfortunately the directory organization won't be like this from 2025.1.
2. The default path may not correspond to the one from the user

JJ-hw (commented):
To be solved : how to handle vivado path. We should support pre - 2025.1 and post.

backends/tfhe-hpu-backend/src/ffi/v80/mod.rs
Bapt-Roux: add in the force push

setup_hpu.sh
Bapt-Roux: yes we should update the path if we change tool version.
However, the version used to load stg1 pdi could be != of the one used to generate the bitstream.

If the user has another path, he could define it before sourcing the script. Those path are for hw-team convenience only

PR #2470, opened 2025-06-20
fix(shortint): wrong LweDimension returned by prf multibit mod switched ct

IceTDrinker:


IceTDrinker:
test oprf_test_uniformity_ci_run_filter will be the verifier for the fix, worked locally

soonum (approved):
catJAM
Thanks !

PR #2471, opened 2025-06-20
chore(gpu): refactor the gpu oom checker

andrei-stoian-zama:
Replace this syntax:

```
     assert!(check_valid_cuda_malloc(
            scalar_bitand_size_on_gpu,
            GpuIndex::new(0)
        ));
```

with new function `check_valid_cuda_malloc_assert_oom` used like:

```
    check_valid_cuda_malloc_assert_oom(
            bitand_size_on_gpu,
            GpuIndex::new(0)
        );
```

Error message printed if OOM is detected:

```
 Not enough memory on GPU 0. Allocating 16400 bytes exceeds total memory: 8318484480 bytes
```

### DIFF
--- a/tfhe/src/core_crypto/gpu/mod.rs
+++ b/tfhe/src/core_crypto/gpu/mod.rs
@@ -1124,6 +1124,24 @@ pub fn synchronize_devices(gpu_count: u32) {
 pub fn check_valid_cuda_malloc(size: u64, gpu_index: GpuIndex) -> bool {
     unsafe { cuda_check_valid_malloc(size, gpu_index.get()) }
 }
+
+/// Check if a memory allocation fits in GPU memory. If it doesn't fit, panic with
+/// a helpful message.
+pub fn check_valid_cuda_malloc_assert_oom(size: u64, gpu_index: GpuIndex) {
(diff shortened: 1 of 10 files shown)


tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -1124,6 +1124,24 @@ pub fn synchronize_devices(gpu_count: u32) {
 pub fn check_valid_cuda_malloc(size: u64, gpu_index: GpuIndex) -> bool {
     unsafe { cuda_check_valid_malloc(size, gpu_index.get()) }
 }
+
+/// Check if a memory allocation fits in GPU memory. If it doesn't fit, panic with
+/// a helpful message.
+pub fn check_valid_cuda_malloc_assert_oom(size: u64, gpu_index: GpuIndex) {
```
agnesLeroy, line 1130: Maybe this function should be `pub(crate)`, I'm not sure users want to panic in this way. :thinking: Wdyt @tmontaigu?

tmontaigu: Being pub is fine, as long the function that just checks is available user will just chose the one they want

agnesLeroy (approved):
Thank you!

PR #2472, opened 2025-06-20
chore(gpu): add cuda debug target for integer tests

andrei-stoian-zama:
Adds a makefile target `test_integer_gpu_debug` that enables Debug compilation of the CUDA backend

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -658,6 +658,14 @@ test_integer_gpu: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --doc --profile $(CARGO_PROFILE) \
 		--features=integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key::
 
+.PHONY: test_integer_gpu_debug # Run the tests of the integer module with Debug flags for CUDA
+test_integer_gpu_debug: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile release_lto_off \
+		--features=integer,gpu-debug -vv -p $(TFHE_SPEC) -- integer::gpu::server_key:: --test-threads=1 --nocapture
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --doc --profile release_lto_off \
+		--features=integer,gpu-debug -p $(TFHE_SPEC) -- integer::gpu::server_key::
+
+
 .PHONY: test_integer_long_run_gpu # Run the long run integer tests on the gpu backend
 test_integer_long_run_gpu: install_rs_check_toolchain install_cargo_nextest
 	BIG_TESTS_INSTANCE="$(BIG_TESTS_INSTANCE)" \
--- a/backends/tfhe-cuda-backend/cuda/CMakeLists.txt
+++ b/backends/tfhe-cuda-backend/cuda/CMakeLists.txt
@@ -52,6 +52,8 @@ endif()
 
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
+else()
+  message("Building CUDA backend in ${CMAKE_BUILD_TYPE}")
 endif()
 
 # Add OpenMP support
--- a/backends/tfhe-cuda-backend/Cargo.toml
+++ b/backends/tfhe-cuda-backend/Cargo.toml
@@ -19,3 +19,4 @@ bindgen = "0.71"
 [features]
 experimental-multi-arch = []
 profile = []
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -655,6 +655,14 @@ test_integer_gpu: install_rs_build_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --doc --profile $(CARGO_PROFILE) \
 		--features=integer,gpu -p $(TFHE_SPEC) -- integer::gpu::server_key::
 
+.PHONY: test_integer_gpu_debug # Run the tests of the integer module with Debug flags for CUDA
+test_integer_gpu_debug: install_rs_build_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile release_lto_off \
+		--features=integer,gpu,tfhe-cuda-backend/debug -p $(TFHE_SPEC) -- integer::gpu::server_key:: --test-threads=4
```
andrei-stoian-zama: should we remove test-threads=4 here ?

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ endif()
 
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
+else()
+  message("Building CUDA backend in ${CMAKE_BUILD_TYPE}")
```
andrei-stoian-zama, line 56: I don't know how to run Cargo while allowing this to print, any idea ?

Makefile
agnesLeroy: --test-threads=1 for debug would make more sense

agnesLeroy (commented):
Thank you @andrei-stoian-zama! You also miss a feature in the tfhe-rs/tfhe Cargo.toml  in the same way that we have
`gpu-profile = ["gpu", "tfhe-cuda-backend/profile"]`

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
IceTDrinker: you likely cannot unless you grab the cmake logs and print it in the build.rs file

agnesLeroy: Normally Cargo prints the cmake output I think :thinking:  Maybe you need cargo clean?

andrei-stoian-zama: I tried the verbose build flag but still doesn't print it.. 

I added the printf in the C++ code

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -35,6 +40,12 @@ cudaStream_t cuda_create_stream(uint32_t gpu_index) {
   cuda_set_device(gpu_index);
   cudaStream_t stream;
   check_cuda_error(cudaStreamCreateWithFlags(&stream, cudaStreamNonBlocking));
+#ifndef NDEBUG
+  if (!g_debug_warning_printed)
```
andrei-stoian-zama: this is not very elegant, as it uses a global variable. 

with the global flag, running the `make test_integer_gpu_debug` target will only print the WARNING once

agnesLeroy: Hmm I'm not sure we want to do it this way. I've tried once to print CMake output in cargo build with no success, but it would be good to be able to make it so that we get the nvcc/gcc warnings displayed as well. So maybe we don't print this for now but keep it for when we manage to make this work properly? :thinking:

andrei-stoian-zama: I did some digging, the `cmake` crate uses `Command::status()` to wait for execution of the `cmake` executable. it appears this method makes the Command inherit stdin/stderr from the parent process, thus `cargo`. I still don't know why the C++ warnings and cmake messages don't show up though.

IceTDrinker: Don't think build.rs prints by default or at least the stdout is captured given some configs are forwarded with magic prints

See e.g. the rerun if example https://doc.rust-lang.org/cargo/reference/build-scripts.html

Does not get printed in your terminal

backends/tfhe-cuda-backend/cuda/CMakeLists.txt
andrei-stoian-zama: I removed the printf

backends/tfhe-cuda-backend/cuda/src/device.cu
andrei-stoian-zama: Adding cargo build -vv  shows build logs

agnesLeroy (approved):
Looks good to me, thank you very much!

PR #2481, opened 2025-06-28
fix(api): add clear error for .min/.max called with ciphertext by val…

swarna1101:
### Issue
When calling `.min()` or `.max()` on an FHE ciphertext and passing another ciphertext **by value** (e.g., `a.min(b)`), the error message was misleading:
```
the trait DecomposableInto<u64> is not implemented for FheUint<FheUint32Id>
```

This did not clearly indicate that a reference was required.

### Fix
Added "decoy" trait implementations for `FheMin<Self>` and `FheMax<Self>` for both `FheUint` and `FheInt` that trigger a clear compile-time error if called with a ciphertext by value.

### What Was Added
- Decoy impls for `.min(self, Self)` and `.max(self, Self)` that use `compile_error!` to provide a clear, actionable message.

### Proof of Fix
Now, calling `.min(b)` (by value) produces:

```
error: Cannot call `.min()` with an `FheUint` by value. Use a reference instead: `.min(&rhs)`
```

This makes it obvious to the user what needs to be fixed.

### DIFF
--- a/tfhe/src/high_level_api/integers/unsigned/ops.rs
+++ b/tfhe/src/high_level_api/integers/unsigned/ops.rs
@@ -277,6 +277,16 @@ where
     }
 }
 
+impl<Id> FheMax<Self> for FheUint<Id>
+where
+    Id: FheUintId,
+{
+    type Output = Self;
+    fn max(&self, rhs: Self) -> Self::Output {
+        self.max(&rhs)
+    }
+}
+
 impl<Id> FheMin<&Self> for FheUint<Id>
 where
     Id: FheUintId,
@@ -327,6 +337,16 @@ where
                 min_cmp.if_then_else(self, rhs)
             }
         })
+    }
+}
+
+impl<Id> FheMin<Self> for FheUint<Id>
+where
+    Id: FheUintId,
+{
+    type Output = Self;
+    fn min(&self, rhs: Self) -> Self::Output {
+        self.min(&rhs)
     }
 }
 
--- a/tfhe/src/high_level_api/integers/signed/ops.rs
+++ b/tfhe/src/high_level_api/integers/signed/ops.rs
@@ -168,6 +168,16 @@ where
     }
 }
 
+impl<Id> FheMax<Self> for FheInt<Id>
+where
+    Id: FheIntId,
+{
+    type Output = Self;
+    fn max(&self, rhs: Self) -> Self::Output {
+        self.max(&rhs)
+    }
+}
+
 impl<Id> FheMin<&Self> for FheInt<Id>
 where
     Id: FheIntId,
@@ -216,6 +226,16 @@ where
                 panic!("Hpu does not support this operation yet.")
             }
         })
+    }
+}
+
+impl<Id> FheMin<Self> for FheInt<Id>
+where
+    Id: FheIntId,
+{
+    type Output = Self;
+    fn min(&self, rhs: Self) -> Self::Output {
+        self.min(&rhs)
     }
 }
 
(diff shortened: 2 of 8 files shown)


tfhe/src/high_level_api/integers/unsigned/ops.rs
```diff
@@ -1199,12 +1219,15 @@ generic_integer_impl_operation!(
...
     ///
-    /// let result = &a / &b;
-    /// let result: u16 = result.decrypt(&client_key);
-    /// assert_eq!(result, 37849u16 / 3u16);
+    /// let (quotient, remainder) = (&a).div_rem(&b);
+    ///
+    /// let quotient: u16 = quotient.decrypt(&client_key);
+    /// assert_eq!(quotient, 23u16 / 3u16);
+    /// let remainder: u16 = remainder.decrypt(&client_key);
+    /// assert_eq!(remainder, 23u16 % 3u16);
```
tmontaigu: Revert this changes

tfhe/tests/test_min_max_by_value.rs
```diff
@@ -0,0 +1,37 @@
...
+    let min_by_ref = a_ref.min(&b_ref);
+    let max_by_ref = a_ref.max(&b_ref);
+    
+    let min_ref_result: u8 = min_by_ref.decrypt(&client_key);
+    let max_ref_result: u8 = max_by_ref.decrypt(&client_key);
+    
+    assert_eq!(min_ref_result, 3);
+    assert_eq!(max_ref_result, 5);
+    
+    println!("✅ Both by-value and by-reference min/max operations work correctly!");
```
tmontaigu: This print is not necessary

tfhe/tests/test_min_max_by_value.rs
```diff
@@ -0,0 +1,37 @@
+use tfhe::prelude::{FheEncrypt, FheDecrypt, FheMin, FheMax};
+use tfhe::{FheUint8, generate_keys, set_server_key, ConfigBuilder};
+
+#[test]
+fn test_min_max_by_value_and_reference() {
+    let config = ConfigBuilder::default().build();
+    let (client_key, server_key) = generate_keys(config);
+    set_server_key(server_key);
+
+    let a = FheUint8::encrypt(5u8, &client_key);
```
tmontaigu: You should generate the input as random values with rand

swarna1101: 👍

tfhe/tests/test_min_max_by_value.rs
swarna1101: 👍

tfhe/src/high_level_api/integers/unsigned/ops.rs
swarna1101: 👍

tfhe/src/high_level_api/integers/unsigned/ops.rs
```diff
@@ -1199,7 +1219,7 @@ generic_integer_impl_operation!(
     /// let (client_key, server_key) = generate_keys(ConfigBuilder::default());
     /// set_server_key(server_key);
     ///
-    /// let a = FheUint16::encrypt(37849u16, &client_key);
```
tmontaigu: This still needs to be reverted

swarna1101: done

tmontaigu (approved):

swarna1101:
can we merge this @tmontaigu ?

PR #2482, opened 2025-06-30
feat(gpu): increase keyswitch occupancy

guillermo-oyarzun:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -60,7 +60,7 @@ keyswitch(Torus *lwe_array_out, const Torus *__restrict__ lwe_output_indexes,
         lwe_array_in, lwe_input_indexes[blockIdx.x], lwe_dimension_in + 1);
 
     if (tid == lwe_dimension_out && threadIdx.y == 0) {
-      local_lwe_out = block_lwe_array_in[lwe_dimension_in];
+      local_lwe_out = -block_lwe_array_in[lwe_dimension_in];
```
agnesLeroy, line 73: Why this minus here?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -93,7 +93,7 @@ keyswitch(Torus *lwe_array_out, const Torus *__restrict__ lwe_output_indexes,
           lwe_acc_out[shmem_index + offset * blockDim.x];
     }
     if (threadIdx.y == 0)
-      block_lwe_array_out[tid] = lwe_acc_out[shmem_index];
+      block_lwe_array_out[tid] = -lwe_acc_out[shmem_index];
```
agnesLeroy, line 106: Same question here

agnesLeroy (commented):
Hey! I'll need some explanation on this PR to understand what's going on :sweat_smile:  
The fact that we skip the call to 
```
auto ksk_block =
            get_ith_block(ksk, i, j, lwe_dimension_out, level_count);
```
is enough to increase occupancy?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: Sure, so it is a trick to accumulate negative values instead of positive ones. This line makes sure that in the corner case we also have a negative value. And line 96  is the output writing that ensures we write back the correct sign again.
So the gains is that using += in line 80 is more efficient than -= in term of registers (a bit of magic probably cause we can use multiply-add instead of two ops). I combined that with pre-calculating the offset, to avoid using 
```
auto ksk_block =
            get_ith_block(ksk, i, j, lwe_dimension_out, level_count);

```
instead i jump directly using the ksk array, i think registers are saved there cause we don't need to recalculate something that is constant. (Otherwise is like having line 76 inside the loop)
We only needed to reduce 2 regs from 34-->32 to achieve 100%, that's why this really silly things make it possible.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: i answered in the previous one

agnesLeroy: Ok, this is magic :sweat_smile: could you write a detailed comment explaining this logic in the kernel please so we don't forget about it?

PR #2484, opened 2025-06-30
feat(hpu): Adding support for modulus switch mean compensation

hcampos-zama:
### PR content/description

This adds support for mean compensation in hardware and in the mockup along with the relevant parameter set to reach a probability of failure below 2**-128.


mockups/tfhe-hpu-mockup/src/modules/regmap.rs
```diff
@@ -24,6 +24,12 @@ pub(crate) struct BpipState {
     pub(crate) timeout: u32,
 }
 
+#[derive(Default)]
+pub(crate) struct KeyswitchState {
+    // Off by default
```
Bapt-Roux: Should be "On by default" to be aligned with HW no ?

Bapt-Roux (approved):
Except a comment regarding the default value of "KeyswitchState", everything look good  to me.

mockups/tfhe-hpu-mockup/src/modules/regmap.rs
hcampos-zama: Yes, thank you. Committed the fix.

hcampos-zama:
I'll then remove all mockup changes and make this PR only about the hardware support for mean compensation.

IceTDrinker (approved):
Approving to unblock but you will likely be missing the bit accuracy of CPU and HPU result for mean compensation for now

PR #2500, opened 2025-07-02
feat(hpu): centered_modswitch

Bapt-Roux:
Add support for centered modswitch with mean compensation in Hpu.
Also add various fix and update v1_3 parameters set.

We are almost ready, only have some final tests to run on board but it doesn't prevent to start the review.
NB: We revert the commit from Arthur for support of modswitch with mean compensation without the shift. Indeed, it's not needed anymore since we support the centered modswitch now.


Bapt-Roux:
Yes I see the issue with the revert commit. I've fix it by prepending "chore:" in front.
I've fixed the issue rose by Pierre et run all the on-board test.
We are good to go after the CI

PR #2507, opened 2025-07-02
chore(bench): reduce number of bit sizes to benchmark

soonum:
This is done to reduce execution time since some values are not useful measure.


IceTDrinker (approved):
Thanks !

PR #2511, opened 2025-07-02
feat(hpu): update HPU benches and use new HPU with improve BW

pgardratzama:


pgardratzama:
@soonum this PR with fix in HPU integer bench & HLAPI bench adding a separate workflow to run HLAPI bench on HPU should be ready for your review

PR #2518, opened 2025-07-03
feat(integer): truncate carry prop on trivial zeros

tmontaigu:
This changes the full_propagate_parallelized to not propagate most significant blocks which are trivial zeros.

This is a small performance improvement, especially interesting when having a bunch of FheUintX data, casted to FheUintY (Y > X) and summing them (e.g. n FheUint2, casted to FheUint32  and doing the sum to get the result on 32 bit)

### DIFF
--- a/tfhe/src/integer/server_key/radix_parallel/mod.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/mod.rs
@@ -227,13 +227,37 @@ impl ServerKey {
         T: IntegerRadixCiphertext,
     {
         let num_blocks = ctxt.blocks().len();
+
+        // Start the propagation on the first block that has carries
         let start_index = ctxt
             .blocks()
             .iter()
             .position(|block| !block.carry_is_empty())
             .unwrap_or(num_blocks);
 
-        let (to_be_cleaned, to_be_propagated) = ctxt.blocks_mut().split_at_mut(start_index);
+        // End the propagation 2 blocks after the last non-trivial zero block.
+        // We end it 2 blocks after because, in the worst case:
+        // 1) the last block (block `n`) 'immediate' carry will be the block n+1
+        // 2) a carry may be propagated from a block `< n` to block `n + 2` because block n is full
+        //
+        // e.g., 2 blocks, 2_2, notation: 0bcarry|msg, block in little endian order
+        // input: [0b11|11, 0b11|11, 0b00|00, 0b00|00, 0b00|00] = (5 * 15) % (1024)
+        // msg  : [0b00|11, 0b00|11, 0b00|00, 0b00|00, 0b00|00] = 15
+        // carry: [0b00|00, 0b00|11, 0b00|11, 0b00|00, 0b00|00] = 60
+        // msg + carry = [0b00|11, 0b00|11+11, 0b00|11, 0b00|00, 0b00|00]
+        // msg + carry = [0b00|00, 0b00|10, 0b00|11+1, 0b00|00, 0b00|00]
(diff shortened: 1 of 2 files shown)


IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/mod.rs` line 267 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518#-OUZhIxvCq65QLLPDbQ9:-OUZhIxvCq65QLLPDbQA:bjglc4l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/b7b425e8451afd0c88ff805fe90ee7bcc4afe8f1/tfhe/src/integer/server_key/radix_parallel/mod.rs#L267)):*
> ```XML
>             // => there is a block with carries
>             // => there is a block non-zero degree
>             .unwrap();
> ```

can there be a case here where we end early but some blocks still need noise cleaning ? a bit like what you have for start ?

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @tmontaigu)

___
*[`tfhe/src/integer/server_key/radix_parallel/mod.rs` line 267 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518#-OUZhIxvCq65QLLPDbQ9:-OUZpEz97-WMHbhjDUav:b4es4m) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/b7b425e8451afd0c88ff805fe90ee7bcc4afe8f1/tfhe/src/integer/server_key/radix_parallel/mod.rs#L267)):*

Yes, and it should be handled by the split_at_mut done later
</blockquote></details>

are we sure about that ? I mean what if blocks beyond the end index needed a noise cleaning, here as we have blocks restricted by the indexing operation could we have blocks that need noise cleaning that are not processed at all ?

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/mod.rs` line 267 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518#-OUZhIxvCq65QLLPDbQ9:-OU_Ewlv04lJCLWqWFV-:b7neqiq) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/b7b425e8451afd0c88ff805fe90ee7bcc4afe8f1/tfhe/src/integer/server_key/radix_parallel/mod.rs#L267)):*

are we sure about that ? I mean what if blocks beyond the end index needed a noise cleaning, here as we have blocks restricted by the indexing operation could we have blocks that need noise cleaning that are not processed at all ?
</blockquote></details>

Ah yes I see, i'll update

IceTDrinker:
*[`tfhe/src/integer/server_key/radix_parallel/mod.rs` line 267 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2518#-OUZhIxvCq65QLLPDbQ9:-OU_Gosd9pWJMT7cFNsq:b-sqg1by) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/b7b425e8451afd0c88ff805fe90ee7bcc4afe8f1/tfhe/src/integer/server_key/radix_parallel/mod.rs#L267)):*

but the probability of having a block with degree zero and noise level > Nominal is extremely low (I don't think its possible in regular use, only by handcrafting it)
</blockquote></details>

I think it can happen with a LUT full of 0s, odd corner case but yeah

PR #2519, opened 2025-07-03
doc(bench): update benchmark results tables

soonum:
All the results are using parameters set with p-fail of 2**-128. CPU tables using parameters set with p-fail 2**-64  are removed. GPU tables for 1xH100 and 2xH100 are now replace with the new hardware standard: 8xH100-SXM5.
HPU results are added to the backend comparison table and integrate latest operations available.


agnesLeroy (commented):
Looks great for GPU, thank you very much @soonum!

IceTDrinker (commented):
Think we are good

ALbur32:
Are the numbers in the table correct?

PR #2534, opened 2025-07-04
chore(ci): add actions write authorizations to be able to use the cache

IceTDrinker:


IceTDrinker:
doesn't fix anything, cache purged still getting 503

PR #2544, opened 2025-07-04
fix(gpu): fix regression on ERC20 throughput

pdroalves:
- transfers for the GPU case should be measured using sequential operations (without rayon!)

This PR partially revert commit fd79c4f97208ce93dfb95944a53a7b69cd0ea711

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1051

### Check-list:

### DIFF
--- a/tfhe-benchmark/benches/high_level_api/erc20.rs
+++ b/tfhe-benchmark/benches/high_level_api/erc20.rs
@@ -71,8 +71,31 @@ where
 
 /// This one also uses a comparison, but it leverages the 'boolean' multiplication
 /// instead of cmuxes, so it is faster
-#[cfg(not(feature = "hpu"))]
+#[cfg(all(feature = "gpu", not(feature = "hpu")))]
 fn transfer_no_cmux<FheType>(
+    from_amount: &FheType,
+    to_amount: &FheType,
+    amount: &FheType,
+) -> (FheType, FheType)
+where
+    FheType: Add<Output = FheType> + CastFrom<FheBool> + for<'a> FheOrd<&'a FheType> + Send + Sync,
+    FheBool: IfThenElse<FheType>,
+    for<'a> &'a FheType:
+        Add<Output = FheType> + Sub<Output = FheType> + Mul<FheType, Output = FheType>,
+{
+    let has_enough_funds = (from_amount).ge(amount);
+
+    let amount = amount * FheType::cast_from(has_enough_funds);
+
+    let new_to_amount = to_amount + &amount;
+    let new_from_amount = from_amount - &amount;
+
+    (new_from_amount, new_to_amount)
+}
+
+/// Parallel variant of [`transfer_no_cmux`].
+#[cfg(not(feature = "hpu"))]
+fn par_transfer_no_cmux<FheType>(
     from_amount: &FheType,
     to_amount: &FheType,
     amount: &FheType,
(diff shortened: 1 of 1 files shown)


tfhe-benchmark/benches/high_level_api/erc20.rs
```diff
@@ -689,7 +771,7 @@ fn main() {
             bench_name,
             "FheUint64",
             "transfer::whitepaper",
-            par_transfer_whitepaper::<FheUint64>,
+            transfer_whitepaper::<FheUint64>,
         );
         cuda_bench_transfer_throughput(
```
agnesLeroy, line 773: I think for all the other cuda throughput cases you need to make the change as well, to use the non-parallel variant.

pdroalves: `transfer_whitepaper` was the only method that had two versions and thus two names, `par_transfer_` and `transfer_`. The others were all parallel version without the `par_` preffix. Now we have two versions of all of them and we use the serial for throughput, with just `transfer_` in the name, and the parallel for latency, with `par_transfer_` as suffix.

tfhe-benchmark/benches/high_level_api/erc20.rs
```diff
@@ -71,7 +71,7 @@ where
 
 /// This one also uses a comparison, but it leverages the 'boolean' multiplication
 /// instead of cmuxes, so it is faster
-#[cfg(not(feature = "hpu"))]
+#[cfg(feature = "gpu")]
```
agnesLeroy: I think you have to keep not(feature = "hpu") here. You can have both:
```
#[cfg(all(feature = "gpu", not(feature = "hpu")))]
```

tfhe-benchmark/benches/high_level_api/erc20.rs
```diff
@@ -95,12 +118,36 @@ where
 
 /// This one uses overflowing sub to remove the need for comparison
 /// it also uses the 'boolean' multiplication
-#[cfg(not(feature = "hpu"))]
+#[cfg(feature = "gpu")]
```
agnesLeroy: Same here

tfhe-benchmark/benches/high_level_api/erc20.rs
```diff
@@ -122,12 +169,36 @@ where
 
 /// This ones uses both overflowing_add/sub to check that both
 /// the sender has enough funds, and the receiver will not overflow its balance
-#[cfg(not(feature = "hpu"))]
+#[cfg(feature = "gpu")]
```
agnesLeroy: Same here

tfhe-benchmark/benches/high_level_api/erc20.rs
agnesLeroy: Aaah ok makes sense! :pray:

agnesLeroy (commented):
Still some concern about the cargo features. After that I think we'll be good.

tfhe-benchmark/benches/high_level_api/erc20.rs
pdroalves: Indeed. That part made me confuse. Thx.

pdroalves: I thought no one would be using `gpu` and `hpu` at the same time, but they could, right?

agnesLeroy: Compilation allows it :slightly_smiling_face:

agnesLeroy (approved):

PR #2549, opened 2025-07-07
chore(integer): improve measurements

tmontaigu:
It seems that in
```rust
bench_group.bench_function(&bench_id, |b| {
  // some code
  b.iter(|| {
      // function to bench
  })
});
```
If we put code in the '// some code' part, it affects the measurements the slower this code is the worse the measurements can be.

For many operations the gap is small (a few ms or no gap), but for the division the gap was around 500ms.

So to reduce this, we move out what we can, moving the keycache access is the most important aspect as it cost around 70ms to 100ms.

A LazyCell is used in order only access the keycache is the bench is not filtered out. Which is the behaviour we had before this commit, and the behaviour we want to keep so that running specific benches via regex selection stay fast.

Also, for clean input benches, we use `iter` instead of `iter_batched` as it makes more sense and should give more accurate results as iter_batched timing include other things that just the timing of the function.

### Check-list:


soonum (commented):
Nice! Do you intend to perform some measurements on throughput benchmarks to check if this improve the behavior ?

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2549#-:-OUdHyNG0kIigE5yN47N:b40x85t)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tmontaigu)

PR #2553, opened 2025-07-08
chore: only run backward compat clippy on x86 machines

IceTDrinker:


nsarlin-zama:
what workfow is currently failing ?

nsarlin-zama (approved):
makes sense!

PR #2563, opened 2025-07-11
chore(bench): write params to file for each zk benchmark on gpu

soonum:
To be parsable each benchmark criterion ID must have their crypto details written to a file.

### DIFF
--- a/tfhe-benchmark/benches/integer/zk_pke.rs
+++ b/tfhe-benchmark/benches/integer/zk_pke.rs
@@ -710,15 +710,21 @@ mod cuda {
                         }
                     }
 
-                    write_to_json::<u64, _>(
-                        &bench_id_verify_and_expand,
-                        shortint_params,
-                        param_name,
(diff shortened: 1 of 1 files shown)


tfhe-benchmark/benches/integer/zk_pke.rs
```diff
@@ -710,15 +710,21 @@ mod cuda {
...
+                    for bench_id in [
+                        bench_id_verify,
+                        bench_id_expand_without_verify,
+                        bench_id_verify_and_expand,
+                    ] {
+                        write_to_json::<u64, _>(
+                            &bench_id,
+                            shortint_params,
+                            param_name,
+                            "pke_zk_verify_and_expand",
```
agnesLeroy: Shouldn't the name change for each bench?

soonum: You're right ! I forgot to handle the display name.

PR #2564, opened 2025-07-11
chore(gpu): add checks on noise levels in debug mode

agnesLeroy:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1063

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/build.rs
+++ b/backends/tfhe-cuda-backend/build.rs
@@ -54,8 +54,7 @@ fn main() {
         }
 
         if cfg!(feature = "debug") {
-            cmake_config.define("CMAKE_BUILD_TYPE", "DEBUG");
-            cmake_config.define("CMAKE_CXX_FLAGS", "-Wuninitialized -O0");
+            cmake_config.define("CMAKE_BUILD_TYPE", "Debug");
         }
 
         // Build the CMake project
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -20,6 +20,36 @@ public:
   static const uint64_t UNKNOWN = std::numeric_limits<uint64_t>::max();
 };
 
+#ifdef DEBUG
+#define CHECK_NOISE_LEVEL(noise_level_expr, msg_mod, carry_mod)                \
+  do {                                                                         \
+    if ((msg_mod) == 2 && (carry_mod) == 2) {                                  \
+      constexpr int max_noise_level = 3;                                       \
+      if ((noise_level_expr) > max_noise_level)                                \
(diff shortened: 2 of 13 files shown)


backends/tfhe-cuda-backend/build.rs
```diff
@@ -54,8 +54,7 @@ fn main() {
         }
 
         if cfg!(feature = "debug") {
-            cmake_config.define("CMAKE_BUILD_TYPE", "DEBUG");
-            cmake_config.define("CMAKE_CXX_FLAGS", "-Wuninitialized -O0");
+            cmake_config.define("CMAKE_BUILD_TYPE", "Debug");
```
andrei-stoian-zama, line 57: I also just stumbled upon this issue, the string is case sensitive compared to the "Debug" literal string so my previous approach wasn't actually working . I will add code to make the comparison case insensitive in another PR

agnesLeroy: Ok thanks, in the meanwhile this version is doing what we want, at least I could ensure the noise asserts are triggered when the debug feature is activated.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -20,6 +20,32 @@ class NoiseLevel {
...
+      constexpr int max_noise_level = 5;                                       \
+      if ((noise_level_expr) > max_noise_level)                                \
+        PANIC("Cuda error: noise exceeds maximum authorized value for 2_2 "    \
+              "parameters");                                                   \
+    } else if ((msg_mod) == 8 && (carry_mod) == 8) {                           \
+      constexpr int max_noise_level = 9;                                       \
+      if ((noise_level_expr) > max_noise_level)                                \
+        PANIC("Cuda error: noise exceeds maximum authorized value for 3_3 "    \
+              "parameters");                                                   \
+    }                                                                          \
```
andrei-stoian-zama, line 45: there are cases when this macro is called on 0, 0 for message modulus and carry modulus and we could want to prevent invalid values being passed too:

maybe add:
```
else if ((msg_mod) == 0 && (carry_mod) ==0) {
  // ok
} else {
 assert(false, "Invalid message modulus or carry modululs")
}
```

agnesLeroy: Hmm I don't think message/carry modulus equal to 0 is a valid case, where do you see we pass this?

agnesLeroy: Ah yes sorry that's what I did for shortint level ops where message & carry modulus are not actually defined.

agnesLeroy: Done

andrei-stoian-zama (approved):
Thanks for the changes!

PR #2566, opened 2025-07-13
Fix version comparison logic for Rust minimum version check

GarmashAlex:
Corrected the logic for comparing Rust versions in the check_cargo_min_ver.sh script. Now, if the current major version is greater than the minimum required, the check passes immediately. If the major versions are equal, the minor version is compared. This prevents false negatives when using newer major Rust versions. Previously, the script could incorrectly fail if the current major version was higher but the minor version was lower than the minimum required.


IceTDrinker:
hello @GarmashAlex if you signed the CLA let us know

PR #2580, opened 2025-07-16
fix(gpu): refactor broadcast_lut() so make it less error prone

pdroalves:
`broadcast_lut()` now assumes that the data that needs to be broadcast resides at `gpu_indexes[0]`.

### Check-list:


backends/tfhe-cuda-backend/cuda/include/device.h
```diff
@@ -26,6 +26,7 @@ inline void cuda_error(cudaError_t code, const char *file, int line) {
     std::abort();                                                              \
   }
 
+uint32_t cuda_get_device();
```
agnesLeroy, line 29: Maybe `cuda_get_current_device` would be a more accurate name?

PR #2591, opened 2025-07-21
chore(gpu): remove h_lut_indexes from int_radix_lut

pdroalves:
That pointer is misleading and unnecessary. It can make one think that its content is used to update the device array, which is actually used in the PBS. This is not the case. This array was meant only to be used as a helper when someone need to update the device array and doesn't want to explicitly allocate host memory.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
@@ -603,13 +603,20 @@ __host__ void integer_radix_apply_univariate_lookup_table_kb(
       cuda_synchronize_stream(streams[i], gpu_indexes[i]);
     }
   }
+  auto h_lut_indexes =
+      static_cast<Torus *>(malloc(num_radix_blocks * sizeof(Torus)));
+  cuda_memcpy_async_to_cpu(h_lut_indexes, lut->get_lut_indexes(0, 0),
+                           num_radix_blocks * sizeof(Torus), streams[0],
+                           gpu_indexes[0]);
+  cuda_synchronize_stream(streams[0], gpu_indexes[0]);
   for (uint i = 0; i < num_radix_blocks; i++) {
-    auto degrees_index = lut->h_lut_indexes[i];
+    auto degrees_index = h_lut_indexes[i];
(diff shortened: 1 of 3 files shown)


bbarbakadze (dismissed):

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -603,13 +603,20 @@ __host__ void integer_radix_apply_univariate_lookup_table_kb(
       cuda_synchronize_stream(streams[i], gpu_indexes[i]);
     }
   }
+  auto h_lut_indexes =
+      static_cast<Torus *>(malloc(num_radix_blocks * sizeof(Torus)));
+  cuda_memcpy_async_to_cpu(h_lut_indexes, lut->get_lut_indexes(0, 0),
```
agnesLeroy, line 608: This we really can't do, as it'll affect performance significantly. I see it in other places as well: we need the indexes on CPU to update degrees actually, and I don't think we can do without it. So I'd say we have to keep h_lut_indexes as it was in int_radix_lut, or do you see a way around this that does not use copy to CPU? Even using cudaMallocHost affects performance significantly we should avoid copy to CPU at all cost.

agnesLeroy (commented):
Hey @pdroalves! So now I remember why we have lut indexes on the CPU, it's actually to update degrees correctly. I don't think we should go the way you propose as it implies to copy data to the CPU which will kill performance.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
pdroalves: Ow. Yeah, you have a point. I don't think we can skip this.

Anyway, this is another example of how this variable is being erratically used: It's treated as a temporary space on utilities functions but this method expects it to actually have those indexes.

Let's forget about this PR and focus on encapsulating the whole lut indexes logic, as suggested on the hackathon.

pdroalves:
In the end this variable is actually needed. Otherwise we would have to allocate and copy memory to the CPU for each LUT we apply.

Let's forget about this PR and focus on encapsulating the whole lut indexes logic, as suggested on the hackathon.

PR #2595, opened 2025-07-23
fix(bench): store correct pfail from params

nsarlin-zama:
pfail in bench metadata was hardcoded to 2^-41, this uses the correct pfail from the parameters


IceTDrinker:
no I mean the old 2^-41 is kind of odd, but that's fine

PR #2601, opened 2025-07-25
feat(gpu): create specialized version of multi-bit pbs tbc

guillermo-oyarzun:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
```diff
@@ -415,15 +415,15 @@ uint64_t scratch_cuda_multi_bit_programmable_bootstrap_64(
...
 
-  if (supports_tbc &&
-      !(input_lwe_ciphertext_count > num_sms / 2 && supports_cg))
-    return scratch_cuda_tbc_multi_bit_programmable_bootstrap<uint64_t>(
-        stream, gpu_index, (pbs_buffer<uint64_t, MULTI_BIT> **)buffer,
-        glwe_dimension, polynomial_size, level_count,
-        input_lwe_ciphertext_count, allocate_gpu_memory);
-  else
+  // if (supports_tbc &&
+  //     !(input_lwe_ciphertext_count > num_sms / 2 && supports_cg))
```
agnesLeroy: The comments can probably be removed? Why don't we need the condition anymore?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
```diff
@@ -489,6 +489,10 @@ uint32_t get_lwe_chunk_size(uint32_t gpu_index, uint32_t max_num_pbs,
   int log2_max_num_pbs = log2_int(max_num_pbs);
   if (log2_max_num_pbs > 13)
     ith_divisor = log2_max_num_pbs - 11;
+#else
+  if (max_num_pbs <= 8) {
+    return num_sms / 2;
```
agnesLeroy, line 500: Why num_sms / 2 here?

agnesLeroy (commented):
Hey @guillermo-oyarzun! Thanks a lot for the changes! There's a lot in this PR :sweat_smile: I guess my main question is whether we could try to duplicate less code, but maybe it's hard to achieve.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
guillermo-oyarzun: so i was assuming the arch >=900 will always support tbc, maybe maybe is better to leave the supports tbc condition, just to be extra safe, the other part of the condition was to switch to cg instead of tbc for certain number of samples, but that is not true anymore

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
guillermo-oyarzun: So, the idea is that when we have few samples, i want to ensure to launch waves capable to use all sms for the keybundle (pbs_acc we don't care cause the number of samples mean only more iterations). So in pbs, the minimal glwe_dim=1, and in keybundle we launch chunk_size*samples*(glwe_dim + 1)² , with num_sms/2, i make sure that we always end up launching the largest multiple of the sms, and wave don't create tailing effect. This of course is not convenient if the number of samples grows much, that's why i limited to 8. Division and addition benefit from this, the rest of operations tend to work with larger number of samples so it fall back to the previous heuristic.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
agnesLeroy: Ah ok, yes maybe I prefer to be extra safe then and keep the tbc support check :smile:

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cu
agnesLeroy: Shouldn't it be num_sms / 4 then?

agnesLeroy: If we launch 4 * chunk_size * samples blocks? (with k + 1 = 2)

guillermo-oyarzun: So initially I thought the same, but the thing is the singleH100 PCIe, which has 114 sms, so 114/4=28.5. So I figured they will always be at least even, so we can allow division by 2, and performance still is good.

agnesLeroy: Aaah ok, maybe adding a comment here to trace why this choice is made would be good then :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -137,6 +137,39 @@ public:
...
+     */
+
+    res_re -= carry_re << base_log;
+    res_im -= carry_im << base_log;
+
+    typecast_torus_to_double(res_re, result[i].x);
+    typecast_torus_to_double(res_im, result[i].y);
+  }
+}
+
```
bbarbakadze: If we are assuming that pbs level is 1 do we need to keep state maybe avoid it might be better for registers

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
```diff
@@ -91,6 +91,23 @@ __device__ inline T init_decomposer_state(T input, uint32_t base_log,
...
+  res >>= 1;
+  constexpr T torus_max = scalar_max<T>();
+  constexpr T mod_mask = torus_max >> non_rep_bit_count;
+  res &= mod_mask;
+  T shifted_random = rounding_bit << (rep_bit_count - 1);
+  T need_balance =
+      (((res - (T)(1)) | shifted_random) & res) >> (rep_bit_count - 1);
+  return res - (need_balance << rep_bit_count);
+}
+
```
bbarbakadze, line 110: maybe we can remove this function too once we hardcode for `decomp_level=1` and this function can be fused with `decompose_and_compress_level_2_2_params`

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
```diff
@@ -96,8 +96,43 @@ template <class params> __device__ void NSMFFT_direct(double2 *A) {
...
+
+      w *= v[i];
+
+      v[i] = u[i] - w;
+      u[i] = u[i] + w;
+      tid = tid + STRIDE;
+    }
+  }
+
+  __syncthreads();
```
bbarbakadze, line 135: can't `__syncwarp` work here? after fft as I understand each thread is accessing memory part of shared memory which was stored by a thread from the same warp?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
```diff
@@ -146,6 +146,134 @@ __global__ void device_multi_bit_programmable_bootstrap_keybundle(
...
+    const Torus *__restrict__ lwe_array_in,
+    const Torus *__restrict__ lwe_input_indexes, double2 *keybundle_array,
+    const Torus *__restrict__ bootstrapping_key, uint32_t lwe_offset,
+    uint32_t lwe_chunk_size, uint32_t keybundle_size_per_input) {
+
+  constexpr uint32_t lwe_dimension = 920;
+  constexpr uint32_t polynomial_size = 2048;
+  constexpr uint32_t grouping_factor = 4;
+  constexpr uint32_t glwe_dimension = 1;
+  constexpr uint32_t level_count = 1;
```
bbarbakadze, line 174: is it going to be like this or will be changed before merging?  maybe not now but later we do some kind of executor in a same way as `cufftdx` is implemented.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
```diff
@@ -146,6 +146,134 @@ __global__ void device_multi_bit_programmable_bootstrap_keybundle(
...
+    // Move from local memory back to shared memory but as complex
+    double2 fft_regs[params::opt / 2];
+    double2 *fft = shared_fft;
+#pragma unroll
+    for (int i = 0; i < params::opt / 2; i++) {
+      fft_regs[i] =
+          make_double2(__ll2double_rn((int64_t)reg_acc[i]) /
+                           (double)std::numeric_limits<Torus>::max(),
+                       __ll2double_rn((int64_t)reg_acc[i + params::opt / 2]) /
+                           (double)std::numeric_limits<Torus>::max());
```
bbarbakadze, line 269: not sure if `(double)std::numeric_limits<Torus>::max()` this entire expression happens on compilation side. you could use a `double const TORUS_MAX`

backends/tfhe-cuda-backend/cuda/src/types/complex/operations.cuh
```diff
@@ -74,4 +74,11 @@ __device__ inline double2 operator*(double a, double2 b) {
   return {__dmul_rn(b.x, a), __dmul_rn(b.y, a)};
 }
 
+__device__ inline double2 shfl_xor_double2(double2 val, int laneMask,
+                                           unsigned mask = 0xFFFFFFFF) {
+  double lo = __shfl_xor_sync(mask, val.x, laneMask);
+  double hi = __shfl_xor_sync(mask, val.y, laneMask);
```
bbarbakadze: maybe rename `lo` and `hi` to `re` and `im`?

bbarbakadze (commented):
Amazing work @guillermo-oyarzun , thanks. just a few comments.

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
guillermo-oyarzun: The problem is that i don't sync before entering into the "warp looping part", and we are rewriting into the same shared memory that is used in the "shared memory looping part". So it could happen (and it did during some tests), that one of the warps arrived here and write on top of shared memory that was used from a warp that was still in the "shared memory loop part".  So in here we write within the warp dimensions, but the other warps might still need to read from that part. It was not happening always,  but with the long run tests i was having this issue. I didn't try adding the sync before the warp zone, maybe it can produce a better result in terms of registers, i will run a small test.

backends/tfhe-cuda-backend/cuda/src/types/complex/operations.cuh
guillermo-oyarzun: you are right that is a better naming.

backends/tfhe-cuda-backend/cuda/src/crypto/torus.cuh
guillermo-oyarzun: yup this was one of the things I wasn't sure about it, so my idea is that at some point we will be able to reuse some of this functions in the normal pbs, if we manage to pass the parameters as constexpr, and then maybe avoid to have this duplicity of functions.

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
guillermo-oyarzun: not sure what do you mean with keeping the state?, we just read it and don't use it anymore.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
guillermo-oyarzun: I didn't give it much of a though to this  part,  so you mean the casting would not happen at compilation time?  you mean using  scalar_max<T>() to calculate the TORUS_MAX instead? can you explain me more?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
guillermo-oyarzun: So the idea will be to wrap up this into a template parameter for instance before merging.  Ideally i would like to have all parameters as constexpr within the cuda backend, so we could enable more easy optimizations, if not we would need some kind of executor as you mention. But for starters, i think I can just wrap it in the template and then other PR that tries to switch to a more generic logic considering all this variables.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit.cuh
bbarbakadze: you could have a `constexpr double` hardcoded, calculate the exact value and hardcode it.

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
bbarbakadze: what I think is that if you hardcode for `pbs_decomp_level = 1` you can directly calculate decomposition without calculating state.

bbarbakadze (approved):

PR #2605, opened 2025-07-29
Am/test/future noise checks 3 noise squash comp

IceTDrinker:
This one was not part of the original atomic patterns tested, but was one I implemented for the noise squashing compression validation before it hit main recently

Edit: some misc small changes should be in separate clearly identifiable commits

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -808,7 +808,7 @@ test_integer_hpu_ci: install_rs_check_toolchain install_cargo_nextest
 test_integer_hpu_mockup_ci: install_rs_check_toolchain install_cargo_nextest
 	source ./setup_hpu.sh --config sim ; \
 	cargo build --release --bin hpu_mockup; \
-    coproc target/release/hpu_mockup --params mockups/tfhe-hpu-mockup/params/tuniform_64b_pfail64_psi64.toml > mockup.log; \
+	coproc target/release/hpu_mockup --params mockups/tfhe-hpu-mockup/params/tuniform_64b_pfail64_psi64.toml > mockup.log; \
 	HPU_TEST_ITER=1 \
 	cargo test --profile devo -p tfhe --features hpu --test hpu -- u32 && \
 	kill %1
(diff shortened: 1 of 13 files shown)


IceTDrinker:
naming for the BSK 128 is not quite correct I guess, standard was in the sense "standard PBS" construct, but in practice the key is in the fourier domain

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605)* status: all files reviewed, 4 unresolved discussions (waiting on @mayeul-zama, @nsarlin-zama, and @soonum)

___
*[`tfhe/src/shortint/server_key/tests/noise_distribution/utils/traits.rs` line 107 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605#-OWzucY987Kg33yBfmw4:-OX7vZxj0MKKEHB4ZZ3Y:b-m0ddlr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/3877827adcf7407d47deb8fb60cc48636052a91d/tfhe/src/shortint/server_key/tests/noise_distribution/utils/traits.rs#L107)):*

I was thinking of the GPU APIs those don't return values for the low level, they take output pointers as input and write into them
</blockquote></details>

given they will be plugging into those same APIs once we have tests setup

mayeul-zama:
*[`tfhe/src/shortint/server_key/tests/noise_distribution/utils/mod.rs` line 427 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605#-OWzsplL-W1G6-1XXEYz:-OX9Entr5RMg6PTh4EWo:bh2kbhb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/3877827adcf7407d47deb8fb60cc48636052a91d/tfhe/src/shortint/server_key/tests/noise_distribution/utils/mod.rs#L427)):*

not sure I follow ?
</blockquote></details>

We could remove the `lwe_per_glwe` argument
And let the caller do `let output  = new_from_glwe(...)[0..  lwe_per_glwe];` if needed

This would make this function more general

mayeul-zama (approved):
:lgtm:
Thanks!

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605#-:-OX9NoSh6gTSspdFFSeV:b-aw02zl)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker, @mayeul-zama, @nsarlin-zama, and @soonum)

___
*[`tfhe/src/shortint/server_key/tests/noise_distribution/utils/mod.rs` line 427 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2605#-OWzsplL-W1G6-1XXEYz:-OX9NfCI2ISdkqin9Vfb:bqy67xj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/3877827adcf7407d47deb8fb60cc48636052a91d/tfhe/src/shortint/server_key/tests/noise_distribution/utils/mod.rs#L427)):*

problem would be that the last slots don't contain the expected message if for example we take a compressed list that cannot be full (because of parameters limiting the amount of slots that are valid) so attempting to decrypt the last slots would be wrong ?

Essentially it means we could return results that have no meaning and I'm not sure I'm a fan of that
</blockquote></details>

Fair enough

IceTDrinker:
conflict after multi bit 128 was merged

PR #2612, opened 2025-07-31
chore(gpu): add valgrind in debug workflow

andrei-stoian-zama:
Adds valgrind to catch mem leak in the debug build workflow

Memory checks CI runs every 3 months and can be run manually

Closes https://github.com/zama-ai/tfhe-rs-internal/issues/1053

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2612)

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cu
@@ -83,6 +83,8 @@ void cleanup_cuda_integer_radix_logical_scalar_shift(
       (int_logical_scalar_shift_buffer<uint64_t> *)(*mem_ptr_void);
 
   mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
+  delete mem_ptr;
+  *mem_ptr_void = nullptr;
 }
 
 void cleanup_cuda_integer_radix_arithmetic_scalar_shift(
@@ -93,4 +95,6 @@ void cleanup_cuda_integer_radix_arithmetic_scalar_shift(
       (int_arithmetic_scalar_shift_buffer<uint64_t> *)(*mem_ptr_void);
 
   mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
+  delete mem_ptr;
+  *mem_ptr_void = nullptr;
 }
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
@@ -5,6 +5,8 @@
 #endif
 #include "ciphertext.h"
 
+#include <stdio.h>
+
 template <typename Torus>
 bool has_support_to_cuda_programmable_bootstrap_cg(uint32_t glwe_dimension,
                                                    uint32_t polynomial_size,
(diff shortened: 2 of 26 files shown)


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -731,6 +733,8 @@ void cleanup_cuda_programmable_bootstrap(void *stream, uint32_t gpu_index,
                                          int8_t **buffer) {
   auto x = (pbs_buffer<uint64_t, CLASSICAL> *)(*buffer);
   x->release(static_cast<cudaStream_t>(stream), gpu_index);
+    printf("RELEASE CUDA PBS CLASSICAL\n");
```
agnesLeroy: I think we want to remove this print

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2612)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @andrei-stoian-zama and @soonum)

___
*[`Makefile` line 691 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2612#-OWUjgWZ54-9Odwu614Q:-OWUjgWZ54-9Odwu614R:bo8l43r) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/876eee17e5167220a29b44038947804df799d3c9/Makefile#L691)):*
> ```Makefile
> .PHONY: test_high_level_api_gpu_debug # Run the tests of the integer module with Debug flags for CUDA
> test_high_level_api_gpu_debug: install_rs_build_toolchain install_cargo_nextest
> 	export RUSTFLAGS="-C target-cpu=x86-64" && \
> ```

just set RUSTFLAGS="" it will use the default
___
*[`scripts/check_memory_errors.sh` line 45 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2612#-OWUjnTn1E4ySbCwSi1l:-OWUjnTn1E4ySbCwSi1m:b3p9w0u) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/876eee17e5167220a29b44038947804df799d3c9/scripts/check_memory_errors.sh#L45)):*
> ```Shell
> fi;
> 
> exit 0
> ```

missing newline at the end of the file it seems install whichever extension for https://editorconfig.org/ for your editor

backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cu
```diff
@@ -93,4 +95,6 @@ void cleanup_cuda_integer_radix_arithmetic_scalar_shift(
       (int_arithmetic_scalar_shift_buffer<uint64_t> *)(*mem_ptr_void);
 
   mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
+  delete mem_ptr;
+  *mem_ptr_void = nullptr;
```
agnesLeroy, line 99: Why do you set mem_ptr_void to nullptr?

.github/workflows/gpu_code_validation_tests.yml
```diff
@@ -22,8 +22,9 @@ env:
 on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
-  pull_request:
-    types: [ labeled ]
+  schedule:
+    # every 3 months
+    - cron: "0 0 1 */3 *"
```
agnesLeroy, line 27: Hmm, I'm wondering if it wouldn't be good to keep compute-sanitizer running on every PR as it also detects warnings, and to have valgrind run every 3 months in another workflow, wdyt?

backends/tfhe-cuda-backend/cuda/src/integer/scalar_shifts.cu
andrei-stoian-zama: just extra safety, in case the pointer is accidentally re-used after free. https://stackoverflow.com/questions/1025589/setting-variable-to-null-after-free

.github/workflows/gpu_code_validation_tests.yml
andrei-stoian-zama: sure, I moved sanitizer to the pcc target

PR #2623, opened 2025-08-01
DO NOT MERGE wip wrappers cuda

IceTDrinker:
This is an attempt at simplifying the indexing logic in the expand cuda code, looks to be nearly working but I can't get around C++ header madness.

This proposal needs a better way of indexing in the flattened list (caching the result in the constructor) at the moment the algorithm re walks the list to find the start of each list which is not good.

My way of writing the wrapper structs naively probably means it's suboptimal for data copy, given some fields like `monomial_degree` for the `compact_lwe_body` are raw uint64_t, most likely meaning it's on the CPU side and CUDA has to make extra work to get the metadata to the GPU, this can likely be easily mitigated by having a pointer to data on the GPU where the proper information would be copied from the host, just a warning here that the code is a "this is the general idea, it needs refinement to fit the max performance requirements of the GPU"

I think this can be pushed further and remove the output pointer from the CUDA kernel and the kernel would only take a "jobs" struct pointer as input coming from the scratch struct. Each thread block fetches its own job with the pointer indexing already computed on the host side and stored in each of the jobs.

I think this is fine as a general idea given the scratch struct allocations are tightly coupled to the actual content of the compact list (said otherwise, the scratch is tailor made for the given compact list and can therefore not be reused for any other list).

Not sure it applies well to other cases given lwe_expand is apparently an outlier, but pushing the logic further, a scratch struct can be turned into a "task" struct that not only holds additional buffers, but also computes the indices for all required inputs by the kernels

cc @pdroalves

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -2,7 +2,7 @@
 #define CUDA_INTEGER_UTILITIES_H
 
 #include "integer.h"
-#include "integer/radix_ciphertext.cuh"
+#include "../../src/integer/radix_ciphertext.cuh"
 #include "integer/radix_ciphertext.h"
 #include "keyswitch/keyswitch.h"
 #include "pbs/programmable_bootstrap.cuh"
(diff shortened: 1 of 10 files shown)


agnesLeroy:
@pdroalves will take a look at this as soon as he has the time :wink:

IceTDrinker:
don't worry @agnesLeroy I discussed with him before closing the PR, the branch is still there :)

PR #2625, opened 2025-08-01
chore: fix noise checks timeout again as there are TWO timeout locations

IceTDrinker:
also change notification logic to warn if the overall result is not success

### DIFF
--- a/.github/workflows/aws_tfhe_noise_checks.yml
+++ b/.github/workflows/aws_tfhe_noise_checks.yml
@@ -55,6 +55,7 @@ jobs:
     name: CPU noise checks
     needs: setup-instance
     runs-on: ${{ needs.setup-instance.outputs.runner-name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/aws_tfhe_noise_checks.yml
```diff
@@ -73,15 +74,15 @@ jobs:
           make test_noise_check
 
       - name: Set pull-request URL
-        if: ${{ failure() && github.event_name == 'pull_request' }}
+        if: ${{ !success() }}
```
mayeul-zama, line 77: Not sure what this is needed
Same for the following changes

IceTDrinker: I need this to get the notification if it did not succeed, for now I could not get any slack notification if the run was cancelled

mayeul-zama (approved):

PR #2630, opened 2025-08-04
chore(gpu): add assert macro

andrei-stoian-zama:
Adds two macros:

- GPU_ASSERT : to be used for logic assumption checking | will be disabled in production
- PANIC_IF_ELSE: to be used to check caller provided values are sane | will run in production

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/device.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/device.cu
@@ -196,14 +196,14 @@ void cuda_memcpy_with_size_tracking_async_gpu_to_gpu(
     return;
   cudaPointerAttributes attr_dest;
   check_cuda_error(cudaPointerGetAttributes(&attr_dest, dest));
-  if (attr_dest.type != cudaMemoryTypeDevice) {
-    PANIC("Cuda error: invalid dest device pointer in copy from GPU to GPU.")
-  }
+  PANIC_IF_FALSE(
+      attr_dest.type == cudaMemoryTypeDevice,
+      "Cuda error: invalid dest device pointer in copy from GPU to GPU.");
   cudaPointerAttributes attr_src;
   check_cuda_error(cudaPointerGetAttributes(&attr_src, src));
-  if (attr_src.type != cudaMemoryTypeDevice) {
-    PANIC("Cuda error: invalid src device pointer in copy from GPU to GPU.")
-  }
+  PANIC_IF_FALSE(
+      attr_src.type == cudaMemoryTypeDevice,
+      "Cuda error: invalid src device pointer in copy from GPU to GPU.");
   cuda_set_device(gpu_index);
   if (attr_src.device == attr_dest.device) {
     check_cuda_error(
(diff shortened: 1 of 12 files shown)


backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -196,14 +196,14 @@ void cuda_memcpy_with_size_tracking_async_gpu_to_gpu(
...
+  PANIC_IF_FALSE(
+      attr_dest.type == cudaMemoryTypeDevice,
+      "Cuda error: invalid dest device pointer in copy from GPU to GPU.");
   cudaPointerAttributes attr_src;
   check_cuda_error(cudaPointerGetAttributes(&attr_src, src));
-  if (attr_src.type != cudaMemoryTypeDevice) {
-    PANIC("Cuda error: invalid src device pointer in copy from GPU to GPU.")
-  }
+  PANIC_IF_FALSE(
+      attr_dest.type == cudaMemoryTypeDevice,
```
andrei-stoian-zama: `attr_src` here

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
```diff
@@ -31,13 +31,13 @@ __host__ void host_expand_without_verification(
...
   auto d_lwe_compact_input_indexes = mem_ptr->d_lwe_compact_input_indexes;
   auto d_body_id_per_compact_list = mem_ptr->d_body_id_per_compact_list;
-  if (sizeof(Torus) == 8) {
-    cuda_lwe_expand_64(streams[0], gpu_indexes[0], expanded_lwes,
-                       lwe_flattened_compact_array_in, lwe_dimension, num_lwes,
-                       d_lwe_compact_input_indexes, d_body_id_per_compact_list);
 
-  } else
-    PANIC("Cuda error: expand is only supported on 64 bits")
+  GPU_ASSERT(sizeof(Torus) == 8,
```
enzodimaria, line 35: Everything looks perfect to me!

I was just wondering why using a GPU_ASSERT here instead of a PANIC_IF_FALSE?

andrei-stoian-zama: the principle behind ASSERT vs PANIC is given in the comments:

```

// The PANIC macro should be used to validate user-inputs to GPU functions
// it will execute in all targets, including production settings
// e.g., cudaMemCopy to the device should check that the destination pointer is
// a device pointer
```

```
// The GPU assert should be used to validate assumptions in algorithms,
// for example, checking that two user-provided quantities have a certain
// relationship or that the size of the buffer  provided to a function is
// sufficient when it is filled with some algorithm that depends on
// user-provided inputs e.g., OPRF corrections buffer should not have a size
// higher than the number of blocks in the datatype that is generated
```

Here we're checking some assumptions about the underlying datatype at compile-type because our cuda kernel only works on uint64. Thus this is not a run-time check to validate caller inputs to the function.

enzodimaria: Ok, I understand!

enzodimaria (approved):

PR #2635, opened 2025-08-06
chore: update check toolchain

IceTDrinker:


IceTDrinker:
nightly compiler is failing (actual panic/bug) on doctest checks for GPU

I guess we'll keep the old toolchain until we have something to manage avx512 attributes

PR #2643, opened 2025-08-08
fix(benches): fix hlapi dex benchmark transfer function

antoniupop:
Small fix: the type of ERC20 transfer was missed and using whitepaper instead of no-cmux in the no-cmux variant.


IceTDrinker (approved):
Thanks for the fix, looks good!

PR #2656, opened 2025-08-13
fix(core): use of deprecated rayon repeatn

nsarlin-zama:
currently the warning is blocking the ci: https://github.com/zama-ai/tfhe-rs/actions/runs/16932199296/job/47980172445?pr=2641

rayon 1.11 is needed because `repeat_n` does not exist in 1.10


IceTDrinker (approved):
Thanks !

PR #2664, opened 2025-08-18
feat(gpu): Implement 128-bit compression and add it to the integer API

pdroalves:
This PR also simplifies the call to 128-bit PBSs. Now we have a single entry point for all sizes.

The same underlying code is shared between 128 and 64-bit compression, only skipping the operations scalar multiplication and PBS in the former.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1067

### Check-list:

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664)

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
@@ -10,6 +10,7 @@ template <typename Torus> struct int_compression {
   Torus *tmp_lwe;
   Torus *tmp_glwe_array_out;
   bool gpu_memory_allocated;
+  uint32_t lwe_per_glwe;
 
   int_compression(cudaStream_t const *streams, uint32_t const *gpu_indexes,
                   uint32_t gpu_count, int_radix_params compression_params,
@@ -21,15 +22,16 @@ template <typename Torus> struct int_compression {
     uint64_t glwe_accumulator_size = (compression_params.glwe_dimension + 1) *
                                      compression_params.polynomial_size;
 
-    tmp_lwe = (Torus *)cuda_malloc_with_size_tracking_async(
+    tmp_lwe = static_cast<Torus *>(cuda_malloc_with_size_tracking_async(
         num_radix_blocks * (compression_params.small_lwe_dimension + 1) *
             sizeof(Torus),
-        streams[0], gpu_indexes[0], size_tracker, allocate_gpu_memory);
-    tmp_glwe_array_out = (Torus *)cuda_malloc_with_size_tracking_async(
-        lwe_per_glwe * glwe_accumulator_size * sizeof(Torus), streams[0],
-        gpu_indexes[0], size_tracker, allocate_gpu_memory);
+        streams[0], gpu_indexes[0], size_tracker, allocate_gpu_memory));
+    tmp_glwe_array_out =
+        static_cast<Torus *>(cuda_malloc_with_size_tracking_async(
+            lwe_per_glwe * glwe_accumulator_size * sizeof(Torus), streams[0],
+            gpu_indexes[0], size_tracker, allocate_gpu_memory));
 
-    size_tracker += scratch_packing_keyswitch_lwe_list_to_glwe_64(
+    size_tracker += scratch_packing_keyswitch_lwe_list_to_glwe<Torus>(
         streams[0], gpu_indexes[0], &fp_ks_buffer,
         compression_params.small_lwe_dimension,
         compression_params.glwe_dimension, compression_params.polynomial_size,
@@ -73,9 +75,6 @@ template <typename Torus> struct int_decompression {
     uint64_t lwe_accumulator_size = (compression_params.glwe_dimension *
                                          compression_params.polynomial_size +
                                      1);
-    decompression_rescale_lut = new int_radix_lut<Torus>(
-        streams, gpu_indexes, gpu_count, encryption_params, 1,
-        num_blocks_to_decompress, allocate_gpu_memory, size_tracker);
 
     tmp_extracted_glwe = (Torus *)cuda_malloc_with_size_tracking_async(
         num_blocks_to_decompress * glwe_accumulator_size * sizeof(Torus),
@@ -87,28 +86,38 @@ template <typename Torus> struct int_decompression {
         num_blocks_to_decompress * lwe_accumulator_size * sizeof(Torus),
         streams[0], gpu_indexes[0], size_tracker, allocate_gpu_memory);
 
-    // Rescale is done using an identity LUT
-    // Here we do not divide by message_modulus
-    // Example: in the 2_2 case we are mapping a 2 bits message onto a 4 bits
-    // space, we want to keep the original 2 bits value in the 4 bits space,
-    // so we apply the identity and the encoding will rescale it for us.
-    auto decompression_rescale_f = [](Torus x) -> Torus { return x; };
+    // rescale is only needed on 64-bit decompression
+    if constexpr (std::is_same_v<Torus, uint64_t>) {
+      decompression_rescale_lut = new int_radix_lut<Torus>(
+          streams, gpu_indexes, gpu_count, encryption_params, 1,
+          num_blocks_to_decompress, allocate_gpu_memory, size_tracker);
 
-    auto effective_compression_message_modulus =
-        encryption_params.carry_modulus;
-    auto effective_compression_carry_modulus = 1;
+      // Rescale is done using an identity LUT
+      // Here we do not divide by message_modulus
+      // Example: in the 2_2 case we are mapping a 2-bit message onto a 4-bit
+      // space, we want to keep the original 2-bit value in the 4-bit space,
+      // so we apply the identity and the encoding will rescale it for us.
+      decompression_rescale_lut = new int_radix_lut<Torus>(
+          streams, gpu_indexes, gpu_count, encryption_params, 1,
+          num_blocks_to_decompress, allocate_gpu_memory, size_tracker);
+      auto decompression_rescale_f = [](Torus x) -> Torus { return x; };
 
-    generate_device_accumulator_with_encoding<Torus>(
-        streams[0], gpu_indexes[0], decompression_rescale_lut->get_lut(0, 0),
-        decompression_rescale_lut->get_degree(0),
-        decompression_rescale_lut->get_max_degree(0),
-        encryption_params.glwe_dimension, encryption_params.polynomial_size,
-        effective_compression_message_modulus,
-        effective_compression_carry_modulus, encryption_params.message_modulus,
-        encryption_params.carry_modulus, decompression_rescale_f,
-        gpu_memory_allocated);
+      auto effective_compression_message_modulus =
+          encryption_params.carry_modulus;
+      auto effective_compression_carry_modulus = 1;
 
-    decompression_rescale_lut->broadcast_lut(streams, gpu_indexes);
+      generate_device_accumulator_with_encoding<Torus>(
+          streams[0], gpu_indexes[0], decompression_rescale_lut->get_lut(0, 0),
+          decompression_rescale_lut->get_degree(0),
+          decompression_rescale_lut->get_max_degree(0),
+          encryption_params.glwe_dimension, encryption_params.polynomial_size,
+          effective_compression_message_modulus,
+          effective_compression_carry_modulus,
+          encryption_params.message_modulus, encryption_params.carry_modulus,
+          decompression_rescale_f, gpu_memory_allocated);
+
+      decompression_rescale_lut->broadcast_lut(streams, gpu_indexes);
+    }
   }
   void release(cudaStream_t const *streams, uint32_t const *gpu_indexes,
                uint32_t gpu_count) {
@@ -118,9 +127,11 @@ template <typename Torus> struct int_decompression {
                                        gpu_indexes[0], gpu_memory_allocated);
     cuda_drop_with_size_tracking_async(tmp_indexes_array, streams[0],
                                        gpu_indexes[0], gpu_memory_allocated);
-
-    decompression_rescale_lut->release(streams, gpu_indexes, gpu_count);
-    delete decompression_rescale_lut;
+    if constexpr (std::is_same_v<Torus, uint64_t>) {
+      decompression_rescale_lut->release(streams, gpu_indexes, gpu_count);
+      delete decompression_rescale_lut;
+      decompression_rescale_lut = nullptr;
+    }
   }
 };
 #endif
--- a/backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cu
@@ -67,7 +67,6 @@ void cuda_integer_decompress_radix_ciphertext_64(
       (cudaStream_t *)(streams), gpu_indexes, gpu_count, lwe_array_out, glwe_in,
       indexes_array, bsks, (int_decompression<uint64_t> *)mem_ptr);
 }
-
 void cleanup_cuda_integer_compress_radix_ciphertext_64(
     void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
     int8_t **mem_ptr_void) {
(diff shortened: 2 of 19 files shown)


nsarlin-zama (commented):
some small comments on the rust part. ~~I think you should also update the high level API to make it accessible ? (in src/high_level_api/compressed_noise_squashed_ciphertext_list.rs)~~ Sorry I see that this is another PR !

> @nsarlin-zama reviewed 11 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-:-OYBkIjYB-F21sIBdZgA:b-gofhr6)* status: 11 of 29 files reviewed, 6 unresolved discussions (waiting on @agnesLeroy, @IceTDrinker, @mayeul-zama, and @tmontaigu)

___
*[`tfhe/src/integer/ciphertext/mod.rs` line 7 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OY7A4rF1LWz1M8hdrfg:-OY7A4rF1LWz1M8hdrfh:bg08s7t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/integer/ciphertext/mod.rs#L7)):*
> ```XML
> mod compressed_ciphertext_list;
> mod compressed_modulus_switched_ciphertext;
> pub mod compressed_noise_squashed_ciphertext_list;
> ```

is this needed ? I would be tempted to keep it similar to the `compressed_ciphertext_list` visibility-wise
___
*[`tfhe/src/integer/gpu/list_compression/server_keys.rs` line 520 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OYBi_kJ5vyPz6mHoNnx:-OYBi_kJ5vyPz6mHoNny:b-rsmksp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/integer/gpu/list_compression/server_keys.rs#L520)):*
> ```XML
>         let uncompressed_len = num_glwes * glwe_mask_size + num_lwes;
>         let ciphertext_modulus_log = ciphertext_modulus.into_modulus_log();
>         // The CPU implementation uses ciphertext_modulus_log instead of self.storage_log_modulus
> ```

Regarding this, when the cpu implem was done the ns compression AP did not include a modswitch, but I added a "no-op" modswitch (from 2^128 to 2^128) to be able to reuse code from the "normal" compression and be a bit more future proof.

In the future the noise squash compression might include a modswitch so we will have to update to add a `storage_log_modulus` param and use this.
___
*[`tfhe/src/shortint/list_compression/server_keys.rs` line 247 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OYBjmV-0dZzWlec4YiV:-OYBjmV-0dZzWlec4YiW:b-qfc9pd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/shortint/list_compression/server_keys.rs#L247)):*
> ```XML
> pub struct NoiseSquashingCompressionKey {
>     pub(crate) packing_key_switching_key: LwePackingKeyswitchKey<Vec<u128>>,
>     pub(crate) lwe_per_glwe: LweCiphertextCount,
> ```

instead of changing the visibility of the fields, it would be better to use the accessor method:
```
   compression_key.key.packing_key_switching_key()
```
___
*[`tfhe/src/integer/gpu/mod.rs` line 906 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OY7DAcr8TnCM3qIpGWm:-OY7DAcr8TnCM3qIpGWn:bzi23t0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/integer/gpu/mod.rs#L906)):*
> ```XML
> /// - [CudaStreams::synchronize] __must__ be called after this function as soon as synchronization
> ///   is required
> pub unsafe fn decompress_integer_radix_async<T: UnsignedInteger, B: Numeric>(
> ```

Does this need to be generic ? If only 64b is supported maybe it's better to remove the generic instead of having a runtime check ? Or maybe it also supports 32b ? (wondering because it calls _64 functions)
___
*[`tfhe/src/integer/gpu/ciphertext/compressed_noise_squashed_ciphertext_list.rs` line 32 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OY7E-l50WTkAEwY520t:-OY7E-l50WTkAEwY520u:b7oyf0j) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/integer/gpu/ciphertext/compressed_noise_squashed_ciphertext_list.rs#L32)):*
> ```XML
> impl Named for CudaCompressedSquashedNoiseCiphertextList {
>     const NAME: &'static str = "integer::gpu::CudaCompressedSquashedNoiseCiphertextList";
> }
> ```

This should not be needed if the type is not serializable
___
*[`tfhe/src/integer/gpu/ciphertext/compressed_noise_squashed_ciphertext_list.rs` line 34 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2664#-OY7EszJ1ZYSbhWaANXv:-OY7EszJ1ZYSbhWaANXw:b-8fwti8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/ffc8ddbb8a2ebd948c264dd18f8186468919da3c/tfhe/src/integer/gpu/ciphertext/compressed_noise_squashed_ciphertext_list.rs#L34)):*
> ```XML
> }
> 
> pub trait CudaSquashedCompressible {
> ```

I would name this `CudaSquashedNoiseCompressible` to be coherent with the other names

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cu
```diff
@@ -85,3 +83,80 @@ void cleanup_cuda_integer_decompress_radix_ciphertext_64(
...
+  mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
+}
+
+void cleanup_cuda_integer_decompress_radix_ciphertext_128(
+    void *const *streams, uint32_t const *gpu_indexes, uint32_t gpu_count,
+    int8_t **mem_ptr_void) {
+
+  int_decompression<__uint128_t> *mem_ptr =
+      (int_decompression<__uint128_t> *)(*mem_ptr_void);
+  mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
```
agnesLeroy, line 164: I think now we also systematically set released pointers to nullptr cc @andrei-stoian-zama (same for the other cleanup functions)

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
```diff
@@ -246,6 +255,9 @@ host_integer_decompress(cudaStream_t const *streams,
                         uint32_t const *h_indexes_array, void *const *d_bsks,
                         int_decompression<Torus> *h_mem_ptr) {
 
+  static_assert(std::is_same_v<Torus, uint64_t> ||
```
agnesLeroy, line 257: Now on main I think we should use GPU_ASSERT for this cc @andrei-stoian-zama

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
```diff
@@ -283,88 +295,111 @@ host_integer_decompress(cudaStream_t const *streams,
...
-        streams[0], gpu_indexes[0], extracted_lwe, extracted_glwe,
-        d_indexes_array_chunk, num_lwes, compression_params.polynomial_size,
-        compression_params.glwe_dimension, compression_params.polynomial_size);
+    if constexpr (std::is_same_v<Torus, uint64_t>)
+      cuda_glwe_sample_extract_64(streams[0], gpu_indexes[0], extracted_lwe,
+                                  extracted_glwe, d_indexes_array_chunk,
+                                  num_lwes, compression_params.polynomial_size,
+                                  compression_params.glwe_dimension,
+                                  compression_params.polynomial_size);
+    else
```
agnesLeroy, line 320: Here we should be exhaustive I think: check if Torus is uint128_t and have a case that calls PANIC if it's neither uint64_t or uint128_t

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
```diff
@@ -283,88 +295,111 @@ host_integer_decompress(cudaStream_t const *streams,
...
-        streams, gpu_indexes, active_gpu_count, (Torus *)d_lwe_array_out->ptr,
-        lut->lwe_indexes_out, lut->lut_vec, lut->lut_indexes_vec, extracted_lwe,
-        lut->lwe_indexes_in, d_bsks, nullptr, lut->buffer,
-        encryption_params.glwe_dimension,
-        compression_params.small_lwe_dimension,
-        encryption_params.polynomial_size, encryption_params.pbs_base_log,
-        encryption_params.pbs_level, encryption_params.grouping_factor,
-        num_blocks_to_decompress, encryption_params.pbs_type, num_many_lut,
-        lut_stride);
+  if constexpr (std::is_same_v<Torus, uint64_t>) {
```
agnesLeroy, line 333: Same here for the if/else, we should check explicitely it's uint128 and if not PANIC

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -2364,11 +2370,15 @@ __host__ void integer_radix_apply_noise_squashing_kb(
...
         ks_level, lwe_array_out->num_radix_blocks);
 
-    execute_pbs_128_async<__uint128_t>(
-        streams, gpu_indexes, active_gpu_count, lwe_after_pbs_vec, lut->lut_vec,
-        lwe_after_ks_vec, bsks, ms_noise_reduction_key, lut->pbs_buffer,
-        small_lwe_dimension, glwe_dimension, polynomial_size, pbs_base_log,
-        pbs_level, lwe_array_out->num_radix_blocks);
+    /// int_noise_squashing_lut doesn't support a different output or lut
+    /// indexing than the trivial
+    execute_pbs_async<uint64_t, __uint128_t>(
```
agnesLeroy: I'm not sure we need to have the two template types: the PBS128 always takes u64 as input and returns u128. This complicates a bit the code so I'm not sure we need it.

agnesLeroy (commented):
Thank you @pdroalves! I am not sure we want to make the changes to the way the PBS is called in this PR, it would be better to keep only the 128 bit compression changes here I believe.

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
pdroalves: Actually we don't need to check if it's 128, since we just skip the PBS in that case. We might want however to PANIC if it's 32 bits.

agnesLeroy: We should PANIC if it's anything else than 64 or 128 bits maybe?

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cu
pdroalves: Ok!

Is this refactoring still going on? I see it's like that in `oprf.cu` not not in, for instance, `bitwise_ops.cu` and `abs.cu`.

agnesLeroy: Ah indeed we haven't merged the change yet: still you can do it here so we don't forget it. The new pattern is:
```
mem_ptr->release((cudaStream_t *)(streams), gpu_indexes, gpu_count);
 delete mem_ptr;
 *mem_ptr_void = nullptr;
```

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
pdroalves: `static_assert` performs this check at compile time. `GPU_ASSERT` does not. I don't think it makes any sense to check this during execution, since we don't want that function to be compiled for `__uint128_t`.

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
pdroalves: `host_integer_decompress` only supports 64 or 128-bit Torus. That's why we have that `static_assert` at the beginning to assert it will not be compiled for anything different than that (as `uint32_t`). So we are kind of checking it exhaustively considering the two possibilities at this point :-P

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
pdroalves: Same answer as before. There are only two possibilities: Torus is `uint64_t` or `__uint128_t`. If it's the former the execute the PBS, otherwise we skip it.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
pdroalves: I moved this entire refactor to https://github.com/zama-ai/tfhe-rs/pull/2692 where we can better discuss this change. 128-bit decompression will not execute a PBS anyway, so there is no need to have this here. This was actually an artifact from when I started working on this and didn't know about this difference.

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
agnesLeroy: Hmm the static_assert is only in debug and here it's a release check so maybe we should have a PANIC instead of static_assert, or we could remove it altogether and do the checks exhaustively at this point?

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cu
pdroalves: Done.

backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
```diff
@@ -118,9 +127,10 @@ template <typename Torus> struct int_decompression {
                                        gpu_indexes[0], gpu_memory_allocated);
     cuda_drop_with_size_tracking_async(tmp_indexes_array, streams[0],
                                        gpu_indexes[0], gpu_memory_allocated);
-
-    decompression_rescale_lut->release(streams, gpu_indexes, gpu_count);
-    delete decompression_rescale_lut;
+    if constexpr (std::is_same_v<Torus, uint64_t>) {
+      decompression_rescale_lut->release(streams, gpu_indexes, gpu_count);
+      delete decompression_rescale_lut;
```
agnesLeroy, line 132: I think in the new pattern we should also do
decompression_rescale_lut = nullptr;

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
pdroalves: Not really. `static_assert()` and `assert()` are completely different. `static_assert()` will be executed even in release mode.

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
agnesLeroy: Ah totally thanks!

agnesLeroy (commented):
Just a minor thing left

backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
agnesLeroy: Yes sorry I was mistaking static_assert and assert

backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
pdroalves: Done.

agnesLeroy (approved):
Thank you very much @pdroalves!

PR #2668, opened 2025-08-20
chore(ci): add permission to github token to release crates

soonum:
When using crates.io trusted publishing feature GitHub token `id-token: write` permission to be able to authenticate the workflow on the registry.


nsarlin-zama (approved):

IceTDrinker:
need rayon repeat_n fix from Nico on main

PR #2672, opened 2025-08-21
fix(shortint): set correct degree for noise squashed decompressed ct

nsarlin-zama:


tfhe/src/shortint/ciphertext/compressed_ciphertext_list.rs
```diff
@@ -229,6 +229,9 @@ impl CompressedSquashedNoiseCiphertextList {
             extracted_lwe.lwe_ciphertext_mut(),
             monomial_degree,
         );
+        extracted_lwe.set_degree(Degree::new(
+            meta.message_modulus.0 * meta.carry_modulus.0 - 1,
```
IceTDrinker: consider using MaxDegree::from_msg_carry_modulus to compute the value and then do a get to put it in the Degree field

tfhe/src/shortint/list_compression/noise_squashing_compression.rs
```diff
@@ -188,11 +190,21 @@ mod test {
             // Compress the ciphertexts in a list
             let compressed = compression_key.compress_noise_squashed_ciphertexts_into_list(&ct);
 
-            // Extract from the list and decrypt
-            let decrypted_values =
-                compression_private_key.unpack_and_decrypt_squashed_noise_ciphertexts(&compressed);
+            let expected_degree = Degree(sks.message_modulus.0 * sks.carry_modulus.0 - 1);
```
IceTDrinker: same remark

tfhe/src/shortint/ciphertext/compressed_ciphertext_list.rs
nsarlin-zama: done

PR #2692, opened 2025-08-27
chore(gpu): refactor the entry points for PBS in the backend

pdroalves:
This PR refactors the entry points for the PBS in the backend. We once again will have a single entry point for any word size, as it was before for 32 and 64 bits now including 128 bits.

Because 128-bit PBS uses different input and output word sizes, `execute_pbs` now takes two template parameters (`InputTorus` and `OutputTorus`). This change breaks compatibility with the `GET_VARIANT_ELEMENT` macro that assumed a single type named `Torus`.

Thus, I replaced it by a inlined function. It:

- deduces the pointer type from the variant (no need to spell a Torus type explicitly),
- is easier to debug and read than a macro,
- is eligible to be inline, so no performance penalty.

### Check-list:


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -342,46 +412,66 @@ void execute_scratch_pbs(cudaStream_t stream, uint32_t gpu_index,
                          uint32_t level_count, uint32_t grouping_factor,
                          uint32_t input_lwe_ciphertext_count, PBS_TYPE pbs_type,
                          bool allocate_gpu_memory, bool allocate_ms_array,
-                         uint64_t &size_tracker) {
```
agnesLeroy, line 346: I think we should keep passing all size_tracker variables by reference, we had errors when passing pointers and we should follow the same pattern everywhere.

guillermo-oyarzun: agree with this, i feel less error prone to pass it by reference.

pdroalves: I think this was changed on a mistake done during some rebase. Sorry, I've rolled it back.

agnesLeroy (approved):
Thx!

PR #2716, opened 2025-09-01
feat(gpu): improve communication scheme

guillermo-oyarzun:

### PR content/description
This PR improves the communications:
- when involve non-trivial indexes.
- fixes multiplication
- broadcast luts only to the involved gpus.
- broadcast only lut indexes.
- remove useless synchronizations, use events when needed.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -324,6 +324,10 @@ template <typename Torus> struct int_radix_lut {
   uint32_t *gpu_indexes;
   bool gpu_memory_allocated;
 
+  cudaEvent_t event_scatter_in;
+  cudaEvent_t event_scatter_out[8];
+  cudaEvent_t event_broadcast;
+
   int_radix_lut(cudaStream_t const *streams, uint32_t const *input_gpu_indexes,
                 uint32_t gpu_count, int_radix_params params, uint32_t num_luts,
                 uint32_t num_radix_blocks, bool allocate_gpu_memory,
@@ -362,6 +366,15 @@ template <typename Torus> struct int_radix_lut {
       cuda_synchronize_stream(streams[i], gpu_indexes[i]);
       buffer.push_back(gpu_pbs_buffer);
     }
+
+    // if(active_gpu_count > 1){
+    event_scatter_in = cuda_create_event(gpu_indexes[0]);
+    event_broadcast = cuda_create_event(gpu_indexes[0]);
+
+    for (int i = 0; i < active_gpu_count; i++) {
+      event_scatter_out[i] = cuda_create_event(gpu_indexes[i]);
+    }
+    //}
 
     // Allocate LUT
     // LUT is used as a trivial encryption and must be initialized outside
@@ -579,7 +592,14 @@ template <typename Torus> struct int_radix_lut {
       cuda_synchronize_stream(streams[i], gpu_indexes[i]);
       buffer.push_back(gpu_pbs_buffer);
     }
+    // if(active_gpu_count > 1){
+    event_scatter_in = cuda_create_event(gpu_indexes[0]);
+    event_broadcast = cuda_create_event(gpu_indexes[0]);
 
+    for (int i = 0; i < active_gpu_count; i++) {
+      event_scatter_out[i] = cuda_create_event(gpu_indexes[i]);
+    }
+    //}
     // Allocate LUT
     // LUT is used as a trivial encryption and must be initialized outside
     // this constructor
@@ -718,9 +738,10 @@ template <typename Torus> struct int_radix_lut {
     auto src_lut = lut_vec[0];
     auto src_lut_indexes = lut_indexes_vec[0];
 
-    cuda_synchronize_stream(streams[0], gpu_indexes[0]);
+    cuda_event_record(event_broadcast, streams[0], gpu_indexes[0]);
     for (uint i = 0; i < active_gpu_count; i++) {
       if (gpu_indexes[i] != gpu_indexes[0]) {
+        cuda_stream_wait_event(streams[i], event_broadcast, gpu_indexes[i]);
         auto dst_lut = lut_vec[i];
         auto dst_lut_indexes = lut_indexes_vec[i];
         cuda_memcpy_with_size_tracking_async_gpu_to_gpu(
@@ -745,7 +766,13 @@ template <typename Torus> struct int_radix_lut {
       cuda_drop_with_size_tracking_async(lut_indexes_vec[i], streams[i],
                                          gpu_indexes[i], gpu_memory_allocated);
     }
-
+    // if( active_gpu_count > 1) {
+    for (uint i = 0; i < active_gpu_count; i++) {
+      cuda_event_destroy(event_scatter_out[i], gpu_indexes[i]);
+    }
+    cuda_event_destroy(event_scatter_in, gpu_indexes[0]);
+    cuda_event_destroy(event_broadcast, gpu_indexes[0]);
+    //}
     cuda_drop_with_size_tracking_async(lwe_indexes_in, streams[0],
                                        gpu_indexes[0], gpu_memory_allocated);
     cuda_drop_with_size_tracking_async(lwe_indexes_out, streams[0],
@@ -1614,6 +1641,7 @@ template <typename Torus> struct int_sum_ciphertexts_vec_memory {
     this->allocated_luts_message_carry = false;
     this->reduce_degrees_for_single_carry_propagation =
         reduce_degrees_for_single_carry_propagation;
+
     setup_index_buffers(streams, gpu_indexes, size_tracker);
     // because we setup_lut in host function for sum_ciphertexts to save memory
     // the size_tracker is topped up here to have a max bound on the used memory
@@ -1661,6 +1689,7 @@ template <typename Torus> struct int_sum_ciphertexts_vec_memory {
     this->current_blocks = current_blocks;
     this->small_lwe_vector = small_lwe_vector;
     this->luts_message_carry = reused_lut;
+
     setup_index_buffers(streams, gpu_indexes, size_tracker);
   }
 
@@ -5909,6 +5938,9 @@ template <typename Torus> struct int_count_of_consecutive_bits_buffer {
                                    allocate_gpu_memory);
     delete ct_prepared;
     ct_prepared = nullptr;
+
+    propagate_mem->release(streams, gpu_indexes, gpu_count);
+    delete propagate_mem;
 
     prepare_mem->release(streams, gpu_indexes, gpu_count);
     delete prepare_mem;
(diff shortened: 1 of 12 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -359,10 +366,22 @@ template <typename Torus> struct int_radix_lut {
       if (i == 0) {
         size_tracker += size;
       }
-      cuda_synchronize_stream(streams[i], gpu_indexes[i]);
+      // cuda_synchronize_stream(streams[i], gpu_indexes[i]);
```
agnesLeroy: We could remove the line instead of commenting. Why wan we avoid synchronization here?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -382,7 +401,7 @@ template <typename Torus> struct int_radix_lut {
       lut_vec.push_back(lut);
       lut_indexes_vec.push_back(lut_indexes);
 
-      cuda_synchronize_stream(streams[i], gpu_indexes[i]);
+      // cuda_synchronize_stream(streams[i], gpu_indexes[i]);
```
agnesLeroy: Same here

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -559,7 +578,6 @@ template <typename Torus> struct int_radix_lut {
 
     ///////////////
     active_gpu_count = get_active_gpu_count(num_radix_blocks, gpu_count);
-    cuda_synchronize_stream(streams[0], gpu_indexes[0]);
```
agnesLeroy, line 562: Why can we void synchronizing here?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -711,54 +734,108 @@ template <typename Torus> struct int_radix_lut {
...
+      }
+      // Ensure the device set at the end of this method is the same as it was
+      // set at the beginning
+      cuda_set_device(active_device);
+    }
+  }
+  // Broadcast luts from device gpu_indexes[0] to all active gpus
+  void broadcast_lut(cudaStream_t const *streams, uint32_t const *gpu_indexes,
+                     uint32_t new_active_gpu_count,
+                     bool broadcast_lut_values = true) {
```
agnesLeroy, line 761: We usually avoid optional arguments because we had many problems with them: one is fine but when we add more it's a mess. So maybe here I would not set true by default, but set it in every call that needs it.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -711,54 +734,108 @@ template <typename Torus> struct int_radix_lut {
...
+                                uint64_t &size_tracker,
+                                bool allocate_gpu_memory) {
+    this->non_trivial_setup = true;
+    // We need to create the auxiliary array only in GPU 0
+    lwe_aligned_vec.resize(active_gpu_count);
+    for (uint i = 0; i < active_gpu_count; i++) {
+      uint64_t size_tracker_on_array_i = 0;
+      auto inputs_on_gpu = std::max(
+          THRESHOLD_MULTI_GPU,
+          get_num_inputs_on_gpu(max_num_radix_blocks, i, active_gpu_count));
```
agnesLeroy, line 803: Why compute this max here? Shouldn't get_num_inputs_on_gpu already return either the threshold or a higher value? If not we should probably modify its logic.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -309,6 +309,8 @@ template <typename Torus> struct int_radix_lut {
   Torus *h_lwe_indexes_out;
   // Enable optimizations if lwe_indexes_(in/out) are trivial
   bool using_trivial_lwe_indexes = true;
+  // To enforce that the non-trivial setup is done
```
agnesLeroy: Here I would be a bit more explicit in naming and in the comment:
```
// This boolean tracks whether non trivial indexes are used for the PBS.
// For example this is the case for the multiplication and expand.
bool non_trivial_indexes = false;
```

agnesLeroy: But this seems redundand with using_trivial_lwe_indexes: couldn't we just reuse that boolean?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -711,54 +734,108 @@ template <typename Torus> struct int_radix_lut {
...
+              streams[i], gpu_indexes[i], gpu_memory_allocated);
+        }
       }
+      // Ensure the device set at the end of this method is the same as it was
+      // set at the beginning
+      cuda_set_device(active_device);
+    }
+  }
+
+  void setup_non_trivial_vector(cudaStream_t const *streams,
```
agnesLeroy: I would name this function `allocate_lwe_vector_for_non_trivial_indexes` maybe?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -1446,9 +1572,11 @@ template <typename Torus> struct int_overflowing_sub_memory {
...
         f_message_acc, gpu_memory_allocated);
 
-    luts_array->broadcast_lut(streams, gpu_indexes);
-    luts_borrow_propagation_sum->broadcast_lut(streams, gpu_indexes);
-    message_acc->broadcast_lut(streams, gpu_indexes);
+    auto active_gpu_count = get_active_gpu_count(num_radix_blocks, gpu_count);
+    luts_array->broadcast_lut(streams, gpu_indexes, active_gpu_count);
+    luts_borrow_propagation_sum->broadcast_lut(streams, gpu_indexes,
+                                               active_gpu_count);
+    message_acc->broadcast_lut(streams, gpu_indexes, active_gpu_count);
```
agnesLeroy, line 1568: Do they all have the same number of blocks? Is it ok to use the same active_gpu_count for all three?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -1744,8 +1882,9 @@ template <typename Torus> struct int_seq_group_prop_memory {
     cuda_memcpy_with_size_tracking_async_to_gpu(
         seq_lut_indexes, h_seq_lut_indexes, num_seq_luts * sizeof(Torus),
         streams[0], gpu_indexes[0], allocate_gpu_memory);
-
-    lut_sequential_algorithm->broadcast_lut(streams, gpu_indexes);
+    auto active_gpu_count = get_active_gpu_count(num_seq_luts, gpu_count);
```
agnesLeroy, line 1874: Is it really num_seq_luts here? Doesn't look like a number of lwes

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -1104,8 +1136,8 @@ void generate_device_accumulator_bivariate_with_factor(
       h_lut, glwe_dimension, polynomial_size, message_modulus, carry_modulus, f,
       factor);
 
-  cuda_synchronize_stream(stream, gpu_index);
-  // copy host lut and lut_indexes_vec to device
+  // cuda_synchronize_stream(stream, gpu_index);
```
agnesLeroy: We can remove the line instead of commenting

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: The problem is that the non trivial vector is not set by default (we need to do it manually), so i wanted to keep a variable that checks if it has been set or not, so we can panic in the gather/scatter if this is not setup.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: I followed the logic of the multi_gpu_alloc_array_async, the only difference is that in here we need to allocate them all in the same gpu

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: yes it is, so the rule i was following is that checking how the lut has been created before, and use the same number of blocks. Here we create it like this, indeed it seems an strange initialization, but it is part of the carry prop. 

lut_sequential_algorithm = new int_radix_lut<Torus>(
        streams, gpu_indexes, gpu_count, params, num_seq_luts, num_seq_luts,
allocate_gpu_memory, size_tracker);

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: i will remove the comment, we can remove the synchronization (maybe won't win much) because at the time it arrives to the push the pointer is already known since we get them from the cpu.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: we can remove it here cause each iteration uses a different stream and everything is executed within the gpu (if we had some interaction with cpu would be different), otherwise we block the cpu thread and the other gpus and seriallize its allocations.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: Yup the main problem is that the number of active gpus might change during the operations so it can not be calculated only once and stored in lut. I choose to kept default true, cause it involved less changes, since only few broadcast calls need to communicate only indexes.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: yup, i just double checked and yes, all have the same.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: Ok sounds reasonable

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: Oh ok, then it was my bad when I introduced the threshold :sweat_smile: we should maybe try to refactor this later to make it so that get_num_inputs_on_gpu returns max(threshold, calculated_inputs)

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
guillermo-oyarzun: we can remove it cause then each gpu/stream allocates independently, so there is no need to wait for the gpu0

agnesLeroy (approved):
Looks great, thank you very much!

PR #2718, opened 2025-09-01
feat(gpu): stream pools

andrei-stoian-zama:
Adds a stream pool per GPU. Stream creation requests are served from the pool. the streams in the pool are never destroyed


agnesLeroy:
Should we close this one?

PR #2720, opened 2025-09-01
fix(gpu):  change mininum number of elements in benches

guillermo-oyarzun:

### PR content/description
The idea is to have a mininum workload based only in the number of radix blocks and specs on gpu.
The number of sms is now read from the gpu instead of using a hardcoded value
### Check-list:

### DIFF
--- a/tfhe-benchmark/src/utilities.rs
+++ b/tfhe-benchmark/src/utilities.rs
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 use std::sync::OnceLock;
 use std::{env, fs};
 #[cfg(feature = "gpu")]
-use tfhe::core_crypto::gpu::get_number_of_gpus;
+use tfhe::core_crypto::gpu::{get_number_of_gpus, get_number_of_sms};
 use tfhe::core_crypto::prelude::*;
 
 #[cfg(feature = "boolean")]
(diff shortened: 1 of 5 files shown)


agnesLeroy (approved):
Thank you very much @guillermo-oyarzun!

tfhe-benchmark/src/utilities.rs
```diff
@@ -431,11 +427,19 @@ pub fn throughput_num_threads(num_block: usize, op_pbs_count: u64) -> u64 {
 
     #[cfg(feature = "gpu")]
     {
-        let total_num_sm = H100_PCIE_SM_COUNT * get_number_of_gpus();
+        let num_sms_per_gpu = get_number_of_sms();
+        let total_num_sm = num_sms_per_gpu * get_number_of_gpus();
+
+        let total_blocks_per_sm = 4u32; // Assume each SM can handle 4 blocks concurrently
+        let total_num_sm = total_blocks_per_sm * total_num_sm;
+        let min_num_waves = 4u64; //Enforce at least 4 waves in the GPU
```
agnesLeroy, line 435: I guess you measured for the bootstrap, and 4 is enough to saturate the GPU, have you tried with more just to check?

guillermo-oyarzun: yup i tried until 6 waves and i was getting the same values than with 4 (217K pbs with 8 gpus), i haven't checked yet how this could affect the integer ops in terms of execution time, i want to test that before merging

agnesLeroy: Yes, there's also keyswitch actually, probably for that one 4 is a bit low

guillermo-oyarzun: actually with keyswitch/keybundle we saturate the GPU before, because we launch many cuda blocks per sample, and also before this fallback option was 200 and now is ~2000. My concern is more the integer ops might take longer in the sense that this new min doesn't consider the number of pbs, but i believe the previous formula would handle those. It is actually hard to have something that covers everything, that's why i tried to change only the fallback part of the formula.

agnesLeroy: Ok, then for integers we should be good I think, the previous formula should always apply and if not this new threshold should be enough.

PR #2724, opened 2025-09-03
chore(csprng): use getrandom as random source for unix seeder

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1112 https://github.com/zama-ai/tfhe-rs-internal/issues/1113

Use [getrandom](https://docs.rs/getrandom/latest/getrandom/index.html) for the unix seeder. This crate will use if possible the `getrandom`/`getentropy` system call which is resilient to fd exhaustion attacks, and is the recommended way to draw system entropy on modern unix systems.


IceTDrinker (commented):
some questions, the ne_bytes comment is just a curiosity of mine given we were trying to make things equivalent between both small and big endian worlds

> @IceTDrinker reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2724#-:-OZEmXC2CHbHroEZq5mN:b-cvhraf)* status: all files reviewed, 2 unresolved discussions (waiting on @mayeul-zama and @tmontaigu)

___
*[`tfhe-csprng/src/seeders/implem/unix.rs` line 40 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2724#-OZEm7-S35d5b-4Wf5rE:-OZEm7-S35d5b-4Wf5rF:b5wkbs1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/721b6b36b573cd9cf0070e6ea54baecc8966cfa1/tfhe-csprng/src/seeders/implem/unix.rs#L40)):*
> ```XML
>     ///
>     /// # Panics
>     /// This may panic if the `getrandom` system call is not available and no file descriptor is
> ```

still an issue with fd then ?
___
*[`tfhe-csprng/src/seeders/implem/unix.rs` line 61 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2724#-OZEmJpf8lIWfUaPhQc2:-OZEmJpf8lIWfUaPhQc3:b-f2dwj6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/721b6b36b573cd9cf0070e6ea54baecc8966cfa1/tfhe-csprng/src/seeders/implem/unix.rs#L61)):*
> ```XML
>     // seeded. See <https://www.2uo.de/myths-about-urandom/> for a rational around random/urandom.
>     getrandom::getrandom(&mut buf).expect("Failed to read entropy from system");
>     u128::from_ne_bytes(buf)
> ```

don't think this survives endianness differences if the pools returned the same bytes

PR #2727, opened 2025-09-03
fix(gpu): add missing broadcast lut

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -1671,8 +1671,6 @@ template <typename Torus> struct int_sum_ciphertexts_vec_memory {
             new int_radix_lut<Torus>(streams, gpu_indexes, gpu_count, params, 2,
                                      pbs_count, true, size_tracker);
         allocated_luts_message_carry = true;
-        auto active_gpu_count =
-            get_active_gpu_count(this->max_total_blocks_in_vec, gpu_count);
         luts_message_carry->allocate_lwe_vector_for_non_trivial_indexes(
             streams, gpu_indexes, gpu_count, this->max_total_blocks_in_vec,
             size_tracker, true);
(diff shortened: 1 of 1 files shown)


agnesLeroy:
Axtually that's because counter_num_blocks is always smaller than our threshold for multi-GPU in the tests we have, so we don't see the problem in the CI. Benches on 256 bit integers are bound to fail for ilog2 though.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -6019,9 +6017,7 @@ template <typename Torus> struct int_prepare_count_of_consecutive_bits_buffer {
         params.glwe_dimension, params.polynomial_size, params.message_modulus,
         params.carry_modulus, generate_bi_lut_lambda, allocate_gpu_memory);
 
-    if (allocate_gpu_memory) {
```
agnesLeroy, line 6022: We never have these if conditions in the rest of the code and I think it's fine, because broadcast lut will only copy anything if the memory was allocated. We'll still be recording the events, but I don't think that'll cause a problem in the case where we're only trying to get the buffer size.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -6334,6 +6330,8 @@ template <typename Torus> struct int_ilog2_buffer {
         this->lut_message_not->get_max_degree(0), params.glwe_dimension,
         params.polynomial_size, params.message_modulus, params.carry_modulus,
         lut_message_lambda, allocate_gpu_memory);
+    auto active_gpu_count = get_active_gpu_count(counter_num_blocks, gpu_count);
```
enzodimaria, line 6333: I forgot these ones, indeed! Thank you

enzodimaria (approved):

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
enzodimaria: I see!
I had some doubt when I wrote the code, that's why I decided to add these checks

guillermo-oyarzun (approved):

PR #2730, opened 2025-09-04
Am/fix/confusing endianness api

IceTDrinker:
`Seed(pub u128)` exposes endianness to the outside world such that a user wanting to use seed 1u128 will have different byte layouts on big/small endian machines

The internal conversions were using bytes in native endian order, fix the conversion for the software generator in the big endian case by using u128::from_le when instantiating an AesKey which itself is oblivious to endianness and should actually be a [u8; 16]

Change the get_system_entropy function to give proper endianness to the bytes that end up in the `Seed(pub u128)` for the same reason.

### DIFF
--- a/tfhe-csprng/src/seeders/implem/unix.rs
+++ b/tfhe-csprng/src/seeders/implem/unix.rs
@@ -58,7 +58,9 @@ fn get_system_entropy() -> u128 {
     // This syscall will use the urandom entropy source but block at startup until it is correctly
     // seeded. See <https://www.2uo.de/myths-about-urandom/> for a rational around random/urandom.
     getrandom::getrandom(&mut buf).expect("Failed to read entropy from system");
-    u128::from_ne_bytes(buf)
+    // For consistency between big and small endian, Seed exposing accidentally the endianness via
+    // the pub u128 field
+    u128::from_le_bytes(buf)
 }
 
 #[cfg(test)]


tfhe-csprng/src/generators/aes_ctr/block_cipher.rs
```diff
@@ -1,8 +1,11 @@
 use crate::generators::aes_ctr::{AES_CALLS_PER_BATCH, BYTES_PER_AES_CALL, BYTES_PER_BATCH};
 
 /// Represents a key used in the AES block cipher.
+///
+/// The u128 endianness should be ignored by implementations and the u128 should be seen as a simple
```
nsarlin-zama, line 5: I think maybe this comment could be a bit more explicit, telling if implementations should handle this in native, big or little endian

tmontaigu (approved):
> @tmontaigu reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2730)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker)

tfhe-csprng/src/generators/aes_ctr/block_cipher.rs
IceTDrinker: fair

IceTDrinker:
fixed the remark from Nico

nsarlin-zama (approved):
thanks!

IceTDrinker:
will run m1 tests as well just to make sure all is well

PR #2731, opened 2025-09-04
fix(gpu): coprocessor bench

andrei-stoian-zama:
- Add custom machine when running the bench manually
- Cuda Profile in throughput mode using manual profiler start on set device

### DIFF
--- a/.github/workflows/coprocessor-benchmark-gpu.yml
+++ b/.github/workflows/coprocessor-benchmark-gpu.yml
@@ -3,6 +3,22 @@ name: coprocessor-benchmark-gpu
 
 on:
   workflow_dispatch:
+    inputs:
+      profile:
+        description: "Instance type"
+        required: true
+        type: choice
+        options:
+          - "l40 (n3-L40x1)"
+          - "4-l40 (n3-L40x4)"
+          - "single-h100 (n3-H100x1)"
+          - "2-h100 (n3-H100x2)"
+          - "4-h100 (n3-H100x4)"
+          - "multi-h100 (n3-H100x8)"
+          - "multi-h100-nvlink (n3-H100x8-NVLink)"
+          - "multi-h100-sxm5 (n3-H100x8-SXM5)"
+          - "multi-h100-sxm5_fallback (n3-H100x8-SXM5)"
+
   schedule:
     # Weekly tests @ 1AM
     - cron: "0 1 * * 6"
@@ -17,7 +33,9 @@ env:
   RUST_BACKTRACE: "full"
   RUST_MIN_STACK: "8388608"
   CHECKOUT_TOKEN: ${{ secrets.REPO_CHECKOUT_TOKEN || secrets.GITHUB_TOKEN }}
-  PROFILE: "multi-h100-sxm5 (n3-H100x8-SXM5)"
+  PROFILE_SCHEDULED_RUN: "multi-h100-sxm5 (n3-H100x8-SXM5)"
+  PROFILE_MANUAL_RUN: ${{ inputs.profile }}
+  IS_MANUAL_RUN: ${{ github.event_name == 'workflow_dispatch' }}
   BENCHMARK_TYPE: "ALL"
   OPTIMIZATION_TARGET: "throughput"
   BATCH_SIZE: "5000"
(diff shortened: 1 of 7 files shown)


backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -5,6 +5,9 @@
 uint32_t cuda_get_device() {
   int device;
   check_cuda_error(cudaGetDevice(&device));
+#ifdef USE_NVTOOLS
+  cudaStartProfile();
+#endif
```
agnesLeroy: Is it ok to call this many times if there are multiple threads calling cuda_get_device? Perhaps doing this like the mempool activation thing would be better? Mempool is merged on main already.

.github/workflows/coprocessor-benchmark-gpu.yml
```diff
@@ -40,15 +54,25 @@ jobs:
       - name: Parse profile
         id: parse_profile
         run: |
+          if [[ ${{ github.event_name == 'workflow_dispatch' }} == true ]]; then
+            PROFILE_RAW="${{ inputs.profile }}"
+          else
+            PROFILE_RAW="${PROFILE}"
```
agnesLeroy: Here you should set PROFILE_RAW to "multi-h100-sxm5 (n3-H100x8-SXM5)" no?

agnesLeroy (commented):
Thank you @andrei-stoian-zama for this change, just some questions :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/src/device.cu
andrei-stoian-zama: The [cuda doc](https://docs.nvidia.com/cuda/archive/9.1/cuda-runtime-api/group__CUDART__PROFILER.html#group__CUDART__PROFILER_1gf536d75bb382356e10e3b4e89f4a5374) says that after the first call to this function, subsequent calls have no effect. Hopefully this is true

agnesLeroy: Ok let's have faith for once :slightly_smiling_face:

.github/workflows/coprocessor-benchmark-gpu.yml
```diff
@@ -17,7 +33,7 @@ env:
   RUST_BACKTRACE: "full"
   RUST_MIN_STACK: "8388608"
   CHECKOUT_TOKEN: ${{ secrets.REPO_CHECKOUT_TOKEN || secrets.GITHUB_TOKEN }}
-  PROFILE: "multi-h100-sxm5 (n3-H100x8-SXM5)"
+  PROFILE_UNATTENDED_RUN: "multi-h100-sxm5 (n3-H100x8-SXM5)"
```
agnesLeroy: Maybe PROFILE_SCHEDULDED_RUN?

ci/slab.toml
```diff
@@ -106,3 +106,9 @@ environment_name = "canada"
 image_name = "Ubuntu Server 22.04 LTS R570 CUDA 12.8"
 flavor_name = "n3-RTX-A6000x1"
 user = "ubuntu"
+
+[backend.hyperstack.4-l40]
+environment_name = "canada"
+image_name = "Ubuntu Server 22.04 LTS R570 CUDA 12.8"
+flavor_name = "n3-L40x4"
+user = "ubuntu"
```
agnesLeroy, line 102: Hmm we already have a profile for this
```
[backend.hyperstack.multi-gpu-test]
environment_name = "canada"
image_name = "Ubuntu Server 22.04 LTS R570 CUDA 12.8"
flavor_name = "n3-L40x4"
user = "ubuntu"
```
although it's badly named. I would recommend using the existing profile here, and in another PR renaming it everywhere (the name dates back to when we only had very few profiles in this file).

.github/workflows/coprocessor-benchmark-gpu.yml
andrei-stoian-zama: it was a bug before, it uses PROFILE_SCHEDULED_RUN now

ci/slab.toml
andrei-stoian-zama: just a few renames were needed, so I did it in this PR

Makefile
```diff
@@ -1004,6 +1004,11 @@ test_high_level_api_gpu: install_rs_build_toolchain install_cargo_nextest
 		--test-threads=4 --features=integer,internal-keycache,gpu,zk-pok -p tfhe \
 		-E "test(/high_level_api::.*gpu.*/)"
 
+.PHONY: test_high_level_build
+test_high_level_build: install_rs_build_toolchain
```
agnesLeroy: Maybe the name should be
test_high_level_api_build_gpu?

andrei-stoian-zama: I went with `test_high_level_build_one_gpu`

agnesLeroy: Why `one`?

PR #2734, opened 2025-09-04
fix(gpu): add missing syncs when releasing scalar ops and returning t…

guillermo-oyarzun:

### PR content/description
Scalar bitop and shifts require a sync when releasing to avoid the corner case when they return without executing anything
Rolledback the refactor in the release of the int_radix_lut
Add missing syncs
### Check-list:


guillermo-oyarzun:
The problem was happening in some scalar operations, some of them check the scalar input an according to that decide to not execute anything, in those case the release was executed right after the scratch, and some pointer erased while still in use. Extra synchronizations were added to the release, it should have rather minor impact in performance.
I rollback the changes in the release of the int_radix_lut just to be extra safe, and added a couple of syncs before we remove additional streams too.

agnesLeroy (approved):

guillermo-oyarzun:
I launched long run tests just to double check, if passes we can merge

PR #2735, opened 2025-09-05
fix(zk): handle limit cases in the four_squares algorithm

nsarlin-zama:

### PR content/description
Fix issue with limit cases in the four_square algorithm. Before this pr, the algorithm would panic/overflow if `v = 2*4^n` (for any n) and run into an infinite loop if `v = k*4^n with k in [6, 10, 14, 22, 34, 38, 110]`.

Added a test to check these limit values

### DIFF
--- a/tfhe-zk-pok/src/four_squares.rs
+++ b/tfhe-zk-pok/src/four_squares.rs
@@ -1,6 +1,8 @@
 use ark_ff::biginteger::arithmetic::widening_mul;
 use rand::prelude::*;
 
+use crate::proofs::ProofSanityCheckMode;
+
 /// Avoid overflows for squares of u64
 pub fn sqr(x: u64) -> u128 {
     let x = x as u128;
@@ -188,21 +190,24 @@ impl Montgomery {
     }
 }
 
-pub fn four_squares(v: u128) -> [u64; 4] {
+pub fn four_squares(v: u128, sanity_check_mode: ProofSanityCheckMode) -> [u64; 4] {
     let rng = &mut StdRng::seed_from_u64(0);
 
-    // In the extreme case where the noise is exactly at the bound, v is 0
-    if v == 0 {
-        return [0; 4];
-    }
+    // Handle limit cases that would trigger an infinite loop
+    match v {
+        0 => return [0; 4],
+        2 => return [1, 1, 0, 0],
+        6 => return [2, 1, 1, 0],
+        _ => {}
+    };
 
     let f = v % 4;
     if f == 2 {
         let b = v.isqrt() as u64;
 
         'main_loop: loop {
-            let x = 2 + rng.gen::<u64>() % (b - 2);
-            let y = 2 + rng.gen::<u64>() % (b - 2);
+            let x: u64 = rng.gen_range(0..=b);
+            let y: u64 = rng.gen_range(0..=b);
 
             let (sum, o) = u128::overflowing_add(sqr(x), sqr(y));
             if o || sum > v {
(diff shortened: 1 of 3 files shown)


tfhe-zk-pok/src/four_squares.rs
```diff
@@ -191,18 +191,21 @@ impl Montgomery {
...
 
     let f = v % 4;
     if f == 2 {
         let b = v.isqrt() as u64;
 
         'main_loop: loop {
-            let x = 2 + rng.gen::<u64>() % (b - 2);
-            let y = 2 + rng.gen::<u64>() % (b - 2);
+            let x: u64 = rng.gen_range(0..=b);
+            let y: u64 = rng.gen_range(0..=b);
```
tmontaigu, line 210: With the changes i believe x and y dont have the same range,
before the range was [2, b] now its [0, b], is it intentional ?

nsarlin-zama: yes if we keep the original range we have to manually handle the v = [6, 10, 14, 22, 34, 38, 110] in the input match because it will trigger an infinite loop.
With [0, b], the algo finds solutions for v = [10, 14, 22, 34, 38, 110] so we only have to have a special case for v = 6. (this was suggested by Olivier from labs):


> Une autre façon de faire semble être de tirer x, y entre 0 et B (inclus) au lieu de (2, B-1) et d'uniquement gérer en entrée les cas V = 0,2 ou 6.

tfhe-zk-pok/src/four_squares.rs
```diff
@@ -272,19 +275,79 @@ pub fn four_squares(v: u128) -> [u64; 4] {
...
+
+        // Sort so that r[0], r[1] are even and r[2], r[3] are odd,
+        // with r[1] < r[0] and r[3] < r[2]
         r.sort_by_key(|&x| {
             if x % 2 == 0 {
+                // -1 is used to distinguish between 0 and 1
                 -1 - ((x / 2) as i64)
             } else {
                 (x / 2) as i64
             }
```
mayeul-zama: Could simply be `if x % 2 == 0 { -x } else {x}` I think

tfhe-zk-pok/src/four_squares.rs
```diff
@@ -272,19 +275,79 @@ pub fn four_squares(v: u128) -> [u64; 4] {
     } else if f == 0 {
         four_squares(v / 4).map(|x| x + x)
     } else {
-        let mut r = four_squares(2 * v);
+        let mut r = four_squares(v.checked_mul(2).unwrap());
```
mayeul-zama: Could add a comment to explain that we look for a solution for `2 * v` and deduce from it a solution for v

And add a proof of this deduction (maybe have it before the return location)

tfhe-zk-pok/src/four_squares.rs
mayeul-zama: Or replace r[0] and r[1] in the return expression and order by `(x % 2 != 0, x)`

tfhe-zk-pok/src/four_squares.rs
```diff
@@ -272,19 +275,79 @@ pub fn four_squares(v: u128) -> [u64; 4] {
...
+        const RAND_TESTS_COUNT: usize = 1000;
+
+        let seed = thread_rng().gen();
+        println!("four_squares seed: {seed:x}");
+        let rng = &mut StdRng::seed_from_u64(seed);
+
+        let limit_values: [u128; 255] = core::array::from_fn(|i| i as u128);
+        for val in limit_values {
+            assert_four_squares(val);
+        }
```
mayeul-zama: Could simply be 
```
        for val in 0..255 {
            assert_four_squares(val as u128);
        }
```

tfhe-zk-pok/src/four_squares.rs
nsarlin-zama: This is not my code, at first I made the same assumption but I guess that the / 2 is to avoids the overflow when casting to i64.

tfhe-zk-pok/src/four_squares.rs
nsarlin-zama: yes of course !

nsarlin-zama:
the test that is currently failing is one where the algorithm is provided with incorrect values on purpose, I am looking for a way to make it pass

tfhe-zk-pok/src/four_squares.rs
mayeul-zama: Then let's cast to i128 :)

tfhe-zk-pok/src/four_squares.rs
nsarlin-zama: I don't have this proof at hand and the goal is to fix the bug quickly to unblock the kms, will add a small comment

tfhe-zk-pok/src/four_squares.rs
```diff
@@ -270,21 +275,80 @@ pub fn four_squares(v: u128) -> [u64; 4] {
...
+        r.sort_by_key(|&x| (x % 2 != 0, x));
+
         [
-            (r[0] + r[1]) / 2,
-            (r[0] - r[1]) / 2,
-            (r[3] + r[2]) / 2,
+            // divide by 2 before addition to avoid overflows
+            (r[1] / 2 + r[0] / 2),
+            (r[1] - r[0]) / 2,
+            (r[3] / 2 + r[2] / 2) + 1,
```
IceTDrinker, line 307: +1 part of the R&D code ?

Do we have test values that trigger all code paths in that function ? at least the branching on the f value

nsarlin-zama: > +1 part of the R&D code ?

The R&D code is:
```python
return ((R[1]+R[0])//2, (R[1]-R[0])//2, (R[3]+R[2])//2, (R[3]-R[2])//2)
```
however that's python, in rust it may overflow on the addition. To remove this overflow possibility I did `a/2+b/2` instead of `(a+b)/2`. We know that r0 and r1 are even, and r2 and r3 are odd. So for r2 and r3 I need to add +1 to the result.

> Do we have test values that trigger all code paths in that function ? at least the branching on the f value

Yes the added test will at least tests all v between 0 and 255, since `f = v % 4` it will trigger all branches

IceTDrinker: perfect

PR #2751, opened 2025-09-08
fix(gpu): avoid out of memory when benchmarking throughput

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/tfhe-benchmark/src/utilities.rs
+++ b/tfhe-benchmark/src/utilities.rs
@@ -421,23 +421,32 @@ pub fn throughput_num_threads(num_block: usize, op_pbs_count: u64) -> u64 {
     let block_multiplicator = (ref_block_count as f64 / num_block as f64).ceil().min(1.0);
     // Some operations with a high serial workload (e.g. division) would yield an operation
     // loading value so low that the number of elements in the end wouldn't be meaningful.
-    let minimum_loading = if num_block < 64 { 0.2 } else { 0.01 };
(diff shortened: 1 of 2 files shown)


tfhe-benchmark/src/utilities.rs
```diff
@@ -437,12 +437,18 @@ pub fn throughput_num_threads(num_block: usize, op_pbs_count: u64) -> u64 {
         let total_num_sm = total_blocks_per_sm * total_num_sm;
         let min_num_waves = 4u64; //Enforce at least 4 waves in the GPU
         let elements_per_wave = total_num_sm as u64 / (num_block as u64);
-
+        // This should ensure that operations with PBS count more than the number of blocks
+        // squared will have only half of the minimum elements to avoid out of memory errors.
+        let min_elements = if op_pbs_count > num_block as u64 * num_block as u64 {
+            (elements_per_wave * min_num_waves / 2) as u64
```
agnesLeroy: We should check if this is enough to have all the "all precision" bench pass, because ilog2 uses more memory than mul I think.

guillermo-oyarzun: yup, i am trying to run the benchmarks now

guillermo-oyarzun: it failed before that, i will try with 1/3, otherwise i will return to our original 200 default in those cases, since we know that one worked well

agnesLeroy (approved):

PR #2757, opened 2025-09-09
chore(ci): run backward compat tests on merge to main

IceTDrinker:
- this is to prime cache and check backward data on merge to main


soonum (approved):
Nice 🎉

PR #2759, opened 2025-09-09
fix(zk): add a size check for the public key

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1129

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2759)


tfhe-zk-pok/src/proofs/mod.rs
```diff
@@ -154,6 +154,8 @@ pub(crate) enum ProofSanityCheckMode {
 /// does not hold.
 #[allow(clippy::too_many_arguments)]
 fn assert_pke_proof_preconditions(
+    a: &[i64],
```
IceTDrinker, line 157: not checked below ?

nsarlin-zama: fixed!

PR #2760, opened 2025-09-09
chore(gpu): refactor the indexing logic for the LWE expand

pdroalves:
This PR is based on https://github.com/zama-ai/tfhe-rs/pull/2623. It simplifies the internal logic for the kernel that executes LWE expand. This is done by wrapping up the input components in structs that store information on where everything lies in global memory. By doing this, we can get rid of all the complicated pointer logic in the kernel and rewrite it in a simpler way on the CPU side. This should not significantly affect performance since the memory access pattern is the same. We add some additional instructions on the CPU side plus one copy to global memory, but that copy should be quite small and negligible in terms of latency.

As additional discussion points for the team:
- I haven't added a scratch function specific to the LWE expand. We run the initialization operations in host_expand_without_verification(), since these are specific to each input. I still run memory allocations only in scratch_cuda_expand_without_verification, though. If we ever need to expose the LWE expand to the Rust side again, we need to rethink that.
- In the aforementioned new copy to global memory, we spend space copying the monomial degree for each body type. I think that was being computed inside the kernel and now needs to be fetched from memory. I don't think that's bad since it is information specific to each body, and by writing it this way everything becomes less error-prone. However, we need to double-check if this doesn't break alignment.
- I haven't included the LWE output in expand_job as @IceTDrinker suggested. That would increase the memory footprint for something that is completely not needed. The other elements replace the copy of many information arrays to the device, but we had nothing related to the output array. We could simplify the way we handle that pointer by introducing a macro if we want to simplify things without paying in performance. For instance:

```c++
auto output_mask = &lwe_array_out[(lwe_dimension + 1) * lwe_output_id];
```

could become

```c++
auto output_mask = &GET_LWE_FROM_ARRAY(lwe_array_out, lwe_dimension, *lwe_output_id);
```

I didn't do it in this PR since it goes completely against what we do in the rest of the backend, but it could be a direction for a bigger refactoring.

BTW, thanks a lot @IceTDrinker for the ideas in this PR. We should surely apply them on other parts of the backend.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1058

### Check-list:

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2760)


backends/tfhe-cuda-backend/build.rs
```diff
@@ -81,6 +81,7 @@ fn main() {
             "cuda/include/ciphertext.h",
             "cuda/include/integer/compression/compression.h",
             "cuda/include/integer/integer.h",
+            "cuda/include/integer/radix_ciphertext.h",
```
agnesLeroy: Why do you add radix_ciphertext.h here? I don't think it's necessary?

backends/tfhe-cuda-backend/cuda/include/zk/zk.h
```diff
@@ -5,14 +5,13 @@
 #include "../pbs/pbs_enums.h"
 #include <stdint.h>
 
-extern "C" {
+#include "zk_utilities.h"
 
 void cuda_lwe_expand_64(void *const stream, uint32_t gpu_index,
```
agnesLeroy: If this function is never called from Rust, how about only having a `host_lwe_expand_64 in zk.cuh`? That's how we do things usually.

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
```diff
@@ -35,9 +35,33 @@ __host__ void host_expand_without_verification(
...
+      h_expand_jobs.push_back(job);
+    }
+  }
+
+  cuda_memcpy_with_size_tracking_async_to_gpu(
+      d_expand_jobs, h_expand_jobs.data(),
+      h_expand_jobs.size() * sizeof(expand_job<Torus>), streams[0],
+      gpu_indexes[0], true);
+
+  cuda_lwe_expand_64(streams[0], gpu_indexes[0], expanded_lwes, d_expand_jobs,
```
agnesLeroy: Here we could simply call `host_lwe_expand` and remove `cuda_lwe_expand_64`.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -272,6 +384,9 @@ template <typename Torus> struct zk_expand_mem {
     cuda_drop_with_size_tracking_async(tmp_ksed_small_to_big_expanded_lwes,
                                        streams[0], gpu_indexes[0],
                                        gpu_memory_allocated);
+    cuda_drop_with_size_tracking_async(d_expand_jobs, streams[0],
+                                       gpu_indexes[0], gpu_memory_allocated);
```
agnesLeroy, line 326: We should synchronize here before calling free to make sure we're not still copying CPU data to the GPU when this free is called.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -22,6 +115,10 @@ template <typename Torus> struct zk_expand_mem {
   uint32_t *d_body_id_per_compact_list;
   bool gpu_memory_allocated;
 
+  uint32_t *num_lwes_per_compact_list;
+  expand_job<Torus> *d_expand_jobs;
+  std::vector<expand_job<Torus>> h_expand_jobs;
```
agnesLeroy: If we're copying h_expand_job to d_expand_job at some point we should not use std::vector. We can't control when these are free'd on the CPU and thus we can't make sure we synchronize before it. This should be a pointer that we allocate and free manually instead, and we should call synchronize before the free.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -272,6 +384,9 @@ template <typename Torus> struct zk_expand_mem {
     cuda_drop_with_size_tracking_async(tmp_ksed_small_to_big_expanded_lwes,
                                        streams[0], gpu_indexes[0],
                                        gpu_memory_allocated);
+    cuda_drop_with_size_tracking_async(d_expand_jobs, streams[0],
+                                       gpu_indexes[0], gpu_memory_allocated);
+    free(num_lwes_per_compact_list);
   }
 };
```
agnesLeroy, line 331: I don't see the removal of the complex indexing section in this file, I was expecting to have it be deleted no?

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
```diff
@@ -35,9 +35,33 @@ __host__ void host_expand_without_verification(
...
+    for (auto lwe_index = 0; lwe_index < list.total_num_lwes; ++lwe_index) {
+      auto job = expand_job(list.get_mask(), list.get_body(lwe_index));
+      h_expand_jobs.push_back(job);
+    }
+  }
+
+  cuda_memcpy_with_size_tracking_async_to_gpu(
+      d_expand_jobs, h_expand_jobs.data(),
+      h_expand_jobs.size() * sizeof(expand_job<Torus>), streams[0],
+      gpu_indexes[0], true);
```
agnesLeroy: Yes so since we're copying h_expand_jobs to GPU here this really shouldn't be a std::vector.

backends/tfhe-cuda-backend/build.rs
pdroalves: Actually I was debugging something I thought was broken in the way we build bindings.rs and then I found this. Don't we need to add this line to check for changes in radix_ciphertext.h?

backends/tfhe-cuda-backend/cuda/include/zk/zk.h
pdroalves: `host_lwe_expand_64` takes a `class param` parameter as template. I use `cuda_lwe_expand_64` to handle it instead of having that huge switch clause inside zk.cuh.

I almost removed that file until I realized this :-/

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
pdroalves: Of course! Thanks.

backends/tfhe-cuda-backend/build.rs
agnesLeroy: No, we don't generate Rust bindings from that file :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
pdroalves: It makes sense. I will refactor to replace the vector by an array.

backends/tfhe-cuda-backend/cuda/include/zk/zk.h
agnesLeroy: Does it really have to take the param as template? If that's so, it would still be better to add this template to host_expand_without_verification and have the polynomial size choice be made in for expand_without_verification.

IceTDrinker (commented):
two comments/question on the indexing logic

> @IceTDrinker reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2760#-:-OZnRcXLCw9F6xqzTy-w:b9ju5om)* status: 5 of 6 files reviewed, 9 unresolved discussions (waiting on @agnesLeroy and @pdroalves)

___
*[`backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h` line 13 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2760#-OZmqraW4hVvLWb3nAi2:-OZmqraW4hVvLWb3nAi3:bekii2y) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/12bcadf2c452f20375eb0f28f2b2b78462f1b2cf/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h#L13)):*
> ```objc
>   Torus *mask;
> 
>   lwe_mask(Torus *mask, uint32_t lwe_dimension)
> ```

missing lwe_dimension info in the struct ?
___
*[`backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h` line 69 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2760#-OZnR2xc023r05jGW7T4:-OZnR2xc023r05jGW7T5:bwirc6n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/12bcadf2c452f20375eb0f28f2b2b78462f1b2cf/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h#L69)):*
> ```objc
>     for (auto i = 0; i < num_compact_lists; ++i) {
>       total_num_lwes += h_num_lwes_per_compact_list[i];
>       if (i + 1 < num_compact_lists)
> ```

is the i+1 required ?

you could likely have

```cpp
curr_list = d_ptr;

for (auto i = 0; i < num_compact_lists; ++i) {
    d_ptr_to_compact_list[i] = curr_list;
    // TODO manage pointer arithmetic correctly 
    curr_list += lwe_dimension = h_num_lwes_per_compact_list[i];
}
```

and I think it should work ?

backends/tfhe-cuda-backend/build.rs
pdroalves: Oh, so I will rollback this.

backends/tfhe-cuda-backend/cuda/include/zk/zk.h
pdroalves: Yeah, we really need it. `lwe_expand` rellies on `polynomial_accumulate_monic_monomial_mul` , which takes `params::opt` as input.

But the idea to move the switch on polynomial_size to `cuda_expand_without_verification_64` is really good. Thanks! I will do it.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
pdroalves: Totally! I forgot about that. I was able to remove `d_lwe_compact_input_indexes` and `d_body_id_per_compact_list` from `zk_expand_mem`. We still need the LWE indexes, though.

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
pdroalves: Done.

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
pdroalves: Done.

backends/tfhe-cuda-backend/cuda/include/zk/zk.h
agnesLeroy: I think now you should be able to remove this function altogether and use only the host_ one

pdroalves: Yes! That's what I did.

agnesLeroy: Strange I still see `void cuda_lwe_expand_64(void *const stream, uint32_t gpu_index, ...` in the github diff here, as if it's not removed.

agnesLeroy: Yes even locally on your branch it's still there in this file:
```
void cuda_lwe_expand_64(void *const stream, uint32_t gpu_index,
                        void *lwe_array_out, expand_job<uint64_t> *d_jobs,
                        uint32_t num_lwes, uint32_t lwe_dimension);
```

pdroalves: It was still in the header. Sorry.

PR #2767, opened 2025-09-11
chore(gpu): multi-gpu debug target

andrei-stoian-zama:
This PR adds a fake multi-gpu target that is only used for debugging and has no impact on normal compilation.

- Add a fake-multi gpu compilation config to enable multi-gpu debugging on single gpu machine. 
   - compiles Rust code in Debug without overflow panics
   - compiles C++ code in Debug
   - compiles GPU kernels in Release to avoid very slow computation preventing debugging
- New short run GPU tests that can be run on a single GPU laptop.
- Small refactoring of long run tests
- Makefile target to build a specific HL test and the new Short run tests in fake-multi-gpu to allow easy debugging in an IDE

### DIFF
--- a/tfhe/src/core_crypto/gpu/mod.rs
+++ b/tfhe/src/core_crypto/gpu/mod.rs
@@ -36,8 +36,31 @@ pub enum PBSMSNoiseReductionType {
 
 impl CudaStreams {
     /// Create a new `CudaStreams` structure with as many GPUs as there are on the machine
+    #[cfg(feature = "gpu-debug-fake-multi-gpu")]
+    pub fn new_multi_gpu() -> Self {
+        let gpu_count = setup_multi_gpu(GpuIndex::new(0)) + 3;
+        assert_eq!(
+            gpu_count, 4,
+            "The fake multi-gpu debug target can only be used on single GPU machines"
+        );
+
+        let mut gpu_indexes = Vec::with_capacity(gpu_count as usize);
+        let mut ptr_array = Vec::with_capacity(gpu_count as usize);
+
+        for _ in 0..gpu_count {
+            ptr_array.push(unsafe { cuda_create_stream(0) });
+            gpu_indexes.push(GpuIndex::new(0));
+        }
+        Self {
+            ptr: ptr_array,
+            gpu_indexes,
+        }
+    }
+
+    #[cfg(not(feature = "gpu-debug-fake-multi-gpu"))]
     pub fn new_multi_gpu() -> Self {
         let gpu_count = setup_multi_gpu(GpuIndex::new(0));
+
         let mut gpu_indexes = Vec::with_capacity(gpu_count as usize);
         let mut ptr_array = Vec::with_capacity(gpu_count as usize);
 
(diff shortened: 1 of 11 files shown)


tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -36,20 +36,46 @@ pub enum PBSMSNoiseReductionType {
...
         let gpu_count = setup_multi_gpu(GpuIndex::new(0));
+
         let mut gpu_indexes = Vec::with_capacity(gpu_count as usize);
         let mut ptr_array = Vec::with_capacity(gpu_count as usize);
 
         for i in 0..gpu_count {
             ptr_array.push(unsafe { cuda_create_stream(i) });
             gpu_indexes.push(GpuIndex::new(i));
         }
+        println!("FAKE MULTI");
```
andrei-stoian-zama: remove this

tfhe/src/core_crypto/gpu/mod.rs
```diff
@@ -36,20 +36,46 @@ pub enum PBSMSNoiseReductionType {
 
 impl CudaStreams {
     /// Create a new `CudaStreams` structure with as many GPUs as there are on the machine
+    #[cfg(feature = "gpu-debug-fake-multi-gpu")]
```
guillermo-oyarzun, line 39: I think it will be good to add a comment here saying that the point here is to simulate a multi-gpu environment having a single GPU

tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
```diff
@@ -25,9 +25,23 @@ fn seeded_shuffle(v: &mut [u32], datagen: &mut DeterministicSeeder<DefaultRandom
 fn make_random_gpu_set(datagen: &mut DeterministicSeeder<DefaultRandomGenerator>) -> CudaGpuChoice {
     // Sample a random subset of 1-N gpus, where N is the number of available GPUs
     // A GPU index should not appear twice in the subset
+    #[cfg(not(feature = "gpu-debug-fake-multi-gpu"))]
     let num_gpus = get_number_of_gpus();
-    //let mut rng = rand::thread_rng();
-    let num_gpus_to_use = (1 + datagen.seed().0 as u32 % (num_gpus - 1)) as usize;
+
+    #[cfg(feature = "gpu-debug-fake-multi-gpu")]
+    let num_gpus_to_use = 2;
```
guillermo-oyarzun, line 32: this is the short long test? why do we use only 2 gpus, is it because of mem constrains?

guillermo-oyarzun (commented):
Looks good to me, I would suggest to add a page in notion explaining it and how to use it

tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
andrei-stoian-zama: yes, 10 operations types x 2 (fake) gpus takes ~6GB on the GPU . we can assume 8GB on debug machines . I can run any number of operations (by default 200) with this setup.

PR #2771, opened 2025-09-12
chore(ci): run cargo audit

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1132

### PR content/description
Not sure if this is really needed with dependabot, but it is fast and we can run it locally if we want.

I ignored the warning about paste being unmaintained because:
- paste is not considered broken but "finished" by its author
- alternatives are from lesser known authors, switch to an alternative would be more risky security wise
- it would be complicated to remove it, and it is also used by some of our dependencies anyway

I could also have left the warnings about unmaintained as "non deny" but this way at least we have to be explicit about if we accept the risk or not on a case by case basis. This can be changed later if it becomes too tedious.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771)

### DIFF
--- a/.github/workflows/cargo_audit.yml
+++ b/.github/workflows/cargo_audit.yml
@@ -1,0 +1,40 @@
+# Run cargo audit
+on:
+  workflow_dispatch:
+  schedule:
+    # runs every day at 4am UTC
+    - cron: '0 4 * * *'
+
+env:
+  CARGO_TERM_COLOR: always
+  ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  CHECKOUT_TOKEN: ${{ secrets.REPO_CHECKOUT_TOKEN || secrets.GITHUB_TOKEN }}
+  SLACK_CHANNEL: ${{ secrets.SLACK_CHANNEL }}
+  SLACK_ICON: https://pbs.twimg.com/profile_images/1274014582265298945/OjBKP9kn_400x400.png
+  SLACK_USERNAME: ${{ secrets.BOT_USERNAME }}
+  SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+  SLACKIFY_MARKDOWN: true
+
+permissions: {}
+
+jobs:
+  audit:
+    name: cargo_audit/audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+        with:
+          persist-credentials: 'false'
+          token: ${{ env.CHECKOUT_TOKEN }}
+
+      - name: Audit dependencies
+        run: |
+          make audit_dependencies
+
+      - name: Slack Notification
+        if: ${{ failure() }}
+        continue-on-error: true
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661
+        env:
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_MESSAGE: "cargo-audit finished with status: ${{ job.status }}. ([action run](${{ env.ACTION_RUN_URL }}))"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -163,6 +163,10 @@ install_tarpaulin: install_rs_build_toolchain
 install_cargo_dylint:
 	cargo install --locked cargo-dylint dylint-link
 
+.PHONY: install_cargo_audit # Check dependencies
+install_cargo_audit:
```
IceTDrinker: likely requires a toolchain to be installed so you can depend on one of the target, either stable or nightly toolchain, to install cargo

.github/workflows/cargo_audit.yml
```diff
@@ -0,0 +1,32 @@
+# Run cargo audit
+name: cargo_audit
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 4 * * *'
```
IceTDrinker, line 6: don't hesitate to put a comment explaining how the cronjob is configured

.github/workflows/cargo_audit.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          make audit_dependencies
+
+      - name: Slack Notification
+        if: ${{ failure() }}
+        continue-on-error: true
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661
+        env:
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_MESSAGE: "cargo-audit finished with status: ${{ job.status }}. (${{ env.ACTION_RUN_URL }})"
```
IceTDrinker: don't hesitate to check `SLACKIFY_MARKDOWN` in other workflows to easily populate the notification message

Makefile
nsarlin-zama: done

.github/workflows/cargo_audit.yml
nsarlin-zama: done

.github/workflows/cargo_audit.yml
nsarlin-zama: done

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @soonum)

___
*[`.github/workflows/cargo_audit.yml` line 40 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771#-O_CoNRkBOR7GXJ13147:-O_CsXyy8HvTXhnI7DRZ:bnza0ts) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/230ce449f34e48adbd2cb960bc84c833c1c2b78b/.github/workflows/cargo_audit.yml#L40)):*

see e.g. .github/workflows/gpu_unsigned_integer_tests.yml:169
</blockquote></details>

but if it works like that it's fine I don't recall exactly what the issue was with the slackify stuff, @soonum might know

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @IceTDrinker and @soonum)

___
*[`.github/workflows/cargo_audit.yml` line 40 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771#-O_CoNRkBOR7GXJ13147:-O_CtS4B8H1mbALAkNAq:b-w5yfot) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/230ce449f34e48adbd2cb960bc84c833c1c2b78b/.github/workflows/cargo_audit.yml#L40)):*

but if it works like that it's fine I don't recall exactly what the issue was with the slackify stuff, @soonum might know
</blockquote></details>

Ah understood! I looked at the notifications for teardown instance and they do not use this.

IceTDrinker (approved):
Thanks !

> @IceTDrinker reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2771#-:-O_D58sI6l69ec2nZHQZ:b-hay2j5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @soonum)

PR #2773, opened 2025-09-15
feat(integer): add KVStore compression and serialization

tmontaigu:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2773)

### DIFF
--- a/tfhe/src/integer/backward_compatibility/ciphertext/mod.rs
+++ b/tfhe/src/integer/backward_compatibility/ciphertext/mod.rs
@@ -5,6 +5,7 @@ use crate::integer::ciphertext::{
     CompressedModulusSwitchedSignedRadixCiphertext, DataKind, SquashedNoiseBooleanBlock,
     SquashedNoiseRadixCiphertext, SquashedNoiseSignedRadixCiphertext,
 };
+use crate::integer::server_key::CompressedKVStore;
 use crate::integer::BooleanBlock;
 #[cfg(feature = "zk-pok")]
 use crate::integer::ProvenCompactCiphertextList;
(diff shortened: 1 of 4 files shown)


tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2773)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @nsarlin-zama)

___
*[`tfhe/src/integer/server_key/radix_parallel/kv_store.rs` line 456 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2773#-O_MsJQf4jndooPa457G:-O_NJ9lB48G7d-383SRD:b-7evjnv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/31eda18cc50354f97065b03b5edcc8132c8e72c6/tfhe/src/integer/server_key/radix_parallel/kv_store.rs#L456)):*

I don't know if it can simply be fixed in this pr, but it is to be noted that this could allow type confusion since the type Key is not part of the name. Ideally we would find a way to encode this in the NAME
</blockquote></details>

YEs it can be 'fixed' in this PR, but also things like FheUint<Id>/FheInt<Id> do not have the Id part in the string so I guess its not that much of a problem

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/kv_store.rs` line 456 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2773#-O_MsJQf4jndooPa457G:-O_SIazvCnnON6ZRha2g:b-xjk3po) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/31eda18cc50354f97065b03b5edcc8132c8e72c6/tfhe/src/integer/server_key/radix_parallel/kv_store.rs#L456)):*

Maybe we can start with an impl for the few ones that are the most likely to be used and then add more later when they are needed ?
On the long run I guess it would make sense to have a named impl for the default types in std so they can be used with `safe_serialize` (I believe Vec was requested at some point)
</blockquote></details>

I added impl for the most common types if keys

Adding impl for std libs is great, but they wont be progamatically composable since a static is required by the trait

PR #2774, opened 2025-09-15
fix(ci): fix serde root crate in tfhe-lints

nsarlin-zama:
closes: 

### PR content/description
serde moved the root crate for the serialize trait from serde to serde_core, so it was not detected anymore


IceTDrinker (approved):
Thanks !

PR #2776, opened 2025-09-16
chore(bench): add crs size in zk-pke benchmark names

soonum:
This is done get more details about the benchmarks when parsing results.


IceTDrinker (approved):
Thanks !

PR #2782, opened 2025-09-16
chore(ci): check if regression workflow should be run early

soonum:
Before, any issue comment or label event would trigger the verify-actor job. Then the next job, prepare-benchmarks, would check if the rest of the workflow should run. Moving this very check in verify-actor ensures the whole workflow to run only if required.


IceTDrinker (approved):
Thanks looks good !

PR #2785, opened 2025-09-17
chore(ci): fix issue_comment trigger event for regression bench

soonum:
One limitation of GitHub, is the fact that `issue_comment` event **only** run on default branch.
So we need to merge this to see if it fixes the triggering issue.


IceTDrinker (approved):
lesgo, you may encounter the wasm-bindgen issue, we should have a fix out fairly soon though

PR #2786, opened 2025-09-17
fix(ci): use precise wasm-bindgen version for the cli

nsarlin-zama:

### PR content/description
wasm bindgen requires an exact match between the cli and the lib versions.
The wasm-bindgen-cli was installed using the version in the Cargo.toml but it might not always match the one that is used as a dependency (due to semver).
Instead we use cargo tree to detect the version to install.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,7 @@ BENCH_CUSTOM_COMMAND:=
 NODE_VERSION=22.6
 BACKWARD_COMPAT_DATA_DIR=utils/tfhe-backward-compat-data
 WASM_PACK_VERSION="0.13.1"
-# We are kind of hacking the cut here, the version cannot contain a quote '"'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -121,7 +120,7 @@ install_cargo_nextest: install_rs_build_toolchain
 # If a version range is specified no need to add the leading ^
 .PHONY: install_wasm_bindgen_cli # Install wasm-bindgen-cli to get access to the test runner
 install_wasm_bindgen_cli: install_rs_build_toolchain
-	cargo +$(RS_BUILD_TOOLCHAIN) install --locked wasm-bindgen-cli --version "$(WASM_BINDGEN_VERSION)"
+	cargo +$(RS_BUILD_TOOLCHAIN) install --locked wasm-bindgen-cli --version "^$(WASM_BINDGEN_VERSION)"
```
IceTDrinker: actually this seems like we don't want that here ? since cargo tree uses the lockfile and we need the exact same version ?

nsarlin-zama: yes you're right, I removed it

IceTDrinker (approved):

PR #2789, opened 2025-09-17
feat(integer): add oprf over any range

mayeul-zama:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)

### DIFF
--- a/tfhe/src/core_crypto/commons/math/random/mod.rs
+++ b/tfhe/src/core_crypto/commons/math/random/mod.rs
@@ -29,7 +29,7 @@ pub use uniform_binary::*;
 pub use uniform_ternary::*;
 
 #[cfg(test)]
-mod tests;
+pub(crate) mod tests;
 
 mod gaussian;
 mod generator;
--- a/tfhe/src/core_crypto/commons/math/random/tests.rs
+++ b/tfhe/src/core_crypto/commons/math/random/tests.rs
@@ -4,6 +4,8 @@ use crate::core_crypto::commons::math::random::{Distribution, RandomGenerable, T
 use crate::core_crypto::commons::math::torus::UnsignedTorus;
 use crate::core_crypto::commons::numeric::{CastFrom, CastInto, Numeric, UnsignedInteger};
 use crate::core_crypto::commons::test_tools::*;
+use itertools::Itertools;
+use std::ops::AddAssign;
 
 fn test_normal_random_three_sigma<T: UnsignedTorus>() {
     //! test if the normal random generation with std_dev is below 3*std_dev (99.7%)
@@ -432,12 +434,18 @@ fn dkw_cdf_bands_width(number_of_samples: usize, confidence_interval: f64) -> f6
     // https://en.wikipedia.org/wiki/Dvoretzky%E2%80%93Kiefer%E2%80%93Wolfowitz_inequality#Building_CDF_bands
     // the true CDF is between the empirical CDF +/- this band width with probability 1 - alpha
     // Said otherwise, the abs diff should be less than that value with high probability
-    fn dkw_cdf_bands_width_formula(sample_size: f64, alpha: f64) -> f64 {
-        f64::sqrt(f64::ln(2.0 / alpha) / (2.0 * sample_size))
-    }
 
     // alpha = 1 - probability of being in the interval
     dkw_cdf_bands_width_formula(number_of_samples as f64, 1.0 - confidence_interval)
+}
+
+pub fn dkw_cdf_bands_width_formula(sample_size: f64, alpha: f64) -> f64 {
+    f64::sqrt(f64::ln(2.0 / alpha) / (2.0 * sample_size))
+}
+
+// https://en.wikipedia.org/wiki/Dvoretzky%E2%80%93Kiefer%E2%80%93Wolfowitz_inequality#The_DKW_inequality
+pub fn dkw_alpha_from_epsilon(sample_size: f64, epsilon: f64) -> f64 {
+    2.0 * (-epsilon * epsilon * (2.0 * sample_size)).exp()
 }
 
 fn test_random_from_distribution_custom_mod<Scalar, D>(
@@ -490,43 +498,22 @@ fn test_random_from_distribution_custom_mod<Scalar, D>(
             }
         }
 
-        let mut cumulative_sums = vec![0u64; distinct_values];
+        let cumulative_bins = cumulate(&bins);
 
-        let mut curr_sum = 0;
-
-        // Compute the cumulative sums
-        for (bin_count, cum_sum) in bins.iter().zip(cumulative_sums.iter_mut()) {
-            curr_sum += bin_count;
-            *cum_sum = curr_sum;
-        }
+        let theoretical_cdf: Vec<f64> = (0..distinct_values)
+            .map(|bin_idx| {
+                let integer_value: Scalar =
+                    distribution.map_usize_to_value(bin_idx, ciphertext_modulus);
+                // CDF for the uniform distribution
+                distribution.cumulative_distribution_function(integer_value, ciphertext_modulus)
+            })
+            .collect();
 
         // Inaccurate if modulus >~ 2^53 / number_of_samples_per_bin, but if that's the case your
         // memory most likely blew up before (or the universe died its heat death)
         let number_of_samples = NUMBER_OF_SAMPLES_PER_VALUE * distinct_values;
 
-        let sup_diff: f64 = cumulative_sums
-            .iter()
-            .copied()
-            .enumerate()
-            .map(|(bin_idx, x)| {
-                // Compute the observed CDF
-                let empirical_cdf = x as f64 / number_of_samples as f64;
-
-                let integer_value: Scalar =
-                    distribution.map_usize_to_value(bin_idx, ciphertext_modulus);
-                // CDF for the uniform distribution
-                let theoretical_cdf = distribution
-                    .cumulative_distribution_function(integer_value, ciphertext_modulus);
-
-                if theoretical_cdf == 1.0 {
-                    assert_eq!(empirical_cdf, 1.0);
-                }
-
-                let diff = empirical_cdf - theoretical_cdf;
-                diff.abs()
-            })
-            .max_by(f64::total_cmp)
-            .unwrap();
+        let sup_diff = sup_diff(&cumulative_bins, &theoretical_cdf);
 
         let upper_bound_for_cdf_abs_diff =
             dkw_cdf_bands_width(number_of_samples, CONFIDENCE_INTERVAL);
@@ -543,6 +530,36 @@ fn test_random_from_distribution_custom_mod<Scalar, D>(
         nok_ratio <= EXPECTED_NOK_RATE_WITH_TOLERANCE,
         "nok_ratio={nok_ratio}"
     );
+}
+
+pub fn sup_diff(cumulative_bins: &[u64], theoretical_cdf: &[f64]) -> f64 {
+    let number_of_samples = *cumulative_bins.last().unwrap();
+
+    cumulative_bins
+        .iter()
+        .copied()
+        .zip_eq(theoretical_cdf.iter().copied())
+        .map(|(x, theoretical_cdf)| {
+            let empirical_cdf = x as f64 / number_of_samples as f64;
+
+            if theoretical_cdf == 1.0 {
+                assert_eq!(empirical_cdf, 1.0);
+            }
+
+            let diff = empirical_cdf - theoretical_cdf;
+            diff.abs()
+        })
+        .max_by(f64::total_cmp)
+        .unwrap()
+}
+
+pub fn cumulate<T: AddAssign + Default + Copy>(bins: &[T]) -> Vec<T> {
+    bins.iter()
+        .scan(T::default(), |sum, x| {
+            *sum += *x;
+            Some(*sum)
+        })
+        .collect()
 }
 
 impl<Scalar: UnsignedInteger + CastFrom<usize> + CastInto<usize>> DistributionTestHelper<Scalar>
--- a/tfhe/src/integer/oprf.rs
+++ b/tfhe/src/integer/oprf.rs
@@ -40,7 +40,6 @@ impl ServerKey {
 
         let sk = &self.key;
 
-        assert!(self.message_modulus().0.is_power_of_two());
         let message_bits_count = self.message_modulus().0.ilog2() as u64;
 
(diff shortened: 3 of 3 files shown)


agnesLeroy:
@mayeul-zama this is a change we should also do on GPU right?

tmontaigu (changes requested):
> @tmontaigu reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker)

___
*[`tfhe/src/integer/oprf.rs` line 245 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XKqIICTLSRlUq7BHc:-O_XKqIJ5zOMbyGlYRRg:bdgt3g8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L245)):*
> ```XML
>                     }
>                 } else {
>                     self.key.create_trivial(0)
> ```

This means we can pull more seed from the seeder that what we actually use?  


I see above we compare against `num_input_random_bits ` and not post_mul_log2_range so
`(excluded_upper_bound as f64).log2()` is not part of the base random bits we generate right ?
___
*[`tfhe/src/integer/oprf.rs` line 252 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XIwyi2zaws5YujOG4:-O_XIwyi2zaws5YujOG5:b-q9nf16) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L252)):*
> ```XML
>         let input = RadixCiphertext::from(in_blocks);
> 
>         let before_shift: crate::integer::ciphertext::BaseRadixCiphertext<
> ```

The explicit type should not be needed
___
*[`tfhe/src/integer/oprf.rs` line 259 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XL88EDdfcjfmwmtHF:-O_XL88FBscUELz-CrYm:bsyx5x1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L259)):*
> ```XML
> 
>         // Adjust the number of leading (MSB) trivial zeros blocks
>         loop {
> ```

I don't like having a loop for this why not do:

```
result.blocks.resize(num_blocks_output as usize, self.key.create_trivial(0));
```

Or if you want to keep the assert 
```
 match result.blocks.len().cmp(&(num_blocks_output as usize)) {
                Ordering::Less => result.blocks.resize(num_blocks_output as usize, self.key.create_trivial(0)),
                Ordering::Equal => {
                    break;
                }
                Ordering::Greater => {
                   
                    while  result.blocks.len() != num_blocks_output {
                         let leading_block = result.blocks.pop().unwap();
                          assert!(leading_block.is_trivial());
                    }
                   // or even result.blocks.drain(num_blocks_output..).for_each(|b| assert(...)
                }
            }

```

mayeul-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker and @tmontaigu)

___
*[`tfhe/src/integer/oprf.rs` line 245 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XKqIICTLSRlUq7BHc:-OaOaYce3GwugrUCLg4s:b-vxj8zx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L245)):*

This means we can pull more seed from the seeder that what we actually use?  


I see above we compare against `num_input_random_bits ` and not post_mul_log2_range so
`(excluded_upper_bound as f64).log2()` is not part of the base random bits we generate right ?
</blockquote></details>

Indeed, but is that problematic?

`excluded_upper_bound as f64).log2()` is indeed not part of the base random bits we generate

We need `num_blocks` after the scalar mul. So we also need `num_blocks` before the scalar mul
___
*[`tfhe/src/integer/oprf.rs` line 252 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XIwyi2zaws5YujOG4:-OaOdzY64ARLyjjH72IB:bb6un1j) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L252)):*

The explicit type should not be needed
</blockquote></details>

Done
___
*[`tfhe/src/integer/oprf.rs` line 259 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-O_XL88EDdfcjfmwmtHF:-OaOdkEU9ug8b8-XLU7m:b866q61) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/801a348696caa149549dcf058a9706c7ccd3a7a2/tfhe/src/integer/oprf.rs#L259)):*

I don't like having a loop for this why not do:

```
result.blocks.resize(num_blocks_output as usize, self.key.create_trivial(0));
```

Or if you want to keep the assert 
```
 match result.blocks.len().cmp(&(num_blocks_output as usize)) {
                Ordering::Less => result.blocks.resize(num_blocks_output as usize, self.key.create_trivial(0)),
                Ordering::Equal => {
                    break;
                }
                Ordering::Greater => {
                   
                    while  result.blocks.len() != num_blocks_output {
                         let leading_block = result.blocks.pop().unwap();
                          assert!(leading_block.is_trivial());
                    }
                   // or even result.blocks.drain(num_blocks_output..).for_each(|b| assert(...)
                }
            }

```
</blockquote></details>

I think I prefer it the way it is

IceTDrinker:
*[`tfhe/src/integer/oprf.rs` line 202 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP0hD5RoRJDifPouD:-OaUu7MO978tefuvHkzG:b-pnmsc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L202)):*

There is a `!`
For a power of 2, the "exact" function should be used instead
</blockquote></details>

I think the ceil_log2 is a ceil only when not an exact power of 2 ? unless there is something I'm missing

mayeul-zama:
*[`tfhe/src/integer/oprf.rs` line 202 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP0hD5RoRJDifPouD:-OaUvAjCDCMujG67IVmB:b-145rew) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L202)):*

I think the ceil_log2 is a ceil only when not an exact power of 2 ? unless there is something I'm missing
</blockquote></details>

When `excluded_upper_bound`is a power of 2, we don't need this algorithm. The functions defined above provide a perfect uniform distribution.
This assert is here to prevent using this function in this case.

A HL wrapper should probably call this function or the one above depending on whether `excluded_upper_bound`is a power of 2.

Not sure what you meant with `ceil_ilog2()`

IceTDrinker (commented):
> @IceTDrinker reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @mayeul-zama and @tmontaigu)

___
*[`tfhe/src/integer/oprf.rs` line 202 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP0hD5RoRJDifPouD:-OaUycel48dR8TZBUI7S:b-7e969s) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L202)):*

I'll add a message if this assert breaks
</blockquote></details>

ah got it, hmm not sure this is a reasonable thing to do ? if we detect it's a power of two we can just forward to the original PRF wdyt ?
___
*[`tfhe/src/integer/oprf.rs` line 205 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP2HY7_n6AJzmSGUj:-OaUyDDU9qjEIkZZZi41:bp216tg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L205)):*

I think most users would always pad to a predefined `num_blocks`. So better do it here.
Wdyt?
</blockquote></details>

well I don't know given it complicates the checks here, in integer its fine to let users do that themselves unless doing it ourselves has a big performance upside, @tmontaigu do you have an opinion on this ?
___
*[`tfhe/src/integer/oprf.rs` line 517 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaUzPhk16Caf75WhpIJ:-OaUzPhk16Caf75WhpIK:bkzvngt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/cb1025db515cdf1f05ac94c4f5c9aa5f289442d3/tfhe/src/integer/oprf.rs#L517)):*
> ```XML
>         let mut density = vec![0_usize; excluded_upper_bound];
> 
>         for i in 0..1 << num_input_random_bits {
> ```

you have the random_input_upper_bound here too nitpick not required to go ahead

mayeul-zama:
*[`tfhe/src/integer/oprf.rs` line 202 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP0hD5RoRJDifPouD:-OaV18gTEQcuGGK31BuY:bbppoj3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L202)):*

ah got it, hmm not sure this is a reasonable thing to do ? if we detect it's a power of two we can just forward to the original PRF wdyt ?
</blockquote></details>

I think in this case, we would ignore some function arguments and "lie" about what happens in the function.
Forcing a user to handle this case differently is safer IMO

A HL wrapper which chooses `num_input_random_bits` given a maximum distribution bias would be a better place to do this branching IMO

IceTDrinker:
*[`tfhe/src/integer/oprf.rs` line 223 at r7](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaZ1PyJ4WStkkPikHJf:-OaZ1PyJ4WStkkPikHJg:b-x8dziu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9bbd2a84683818360719d1db0a76581d5d19e577/tfhe/src/integer/oprf.rs#L223)):*
> ```XML
>             .collect();
> 
>         let mut in_blocks = seeds
> ```

can't we use the standard integer PRF for this with the correct number of bits directly ? :thinking:

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)* status: 2 of 3 files reviewed, 5 unresolved discussions (waiting on @IceTDrinker and @mayeul-zama)

___
*[`tfhe/src/integer/oprf.rs` line 205 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP2HY7_n6AJzmSGUj:-OaZ1OD71iOxn2EJifBC:bvaoame) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L205)):*

no probably not worth the effort to have several functions that do that
</blockquote></details>

It also means users could give a number of output block that is lower that what is needed to represent the range [0..excluded_bound[

mayeul-zama:
*[`tfhe/src/integer/oprf.rs` line 205 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaPP2HY7_n6AJzmSGUj:-OaZ2kWsBkrIV_GmObUJ:b1wi8kn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7c0fa531430152f23f49cfeb944ade388e18cc42/tfhe/src/integer/oprf.rs#L205)):*

It also means users could give a number of output block that is lower that what is needed to represent the range [0..excluded_bound[
</blockquote></details>

Yes, that's why there's the following assert

IceTDrinker (commented):
Looks mostly good just two details to make it easier to read and reason about

> @IceTDrinker reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-:-OaZNn141MN-dSLIwioT:br8gw4a)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @tmontaigu)

___
*[`tfhe/src/integer/oprf.rs` line 206 at r8](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaZLzqPDBAFyOXRXvc_:-OaZLzqPDBAFyOXRXvca:b-61js3x) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f71bbffda3f134d9ed9f84e597946c2e6eb91750/tfhe/src/integer/oprf.rs#L206)):*
> ```XML
>         );
> 
>         let num_bits_output = num_blocks_output * self.message_modulus().0;
> ```

that's not a bit count, to be checked
___
*[`tfhe/src/integer/oprf.rs` line 209 at r8](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789#-OaZMpDKCosmf0qqqWgy:-OaZMpDKCosmf0qqqWgz:bcfhk8e) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f71bbffda3f134d9ed9f84e597946c2e6eb91750/tfhe/src/integer/oprf.rs#L209)):*
> ```XML
>         assert!((excluded_upper_bound as f64) < 2_f64.powi(num_bits_output as i32));
> 
>         let post_mul_log2_range =
> ```

let's call this post_mul_bit_count or something equivalent

which means we need `(excluded_upper_bound as f64).log2().ceil()` to have the next power of 2

the division below keeps the ceil as discussed

tmontaigu (approved):
> @tmontaigu reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2789)* status: all files reviewed, 1 unresolved discussion (waiting on @IceTDrinker)

PR #2799, opened 2025-09-22
chore(ci): fix audit workflow name

nsarlin-zama:

### PR content/description
I forgot to set the name of the workflow and now it displays weirdly


soonum (changes requested):
According to the workflow naming standard, the top-level name should be the same than the file name, thus `cargo_audit`.

PR #2809, opened 2025-09-23
chore(gpu): GPU deterministic long run test

andrei-stoian-zama:
Makes the long run test deterministic with regards to GPU key GPU association.

Speeds up long run tests by avoiding keygen for each operation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2809)

### DIFF
--- a/tfhe/src/high_level_api/global_state.rs
+++ b/tfhe/src/high_level_api/global_state.rs
@@ -287,7 +287,7 @@ mod gpu {
     }
 
     impl CudaGpuChoice {
-        pub(in crate::high_level_api) fn build_streams(self) -> CudaStreams {
+        pub(crate) fn build_streams(self) -> CudaStreams {
             match self {
                 Self::Single(idx) => CudaStreams::new_single_gpu(idx),
                 Self::Multi => CudaStreams::new_multi_gpu(),
--- a/tfhe/src/high_level_api/mod.rs
+++ b/tfhe/src/high_level_api/mod.rs
@@ -53,6 +53,8 @@ use backward_compatibility::compressed_ciphertext_list::SquashedNoiseCiphertextS
 pub use config::{Config, ConfigBuilder};
 #[cfg(feature = "gpu")]
 pub use global_state::CudaGpuChoice;
+#[cfg(feature = "gpu")]
+pub use global_state::CustomMultiGpuIndexes;
 pub use global_state::{set_server_key, unset_server_key, with_server_key_as_context};
 
 pub use integers::{
--- a/tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
@@ -1,28 +1,55 @@
-use crate::core_crypto::gpu::vec::GpuIndex;
+use crate::core_crypto::commons::generators::DeterministicSeeder;
 use crate::core_crypto::gpu::{get_number_of_gpus, CudaStreams};
 use crate::integer::gpu::ciphertext::boolean_value::CudaBooleanBlock;
 use crate::integer::gpu::ciphertext::{CudaSignedRadixCiphertext, CudaUnsignedRadixCiphertext};
 use crate::integer::gpu::server_key::radix::tests_unsigned::GpuContext;
 use crate::integer::gpu::CudaServerKey;
-use crate::integer::server_key::radix_parallel::tests_cases_unsigned::FunctionExecutor;
-use crate::integer::{
-    BooleanBlock, RadixCiphertext, RadixClientKey, ServerKey, SignedRadixCiphertext, U256,
-};
-use rand::seq::SliceRandom;
-use rand::Rng;
-use std::sync::Arc;
+use crate::integer::server_key::radix_parallel::tests_long_run::OpSequenceFunctionExecutor;
+use crate::integer::{BooleanBlock, RadixCiphertext, RadixClientKey, SignedRadixCiphertext, U256};
+use crate::{CompressedServerKey, CudaGpuChoice, CustomMultiGpuIndexes, GpuIndex};
+use tfhe_csprng::generators::DefaultRandomGenerator;
+use tfhe_csprng::seeders::Seeder;
 
 pub(crate) mod test_erc20;
 pub(crate) mod test_random_op_sequence;
 pub(crate) mod test_signed_erc20;
 pub(crate) mod test_signed_random_op_sequence;
 
-pub(crate) struct GpuMultiDeviceFunctionExecutor<F> {
+/// Fisher-Yates shuffle using the seeded random number generator
+fn seeded_shuffle(v: &mut [u32], datagen: &mut DeterministicSeeder<DefaultRandomGenerator>) {
+    for i in (1..v.len()).rev() {
+        let j = datagen.seed().0 as usize % (i + 1);
+        v.swap(i, j);
+    }
+}
+fn make_random_gpu_set(datagen: &mut DeterministicSeeder<DefaultRandomGenerator>) -> CudaGpuChoice {
+    // Sample a random subset of 1-N gpus, where N is the number of available GPUs
+    // A GPU index should not appear twice in the subset
+    let num_gpus = get_number_of_gpus();
+    //let mut rng = rand::thread_rng();
+    let num_gpus_to_use = (1 + datagen.seed().0 as u32 % (num_gpus - 1)) as usize;
+    let mut all_gpu_indexes: Vec<u32> = (0..num_gpus).collect();
+    seeded_shuffle(&mut all_gpu_indexes, datagen);
+
+    let gpu_indexes_to_use = &all_gpu_indexes[..num_gpus_to_use];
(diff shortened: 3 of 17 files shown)


tfhe/src/high_level_api/global_state.rs
```diff
@@ -287,7 +287,7 @@ mod gpu {
     }
 
     impl CudaGpuChoice {
-        pub(in crate::high_level_api) fn build_streams(self) -> CudaStreams {
+        pub fn build_streams(self) -> CudaStreams {
```
agnesLeroy: I think this should be pub(crate) instead of pub

tfhe/src/high_level_api/mod.rs
```diff
@@ -53,6 +53,8 @@ use backward_compatibility::compressed_ciphertext_list::SquashedNoiseCiphertextS
 pub use config::{Config, ConfigBuilder};
 #[cfg(feature = "gpu")]
 pub use global_state::CudaGpuChoice;
+#[cfg(feature = "gpu")]
+pub use global_state::CustomMultiGpuIndexes;
```
agnesLeroy, line 57: Do you really need to add this pub use?

tfhe/src/integer/server_key/radix_parallel/tests_long_run/mod.rs
```diff
@@ -61,10 +104,10 @@ impl RadixEncryptable for i64 {
     }
 }
 
-struct RandomOpSequenceDataGenerator<P, C> {
+pub struct RandomOpSequenceDataGenerator<P, C> {
```
agnesLeroy: pub(crate)

tfhe/src/integer/server_key/radix_parallel/tests_long_run/mod.rs
```diff
@@ -75,7 +118,7 @@ impl<
         C: IntegerCiphertext,
     > RandomOpSequenceDataGenerator<P, C>
 {
-    fn new(total_num_ops: usize, cks: &RadixClientKey) -> Self {
+    pub fn new(total_num_ops: usize, cks: &RadixClientKey) -> Self {
```
agnesLeroy: pub(crate)

tfhe/src/integer/server_key/radix_parallel/tests_long_run/mod.rs
```diff
@@ -112,7 +155,7 @@ impl<
         }
     }
 
-    fn get_seed(&self) -> Seed {
+    pub fn get_seed(&self) -> Seed {
```
agnesLeroy: pub(crate)

tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
```diff
@@ -31,34 +60,37 @@ impl<F> GpuMultiDeviceFunctionExecutor<F> {
...
-        let gpu_indexes: Vec<GpuIndex> = gpu_indexes_to_use
-            .iter()
-            .map(|idx| GpuIndex::new(*idx))
-            .collect();
-        println!("Setting up server key on GPUs: [{gpu_indexes_to_use:?}]");
-
-        let streams = CudaStreams::new_multi_gpu_with_indexes(&gpu_indexes);
-
+impl<F> OpSequenceGpuMultiDeviceFunctionExecutor<F> {
+    pub(crate) fn setup_from_cpu_keys(&mut self, cks: &RadixClientKey, _sks: &Arc<ServerKey>) {
```
agnesLeroy: Why do you need two different functions, couldn't you have only one setup function that does the setup from gpu keys? Couldn't we use the CompressedServerKey for CPU setup as well?

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
```diff
@@ -619,6 +624,106 @@ where
...
+    }
+}
+
+/// For 4-ary operations
+impl<F, I1, I2, I3, I4, O> OpSequenceFunctionExecutor<(I1, I2, I3, I4), O>
+    for CpuFunctionExecutor<F>
+where
+    F: Fn(&ServerKey, I1, I2, I3, I4) -> O,
+{
+    fn setup(&mut self, _cks: &RadixClientKey, sks: Arc<ServerKey>) {
```
agnesLeroy: Same for all these function executors, it seems strange to have setup & setup_gpu which are different. If possible the CPU setup should be done using the CompressedServeryKey I think. Btw the setup function takes Arc<ServerKey> as input, don't we need to pass Arc<CompressedServerKey> to avoid copying the key around? Maybe not :thinking:

tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
andrei-stoian-zama: the CompressedServerKey is not handled by the key-cache. thus to keep the previous logic for CPU (using KeyCache)  I added two functions

agnesLeroy: But then do we really need to use the key cache for the CPU test? I think it'll be fine to do the same as the GPU because we'll only be generating one key for this test.

tfhe/src/integer/server_key/radix_parallel/tests_unsigned/mod.rs
andrei-stoian-zama: I see what you mean! maybe @IceTDrinker can you confirm it's ok not to use the server key in the key cache in CPU long run tests ? we'd be generating a compressedServerKey from the client key stored in the key cache

IceTDrinker: the main usage for the key cache was to avoid having each test doing keygen and saturating both CPU and RAM during tests (which would often result in OOM crashes)

So if there aren't too many long run tests running at once should be fine I guess a tiny bit wasteful but probably manageable could take the key by value to avoid keeping both the compressed key and decompressed key in memory

IceTDrinker: also keygen used to be über slow which is not the case anymore

tfhe/src/high_level_api/mod.rs
andrei-stoian-zama: yes, since I can't import

`use crate::high_level_api::global_state::CustomMultiGpuIndexes;`

because `global_state` is private and I'd rather not change the whole module visibility

agnesLeroy: Maybe it should be `pub(crate) use` instead then?

IceTDrinker: you should be able to import it without pub if you are able to import it with pub 🧐

tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
andrei-stoian-zama: A second issue is that we need to pass an additional argument: the gpu choice. That doesn't make sense for the Cpu executor. The other option is to pass the RNG as an Option<>. It's quite a pain since the RNG needs to be passed as a pointer (Arc?) and stored in the executor - all executors must share the same rng.

IceTDrinker: it may just mean that we need to have a CompressedServerKey KeyCache

tfhe/src/high_level_api/mod.rs
andrei-stoian-zama: I tried the various options:
`pub(crate) use` -> clippy says unused import
`use` -> cannot find CustomMultiGpuIndexes

since CudaGpuChoice above is `pub` can't this one be too ?

agnesLeroy: Ah that struct is already pub, sorry so it's fine to do `pub use` I think

IceTDrinker: unused means you need gate it with gpu potetntially

IceTDrinker: pub use is intended to allow outside users to import it, if you don't want that you can have a pub(crate) or private import but with a feature gate, when this happens we tend to create a dedicated module to have only a single cfg on that module and then use the import as needed

tfhe/src/integer/gpu/server_key/radix/tests_signed/mod.rs
```diff
@@ -643,3 +646,1211 @@ where
         d_block.to_boolean_block(&context.streams)
     }
 }
+
+pub(crate) struct GpuMultiDeviceFunctionExecutor<F> {
```
agnesLeroy, line 650: All the unsigned related function executors should be placed in tests_unsigned/mod.rs please. Here there should only be executors for signed types.

agnesLeroy (commented):
Looks much better, thank you @andrei-stoian-zama! Just one comment, and I think we'll be good.  There's also the question  of `pub(use) global_state::CustomMultiGpuIndexes;` that remains.

PR #2822, opened 2025-09-25
chore(ci): split cargo builds into several jobs

soonum:
Post-commit checks was the bottleneck regarding running duration.
It's now split into 7 batches to improve parallelism.
Builds that are specific to Ubuntu are run in their own jobs, so that only `build_tfhe_full` recipe call remains in the OS matrix.

A final check is performed to ensure all the checks have passed, this very job is used as branch protection rule.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2822)

### DIFF
--- a/.github/workflows/cargo_build.yml
+++ b/.github/workflows/cargo_build.yml
@@ -18,17 +18,95 @@ permissions:
   contents: read
 
 jobs:
-  cargo-builds:
-    name: cargo_build/cargo-builds (bpr)
-    runs-on: ${{ matrix.os }}
+  prepare-parallel-pcc-matrix:
+    name: cargo_build/prepare-parallel-pcc-matrix
+    runs-on: ubuntu-latest
+    outputs:
+      matrix_command: ${{ steps.set-pcc-commands-matrix.outputs.commands }}
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+        with:
+          persist-credentials: "false"
+          token: ${{ env.CHECKOUT_TOKEN }}
 
+      # Fetch all the Make recipes that start with `pcc_batch_`
+      - name: Set pcc commands matrix
+        id: set-pcc-commands-matrix
+        run: |
+          COMMANDS=$(grep -oE '^pcc_batch_[^:]*:' Makefile | sed 's/:/\"/; s/^/\"/' | paste -sd,)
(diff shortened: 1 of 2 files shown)


.github/workflows/cargo_build.yml
```diff
@@ -18,17 +18,76 @@ permissions:
...
 jobs:
-  cargo-builds:
-    name: cargo_build/cargo-builds (bpr)
-    runs-on: ${{ matrix.os }}
+  parallel-pcc-cpu:
+    name: cargo_build/parallel-pcc-cpu
+    runs-on: large_ubuntu_16
+    strategy:
+      matrix:
+        batch: [ 1, 2, 3, 4, 5, 6, 7 ]
```
IceTDrinker: we can do better and query the Makefile or grep the makefile to know how many batches there are and automate the matrix population (see ntt workflow for dynamic matrices) 😉

IceTDrinker:
> From 1 hours 25 mins down to 16 mins 🎉

image

.github/workflows/cargo_build.yml
IceTDrinker: if easier write the command directly in the Lakefile and have the output of the Makefile command be stored in the github output by the workflow

soonum: Way more convenient this way and reduce maintenance issues if a new pcc recipe is created.

soonum (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2822)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @IceTDrinker)

___
*[`.github/workflows/cargo_build.yml` line 218 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/2822#-Oa0hgI-2ycJcnFoXgYV:-Oa497kZAWnjVfo59yEt:b-ae6y3l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/80287e98588a44c25ae141b55f6fd0fa68f6ef03/.github/workflows/cargo_build.yml#L218)):*

wondering about making it easy to see what build failed. Before we had only essentially one job, now we have several and I wonder if we want to make it easy to find by printing the failed job runs URLs in this error message
</blockquote></details>

All jobs are displayed in the check suite, so if one fails, it will be shown in the `Failure` section.

PR #2823, opened 2025-09-25
chore(gpu): get decompress size on gpu without calling on_gpu

agnesLeroy:

### PR content/description

In get_decompression_size_on_gpu, we used to do:
```
self.inner.on_gpu(streams).get_decompression_size_on_gpu(
                            index,
                            decompression_key,
                            streams,
```
The problem is that calling `on_gpu()` allocates and copies data to the GPU. 
It shouldn't be necessary to do this only to query decompression size. 
Also, this means the allocated GPU variable gets dropped at the end of `get_decompression_size_on_gpu` which means a
`cuda_device_synchronize` gets called, and this affects performance in the coprocessor.

This PR introduces an entry point for `get_decompression_size_on_gpu` when the input is on the CPU, using the CPU metadata to compute the size directly.

### Check-list:

### DIFF
--- a/tfhe/src/high_level_api/compressed_ciphertext_list.rs
+++ b/tfhe/src/high_level_api/compressed_ciphertext_list.rs
@@ -622,20 +622,32 @@ impl CiphertextList for CompressedCiphertextList {
         global_state::with_internal_keys(|key| {
             if let InternalServerKey::Cuda(cuda_key) = key {
                 let streams = &cuda_key.streams;
-                cuda_key
-                    .key
-                    .decompression_key
-                    .as_ref()
-                    .ok_or_else(|| {
-                        crate::Error::new("Compression key not set in server key".to_owned())
-                    })
-                    .map(|decompression_key| {
-                        self.inner.on_gpu(streams).get_decompression_size_on_gpu(
-                            index,
-                            decompression_key,
-                            streams,
-                        )
-                    })
+                match &self.inner {
+                    InnerCompressedCiphertextList::Cpu(ct_list) => cuda_key
+                        .key
+                        .decompression_key
+                        .as_ref()
+                        .ok_or_else(|| {
+                            crate::Error::new("Compression key not set in server key".to_owned())
+                        })
+                        .map(|decompression_key| {
+                            ct_list.get_decompression_size_on_gpu(index, decompression_key, streams)
+                        }),
+                    InnerCompressedCiphertextList::Cuda(cuda_ct_list) => cuda_key
+                        .key
+                        .decompression_key
+                        .as_ref()
+                        .ok_or_else(|| {
+                            crate::Error::new("Compression key not set in server key".to_owned())
+                        })
+                        .map(|decompression_key| {
+                            cuda_ct_list.get_decompression_size_on_gpu(
+                                index,
+                                decompression_key,
+                                streams,
+                            )
+                        }),
+                }
             } else {
                 Ok(Some(0))
             }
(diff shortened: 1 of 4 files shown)


tfhe/src/high_level_api/compressed_ciphertext_list.rs
```diff
@@ -622,20 +622,32 @@ impl CiphertextList for CompressedCiphertextList {
...
+                            crate::Error::new("Compression key not set in server key".to_owned())
+                        })
+                        .map(|decompression_key| {
+                            cuda_ct_list.get_decompression_size_on_gpu(
+                                index,
+                                decompression_key,
+                                streams,
+                            )
+                        }),
+                }
```
tmontaigu, line 650: fetch the key before: 
```rust
let decompression_key = cuda_key
                        .key
                        .decompression_key
                        .as_ref()
                        .ok_or_else(|| {
                            crate::error!("Compression key not set in server key")
                        })?;

match &self.inner  {
 Cpu => Ok(Some(...))
 Gpu => Ok(Some(...))
}

```

tfhe/src/integer/ciphertext/compressed_ciphertext_list.rs
```diff
@@ -176,6 +180,42 @@ impl CompressedCiphertextList {
...
+    #[cfg(feature = "gpu")]
+    pub fn get_decompression_size_on_gpu(
+        &self,
+        index: usize,
+        decomp_key: &CudaDecompressionKey,
+        streams: &CudaStreams,
+    ) -> Option<u64> {
+        self.get_blocks_of_size_on_gpu(index, decomp_key, streams)
+    }
+    #[allow(clippy::unnecessary_wraps)]
```
tmontaigu: I would agree with clippy that this function shoud return u64, not Option<u64>

agnesLeroy: Actually in this function I do:
```
let preceding_infos = self.info.get(..index)?;
```
and I can't do that if I don't return an option.

tfhe/src/high_level_api/compressed_ciphertext_list.rs
agnesLeroy: Clippy tells me:
```
error[E0277]: the `?` operator can only be used on `Option`s, not `Result`s, in a closure that returns `Option`                                                                                                    
   --> tfhe/src/high_level_api/compressed_ciphertext_list.rs:632:27                                                                                                                                                
    |                                                                                                                                                                                                              
622 |         global_state::with_internal_keys(|key| {                                                                                                                                                             
    |                                          ----- this function returns an `Option`                                                                                                                             
...                                                                                                                                                                                                                
632 |                         })?;
    |                           ^ use `.ok()?` if you want to discard the `std::result::Result<std::convert::Infallible, error::Error>` error information
    ```

tfhe/src/integer/ciphertext/compressed_ciphertext_list.rs
agnesLeroy: Actually the `#[allow(clippy::unnecessary_wraps)]` doesn't seem necessaryn clippy doesn't complain without it. I don't know why it was there.

tmontaigu: Well actually, what is the error clippy gives you here ? because the Option is needed here

agnesLeroy: There is no error :sweat_smile: I just removed the allow

andrei-stoian-zama (approved):

PR #2828, opened 2025-09-26
feat(gpu): Implements optimized division algorithm for message_2_carry_2

bbarbakadze:
### PR content/description
Implements optimized division algorithm for message_2_carry_2 at least 4 gpus are available
### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
+++ b/backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
@@ -84,6 +84,11 @@ public:
     return CudaStreams(_streams, _gpu_indexes, 1);
   }
 
+  // Returns a subset containing only the ith stream
+  CudaStreams subset(int i) const {
+    return CudaStreams(&_streams[i], &_gpu_indexes[i], 1);
+  }
+
   // Synchronize all the streams in the set
   void synchronize() const {
     for (uint32_t i = 0; i < _gpu_count; i++) {
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -1065,6 +1065,10 @@ template <typename InputTorus> struct int_noise_squashing_lut {
   }
 };
 
+// Forward declarations for operation buffers
+template <typename Torus> struct int_sub_and_propagate;
+template <typename Torus> struct int_bitop_buffer;
+
 template <typename Torus> struct int_bit_extract_luts_buffer {
   int_radix_params params;
   int_radix_lut<Torus> *lut;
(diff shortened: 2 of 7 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -4237,6 +4241,842 @@ template <typename Torus> struct int_comparison_buffer {
...
+  uint64_t tmp_size_tracker = 0;
+
+  // allocate and initialize if needed, temporary arrays used to calculate
+  // cuda integer div_rem_2_2 operation
+  void init_temporary_buffers(CudaStreams streams, uint32_t num_blocks,
+                              bool allocate_gpu_memory,
+                              uint64_t &size_tracker) {
+
+    // more than one block temporary arrays
+    tmp_gpu_0 = new CudaRadixCiphertextFFI;
```
andrei-stoian-zama, line 4339: could it be written as :

```
CudaRadixCiphertextFFI** temp_arrays_num_blocks_plus_1 = { tmp_gpu_0, d3, ... };
for (auto tmp_array : temp_arrays_num_blocks_plus_1) {
   *tmp_array = new ..;
      create_zero_radix_ciphertext_async<Torus>(
        streams.stream(0), streams.gpu_index(0), *tmp_array, num_blocks + 1,
        params.big_lwe_dimension, size_tracker, allocate_gpu_memory);
 }
```

it would make the code a lot more compact

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -4237,6 +4241,842 @@ template <typename Torus> struct int_comparison_buffer {
...
+  CudaRadixCiphertextFFI *comparison_blocks_1; // num_blocks
+  CudaRadixCiphertextFFI *comparison_blocks_2; // num_blocks
+  CudaRadixCiphertextFFI *comparison_blocks_3; // num_blocks
+  CudaRadixCiphertextFFI *cmp_1;               // boolean block
+  CudaRadixCiphertextFFI *cmp_2;               // boolean block
+  CudaRadixCiphertextFFI *cmp_3;               // boolean block
+  CudaRadixCiphertextFFI *c0;                  // single block
+  CudaRadixCiphertextFFI *q1;                  // single block
+  CudaRadixCiphertextFFI *q2;                  // single block
+  CudaRadixCiphertextFFI *q3;                  // single block
```
agnesLeroy, line 4311: Looking at this list I'm wondering if we couldn't avoid allocating so much data by reusing some of it, but for now i's fine.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -4237,6 +4241,842 @@ template <typename Torus> struct int_comparison_buffer {
...
+        }
+      }
+      cuda_memcpy_with_size_tracking_async_to_gpu(
+          second_indexes_for_overflow_sub_gpu_0[nb - 1], h_lut_indexes,
+          nb * sizeof(Torus), streams.stream(0), streams.gpu_index(0),
+          allocate_gpu_memory);
+      cuda_memcpy_with_size_tracking_async_to_gpu(
+          scalars_for_overflow_sub_gpu_0[nb - 1], h_scalar, nb * sizeof(Torus),
+          streams.stream(0), streams.gpu_index(0), allocate_gpu_memory);
+    }
```
agnesLeroy, line 4810: You should synchronize stream 0 before freeing to ensure memcpy's are over.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -4237,6 +4241,842 @@ template <typename Torus> struct int_comparison_buffer {
...
+      cuda_drop_with_size_tracking_async(
+          first_indexes_for_overflow_sub_gpu_2[i], streams.stream(2),
+          streams.gpu_index(2), gpu_memory_allocated);
+      cuda_drop_with_size_tracking_async(
+          second_indexes_for_overflow_sub_gpu_2[i], streams.stream(2),
+          streams.gpu_index(2), gpu_memory_allocated);
+      cuda_drop_with_size_tracking_async(
+          scalars_for_overflow_sub_gpu_2[i], streams.stream(2),
+          streams.gpu_index(2), gpu_memory_allocated);
+    }
```
agnesLeroy, line 5063: Same here you should synchronize gpu 0, 1, 2 before freeing these

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -4237,6 +4241,842 @@ template <typename Torus> struct int_comparison_buffer {
   }
 };
 
+// used only when 4 gpus are available
```
andrei-stoian-zama, line 4244: could this be moved to a new file so we avoid adding new stuff to this already huge file ?

agnesLeroy (commented):
Thanks a lot for this PR @bbarbakadze, this is neat! Only a few comments to make :slightly_smiling_face:

backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
```diff
@@ -31,6 +32,482 @@ __host__ uint64_t scratch_cuda_integer_div_rem_kb(
   return size_tracker;
 }
 
+template <typename Torus>
+__host__ void host_unsigned_integer_div_rem_kb_block_by_block_2_2(
```
andrei-stoian-zama, line 36: could this function be split into several subfunctions with descriptive names ?

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
```diff
@@ -84,6 +84,11 @@ struct CudaStreams {
     return CudaStreams(_streams, _gpu_indexes, 1);
   }
 
+  // Returns a subset containing only the ith stream
+  CudaStreams subset(int i) const {
```
andrei-stoian-zama, line 88: if you add this then I suggest removing `subset_first_gpu` in a second PR, and replacing all usages with `subset(0)`

andrei-stoian-zama (changes requested):
I suggest some improvements for readability.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: the same could be done for deletion

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: We need to think about what we do with this file more generally: probably we need to split it altogether to follow the same structure as cuh/cu files (but not now)

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
agnesLeroy: We discussed it this morning and were thinking of doing it after we merge this one. Also a better name would be get_stream instead of subset.

backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
agnesLeroy: Not sure this is a good idea, I quite like the fact that it's very linear and easy to follow :thinking: 
@bbarbakadze wdyt?

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
andrei-stoian-zama: `get_stream` could be understood as returning a cudaStream_t while `subset` returns CudaStreams

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: this is the only one left (you can ignore all the other comments)

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
agnesLeroy: `subset` makes it feel mike it's a slice it will return though, if it only returns a single element maybe simply `get` is  better.

agnesLeroy: Or `get_from_index`

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
bbarbakadze: compred to previous implementation I tried to reuse some memory that was obvious that it could be reuse, there are probably more to be reused, maybe later we can do.

backends/tfhe-cuda-backend/cuda/src/integer/div_rem.cuh
bbarbakadze: I agree to Agnes, and I tried to have similar structure as CPU. They have one big function too.

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
bbarbakadze: maybe keep subset but instead of always returning 1 gpu, lets make it a parameter and pass how many gpu we want to be returned, but maybe in another pr.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
bbarbakadze: @andrei-stoian-zama 
Then I would need another array to specify which gpu to use for each of them. and this array should be a member array because we need it in release too, in release we need to know which gpu to use for releasing.

if you think it will still be better to read I can modify it.

agnesLeroy (approved):

PR #2835, opened 2025-09-29
feat: add KVStore to the high level api

tmontaigu:
* Added Value type name to crate::integer::KVStore impl of Named trait as well as a bool to check we deserialize the correct value type (Radix vs SignedRadix)
* Add KVStore to high_level_api
* Add KVStore hlapi benches
* Remove specialized `[add,mul,sub]_to_slot` as `map` is now the intended API.
    - mul_to_slot was way slower than using `map`
    - add/mul_to_slot were a bit faster (~5% latency-wise), but returned less information (no old_value, no new_value, no boolean to check) if the key matched
    - Some known improvement can be made to map, which should result in it being better than add/sub_to_slot
* Add FheIntegerType trait to make the KVStore generic over FheUint/FheInt, and should make GPU integration "easy"


soonum (commented):
> @soonum reviewed 11 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835)* status: 11 of 16 files reviewed, 4 unresolved discussions (waiting on @IceTDrinker)

___
*[`tfhe-benchmark/benches/high_level_api/bench.rs` line 250 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaKu1Ok4mbdFI5fzvve:-OaKu1Ol25RHsouK6T8d:b-nvw111) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe-benchmark/benches/high_level_api/bench.rs#L250)):*
> ```XML
> impl TypeDisplay for u128 {}
> 
> impl<Id: FheUintId> TypeDisplay for tfhe::FheUint<Id> {
> ```

No `tfhe::` here comparing to the impl below (`Id: tfhe::FheIntId`) ?
___
*[`tfhe-benchmark/benches/high_level_api/bench.rs` line 286 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaKvDMy7Qr-kKxJi6vY:-OaKuwj4CSsK1mtIYH2F:bwv6jig) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe-benchmark/benches/high_level_api/bench.rs#L286)):*
> ```XML
> }
> 
> fn get_factor_for_op<Op>(op: Op, cks: &ClientKey) -> usize
> ```

Is this function is a way to define a throughput heuristic ?
If so, maybe integrate it in the `src/utilities.rs` file, if applicable
___
*[`tfhe-benchmark/benches/high_level_api/bench.rs` line 311 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaKwOtwAymdtkQ68IGT:-OaKwOtwAymdtkQ68IGU:b3xdpa7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe-benchmark/benches/high_level_api/bench.rs#L311)):*
> ```XML
> 
>     // Then find how many ops we should do to have full occupancy
>     let factor = time_for_op.as_millis().div_ceil(time_if_full_occupancy);
> ```

I suppose you've checked visually with htop that we are at full occupancy during the run ?
___
*[`tfhe-benchmark/benches/high_level_api/bench.rs` line 352 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaL2c3x3cPKosy31eKB:-OaL2c3y4_0UIXoUJZgu:btb4c1f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe-benchmark/benches/high_level_api/bench.rs#L352)):*
> ```XML
>             let value_to_add = Value::encrypt(value, cks);
> 
>             c.bench_function(&format_id_bench("Get"), |b| {
> ```

Naming will have to change once we want the results in database.
No need to do it now though.

tmontaigu:
*[`tfhe-benchmark/benches/high_level_api/bench.rs` line 366 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaL0kfH2Fn_dvxI-yvP:-OaOhr5xETUvNb0kUHm2:baku83u) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe-benchmark/benches/high_level_api/bench.rs#L366)):*

Just want to understand what this measures since it runs the identity ?
</blockquote></details>

It measures the core of the map operation without having the cost of the map operation

tmontaigu (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835)* status: all files reviewed, 9 unresolved discussions (waiting on @IceTDrinker and @soonum)

___
*[`tfhe/src/high_level_api/kv_store.rs` line 126 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaL7sJ_A5nkZhpSEKom:-OaU7CNUBqCLBlAwFZMx:b24v9a7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe/src/high_level_api/kv_store.rs#L126)):*

what about impling Index for the clear variant ?
</blockquote></details>

It would work for the integer thing, but not the hlapi one because Index requires to return a `&T` and in the high level API since we need to wrap the Radix into a FheUint we cant return a `&`
___
*[`tfhe/src/high_level_api/kv_store.rs` line 281 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaL8DSd8SmbW7FSMI-a:-OaU9u_U99UfiHJByH3p:b-9xps1d) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe/src/high_level_api/kv_store.rs#L281)):*

even for the tags ? 

Re thinking about this rerand metadata should be a non problem
</blockquote></details>

The HLAPI has a tag, but its not used to tag ciphertext extracted from it are tagged with the tag of the server key that is used

Using the HL compressed list here would allow to not manually recheck the block count here, but then we would be getting FheUint/FheInt out of the list which we would deconstruct back into plain integer radix (not that its wrong)

Also, should the KVStore have its own Tag ? If so should the tag be used to tag resulting ciphertex (instead of using the server_key tag )?
___
*[`tfhe/src/high_level_api/integers/mod.rs` line 70 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2835#-OaKzQZ58KsEETJ1_K2O:-OaU7uhTAPnt-Y4tqOeQ:bq8x62l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4b7b596b634ef2e6455df3deac9841edb32e24a4/tfhe/src/high_level_api/integers/mod.rs#L70)):*

ok, did we make sure this one is not leaking or is sealed ?

lots of traits which should have been impl details cause annoyances because we did not seal them/made them un-implementable for external suff
</blockquote></details>

I will seal it, But this wont prevent users from being able to call functions of the trait unless I also do the private token struct trick. This trait is meant to be usable in bounds of users' functions

Wish rust had more granularity the visibility of traits elements

PR #2836, opened 2025-09-29
fix(shortint): accept trivial ciphertexts for rerand

IceTDrinker:
- make sure to set their noise to NOMINAL once rerand is done


nsarlin-zama (approved):
That's better indeed!

PR #2844, opened 2025-10-01
Tm/move rerand ksk params

tmontaigu:
Re-Randomization is something that requires a
dedicated public key.

Thus we move the parameters of the KSK
into the struct for dedicated PKE parameters

BREAKING CHANGE: This is breaking change regarding the latest alpha
released. But MetaParameters did not seem to be used directly in
fhevm/kms

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2844)

### DIFF
--- a/tfhe/src/shortint/parameters/meta.rs
+++ b/tfhe/src/shortint/parameters/meta.rs
@@ -11,6 +11,9 @@ pub struct DedicatedCompactPublicKeyParameters {
     /// Parameters used to key switch from the compact public key
     /// parameters to compute parameters
     pub ksk_params: ShortintKeySwitchingParameters,
+    /// Parameters to key switch from the compact public key
+    /// to rerand state
+    pub re_randomization_parameters: Option<ShortintKeySwitchingParameters>,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


tfhe/src/shortint/parameters/meta.rs
```diff
@@ -25,6 +28,4 @@ pub struct MetaParameters {
     pub compression_parameters: Option<CompressionParameters>,
     /// Parameters for noise squashing
     pub noise_squashing_parameters: Option<MetaNoiseSquashingParameters>,
-    /// Parameters for ciphertext re-randomization
-    pub re_randomization_parameters: Option<ShortintKeySwitchingParameters>,
```
mayeul-zama, line 29: Is this parameter used nowhere in the codebase at the moment?

tmontaigu: Its moved into the `DedicatedCompactPublicKeyParameters` struct above as re-rand is done on public key encrypted ciphertexts

mayeul-zama: Yes, I meant, it's stored by never used, right?

tmontaigu: We don't use it, its more meant for a user to use that with the added `From<MetaParameters> for Config`
and by use we mean fhevm

IceTDrinker (approved):
Thanks!

> @IceTDrinker reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2844#-:-OatGJzVCgwLbxcUSM3k:b-ucfknh)* status: all files reviewed, 1 unresolved discussion (waiting on @mayeul-zama)

tfhe/src/shortint/parameters/meta.rs
IceTDrinker: it will be used through the HL Config for rerand

PR #2784, opened 2025-09-16
rerand

nsarlin-zama:
closes: _please link all relevant issues_

### PR content/description
Adds ciphertext re-randomization using public key encryption of zero

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2784)

PR #2847, opened 2025-10-01
chore(core): fix encryption of single lwe to use the noise generator

nsarlin-zama:
This is a follow up of https://github.com/zama-ai/tfhe-rs/pull/2784/commits/958fc48d00afba769dc63a7fe932071c6c84f9c7
but applied to the encryption of a single lwe.


tfhe/src/core_crypto/commons/generators/encryption/mod.rs
```diff
@@ -197,6 +197,7 @@ impl<G: ByteRandomGenerator> EncryptionRandomGenerator<G> {
     }
 
     // Fills the input slice with random noise, using the random generator.
+    #[cfg(test)]
```
IceTDrinker, line 200: why ?

nsarlin-zama: this is now only used in a test so clippy complains about an unused function without this attribute

IceTDrinker: huh ok

IceTDrinker: I guess the custom mod variant might be used now

IceTDrinker (approved):
Thanks !

PR #2862, opened 2025-10-06
fix(hpu): fix HPU throughput bench & ilog2/modulus operations in hpu_test

pgardratzama:


pgardratzama:
HLAPI bench is creating some new failure: HPU does not support KVStore yet

PR #2867, opened 2025-10-06
chore(hlapi): export CompressedKVStore

tmontaigu:
Without this, users cannot use the CompressedKVStore type that is required in the KVStore deserialization


IceTDrinker (approved):
Thanks!

PR #2871, opened 2025-10-07
chore(backward): add data for ks32 client key, server key and ct

nsarlin-zama:
### PR content/description
Adds data for ks32 client key, server key and ciphertext. This should have been done in 1.3 but has been forgotten, so the data is added in the 1_3 folder.

The noise squashing key ks32 will come in a following pr, as it is for 1.4.

Some data inside 1_4 has been re-generated, certainly because the types have changed between the time they were generated and now. But the generation looks stable so it must not be a determinism issue.


tmontaigu (approved):
> @tmontaigu reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2871)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @IceTDrinker)

PR #2880, opened 2025-10-10
chore(docs): fix docs for docs.rs

tmontaigu:
doc_auto_cfg is no longer available in nightly >= 1.92

This prevents the docs to be build on docs.rs, as docs.rs uses the latest nightly

This commit also make the `make doc` target use the lastest nightly so that we can catch these errors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,14 @@ install_rs_check_toolchain:
 	( echo "Unable to install $(RS_CHECK_TOOLCHAIN) toolchain, check your rustup installation. \
 	Rustup can be downloaded at https://rustup.rs/" && exit 1 )
 
+.PHONY: install_rs_latest_nightly_toolchain # Install the nightly toolchain used to build docs using same version as docs.rs
+# We don't check that it exists, because we always want the latest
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -87,6 +87,14 @@ install_rs_check_toolchain:
 	( echo "Unable to install $(RS_CHECK_TOOLCHAIN) toolchain, check your rustup installation. \
 	Rustup can be downloaded at https://rustup.rs/" && exit 1 )
 
+.PHONY: install_rs_latest_nightly_toolchain # Install the nightly toolchain used to build docs using same version as doc.rs
```
IceTDrinker: doc.rs -> docs.rs

Makefile
```diff
@@ -1711,7 +1719,7 @@ pcc_batch_5: clippy_tfhe_lints check_compile_tests clippy_backward_compat_data
 
 .PHONY: pcc_batch_6  # duration: 4'50'' (shortest one, extend it with further checks)
```
tmontaigu: update with new timings after CI is done

PR #2890, opened 2025-10-13
fix(js): catch undefined variant using Option<>

tmontaigu:
In the JS ShortintParametersName, users could
make typo in the variant used e.g:
`ShortintParametersName.PARAM_MESSAGE_2_CARRY128`

In JS this returns `undefined` which is then later casted to an int and it becomes 0, leading to match the first variant

We modify the input to receive an `Option<ShortintParametersName>` as it seems to allow us to catch the `undefined` and return a proper error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1736,6 +1736,7 @@ pcc_batch_1:
 pcc_batch_2:
 	$(call run_recipe_with_details,clippy)
 	$(call run_recipe_with_details,clippy_all_targets)
+	$(call run_recipe_with_details,check_fmt_js)
 
 .PHONY: pcc_batch_3 # duration: 6'50''
 pcc_batch_3:


IceTDrinker:
We have a cache for npm stuff, or we can create a new batch

Makefile
```diff
@@ -1736,6 +1736,7 @@ pcc_batch_1:
 pcc_batch_2:
 	$(call run_recipe_with_details,clippy)
 	$(call run_recipe_with_details,clippy_all_targets)
+	$(call run_recipe_with_details,check_fmt_js)
```
IceTDrinker, line 1739: will crash if we don't have nvm ?

IceTDrinker: well looks like we are lucky and nvm is available on the github runners, all good

PR #2895, opened 2025-10-14
chore(c-api): add missing docs for some C function

tmontaigu:
Add explanations on the fact that config and config buider do not need to be freed/


tfhe/src/c_api/high_level_api/keys.rs
```diff
@@ -70,6 +82,12 @@ pub unsafe extern "C" fn generate_keys(
     })
 }
 
+/// Sets the given ServerKey into local storage.
+///
+/// Computations done after this call will use the given server key
+///
+/// This increases the reference count of the key, so the caller should still
+/// free the `server_key` and also use `unset_server_key` to fully delete it.
```
nsarlin-zama: shouldn't we expose a `free_server_key` function that just does a Box::from_raw but is guaranteed to use the same allocator ?

tmontaigu: Yes we should have detrou_server_key to decrease the ref count

tmontaigu: destroy_server_key already existed, I just added some unset_server_key() in the tests as it should also be used

PR #2897, opened 2025-10-14
feat(gpu): KS32 support in GPU param sets, core-crypto GPU bindings, benches and tests

andrei-stoian-zama:
Change LWE keys to 32b and output LWEs to 32b, while input stays 64b.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2897)


backends/tfhe-cuda-backend/cuda/include/keyswitch/keyswitch.h
```diff
@@ -5,12 +5,12 @@
 
 extern "C" {
 
-void cuda_keyswitch_lwe_ciphertext_vector_32_32(
+/*void cuda_keyswitch_lwe_ciphertext_vector_32_32(
```
agnesLeroy: You could remove it instead of commenting?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cu
```diff
@@ -5,7 +5,7 @@
 /* Perform keyswitch on a batch of 32 bits input LWE ciphertexts.
  * Head out to the equivalent operation on 64 bits for more details.
  */
-void cuda_keyswitch_lwe_ciphertext_vector_32_32(
+/*void cuda_keyswitch_lwe_ciphertext_vector_32_32(
```
agnesLeroy: Same here, if it's not used let's remove it altogether?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -46,10 +46,23 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
   return ptr;
 }
 
+template <typename T>
+__device__ T closest_repr(T input, uint32_t base_log, uint32_t level_count) {
+  T minus_2 = static_cast<T>(-2);
```
agnesLeroy, line 50: Maybe @IceTDrinker would be good if you could double check this logic: it looks ok to me compared to the CPU implementation of closest_representable (`tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs:25`) but these things always come with some subtleties.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -165,16 +183,36 @@ __global__ void keyswitch_gemm_copy_negated_message_with_indices(
...
+
+    // Power of two are encoded in the MSBs of the types so we need to scale
+    // the type to the other one without having to worry about the moduli
+    Torus input_to_output_scaling_factor =
+        (sizeof(Torus) - sizeof(KSTorus)) * 8;
+
+    auto rounded_downscaled_body =
+        (KSTorus)(body_out >> input_to_output_scaling_factor);
+
+    body_out = -rounded_downscaled_body;
```
agnesLeroy, line 230: Same here @IceTDrinker, maybe someone else from the CPU team knows better about KS32 but I would appreciate to have an external pair of eyes to confirm this is the correct logic.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -209,17 +247,6 @@ __global__ void keyswitch_zero_output_with_output_indices(
   buffer_in[val_idx] = 0;
 }
 
-template <typename T>
-__device__ T closest_repr(T input, uint32_t base_log, uint32_t level_count) {
```
agnesLeroy: Oh so @andrei-stoian-zama you moved this function actually, we already had it implemented. I see you changed a bit the implementation to do a static cast on -2, I think it's equivalent but better write it explicitly probably.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -446,22 +479,22 @@ __host__ void host_gemm_keyswitch_lwe_ciphertext_vector(
...
+      tgemm<KSTorus, BLOCK_SIZE_GEMM_KS, THREADS_GEMM_KS>
           <<<grid_gemm, threads_gemm, shared_mem_size, stream>>>(
               num_samples, (lwe_dimension_out + 1), lwe_dimension_in, d_mem_0,
               ksk + li * ksk_block_size, stride_KSK_buffer, lwe_array_out,
               lwe_dimension_out + 1);
       check_cuda_error(cudaGetLastError());
 
     } else {
-      tgemm_with_indices<Torus, BLOCK_SIZE_GEMM_KS, THREADS_GEMM_KS>
+      tgemm_with_indices<KSTorus, Torus, BLOCK_SIZE_GEMM_KS, THREADS_GEMM_KS>
```
agnesLeroy, line 525: Small detail but maybe it's better to always keep the same order for template arguments, Torus then KSTorus.

backends/tfhe-cuda-backend/cuda/src/linearalgebra/multiplication.cuh
```diff
@@ -203,11 +203,11 @@ __global__ void tgemm(uint M, uint N, uint K, const Torus *A, const Torus *B,
 // This code is adapted by generalizing the 1d block-tiling
 // kernel from https://github.com/siboehm/SGEMM_CUDA
 // to any matrix dimension
-template <typename Torus, int BLOCK_SIZE, int THREADS>
+template <typename Torus, typename IndicesType, int BLOCK_SIZE, int THREADS>
```
agnesLeroy, line 206: Why an IndicesType here? Couldn't we hard code indices type to uint32_t btw?

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -344,6 +345,7 @@ mod cuda {
 
     fn cuda_keyswitch_classical_and_gemm<
         Scalar: UnsignedTorus + CastInto<usize> + CastFrom<u64> + Serialize,
+        KSKSCalar: UnsignedTorus + CastFrom<Scalar>,
```
agnesLeroy: KeyswitchScalar would be a better name I think?

Makefile
```diff
@@ -16,7 +16,7 @@ FAST_TESTS?=FALSE
 BIT_SIZES_SET?=ALL
 NIGHTLY_TESTS?=FALSE
 BENCH_OP_FLAVOR?=DEFAULT
-BENCH_TYPE?=latency
+BENCH_TYPE?=throughput
```
agnesLeroy: No need to change this I guess 🙂

backends/tfhe-cuda-backend/cuda/src/linearalgebra/multiplication.cuh
andrei-stoian-zama: Indices are 64-bit, yes, we could hard-code that!

agnesLeroy: We had checked a while back on PBS and KS kernels that indices types don't affect performance, so u64 seems fine. u32 would probably be enough though, we'd run out of memory way before the max of u32.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: the static cast was needed otherwise there's a warning which triggers pcc to fail

tfhe/src/shortint/atomic_pattern/compressed/ks32.rs
```diff
@@ -73,6 +73,20 @@ impl CompressedKS32AtomicPatternServerKey {
         }
     }
 
+    pub fn into_raw_parts(
```
andrei-stoian-zama: should be removed and replaced with two methods: eg. `.key_switching_key()`

backends/tfhe-cuda-backend/cuda/src/linearalgebra/multiplication.cuh
andrei-stoian-zama: then the change must be done all the way up to the Rust code:

```           
let lwe_indexes = lwe_indexes_usize
                .iter()
                .map(|&x| <usize as CastInto<Scalar>>::cast_into(x))
                .collect_vec();
```
must be changed to `u64` in tests / benches.

if that's ok, I can do that

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
IceTDrinker: proposal for this : make those functions both device and hosts and generate tests vectors from rust or reproduce rust tests, we will never be as certain if we don't have the same test coverage I would think (including edge cases)

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
IceTDrinker: I don't understand the sign flip logic when comparing to the CPU code

IceTDrinker (commented):
The sign flip logic in the KS with type change in the cuda backend does not make sense at the moment to me @andrei-stoian-zama if you can comment on why you flip the body sign, I don't see the same thing in the CPU code

what we do for CPU as a first step is :

output_mask = 0
output_body = scale_down(round(input_body))

here you seem to flip the sign when setting the output body

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: great idea! could you point out those rust tests that I should port to GPU ?

backends/tfhe-cuda-backend/cuda/src/linearalgebra/multiplication.cuh
agnesLeroy: Yes it's ok :+1:

tfhe/src/core_crypto/algorithms/test/mod.rs
```diff
@@ -538,4 +559,6 @@ macro_rules! create_parameterized_test_with_non_native_parameters {
     };
 }
 
+#[cfg(feature = "gpu")]
```
mayeul-zama: Usually at the top of the file
I guess the previous `use` here was put here by mistake

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
IceTDrinker: anything marked #[test] in tfhe/src/core_crypto/commons/math/decomposition

IceTDrinker: can also take some of the doc test examples if that makes sense

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -165,16 +183,36 @@ __global__ void keyswitch_gemm_copy_negated_message_with_indices(
...
+  if constexpr (std::is_same_v<KSTorus, Torus>) {
+    body_out = -body_in;
+  } else {
+    body_out = closest_repr(
+        lwe_in[lwe_in_idx * (lwe_dimension_in + 1) + lwe_dimension_in],
+        sizeof(KSTorus) * 8, 1);
+
+    // Power of two are encoded in the MSBs of the types so we need to scale
+    // the type to the other one without having to worry about the moduli
+    Torus input_to_output_scaling_factor =
```
IceTDrinker, line 224: we have an assert in the rust code to only manage going from a big torus to a small KS torus, may want to have some template static asserts in C++ to verify the condition (avoids having to think about all the possible cases that may exist)

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: it's an artifact of the fact that Torus is the type of the indices. the function signature use:

```
template <typename Torus, typename IndicesType, int BLOCK_SIZE, int THREADS>
__global__ void tgemm_with_indices(uint M, uint N, uint K, const Torus *A,
```

it kind of makes sense that the Torus dtype (which will be set to KSTorus by the caller) is the first one and Indices is the second type. When moving indices to uint32 this second typename `IndicesType` will disappear.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: I added 
`static_assert(sizeof(Torus) >= sizeof(KSTorus), "Cannot compile keyswitch with given input/output dtypes");`

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: I added `test_closest_representable_gpu` and `test_round_to_closest_representable_gpu` in `tfhe/src/core_crypto/gpu/algorithms/test/lwe_keyswitch.rs`

tfhe/src/shortint/atomic_pattern/compressed/ks32.rs
agnesLeroy: This can be removed, I still see it but it's not used 🤔

andrei-stoian-zama:
> some gpu/silent shortint feature are in core in your latest patch
> 

Thanks. I think I now removed shortint references from core. Could you please check again?

With regards to the decomposers, I only implemented the "signedDecomposer" decomposition logic on GPU, as that's the one used in the KS32 implementation on CPU. Thus I only test SignedDecomposer. Would all decomposers pass the two tests I implemented ?

tfhe/src/shortint/atomic_pattern/compressed/ks32.rs
IceTDrinker: agreed

agnesLeroy: @andrei-stoian-zama it's still here

IceTDrinker: it's still here @andrei-stoian-zama the into_raw_parts function has not been removed

IceTDrinker:
*[`tfhe/src/high_level_api/keys/server.rs` line 491 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/2897#-OgHuLUp9xnGJ3a1t5c7:-OgHuLUq1_lh8W6-4RPB:bjejthz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9851f262caf50ea69dd65d8296a690b8cade140a/tfhe/src/high_level_api/keys/server.rs#L491)):*
> ```XML
>     pub fn gpu_indexes(&self) -> &[GpuIndex] {
>         match &self.key.key.key_switching_key {
>             CudaDynamicAtomicPatternKeySwitchingKey::KeySwitch32(ksk_32) => {
> ```

Still not the name I had in mind

CudaDynamicKeySwitchingKey

here it's more confusing I would think given DynamicAtomicPattern has another meaning for CPU

You can keep the original name, I was just merely pointing out the naming is maybe not the most accurate :)

PR #2898, opened 2025-10-14
chore(ci): echo post-commit checks sub-recipe names

soonum:
This is done to improve readability in case of recipe failure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,12 @@ define COVERAGE_EXCLUDED_FILES
 --exclude-files tfhe/examples/utilities/*
 endef
 
+# Prints out recipe name at the beginning of the execution and print it out again at the end if a failure occurs.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -72,6 +72,12 @@ define COVERAGE_EXCLUDED_FILES
 --exclude-files tfhe/examples/utilities/*
 endef
 
+# Prints out recipe name at the beginning of the execution and print it out again at if a failure occurs.
```
IceTDrinker: missing word after at

Makefile
```diff
@@ -72,6 +72,12 @@ define COVERAGE_EXCLUDED_FILES
...
+define run_recipe_with_details
```
mayeul-zama, line 76: Could use a shorter name
Maybe `print_name_run`

PR #2900, opened 2025-10-14
chore(gpu): coprocessor bench workflow

andrei-stoian-zama:
Fixes the installation of smart contracts in the coproc CI

### DIFF
--- a/.github/workflows/benchmark_gpu_coprocessor.yml
+++ b/.github/workflows/benchmark_gpu_coprocessor.yml
@@ -232,8 +232,9 @@ jobs:
           ls
           pwd
           cp ./host-contracts/.env.example ./host-contracts/.env
-          npm --prefix ./host-contracts ci --include=optional
-          cd host-contracts && npm install && npm run deploy:emptyProxies && npx hardhat compile
+          cd ./host-contracts 
+          npm ci --include=optional
+          npm install && npm run deploy:emptyProxies && npx hardhat compile
         working-directory: fhevm/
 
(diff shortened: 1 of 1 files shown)


backends/tfhe-cuda-backend/cuda/src/aes/aes.cuh
```diff
@@ -105,11 +105,11 @@ aes_xor(CudaStreams streams, int_aes_encrypt_buffer<Torus> *mem,
  * result.
  *
  */
-template <typename Torus>
+template <typename Torus, typename KSTorus>
```
agnesLeroy: This shouldn't be part of this PR i believe

.github/workflows/benchmark_gpu_coprocessor.yml
```diff
@@ -232,8 +232,9 @@ jobs:
           ls
           pwd
           cp ./host-contracts/.env.example ./host-contracts/.env
-          npm --prefix ./host-contracts ci --include=optional
-          cd host-contracts && npm install && npm run deploy:emptyProxies && npx hardhat compile
+          cd ./host-contracts 
+          npm ci --include=optional
+          npm install && npm run deploy:emptyProxies && npx hardhat compile
```
agnesLeroy, line 237: There was still a failure in the bench last time I checked, is this really enough?

andrei-stoian-zama: the last failure was a compilation error due to the params struct that wasn't updated in coprocessor (I pushed updates to fhevm for that)

PR #2901, opened 2025-10-15
chore(backward): backward compat data targeted generation

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1029

### PR content/description
Split backward data crate in multiple crates. We have one crate for each tfhe-rs version, that way we don't have a Cargo.toml with multiple tfhe-rs dependencies. It also allows generating data only for a single tfhe-rs version, instead of re-generating everything.

To simplify the process of adding the crate for a new version, a jinja template has been created. We can instantiate it using `make new_backward_compat_crate` if it does not already exist. This commit shows what this would do for 1.5: https://github.com/zama-ai/tfhe-rs/commit/9a3c5a531bbe2281258a95e7dc6b3045447fa1c0 (`tfhe/Cargo.toml` update is manual).

To generate the data, we still use `make gen_backward_compat_data`. However, to generate only for a single version we can now do `TFHE_VERSION=1.4 make gen_backward_compat_data`.

Maybe this can by default only generate for the current version and use something like `TFHE_VERSION=all make gen_backward_compat_data` to generate everything, wdyt ?

During the review, please also check the doc and tell me if it looks correct/understandable/etc.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901)


utils/tfhe-backward-compat-data/crates/add_new_version/src/main.rs
```diff
@@ -0,0 +1,236 @@
+//! Instantiate a new data generation crate for a version of TFHE-rs.
+//! The global `generate` crate will also be updated with this version.
```
nsarlin-zama: this is not true anymore, comment must be fixed

utils/tfhe-backward-compat-data/README.md
```diff
@@ -188,15 +204,16 @@ impl TestedModule for Hl {
...
 ```
 
-## Adding a new tfhe-rs release
-To add data for a new released version of tfhe-rs, you should first add a dependency to that version in the `Cargo.toml` of this project. This dependency should only be enabled with the `generate` feature to avoid conflicts during testing.
-
-You should then implement the `TfhersVersion` trait for this version. You can use the code in `data_0_6.rs` as an example.
-
-## Using the test data
-The data is stored using git-lfs, but they are not pulled by default. You need to pull them by running:
+## Adding a new tfhe-rs version
```
mayeul-zama, line 209: Could explain precisely which version should be used when we are between releases

It should be the next version name (which must have been updated in `tfhe/Cargo.toml`)

utils/tfhe-backward-compat-data/crates/generate_1_4/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+tfhe = { features = [
+    "boolean",
+    "integer",
+    "shortint",
+    "zk-pok",
+    "experimental-force_fft_algo_dif4",
+], path = "../../../../tfhe" }
+tfhe-versionable = { path = "../../../tfhe-versionable" }
+
+# Uncomment this and remove the lines above once the current tfhe-rs version has been released
```
mayeul-zama: Should be uncommented now

utils/tfhe-backward-compat-data/README.md
nsarlin-zama: added some explanation about that

utils/tfhe-backward-compat-data/crates/generate_1_4/Cargo.toml
nsarlin-zama: done

utils/tfhe-backward-compat-data/crates/add_new_version/src/main.rs
IceTDrinker: did you have time to update ?

nsarlin-zama: done

mayeul-zama (commented):
:lgtm:
Thanks!

> @mayeul-zama reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901#-:-OblNt3x1iSUyUjyGrZR:b-aw02zl)* status: 50 of 52 files reviewed, 6 unresolved discussions (waiting on @IceTDrinker and @soonum)

IceTDrinker (commented):
> @IceTDrinker reviewed 2 of 4 files at r2, 13 of 14 files at r3, 8 of 8 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901)* status: all files reviewed, 2 unresolved discussions (waiting on @soonum)

___
*[`utils/tfhe-backward-compat-data/crates/add_new_version/src/main.rs` line 34 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901#-OblNuhiC1ZGc-zi1gJX:-OblNuhiC1ZGc-zi1gJY:b-h573mm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/267e83d01564223dd4b7d103fcdbedf6c4f8cea9/utils/tfhe-backward-compat-data/crates/add_new_version/src/main.rs#L34)):*
> ```XML
> impl CrateVersion {
>     fn exact(&self) -> String {
>         format!("{}", self.0)
> ```

last annoyance from me: if we have alphas, don't we prefer to have major, minor patch ?

so that the extra bit of version is not kept later on ?
___
*[`utils/tfhe-backward-compat-data/crates/generate_0_11/src/lib.rs` line 10 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901#-OblNPqA2hmxeFTnHbf7:-OblNPqA2hmxeFTnHbf8:bitud39) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7f95262f4bb983e6bc7ca2b5333a9aa6f54d682b/utils/tfhe-backward-compat-data/crates/generate_0_11/src/lib.rs#L10)):*
> ```XML
> use tfhe::zk::{CompactPkeCrs, ZkComputeLoad, ZkMSBZeroPaddingBitCount};
> use tfhe::{
>     ClientKey, CompactPublicKey, ProvenCompactCiphertextList, Seed, ServerKey, set_server_key,
> ```

seeing this, are the crates in the check_fmt make target ?

IceTDrinker (approved):
Alright that's a big chunk of work but that will be extremely useful, big thank you @nsarlin-zama !



> @IceTDrinker reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2901#-:-OblSC7X91RZuQ9MYqYb:b72bwpi)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @soonum)

PR #2906, opened 2025-10-15
chore: don't import deprecated GenericArray use the aes crate Block instead

IceTDrinker:
- allow deprecated methods for now since aes 0.9 is not out yet


nsarlin-zama (approved):
thanks!

PR #2911, opened 2025-10-16
chore(bench): remove trivial encryptions

tmontaigu:
It makes benches not accurate


IceTDrinker:
thanks good to merge as is (don't want to retrigger the CI)

IceTDrinker (commented):
(adding a comment via review so that this appears as solved on my end 🙄 )

good to merge as is to not retrigger the CI

PR #2913, opened 2025-10-16
chore(ci): use trusted publishing for npm packages

soonum:


.github/workflows/make_release_tfhe.yml
IceTDrinker: no more token use for all npm publish normally I think ?

.github/workflows/make_release_tfhe.yml
```diff
@@ -83,6 +83,12 @@ jobs:
         run: |
           make build_web_js_api_parallel
 
+      - name: Authenticate on NPM
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
+        with:
+          node-version: '24'
```
IceTDrinker: it does not match our node version I believe (22) is that a problem ?

soonum: Let's use the same Node version indeed.

IceTDrinker: up to you :

proposal: make it dynamic by fetching the info in the makefile

can be done now or in a later PR, as you wish

IceTDrinker (approved):
Thanks!

Let's do the autodetection later :)

.github/workflows/make_release_tfhe.yml
```diff
@@ -83,11 +83,16 @@ jobs:
         run: |
           make build_web_js_api_parallel
 
+      - name: Authenticate on NPM
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
```
IceTDrinker, line 87: is that enough to authenticate btw ?

IceTDrinker: yep seems to be all it takes from the docs, thanks !

soonum: Yes it uses the GITHUB_TOKEN with the OICD permission to do so.

PR #2914, opened 2025-10-16
fix(gpu): fix bug in are_all_comparison_blocks_true when number of blocks is 0

bbarbakadze:
### PR content/description

fix bug in are_all_comparison_blocks_true when number of blocks is 0

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
@@ -76,6 +76,13 @@ __host__ void are_all_comparisons_block_true(
   auto message_modulus = params.message_modulus;
   auto carry_modulus = params.carry_modulus;
 
+  if (num_radix_blocks == 0) {
+    set_single_scalar_trivial_radix_ciphertext_async<Torus>(
+        streams.stream(0), streams.gpu_index(0), lwe_array_out, 1,
+        message_modulus, carry_modulus);
+    return;
+  }
+
   auto are_all_block_true_buffer =
       mem_ptr->eq_buffer->are_all_block_true_buffer;
   auto tmp_out = are_all_block_true_buffer->tmp_out;
--- a/backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cuh
@@ -207,6 +207,33 @@ __host__ void set_trivial_radix_ciphertext_async(
   }
 }
 
+// set single trivial value for a radix ciphertext
+template <typename Torus>
+__host__ void set_single_scalar_trivial_radix_ciphertext_async(
+    cudaStream_t stream, uint32_t gpu_index,
+    CudaRadixCiphertextFFI *lwe_array_out, Torus const scalar,
+    Torus message_modulus, Torus carry_modulus) {
+
+  set_zero_radix_ciphertext_slice_async<Torus>(
+      stream, gpu_index, lwe_array_out, 0, lwe_array_out->num_radix_blocks);
+  if (lwe_array_out->num_radix_blocks == 0)
+    return;
+
+  // Value of the shift we multiply our messages by
+  // If message_modulus and carry_modulus are always powers of 2 we can simplify
+  // this
+  auto nbits = sizeof(Torus) * 8;
+  Torus delta = (static_cast<Torus>(1) << (nbits - 1)) /
+                (message_modulus * carry_modulus);
+
+  cuda_set_value_async<Torus>(stream, gpu_index,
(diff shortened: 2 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -76,6 +76,13 @@ __host__ void are_all_comparisons_block_true(
   auto message_modulus = params.message_modulus;
   auto carry_modulus = params.carry_modulus;
 
+  if (num_radix_blocks == 0) {
+    set_single_trivial_radix_ciphertext_async<Torus>(
+        streams.stream(0), streams.gpu_index(0), lwe_array_out, 1,
+        message_modulus, carry_modulus);
```
agnesLeroy, line 82: No need to synchronize here?

agnesLeroy: How could num_radix_blocks be 0?

agnesLeroy: Ah in division right?

backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cuh
```diff
@@ -207,6 +207,33 @@ __host__ void set_trivial_radix_ciphertext_async(
   }
 }
 
+// set trivial value for a single radix ciphertext
+template <typename Torus>
+__host__ void set_single_trivial_radix_ciphertext_async(
```
agnesLeroy: Why do you define a new function? We have set_trivial_radix_ciphertext_async, couldn't you reuse it?

bbarbakadze: `set_trivial_radix_ciphertext_async` requires array of scalars, it means one extra copy to device, all I need is single scalar value.

agnesLeroy: Ok so the comment and function name should be edited I think, how about:
```
// set single trivial value for a radix ciphertext
```
for the comment, and 
```
set_single_scalar_trivial_radix_ciphertext_async
```
for the function name?

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
bbarbakadze: Added the synchronization before return.

backends/tfhe-cuda-backend/cuda/src/integer/radix_ciphertext.cuh
bbarbakadze: done

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -76,6 +76,14 @@ __host__ void are_all_comparisons_block_true(
   auto message_modulus = params.message_modulus;
   auto carry_modulus = params.carry_modulus;
 
+  if (num_radix_blocks == 0) {
+    set_single_scalar_trivial_radix_ciphertext_async<Torus>(
+        streams.stream(0), streams.gpu_index(0), lwe_array_out, 1,
+        message_modulus, carry_modulus);
+    cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
```
agnesLeroy: No need to synchronize here: the release function calls synchronize already:
```
  void release(CudaStreams streams) {
    release_radix_ciphertext_async(streams.stream(0), streams.gpu_index(0),
                                   tmp_out, gpu_memory_allocated);
    release_radix_ciphertext_async(streams.stream(0), streams.gpu_index(0),
                                   tmp_block_accumulated, gpu_memory_allocated);
    is_max_value->release(streams);
    delete is_max_value;
    delete tmp_out;
    delete tmp_block_accumulated;
    cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
    free(preallocated_h_lut);
  }
```

agnesLeroy: Btw I only synchronize stream 0 because up till now this stream was the main one: all others were only used for multi-gpu parts of the computations and they should always be sync'ed with stream 0 after a multi-gpu call. Do you think this is ok for the new division? I assumed it would because in your implementation the different streams also wait for stream 0. But maybe it would be safer to sync all streams in the release functions?

bbarbakadze: in new div implementation memory is created for only one gpu and stream[0] will be the one to be synchronized so it is fine.

agnesLeroy (approved):
Good, let's wait for AWS to be back on to merge (gpu pcc can't run atm)

PR #2918, opened 2025-10-16
feat(gpu): use gemm ks in AES

andrei-stoian-zama:
- Uses GEMM-based KS in AES when large batches of LWEs need to be keyswitched
- Implements GEMM KS with non-trivial indexes
- Improves KS GPU bench and KS GPU test

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2918)


backends/tfhe-cuda-backend/cuda/include/integer/rerand_utilities.h
```diff
@@ -56,6 +60,8 @@ template <typename Torus> struct int_rerand_mem {
 
     cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
 
+    streams.synchronize();
```
agnesLeroy: We had a sync on stream 0 at the line just above, maybe it's enough?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -10,8 +12,29 @@
...
+  Torus *d_buffer;
+  uint64_t num_lwes;
+  uint32_t lwe_dimension;
+};
+
+template <typename Torus>
+uint64_t scratch_cuda_keyswitch_size(uint32_t lwe_dimension_in,
+                                     uint32_t lwe_dimension_out,
+                                     uint32_t num_lwes) {
+  return (uint64_t)num_lwes * std::max(lwe_dimension_in, lwe_dimension_out) *
```
agnesLeroy: Here we could keep `lwe_dimension_in` which is the large one, instead of the max

backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -389,13 +389,16 @@ __host__ void host_integer_partial_sum_ciphertexts_vec(
                                       needs_processing);
 
     auto active_streams = streams.active_gpu_subset(total_ciphertexts);
+    GPU_ASSERT(total_ciphertexts <= mem_ptr->luts_message_carry->num_blocks,
```
agnesLeroy, line 392: Should be a PANIC_IF_FALSE in my opinion, GPU_ASSERT will only trigger in debug mode

backends/tfhe-cuda-backend/cuda/src/device.cu
```diff
@@ -1,4 +1,5 @@
 #include "device.h"
+
```
agnesLeroy: No need for new line

backends/tfhe-cuda-backend/cuda/include/aes/aes_utilities.h
```diff
@@ -37,6 +37,7 @@ template <typename Torus> struct int_aes_lut_buffers {
     auto active_streams_and_lut = streams.active_gpu_subset(
         SBOX_MAX_AND_GATES * num_aes_inputs * sbox_parallelism);
     this->and_lut->broadcast_lut(active_streams_and_lut);
+    this->and_lut->setup_gemm_batch_ks_temp_buffers(size_tracker);
```
agnesLeroy, line 40: Looks strange to have to set this up manually, but at least we're sure it can only be activated in AES, so let's keep it like this for now.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -443,6 +446,28 @@ struct int_radix_lut_custom_input_output {
                                allocate_gpu_memory);
   }
 
+  void setup_gemm_batch_ks_temp_buffers(uint64_t &size_tracker) {
+
+    auto inputs_on_gpu = std::max(
```
agnesLeroy: This formula has now changed on main, it is:
```
auto num_blocks_on_gpu =
          std::min((int)num_radix_blocks,
                   std::max(THRESHOLD_MULTI_GPU,
                            get_num_inputs_on_gpu(num_radix_blocks, i,
                                                  active_streams.count())));
```

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -22,6 +45,202 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+}
+
+// Decompose an array of LWEs with indirection through lwe_input_indices
+// The LWE array can contain total_lwe LWEs where total_lwe can be different
+// from num_lwe. The maximum index should be <= total_lwe. num_lwe is the number
+// of LWEs to decompose The output buffer should have space for num_lwe LWEs.
+// These will be sorted according to the input indices.
+template <typename Torus>
+__global__ void decompose_vectorize_init_with_indices(
+    Torus const *lwe_in, const Torus *__restrict__ lwe_input_indices,
```
agnesLeroy, line 91: Ideally we should have output indexes as well, since the bootstrap logic makes it possible to shuffle inputs but also outputs.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -22,6 +45,202 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+
+  if (lwe_idx >= num_lwe || lwe_sample_idx >= lwe_dimension)
+    return;
+
+  // Input LWE array is [mask_0, .., mask_lwe_dim, message] and
+  // we only decompose the mask. Thus the stride for reading
+  // is lwe_dimension + 1, while for writing it is lwe_dimension
+  auto read_val_idx =
+      lwe_input_indices[lwe_idx] * (lwe_dimension + 1) + lwe_sample_idx;
+  auto write_val_idx = lwe_idx * lwe_dimension + lwe_sample_idx;
```
agnesLeroy, line 108: Should use output indices

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -22,6 +45,202 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+
+  Torus mod_b_mask = (1ll << base_log) - 1ll;
+
+  buffer_in[val_idx] = decompose_one<Torus>(state, mod_b_mask, base_log);
+  __syncthreads();
+  buffer_in[state_idx] = state;
+}
+
+template <typename Torus>
+__global__ void keyswitch_gemm_copy_message(const Torus *lwe_in, Torus *lwe_out,
```
agnesLeroy: Maybe `keyswitch_gemm_copy_negated_message`?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -22,6 +45,202 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+
+  if (lwe_id >= num_lwes)
+    return;
+
+  lwe_out[lwe_id * (lwe_dimension_out + 1) + lwe_dimension_out] =
+      -lwe_in[lwe_id * (lwe_dimension_in + 1) + lwe_dimension_in];
+}
+
+template <typename Torus>
+__global__ void keyswitch_gemm_copy_message_with_indices(
```
agnesLeroy: `keyswitch_gemm_copy_negated_message_with_indices`

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -22,6 +45,202 @@ __device__ Torus *get_ith_block(Torus *ksk, int i, int level,
...
+  auto lwe_sample_idx = blockIdx.x * blockDim.x + threadIdx.x;
+  // index of the LWE sample in the LWE ct
+  auto lwe_idx = blockIdx.y * blockDim.y + threadIdx.y;
+
+  if (lwe_idx >= num_lwe || lwe_sample_idx >= lwe_size)
+    return;
+
+  auto val_idx = lwe_output_indices[lwe_idx] * lwe_size + lwe_sample_idx;
+
+  buffer_in[val_idx] = 0;
```
agnesLeroy, line 209: Would this work inplace really if indices overlap? 🤔

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -142,14 +361,160 @@ __host__ void host_keyswitch_lwe_ciphertext_vector(
...
+    cuda_memset_async(lwe_array_out, 0,
+                      num_samples * (lwe_dimension_out + 1) * sizeof(Torus),
+                      stream, gpu_index);
+  } else {
+    // gemm to ks the individual LWEs to GLWEs
+    dim3 grid_zero(CEIL_DIV(lwe_dimension_out + 1, BLOCK_SIZE_DECOMP),
+                   CEIL_DIV(num_samples, BLOCK_SIZE_DECOMP));
+    dim3 threads_zero(BLOCK_SIZE_DECOMP, BLOCK_SIZE_DECOMP);
+
+    keyswitch_zero_output_with_output_indices<Torus>
```
agnesLeroy, line 355: I think you could just memset to 0 in the case with indices no?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -142,14 +361,160 @@ __host__ void host_keyswitch_lwe_ciphertext_vector(
...
+  // lwe_array_out is num_samples x (lwe_dimension_out + 1). copy the bodies
+  // lwe_array_in[:,lwe_dimension_in] to lwe_array_out[:,lwe_dimension_out]
+  // and negate
+  if (uses_trivial_indices) {
+    keyswitch_gemm_copy_message<Torus><<<grid_copy, threads_copy, 0, stream>>>(
+        lwe_array_in, lwe_array_out, lwe_dimension_in, num_samples,
+        lwe_dimension_out);
+    check_cuda_error(cudaGetLastError());
+  } else {
+    keyswitch_gemm_copy_message_with_indices<Torus>
```
agnesLeroy: Couldn't we keep only this version? I don't expect it should affect performance to copy with indices

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -142,14 +361,160 @@ __host__ void host_keyswitch_lwe_ciphertext_vector(
...
+  // here that the LWE dimension is a multiple of the block size
+  dim3 grid_decomp(CEIL_DIV(num_samples, BLOCK_SIZE_DECOMP),
+                   CEIL_DIV(lwe_dimension_in, BLOCK_SIZE_DECOMP));
+  dim3 threads_decomp(BLOCK_SIZE_DECOMP, BLOCK_SIZE_DECOMP);
+
+  uint32_t shared_mem_size = get_shared_mem_size_tgemm<Torus>();
+  // Shared memory requirement is 4096, 8192, and 16384 bytes respectively for
+  // 32, 64, and 128-bit Torus elements
+  // Sanity check: the shared memory size is a constant defined by the algorithm
+  GPU_ASSERT(shared_mem_size <= 1024 * sizeof(Torus),
```
agnesLeroy, line 386: Should be PANIC_IF_FALSE in my opinion

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -142,14 +361,160 @@ __host__ void host_keyswitch_lwe_ciphertext_vector(
...
+  dim3 threads_gemm(BLOCK_SIZE_GEMM_KS * THREADS_GEMM_KS);
+
+  // decompose first level (skips the body in the input buffer)
+  if (uses_trivial_indices) {
+    decompose_vectorize_init<Torus><<<grid_decomp, threads_decomp, 0, stream>>>(
+        lwe_array_in, fp_tmp_buffer, lwe_dimension_in, num_samples, base_log,
+        level_count);
+    check_cuda_error(cudaGetLastError());
+  } else {
+    decompose_vectorize_init_with_indices<Torus>
```
agnesLeroy: Same for this kernel, we might be able to keep only this variant with indices

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -142,14 +361,160 @@ __host__ void host_keyswitch_lwe_ciphertext_vector(
...
+  // gemm to ks the individual LWEs to GLWEs
+  dim3 grid_negate(CEIL_DIV(lwe_dimension_out + 1, BLOCK_SIZE_DECOMP),
+                   CEIL_DIV(num_samples, BLOCK_SIZE_DECOMP));
+  dim3 threads_negate(BLOCK_SIZE_DECOMP, BLOCK_SIZE_DECOMP);
+  // Negate all outputs in the LWE
+  if (uses_trivial_indices) {
+    keyswitch_negate<Torus><<<grid_negate, threads_negate, 0, stream>>>(
+        lwe_array_out, lwe_dimension_out + 1, num_samples);
+  } else {
+    keyswitch_negate_with_output_indices<Torus>
```
agnesLeroy: Here as well most probably

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -164,12 +529,39 @@ void execute_keyswitch_async(CudaStreams streams,
...
 
-    // Compute Keyswitch
-    host_keyswitch_lwe_ciphertext_vector<Torus>(
-        streams.stream(i), streams.gpu_index(i), current_lwe_array_out,
-        current_lwe_output_indexes, current_lwe_array_in,
-        current_lwe_input_indexes, ksks[i], lwe_dimension_in, lwe_dimension_out,
-        base_log, level_count, num_samples_on_gpu);
+    if (!fp_tmp_buffer.empty() &&
+        num_samples_on_gpu >= get_threshold_ks_gemm()) {
+      GPU_ASSERT(fp_tmp_buffer.size() >= streams.count(),
```
agnesLeroy, line 483: I would use PANIC_IF_FALSE instead of asserts, otherwise we won't get nice failures at runtime without debug flags.

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -423,12 +423,89 @@ mod cuda {
...
+                                    true,
                                     &streams,
                                 );
+
                                 black_box(&mut ct_gpu);
                             })
                         });
                     }
+
+                    for uses_simple_indices in [false, true] {
```
agnesLeroy: Maybe this additional bench should be just the one with non-trivial indices, since the case with trivial indexes is handled just above?

agnesLeroy: Also I see you only added the non trivial indexes case for latency but probably what makes more sense is to add it for thorughput for gemm keyswitch?

Makefile
```diff
@@ -781,7 +781,7 @@ build_debug_integer_short_run_gpu: install_rs_check_toolchain install_cargo_next
 	RUSTFLAGS="$(RUSTFLAGS)" cargo $(CARGO_RS_BUILD_TOOLCHAIN) test --profile debug_lto_off \
 		--features=integer,gpu-debug-fake-multi-gpu -p tfhe -- integer::gpu::server_key::radix::tests_long_run::test_random_op_sequence::test_gpu_short_random --list
 	@echo "To debug fake-multi-gpu short run tests run:"
-	@echo "TFHE_RS_TEST_LONG_TESTS_MINIMAL=TRUE <executable> integer::gpu::server_key::radix::tests_long_run::test_random_op_sequence::test_gpu_short_random_op_sequence_param_gpu_multi_bit_group_4_message_2_carry_2_ks_pbs_tuniform_2m128 --nocapture"
+	@echo "TFHE_RS_LONGRUN_TESTS_SEED=<SEED_FROM_CI> TFHE_RS_TEST_LONG_TESTS_MINIMAL=TRUE <executable> integer::gpu::server_key::radix::tests_long_run::test_random_op_sequence::test_gpu_short_random_op_sequence_param_gpu_multi_bit_group_4_message_2_carry_2_ks_pbs_tuniform_2m128 --nocapture"
```
agnesLeroy, line 779: Maybe we shouldn't do this change here?

tfhe/src/core_crypto/gpu/algorithms/test/lwe_keyswitch.rs
```diff
@@ -19,6 +21,57 @@ fn lwe_encrypt_ks_decrypt_custom_mod<Scalar: UnsignedTorus + CastFrom<usize>>(
     let ks_decomp_base_log = params.ks_base_log;
     let ks_decomp_level_count = params.ks_level;
 
+    base_lwe_encrypt_ks_decrypt_custom_mod(
```
agnesLeroy, line 24: So here we only test one parameter set with one decomposition level, we should probably test for more levels. 
Params are here: `tfhe/src/core_crypto/algorithms/test/mod.rs:56`

Maybe you could instantiate tests with
```
create_gpu_multi_bit_parameterized_test!(lwe_encrypt_ks_decrypt_custom_mod);
```
Because then we would test 
```
MULTI_BIT_2_2_2_PARAMS,
MULTI_BIT_2_2_3_PARAMS,
MULTI_BIT_3_3_2_PARAMS,
MULTI_BIT_3_3_3_PARAMS
```
and some of them have 2 levels (3_3_2 for example)

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: this kernel only decomposes the input to a temporary buffer. the gemm kernel takes care of output indices

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: I assumed indices are unique, both input and output, are they not ?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: Ah yes ok I had missed that

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: They should be, actually. We don't have a check to ensure that though. Maybe something we should add when generating indices.

tfhe/src/core_crypto/gpu/algorithms/test/lwe_keyswitch.rs
andrei-stoian-zama: we have `lwe_encrypt_ks_decrypt_custom_mod_mb` which is instantiated with multi-bit params. it calls `base_lwe_encrypt_ks_decrypt_custom_mod` (also called by the non-multi-bit params test) which is the function that I updated to use all combinations of indices/sizes

agnesLeroy: Ok sorry I missed the line where you add the multi-bit tests.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: Currently GPU_ASSERT is enabled in `release` as well. I think GPU_ASSERT should be used to test algorithm assumptions that do not rely on user inputs - i.e. the algorithm is correctly implemented. Here I'm checking that sufficient buffers (one per gpu) were created by the programmer - no user input is relied upon. In the future, we can disable GPU_ASSERT in release when putting GPU tfhe-rs in production. 

Let me know if that is consistent, we can take a different approach if needed.

backends/tfhe-cuda-backend/cuda/include/aes/aes_utilities.h
andrei-stoian-zama: In this way we're sure which ops use GEMM KS instead of relying on some automatic mechanism (solely based on LWE batch size).

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: We can't do it with only 1 memset, because the code is potentially setting only some LWEs (e.g indices 1,5,10 out of a buffer of 16 LWEs) of the output LWE array to 0. So either I do many `memset` calls (I fear launch contention for this) or I use a custom kernel.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: we use `decompose_vectorize_init` in packing keyswitch. I can only call the one with indices here, but I'll have to keep the trivial indexes version anyway, unless I change packing KS.

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
andrei-stoian-zama: I removed the non-indices one

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: Hmm in case we call this with indexes that span only a subset of input LWEs, we should actually copy the input to the output array for indexes that are untouched? How is this handled in the PBS?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: Ah yes ok, let's not change packing keyswitch then and keep this one :+1:

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: Ah ok, I thought GPU_ASSERT would only trigger an error in debug mode. If it also triggers in release mode then ok let's keep it like this. Not sure we'll convert these asserts to debug only any time soon, it feels much safer to have them systematically and shouldn't be introducing any slowdown compared to FHE computations.

Makefile
andrei-stoian-zama: it's only a comment that I improved to help me out during debugging

backends/tfhe-cuda-backend/cuda/include/integer/rerand_utilities.h
andrei-stoian-zama: you're right, this was something I forgot to remove.

Makefile
agnesLeroy: Ah yes ok makes sense

tfhe-benchmark/benches/core_crypto/ks_bench.rs
agnesLeroy: For latency you still run twice the trivial indexes case it seems

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
agnesLeroy: In the PBS case we just memset the whole output to 0. I'm not sure we have a use case right now with indexes that don't span the whole output buffer. Anyway it would be up to the user to copy the input to the output at the right places after the operation, that makes sense to me.

agnesLeroy: So as discussed, you're right to have this kernel we need to keep it :+1:

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -423,120 +423,148 @@ mod cuda {
...
-                                    );
-                                    CudaLweCiphertextList::from_lwe_ciphertext_list(
-                                        &input_ks_list,
-                                        &local_streams[i],
-                                    )
-                                })
-                                .collect::<Vec<_>>();
+                    for uses_simple_indices in [false, true] {
+                        bench_id = format!(
+                            "{bench_name}::throughput::simple_{uses_simple_indices}::{name}"
```
agnesLeroy: In the bench name you can specify whether it's gemm ks or classical keyswitch that's benched

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -423,120 +423,148 @@ mod cuda {
...
-                                        big_lwe_sk.lwe_dimension().to_lwe_size(),
-                                        params.ciphertext_modulus.unwrap(),
-                                    );
-                                    CudaLweCiphertextList::from_lwe_ciphertext_list(
-                                        &input_ks_list,
-                                        &local_streams[i],
-                                    )
-                                })
-                                .collect::<Vec<_>>();
+                    for uses_simple_indices in [false, true] {
```
agnesLeroy: I would rename simple -> trivial to match the naming used in the backend. Also we'd need to have 2 throughput benches: 1 with classical KS, 1 with GEMM KS. Here it seems to me we only run the gemm one.

andrei-stoian-zama: PCC check won't let me use the word "trivial". I can maybe use "triv" or "trivi_al" to fool it

agnesLeroy: Ah :sob: ok let's keep simple then :+1:

tfhe-benchmark/benches/core_crypto/ks_bench.rs
```diff
@@ -423,120 +423,148 @@ mod cuda {
...
+                                        .for_each(|(((i, input_ct), output_ct), local_stream)| {
+                                            cuda_keyswitch_lwe_ciphertext(
+                                                gpu_keys_vec[i].ksk.as_ref().unwrap(),
+                                                input_ct,
+                                                output_ct,
+                                                &cuda_indexes_vec[i].d_input,
+                                                &cuda_indexes_vec[i].d_output,
+                                                true,
+                                                local_stream,
+                                                true,
```
agnesLeroy: I think one comment got lost: for the throughput bench we'd need to bench both gemm & classical KS, with different names for the benchmarks.

tfhe-benchmark/benches/core_crypto/ks_bench.rs
andrei-stoian-zama: it's fixed now I think

PR #2920, opened 2025-10-16
chore(docs): update benchmark results for all backends

soonum:
This also removes tables in PBS benchmarks for failure probability of 2**-40.


IceTDrinker (approved):

agnesLeroy:
We should also merge this on main no?

PR #2923, opened 2025-10-17
chore(ci): show action run url on custom regression benchmark

soonum:
When triggered via issue comment, a performance regression benchmark doesn't appear in the check suite. To ease tracking, a comment is created which display URL of the workflow run.


IceTDrinker (approved):
thanks !

IceTDrinker:
ah this needs to be merged to be active

PR #2929, opened 2025-10-20
chore: use pull request permissions to be able to post comments

IceTDrinker:
- the "print URL" comment failed, while the "bench failed" comment worked difference is the former has an issues write permissions, and the latter a pull request write permissions

see https://github.com/zama-ai/tfhe-rs/pull/2928


IceTDrinker:
fix confirmed https://github.com/zama-ai/tfhe-rs/pull/2928#issuecomment-3421733716

PR #2932, opened 2025-10-20
chore(gpu): remove device synchronize in drop for cudavec

agnesLeroy:

### Check-list:

### DIFF
--- a/tfhe/src/core_crypto/gpu/vec.rs
+++ b/tfhe/src/core_crypto/gpu/vec.rs
@@ -1,6 +1,6 @@
 use super::get_number_of_gpus;
 use crate::core_crypto::gpu::slice::{CudaSlice, CudaSliceMut};
-use crate::core_crypto::gpu::{synchronize_device, CudaStreams};
+use crate::core_crypto::gpu::CudaStreams;
 use crate::core_crypto::prelude::Numeric;
 use std::collections::Bound::{Excluded, Included, Unbounded};
 use std::ffi::c_void;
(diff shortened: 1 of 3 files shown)


tfhe/src/core_crypto/gpu/vec.rs
```diff
@@ -485,12 +485,10 @@ unsafe impl<T> Send for CudaVec<T> where T: Send + Numeric {}
 unsafe impl<T> Sync for CudaVec<T> where T: Sync + Numeric {}
 
 impl<T: Numeric> Drop for CudaVec<T> {
-    /// Free memory for pointer `ptr` synchronously
+    /// Free memory on GPU for pointers in `ptr`
     fn drop(&mut self) {
         for (i, &ptr) in self.ptr.iter().enumerate() {
-            // Synchronizes the device to be sure no stream is still using this pointer
             let gpu_index = self.gpu_indexes[i];
-            synchronize_device(gpu_index.0);
```
guillermo-oyarzun, line 493: Can we remove the synchronize device from the clone too?

agnesLeroy: I don't see how, we use different streams for to_cpu, to_gpu and the computation in the HL API

agnesLeroy: We had to implement Clone to satisfy some Rust trait but actually it should never be called, we can call duplicate which takes streams as input and doesn't synchronize device

guillermo-oyarzun: ahh you are right! i wasn't considering that case.

PR #2934, opened 2025-10-20
feat(gpu): extend specialized version to classical pbs

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
@@ -43,9 +43,8 @@ bool has_support_to_cuda_programmable_bootstrap_tbc(
                                       max_shared_memory);
   case 2048:
     return supports_thread_block_clusters_on_classic_programmable_bootstrap<
-        Torus, AmortizedDegree<2048>>(num_samples, glwe_dimension,
-                                      polynomial_size, level_count,
-                                      max_shared_memory);
+        Torus, Degree<2048>>(num_samples, glwe_dimension, polynomial_size,
+                             level_count, max_shared_memory);
   case 4096:
     return supports_thread_block_clusters_on_classic_programmable_bootstrap<
         Torus, AmortizedDegree<4096>>(num_samples, glwe_dimension,
@@ -96,7 +95,7 @@ uint64_t scratch_cuda_programmable_bootstrap_tbc(
         glwe_dimension, polynomial_size, level_count,
         input_lwe_ciphertext_count, allocate_gpu_memory, noise_reduction_type);
   case 2048:
-    return scratch_programmable_bootstrap_tbc<Torus, AmortizedDegree<2048>>(
+    return scratch_programmable_bootstrap_tbc<Torus, Degree<2048>>(
         static_cast<cudaStream_t>(stream), gpu_index, pbs_buffer, lwe_dimension,
         glwe_dimension, polynomial_size, level_count,
         input_lwe_ciphertext_count, allocate_gpu_memory, noise_reduction_type);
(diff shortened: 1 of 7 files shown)


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -96,7 +95,7 @@ uint64_t scratch_cuda_programmable_bootstrap_tbc(
         glwe_dimension, polynomial_size, level_count,
         input_lwe_ciphertext_count, allocate_gpu_memory, noise_reduction_type);
   case 2048:
-    return scratch_programmable_bootstrap_tbc<Torus, AmortizedDegree<2048>>(
+    return scratch_programmable_bootstrap_tbc<Torus, Degree<2048>>(
```
agnesLeroy, line 98: So it's better with Degree than AmortizedDegree?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_tbc_classic.cuh
```diff
@@ -340,14 +539,42 @@ __host__ void host_programmable_bootstrap_tbc(
...
-        supports_dsm, num_many_lut, lut_stride, buffer->noise_reduction_type));
+    if (polynomial_size == 2048 && level_count == 1 && glwe_dimension == 1 &&
+        base_log == 23) {
+      uint64_t full_sm_2_2 =
+          get_buffer_size_full_sm_programmable_bootstrap_tbc_2_2_params<Torus>(
+              polynomial_size);
+
+      config.dynamicSmemBytes = full_sm_2_2;
+
+      check_cuda_error(cudaFuncSetAttribute(
```
agnesLeroy, line 532: Aren't we setting these twice?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -159,6 +159,33 @@ __device__ void multiply_by_monomial_negacyclic_and_sub_polynomial(
...
+ * Receives num_poly  concatenated polynomials of type T. For each:
+ *
+ * Performs result_acc = acc * (X^ä - 1) - acc
+ * takes single buffer as input and returns a single rotated buffer
+ * result_acc must be in registers
+ * acc must be in shared memory
+ *  By default, it works on a single polynomial.
+ */
+template <typename T, int elems_per_thread, int block_size>
+__device__ void multiply_by_monomial_negacyclic_and_sub_polynomial_in_regs(
```
agnesLeroy, line 172: We didn't have this one already? I feel like i've seen this before :thinking:

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
guillermo-oyarzun: I did test it for erc20 mainly, and it seemed that 512 threads is better. Basically using 256 (amortizedDegree) reduces the occupancy

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
guillermo-oyarzun: So this actually doesn't exists in the multi-bit, since it is part of the keybundle and that one is very tuned so it is not fully reusable. i just did a small change with respect to the one in standard classical only to keep everything in registers as i did in multi-bit specialized. This is one of the functions that we might optimize further by using ideas of the keybundle, but i leave those for a future if classical is ever needed, because the ideas can not just be copy pasted since would involve changes in the next functions too.

agnesLeroy (approved):

PR #2935, opened 2025-10-20
Am/test/ks32 rerand noise

IceTDrinker:
Mostly : new trait to manage uncorrelated noise additions for noise simulation

Add primitives "with_engine" in shortint to avoid thread borrow crashes

Make things work for KS32

Add ReRand AP (has sanity check) added a function to generate the CPK 0 for ReRand which allows to verify that implem match

Removed functions for sanity checks that were just doing a PBS after a chain of ops, moved the PBS to be done manually to save on overall quantity of lines

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2935)


IceTDrinker:
as discussed I only rebased + made a small change, the larger ones will be done in a follow up PR to come soon

PR #2938, opened 2025-10-23
fix: Tag propagation in XofKeySet

tmontaigu:
Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/1181


IceTDrinker (approved):
Thanks !

Let's plan a backport for 1.4, wdyt ? if so prepare a 1.4.2 I would think

PR #2958, opened 2025-10-29
chore(ci): refactor hpu benchmarks workflows

soonum:
Following the same pattern as GPU benchmarks, HPU benchmarks rely on a common workflow. All the manual launches via workflow_dispatch event are now done in one place. That way, one doesn't have to browse the workflow tree to find the right HPU benchmark to trigger.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2958)


pgardratzama (approved):
looks great but can I try it?
poor github does give me only the workflow from main so it is difficult to test these new workflow before we merge...
also we did not solve the issue on hpu integer bench which cannot not publish its result to grafana because of conflict on left_shift vs shift_left

PR #2962, opened 2025-10-29
fix(shortint): avoid to crash when thread engine is reused

nsarlin-zama:

### PR content/description
We used to crash when shortint engine was reused in the same thread. However this may happen if we do rayon in rayon with an engine borrow in between, eg: par_iter -> engine borrow_mut -> par_iter.
In that case, we return a fresh engine instead.

I added a test for this that crashes before this pr.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2962)

### DIFF
--- a/tfhe/src/shortint/engine/mod.rs
+++ b/tfhe/src/shortint/engine/mod.rs
@@ -311,7 +311,27 @@ impl ShortintEngine {
     where
         F: FnOnce(&mut Self) -> R,
     {
-        LOCAL_ENGINE.with(|engine_cell| func(&mut engine_cell.borrow_mut()))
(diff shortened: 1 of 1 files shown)


IceTDrinker (commented):
could you try running a regression benchmark just so that we get familiarized with the process ?

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2962#-:-OcoV3u60FRzeSPrn_Nu:bcyaxx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mayeul-zama and @tmontaigu)

PR #2972, opened 2025-11-03
fix: remove redundant Clone bound from get()

VolodymyrBg:
The Clone bound on K in KeyCache<P, K, S>::get was unnecessary.
The method forwards &mut K::from to get_with_closure and returns SharedKey<K>,which is Arc<OnceLock<K)>-backed and does not require K: Clone. Removing the bound relaxes trait requirements without changing behavior or API guarantees.

Notes:
- In-memory cache holds SharedKey<K>, not K, so no cloning occurs.
- PersistentStorage only needs (de)serialization, not cloning.


IceTDrinker:
done in https://github.com/zama-ai/tfhe-rs/pull/2975

PR #2974, opened 2025-11-03
chore: make typos version fixed

IceTDrinker:
- add a script to properly install the correct version

### DIFF
--- a/scripts/install_typos.sh
+++ b/scripts/install_typos.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+
+set -e
+
+rust_toolchain=""
+required_typos_version=""
+
+function usage() {
+    echo "$0: install typos-cli"
+    echo
+    echo "--help                    Print this message"
+    echo "--rust-toolchain          The toolchain to check the version for with leading"
+    echo "--typos-version           Version of typos-cli to install"
+    echo
+}
+
+while [ -n "$1" ]
+do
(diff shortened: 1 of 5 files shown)


scripts/install_typos.sh
```diff
@@ -0,0 +1,64 @@
...
+
+ver_string="$(typos --version | cut -d ' ' -f 2)"
+
+ver_major="$(echo "${ver_string}" | cut -d '.' -f 1)"
+ver_minor="$(echo "${ver_string}" | cut -d '.' -f 2)"
+
+min_ver_major="$(echo "${required_typos_version}" | cut -d '.' -f 1)"
+min_ver_minor="$(echo "${required_typos_version}" | cut -d '.' -f 2)"
+
+if [[ "${ver_major}" -gt "${min_ver_major}" ]]; then
```
mayeul-zama, line 57: Shouldn't we at least print a warning if the version is not exactly the same?
New versions may allow words that were not allowed before

IceTDrinker: true but that should not be a problem ? given the CI will always have the version we set in the Makefile ?

mayeul-zama: To reproduce locally
Otherwise no need to make any version check

IceTDrinker: new versions will be more strict so I think it's still fine

PR #2978, opened 2025-11-04
docs: add documentation for hash_gamma generation in domain separator upgrades

crStiv:
Add documentation for hash_gamma generation in domain separator upgrades.

Documents why specific domain separators are hashed in a fixed order to generate deterministic hash_gamma values. 
This prevents accidental modifications that could  break backward compatibility and ensures reproducible proof verification when 
upgrading V0 to V1 domain separators.

### DIFF
--- a/tfhe-zk-pok/src/backward_compatibility/mod.rs
+++ b/tfhe-zk-pok/src/backward_compatibility/mod.rs
@@ -151,10 +151,24 @@ pub struct SerializablePKEv2DomainSeparatorsV0 {
 impl Upgrade<SerializablePKEv2DomainSeparators> for SerializablePKEv2DomainSeparatorsV0 {
     type Error = Infallible;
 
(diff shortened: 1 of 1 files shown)


crStiv:
Hi @nsarlin-zama, thanks for the review!

I added documentation because the current code has implicit constraints that aren't obvious. 
The order of fields in the hash is fixed for reproducibility, and the field selection (excluding PKEv2-specific fields) isn't documented. Without this context, someone might accidentally change the order or add/remove fields, which would break compatibility with existing V0 data.

If you think it's too much, I can trim it down to just the critical parts. What do you think?

nsarlin-zama:
Thanks, but I think the existing comment (that mentions reproducibility) is sufficient for our purposes. In general, the implementation of the `upgrade()` method for all data types should not be modified without proper justification. Therefore, this should raise suspicions upon review.

PR #2985, opened 2025-11-06
Am/test/noise check compression

IceTDrinker:
added a pfail estimate script in the scripts dir if people want to have a quick tool to check and see what some results might be when doing some noise measurements

added tests for the compression atomic pattern, had to update some structs to be more dynamic for noise simulation (like being able to handle ciphertext moduli on u64 or u128, notably for the packing ksk)

one change compared to other tests is the iteration count which are adapted here to generate the same number of keys as the classic PBS ones for the pfail tests, noise checks are still lower in overall keys being generated because some confidence interval estimation functions can return NaN for large amounts of samples

lifted restrictions on the confidence interval, if the computation does not return finite f64 values we panic with an indication of what should be done to the user

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2985)

### DIFF
--- a/tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs
+++ b/tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs
@@ -2,8 +2,8 @@ use super::utils::noise_simulation::*;
 use super::utils::traits::*;
 use super::utils::{
     expected_pfail_for_precision, mean_and_variance_check, normality_check, pfail_check,
-    precision_with_padding, DecryptionAndNoiseResult, NoiseSample, PfailAndPrecision,
-    PfailTestMeta, PfailTestResult,
+    precision_with_padding, update_ap_params_msg_and_carry_moduli, DecryptionAndNoiseResult,
+    NoiseSample, PfailAndPrecision, PfailTestMeta, PfailTestResult,
 };
 use super::{should_run_short_pfail_tests_debug, should_use_single_key_debug};
 use crate::shortint::atomic_pattern::AtomicPattern;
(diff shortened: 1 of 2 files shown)


mayeul-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/2985)* status: 14 of 15 files reviewed, 3 unresolved discussions (waiting on @IceTDrinker)

___
*[`-- commits` line 12 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2985#-OdxgvPZAxB4FEgltnnl:-Oe1hSnT9F5rP3RUOrDb:b32w1t):*

yep, I opened the issue and updated the commit, will repush with the other changes
</blockquote></details>

Ok
___
*[`tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs` line 166 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2985#-Odxu9B89fvHgXBG6RKq:-Oe1hClI-v0672UfMs4T:bb74ce2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2e181bf6ceaa07bff71ad9a33148b1b17d6d0573/tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs#L166)):*

fair, is it ok with you if it's done in a follow up PR where I would do it for the various noise tests instead of making this PR even bigger than it is ?
</blockquote></details>

Sure
___
*[`tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs` line 642 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/2985#-OdxwVgB3Mgav8DfZTAm:-Oe1hRt0506XHuMalcOJ:b32w1t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2e181bf6ceaa07bff71ad9a33148b1b17d6d0573/tfhe/src/shortint/server_key/tests/noise_distribution/br_dp_packingks_ms.rs#L642)):*

I could add a test primitive to be able to update the message mod yes
</blockquote></details>

Ok

PR #2987, opened 2025-11-06
chore: comment is erroneous and could lead to confusion

IceTDrinker:
- function is document and behavior is checked in a test

### DIFF
--- a/tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs
+++ b/tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs
@@ -176,7 +176,9 @@ where
         let mod_mask = Scalar::MAX >> (Scalar::BITS - rep_bit_count);
         res &= mod_mask;
         // Control bit about whether we should balance the state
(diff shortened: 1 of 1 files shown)


tfhe/src/core_crypto/commons/math/decomposition/decomposer.rs
```diff
@@ -176,7 +176,8 @@ where
         let mod_mask = Scalar::MAX >> (Scalar::BITS - rep_bit_count);
         res &= mod_mask;
         // Control bit about whether we should balance the state
-        // This is equivalent to res > 2^(base_log * l) || (res == 2^(base_log * l) && random == 1)
+        // This is equivalent to:
+        // res > 2^(base_log * l) / 2 || (res == 2^(base_log * l) / 2 && random == 1)
```
mayeul-zama: IMO Would be easier to read as
"res > B / 2 || (res == B / 2 && random == 1)
    with B = 2^(base_log * l)"

IceTDrinker: can do and you are right it's more readable

PR #2989, opened 2025-11-07
refactor(gpu): match_value to backend

enzodimaria:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
@@ -1,0 +1,595 @@
+#pragma once
+#include "integer/comparison.h"
+#include "integer/radix_ciphertext.cuh"
+#include "integer_utilities.h"
+#include <functional>
+#include <vector>
+
+const uint32_t MAX_STREAMS_FOR_VECTOR_FIND = 10;
+
+template <typename Torus> struct int_equality_selectors_buffer {
+  int_radix_params params;
+  bool allocate_gpu_memory;
+  uint32_t lut_stride;
+
+  uint32_t num_possible_values;
+  int_radix_lut<Torus> *comparison_luts;
+  CudaRadixCiphertextFFI *tmp_many_luts_output;
+
+  CudaStreams active_streams;
+  std::vector<CudaStreams> sub_streams_vec;
+  cudaEvent_t incoming_event;
+  std::vector<std::vector<cudaEvent_t>> outgoing_events;
+  uint32_t num_streams;
+
+  std::vector<CudaRadixCiphertextFFI *> tmp_block_comparisons_vec;
+  std::vector<int_comparison_buffer<Torus> *> reduction_buffers;
+
+  int_equality_selectors_buffer(CudaStreams streams, int_radix_params params,
+                                uint32_t num_possible_values,
+                                uint32_t num_blocks, bool allocate_gpu_memory,
+                                uint64_t &size_tracker) {
+    this->params = params;
+    this->allocate_gpu_memory = allocate_gpu_memory;
+    this->num_possible_values = num_possible_values;
+
+    uint32_t num_streams_to_use =
+        std::min((uint32_t)MAX_STREAMS_FOR_VECTOR_FIND, num_possible_values);
+    if (num_streams_to_use == 0)
+      num_streams_to_use = 1;
+
+    this->num_streams = num_streams_to_use;
+
+    this->active_streams = streams.active_gpu_subset(num_blocks);
+
+    incoming_event = cuda_create_event(streams.gpu_index(0));
+
+    sub_streams_vec.resize(num_streams_to_use);
+    outgoing_events.resize(num_streams_to_use);
+
+    for (uint32_t i = 0; i < num_streams_to_use; i++) {
+      sub_streams_vec[i].create_on_same_gpus(active_streams);
+      outgoing_events[i].resize(active_streams.count());
+      for (uint32_t j = 0; j < active_streams.count(); j++) {
+        outgoing_events[i][j] = cuda_create_event(active_streams.gpu_index(j));
+      }
+    }
+
+    uint32_t ciphertext_modulus = params.message_modulus * params.carry_modulus;
+    uint32_t box_size = params.polynomial_size / ciphertext_modulus;
+    lut_stride = (ciphertext_modulus / params.message_modulus) * box_size;
+
+    this->comparison_luts = new int_radix_lut<Torus>(
+        streams, params, 1, num_blocks, params.message_modulus,
+        allocate_gpu_memory, size_tracker);
+
+    std::vector<std::function<Torus(Torus)>> fns;
+    fns.reserve(params.message_modulus);
+    for (uint32_t i = 0; i < params.message_modulus; i++) {
+      fns.push_back([i](Torus x) -> Torus { return (x == i); });
+    }
+
+    generate_many_lut_device_accumulator<Torus>(
+        streams.stream(0), streams.gpu_index(0),
+        this->comparison_luts->get_lut(0, 0),
+        this->comparison_luts->get_degree(0),
+        this->comparison_luts->get_max_degree(0), params.glwe_dimension,
+        params.polynomial_size, params.message_modulus, params.carry_modulus,
+        fns, allocate_gpu_memory);
+
+    fns.clear();
+
+    this->comparison_luts->broadcast_lut(active_streams);
+
+    this->tmp_many_luts_output = new CudaRadixCiphertextFFI;
+    create_zero_radix_ciphertext_async<Torus>(
+        streams.stream(0), streams.gpu_index(0), this->tmp_many_luts_output,
+        params.message_modulus * num_blocks, params.big_lwe_dimension,
+        size_tracker, allocate_gpu_memory);
+
+    this->tmp_block_comparisons_vec.resize(this->num_streams);
+    this->reduction_buffers.resize(this->num_streams);
+    for (uint32_t j = 0; j < this->num_streams; j++) {
+      this->tmp_block_comparisons_vec[j] = new CudaRadixCiphertextFFI;
+      create_zero_radix_ciphertext_async<Torus>(
+          streams.stream(0), streams.gpu_index(0),
+          this->tmp_block_comparisons_vec[j], num_blocks,
+          params.big_lwe_dimension, size_tracker, allocate_gpu_memory);
+
+      this->reduction_buffers[j] = new int_comparison_buffer<Torus>(
+          sub_streams_vec[j], COMPARISON_TYPE::EQ, params, num_blocks, false,
+          allocate_gpu_memory, size_tracker);
+    }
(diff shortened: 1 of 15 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+
+  int_equality_selectors_buffer(CudaStreams streams, int_radix_params params,
+                                uint32_t num_possible_values,
+                                uint32_t num_blocks, bool allocate_gpu_memory,
+                                uint64_t &size_tracker) {
+    this->params = params;
+    this->allocate_gpu_memory = allocate_gpu_memory;
+    this->num_possible_values = num_possible_values;
+
+    uint32_t N = num_blocks;
```
agnesLeroy: N is usually used to denote the polynomial size. Also, we use lower case variables everywhere (except for Class names or Enums). Here I don't think there is a need to have a variable to wrap num_blocks you could probably just remove it.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+    uint32_t K = params.message_modulus;
```
agnesLeroy: Same here for the `K` notations. You could keep a variable named `message_modulus` to have a shorter name.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+
+    uint32_t block_modulus = params.message_modulus * params.carry_modulus;
```
agnesLeroy: This is usually named `ciphertext_modulus` or `total_modulus`. You actually use total_modulus a little bit further down so you could also use this name here.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+    this->num_packed_blocks = num_blocks;
+    this->max_packed_value = params.message_modulus * params.message_modulus;
+
+    this->max_num_many_luts =
+        (params.message_modulus * params.carry_modulus) / 2;
+    if (this->max_num_many_luts == 0) {
+      this->max_num_many_luts = 1;
+    }
+
+    uint32_t N = 1;
```
agnesLeroy: If N is hardcoded to 1 you could maybe remove the variable?

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+    uint32_t K = this->max_num_many_luts;
```
agnesLeroy: K -> max_num_many_luts

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,362 @@
...
+      }
+    }
+
+    release_radix_ciphertext_async(streams.stream(0), streams.gpu_index(0),
+                                   this->packed_selectors_ct,
+                                   this->allocate_gpu_memory);
+
+    cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
+
+    delete[] this->selectors_list;
```
agnesLeroy, line 591: You could delete these before the synchronize, it's only when we free CPU memory that we have to make sure it's done after sync, to be sure any CPU->GPU copy is over before returning.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+
+  uint32_t num_possible_values = mem_ptr->num_possible_values;
+  uint32_t message_modulus = mem_ptr->params.message_modulus;
+
+  integer_radix_apply_many_univariate_lookup_table<Torus>(
+      streams, mem_ptr->tmp_many_luts_output, lwe_array_in, bsks,
+      (Torus *const *)ksks, mem_ptr->comparison_luts, message_modulus,
+      mem_ptr->lut_stride);
+
+  for (uint32_t i = 0; i < num_possible_values; i++) {
```
agnesLeroy, line 37: Here we could reduce the latency by operating in parallel on this loop. 
1. In this PR you could add to the compute_equality_selectors mem structure an array of streams (can be std::vector) of size `min(num_possible_values, max_num_streams)` with `max_num_streams = 10` for example.
Each possible value `i` would be executed on a different stream. You can check in other functions (overflowing subtraction for example) how we manage events to be sure streams remain synced. 

2. To go further we could even split the different possible values on different GPUs (so executing for each possible value on 1 GPU). Then, if there are less possible values than available GPUs maybe we could use multiple GPUs somehow, it's a bit more complicated. You could go with item 1. only at first and we could revisit this later for multi-GPU.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+    Torus *const *ksks) {
+
+  uint32_t num_packed_blocks = mem_ptr->num_packed_blocks;
+  uint32_t max_packed_value = mem_ptr->max_packed_value;
+  uint32_t max_num_many_luts = mem_ptr->max_num_many_luts;
+
+  std::vector<std::function<Torus(Torus)>> fns;
+  fns.reserve(max_num_many_luts);
+
+  for (uint32_t i = 0; i < num_possible_values; i++) {
```
agnesLeroy, line 123: It's the same idea here, all possible values can be operated on in parallel.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+    CudaRadixCiphertextFFI const *lwe_array_in_list,
+    uint32_t num_possible_values, const uint64_t *h_decomposed_cleartexts,
+    int_possible_results_buffer<Torus> *mem_ptr, void *const *bsks,
+    Torus *const *ksks) {
+
+  uint32_t num_packed_blocks = mem_ptr->num_packed_blocks;
+  uint32_t max_packed_value = mem_ptr->max_packed_value;
+  uint32_t max_num_many_luts = mem_ptr->max_num_many_luts;
+
+  std::vector<std::function<Torus(Torus)>> fns;
```
agnesLeroy: This needs to be deleted before returning. Btw, our compute sanitizer check does not fail in case of leak at the moment: could you check the CI run to be sure there's nothing there please? I'm trying to spot things here but it'll be better to make sure with the CI run. We need to fix this behavior asap but haven't had time yet.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+        if (packed_block_value >= max_packed_value) {
+          PANIC("Cuda error: block value in create_possible_results "
+                "exceeds max packed value");
+        }
+        fns.push_back([packed_block_value](Torus x) -> Torus {
+          return (x == 1) * packed_block_value;
+        });
+      }
+
+      generate_many_lut_device_accumulator<Torus>(
```
agnesLeroy: Calling generate_device_accumulator and broadcast_lut here will degrade throughput on a multi-GPU machine. Is there a way to generate the LUTs in the scratch? it would already be a bit better.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+        2 * index + 1, message_ct, index, index + 1);
+
+    if (2 * index + 1 < lwe_array_out->num_radix_blocks) {
+      copy_radix_ciphertext_slice_async<Torus>(
+          streams.stream(0), streams.gpu_index(0), lwe_array_out, 2 * index + 1,
+          2 * index + 2, carry_ct, index, index + 1);
+    }
+  }
+
+  cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
```
agnesLeroy: Do you synchronize here because there's no scratch/cleanup for this function? If that's the case you should do it in the .cu file right after the call to this host function instead: if this host function is called by another one synchronizing here won't be necessary and could degrade performance.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+  CudaRadixCiphertextFFI *message_ct = mem_ptr->message_ct;
+  CudaRadixCiphertextFFI *carry_ct = mem_ptr->carry_ct;
+
+  int_radix_lut<Torus> *identity_lut = mem_ptr->identity_lut;
+  int_radix_lut<Torus> *message_extract_lut = mem_ptr->message_extract_lut;
+  int_radix_lut<Torus> *carry_extract_lut = mem_ptr->carry_extract_lut;
+
+  uint32_t num_chunks = (num_input_ciphertexts + chunk_size - 1) / chunk_size;
+
+  for (uint32_t chunk_idx = 0; chunk_idx < num_chunks - 1; chunk_idx++) {
```
agnesLeroy: These two loops can also be executed in parallel (see `tfhe/src/integer/server_key/radix_parallel/vector_find.rs:1285`). Would be good to use different streams to do this.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -0,0 +1,300 @@
...
+
+  if (mem_ptr->max_output_is_zero) {
+    host_integer_is_at_least_one_comparisons_block_true<Torus>(
+        streams, lwe_array_out_boolean, mem_ptr->packed_selectors_ct,
+        mem_ptr->at_least_one_true_buffer, bsks, (Torus **)ksks,
+        mem_ptr->num_matches);
+    return;
+  }
+
+  host_aggregate_one_hot_vector<Torus>(
```
agnesLeroy, line 431: If max_output_is_zero you could avoid allocating memory for  aggregate_one_hot_vector and is_at_least_one_comparisons_block_true

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -0,0 +1,567 @@
...
+
+    this->possible_results_buffer->release(streams);
+    delete this->possible_results_buffer;
+
+    if (!max_output_is_zero) {
+      this->aggregate_buffer->release(streams);
+      delete this->aggregate_buffer;
+    }
+
+    this->at_least_one_true_buffer->release(streams);
```
agnesLeroy, line 571: This is in the if as well no?

enzodimaria: You're right, problem solved !

PR #2994, opened 2025-11-10
chore(ci): fix cpu weekly benchmarks schedule groups handling

soonum:
The steps responsible for setting the OP_FLAVOR and ALL_PRECISION variables were never executed due to usage of non-existing env variable.
This causes OP_FLAVOR value to be null and thus would trigger error on benchmarks that doesn't handle unknown values for BENCH_OP_FLAVOR.

Also fixes filename to parse for additional boolean benchmark.

### DIFF
--- a/.github/workflows/benchmark_cpu_weekly.yml
+++ b/.github/workflows/benchmark_cpu_weekly.yml
@@ -46,13 +46,13 @@ jobs:
           echo "is_quarterly_bench=${{ github.event.schedule == '0 4 25 MAR,JUN,SEP,DEC *' }}" >> "${GITHUB_OUTPUT}"
 
       - name: Weekly benchmarks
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark_cpu_weekly.yml
```diff
@@ -190,7 +190,7 @@ jobs:
     with:
       command: boolean
       additional_recipe: measure_boolean_key_sizes
-      additional_file_to_parse: boolean_key_size.csv
+      additional_file_to_parse: boolean_key_sizes.csv
```
IceTDrinker, line 193: file name has been validated ?

soonum: Here's the name in the codebase:
`let results_file = Path::new("boolean_key_sizes.csv");`

IceTDrinker (approved):
Thanks !

PR #2995, opened 2025-11-10
chore(bench): fix naming order for erc20 hpu benchmarks

soonum:


pgardratzama (approved):
not yet understood what it changes but let's try

PR #3000, opened 2025-11-13
fix(core): fix decomposition algorithm not matching the theory

IceTDrinker:
- problem arose from a shift being done on an unsigned value which did not keep the signed characteristics of the represented signed value
- introduce an arithmetic_shift on the UnsignedInteger trait with a blanket implementation
- add the edge case which revelead the issue
- the asm has been verified to only change for the shift operation being applied, meaning no performance regression will occurr


IceTDrinker:
> Test the fix on several HPU tests that were failing due to this issue. Now they pass !

Awesome ! Thanks :)

PR #3002, opened 2025-11-14
fix(gpu): fix decomposition algorithm not matching the theory

agnesLeroy:

### Check-list:

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3002)

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
@@ -1,10 +1,7 @@
-#ifndef CNCRT_CRYPTO_CUH
-#define CNCRT_CRYPTO_CUH
-
+#pragma once
 #include "crypto/torus.cuh"
 #include "device.h"
 #include "fft128/f128.cuh"
-#include <cstdint>
 
 /**
  * GadgetMatrix implements the iterator design pattern to decompose a set of
@@ -16,7 +13,20 @@
  * This class always decomposes the entire set of num_poly polynomials.
  * By default, it works on a single polynomial.
  */
-#pragma once
+
+// Define explicitly an arithmetic shift right with a cast to signed
+template <typename T> __device__ T signed_shift_right(T value, int base_log) {
+  if constexpr (sizeof(T) == 4) {
+    return static_cast<T>(static_cast<int32_t>(value) >> base_log);
+  } else if constexpr (sizeof(T) == 8) {
+    return static_cast<T>(static_cast<int64_t>(value) >> base_log);
+  } else if constexpr (sizeof(T) == 16) {
+    return static_cast<T>(static_cast<__int128_t>(value) >> base_log);
+  } else {
+    return value >> base_log; // fallback for unusual sizes
+  }
+}
+
 template <typename T, class params> class GadgetMatrix {
 private:
   uint32_t level_count;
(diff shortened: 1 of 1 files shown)


backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -168,7 +168,16 @@ __device__ void decompose_and_compress_level_2_2_params(double2 *result,
 template <typename Torus>
 __device__ Torus decompose_one(Torus &state, Torus mask_mod_b, int base_log) {
   Torus res = state & mask_mod_b;
-  state >>= base_log;
+  if constexpr (sizeof(Torus) == 4) {
```
guillermo-oyarzun: Shouldn't this be applied to the decomposition in the pbs too?

IceTDrinker: all decomposition algorithms should follow this

agnesLeroy: Ah totally I missed them because the variable isn't named state there. I'll make the changes.

IceTDrinker (changes requested):
I think written as is the code might be wrong ? because the input values are not assigned to while before they were with the >>= operator ?

> @IceTDrinker reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3002#-:-Oe0t3-BE0x9Ldrvwvyv:b-9yba72)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @guillermo-oyarzun)

___
*[`backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh` line 26 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3002#-Oe0snup131fWv49Scju:-Oe0snup131fWv49Scjv:b-1ha8tu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f86ef4fe168a2725282a7be98409252c45dbd6b4/backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh#L26)):*
> ```Cuda
>     return static_cast<T>(static_cast<__int128_t>(value) >> base_log);
>   } else {
>     return value >> base_log; // fallback for unusual sizes
> ```

the fallback will fail ? would it be better to have a panic maybe ?
___
*[`backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh` line 74 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3002#-Oe0swM46FWDE_W7xDxI:-Oe0swM46FWDE_W7xDxJ:b-xfags9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f86ef4fe168a2725282a7be98409252c45dbd6b4/backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh#L74)):*
> ```Cuda
>       T res_im = *input2 & mask_mod_b;
> 
>       signed_shift_right<T>(*input1, base_log); // Update state
> ```

I'm afraid this is wrong, it is not assigning to input1 ? same below for input 2 ?

this would do nothing ?

IceTDrinker (approved):
After discussing good to merge as is, testing being tricky, and the fallback case normally never will be hit.

thanks !

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3002#-:-Oe1ep6qCpQlFDGDxlfD:b-q0ngu)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @agnesLeroy and @guillermo-oyarzun)

agnesLeroy:
@IceTDrinker as discussed:
- the decomposition is not exposed in a separate kernel so we don't have a test for it (maybe @guillermo-oyarzun in the noise tests you do?)
- the fallback is slightly wrong but we should most probably not integrate another type than u32, u64, u128 so I think it's alright to have it like this

PR #3010, opened 2025-11-17
chore(ci): use builds result to pass branch protection

soonum:
Instance teardown might fail regardless the result of build commands. To make CI green in pull-request we need to check the result of the build commands instead of the whole workflow result. Doing so ensure a PR could pass branch protection rules even if the teardown fails.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3010)

### DIFF
--- a/.github/workflows/cargo_build.yml
+++ b/.github/workflows/cargo_build.yml
@@ -141,22 +141,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check all builds success
-        if: needs.parallel-pcc-cpu.result == 'success' &&
-          needs.pcc-hpu.result == 'success' &&
-          needs.build-tfhe-full.result == 'success' &&
-          needs.build.result == 'success' &&
-          needs.build-layers.result == 'success' &&
-          needs.build-c-api.result == 'success'
+        if: needs.parallel-pcc-cpu.outputs.builds-result == 'success' &&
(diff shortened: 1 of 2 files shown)


IceTDrinker (commented):
just a question

> @IceTDrinker reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3010#-:-OeGtPibELN3ia2NJtW9:b-94nkfg)* status: all files reviewed, 1 unresolved discussion (waiting on @soonum)

___
*[`.github/workflows/cargo_build_common.yml` line 205 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3010#-OeGsaG1EZ_2By4E1suf:-OeGsaG1EZ_2By4E1sug:b-tkcxme) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2ee15eb79459417ad26a9b34460b93360ddcf273/.github/workflows/cargo_build_common.yml#L205)):*
> ```YAML
>         if: ${{ always() && needs.setup-instance.result == 'success' }}
>         run: | # zizmor: ignore[template-injection] this context variable is safe
>           echo "result=${{ job.status }}" >> "${GITHUB_OUTPUT}"
> ```

isn't if always() enough ? given we depend on the setup instance job ? I'm guessing this might be equivalent

I'm guessing we need the dedicated output otherwise the status includes the teardown status ?

PR #3026, opened 2025-11-20
chore(ci): remove check on secret availability for params checks

soonum:
This is needed only when the triggering event is pull_request and PR comes from an external contributor.


nsarlin-zama (approved):
thanks a lot!

PR #3031, opened 2025-11-20
test(core): add exhaustive decomposition test on several levels

IceTDrinker:
- also add a verification that recomposition works in the previously added edge case test

### DIFF
--- a/tfhe/src/core_crypto/commons/math/decomposition/tests.rs
+++ b/tfhe/src/core_crypto/commons/math/decomposition/tests.rs
@@ -230,11 +230,54 @@ fn test_decomposition_edge_case_sign_handling() {
     let decomposer = SignedDecomposer::new(DecompositionBaseLog(17), DecompositionLevelCount(3));
     let val: u64 = 0x8000_00e3_55b0_c827;
 
+    let rounded = decomposer.closest_representable(val);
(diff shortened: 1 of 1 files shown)


tfhe/src/core_crypto/commons/math/decomposition/tests.rs
```diff
@@ -230,11 +230,49 @@ fn test_decomposition_edge_case_sign_handling() {
...
+#[test]
+fn test_recompose_exhaustive() {
+    let base_log = DecompositionBaseLog(10);
+    let level = DecompositionLevelCount(3);
+    let decomposer = SignedDecomposer::new(base_log, level);
+
+    assert!(
+        decomposer.level_count().0 > 1,
+        "This test expects more than 1 level in the decomposer"
+    );
```
JJ-hw, line 254: Why is this restriction necessary ?
Considering the test range (u32), maybe the constraint is that `base * level` < 32.

IceTDrinker: we have both, we already have a test for balanced decomp over one level (was a separate issue that we fixed a while ago, so here I want to make sure we have strictly more than one level)

you are right on the limitation you mention, I can add it to be safe

PR #3043, opened 2025-11-24
fix(gpu): change a type to avoid possible issues when compression is executed for a large batch of LWEs

pdroalves:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
@@ -136,9 +136,9 @@ host_integer_compress(CudaStreams streams,
   }
 
   // Modulus switch
-  int size = num_glwes * compression_params.glwe_dimension *
-                 compression_params.polynomial_size +
-             glwe_array_out->total_lwe_bodies_count;
+  uint32_t size = num_glwes * compression_params.glwe_dimension *
+                      compression_params.polynomial_size +
+                  glwe_array_out->total_lwe_bodies_count;
 
   host_modulus_switch_inplace<Torus>(streams.stream(0), streams.gpu_index(0),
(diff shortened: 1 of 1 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/compression/compression.cuh
```diff
@@ -136,7 +136,7 @@ host_integer_compress(CudaStreams streams,
   }
 
   // Modulus switch
-  int size = num_glwes * compression_params.glwe_dimension *
+  auto size = num_glwes * compression_params.glwe_dimension *
```
guillermo-oyarzun: uint32_t  seems to be fine to capture the result of that mul, otherwise we might need to cast one element to uint64_t

agnesLeroy: This looks ok indeed, the first prod will always remain small and total_lwe_bodies_count would have to be insanely big to make int insufficient for this 🤔

pdroalves: `int` is not terrible. We still can fit a lot of LWEs in there before overflowing. However, we don't usually use that type in the rest of the library and doubling the space for `size` would not be bad.

guillermo-oyarzun: yup i am a bit paranoid with the int too, prefer the uint32_t at least.

agnesLeroy: Yes let's go for uint32_t instead of int or auto

pdroalves: Ok!
Done.

PR #3057, opened 2025-11-27
chore(ci): add placeholders for whitepaper benchmarks

soonum:
This is done to be able to test the workflow in a development branch.


soonum:
> Just so I understand, why couldn't this use the existing placeholder workflow ?

In this case I could have used it indeed.
The main purpose for it is to have all my run attempt at the same location than the final workflow file.

nsarlin-zama (approved):

nsarlin-zama:
Understood :)

PR #3060, opened 2025-11-27
chore: remove expand special case

IceTDrinker:


IceTDrinker:
ok pcc red and backward compat tests seem to rely on the expand without server keys...

IceTDrinker:
Maybe this PR is not worth the trouble for now, will see tomorrow

PR #3086, opened 2025-12-05
refactor(gpu): creating benchmarks for match_value

enzodimaria:

### Check-list:

### DIFF
--- a/tfhe-benchmark/benches/integer/vector_find.rs
+++ b/tfhe-benchmark/benches/integer/vector_find.rs
@@ -1,0 +1,107 @@
+use benchmark::params_aliases::BENCH_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+use benchmark::utilities::{write_to_json, OperatorType};
+use criterion::{black_box, Criterion};
+use tfhe::integer::keycache::KEY_CACHE;
+use tfhe::integer::{IntegerKeyKind, RadixClientKey};
+use tfhe::keycache::NamedParam;
+use tfhe::shortint::AtomicPatternParameters;
+use tfhe::MatchValues;
+
+fn match_value_scenarios() -> Vec<(usize, usize, usize)> {
+    vec![(64, 10, 32), (8, 256, 4)]
+}
+
+pub fn match_value(c: &mut Criterion) {
+    let bench_name = "integer::match_value";
+    let mut group = c.benchmark_group(bench_name);
+    group.sample_size(15);
+
+    let param = BENCH_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+    let atomic_param: AtomicPatternParameters = param.into();
+    let param_name = param.name();
+
+    let (cpu_cks, cpu_sks) = KEY_CACHE.get_from_params(atomic_param, IntegerKeyKind::Radix);
+
+    for (bits, num_elements, num_blocks) in match_value_scenarios() {
+        let bench_id = format!("{bench_name}::{param_name}::{bits}bit_{num_elements}elements");
+
+        let cks = RadixClientKey::from((cpu_cks.clone(), num_blocks));
+
+        let mapping_data: Vec<(u64, u64)> = (0..num_elements as u64).map(|i| (i, i)).collect();
+        let match_values = MatchValues::new(mapping_data).expect("Invalid match values");
+
+        let input_val = 1u64;
+        let ct_input = cks.encrypt(input_val);
+
+        group.bench_function(&bench_id, |b| {
+            b.iter(|| {
+                black_box(cpu_sks.match_value_parallelized(&ct_input, &match_values));
+            })
+        });
+
+        write_to_json::<u64, _>(
+            &bench_id,
+            atomic_param,
+            param.name(),
+            "match_value",
+            &OperatorType::Atomic,
+            bits as u32,
+            vec![atomic_param.message_modulus().0.ilog2(); bits],
+        );
+    }
+    group.finish();
+}
+
+#[cfg(feature = "gpu")]
+pub mod cuda {
+    use super::*;
+    use tfhe::core_crypto::gpu::CudaStreams;
+    use tfhe::integer::gpu::ciphertext::CudaUnsignedRadixCiphertext;
+    use tfhe::integer::gpu::CudaServerKey;
+
+    pub fn cuda_match_value(c: &mut Criterion) {
+        let bench_name = "integer::cuda::match_value";
+        let mut group = c.benchmark_group(bench_name);
+        group.sample_size(15);
+
+        let param = BENCH_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+        let atomic_param: AtomicPatternParameters = param.into();
+        let param_name = param.name();
+
+        let streams = CudaStreams::new_multi_gpu();
+
+        let (cpu_cks, _) = KEY_CACHE.get_from_params(atomic_param, IntegerKeyKind::Radix);
+        let sks = CudaServerKey::new(&cpu_cks, &streams);
+
+        for (bits, num_elements, num_blocks) in match_value_scenarios() {
+            let bench_id = format!("{bench_name}::{param_name}::{bits}bit_{num_elements}elements");
+
+            let cks = RadixClientKey::from((cpu_cks.clone(), num_blocks));
+
+            let mapping_data: Vec<(u64, u64)> = (0..num_elements as u64).map(|i| (i, i)).collect();
+            let match_values = MatchValues::new(mapping_data).expect("Invalid match values");
+
+            let input_val = 1u64;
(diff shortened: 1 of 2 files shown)


tfhe-benchmark/Cargo.toml
```diff
@@ -133,6 +133,12 @@ path = "benches/integer/aes.rs"
 harness = false
 required-features = ["integer", "internal-keycache"]
 
+[[bench]]
```
agnesLeroy: I don't think we add a new bench target in Cargo for this one

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,217 @@
...
+criterion_group!(match_value_group, match_value);
+
+#[cfg(feature = "gpu")]
+criterion_group!(gpu_match_value_group, cuda::cuda_match_value);
+
+fn main() {
+    #[cfg(feature = "gpu")]
+    gpu_match_value_group();
+
+    match_value_group();
```
soonum: If you do that you'll run both GPU and CPU bench on a machine.
Is it what we want ?

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,217 @@
...
+        let mapping_data: Vec<(u64, u64)> = (0..num_elements as u64).map(|i| (i, i)).collect();
+        let match_values = MatchValues::new(mapping_data).expect("Invalid match values");
+
+        let input_val = 1u64;
+        let ct_input = cks.encrypt(input_val);
+
+        let mut group = c.benchmark_group(&bench_id);
+        group.sample_size(15);
+
+        group.bench_function("latency", |b| {
```
soonum: Prefer the `bench_id` as a identifier rather than `"latency"`, even if criterion will most likely truncate the `bench_id` string.

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let (cpu_cks, cpu_sks) = KEY_CACHE.get_from_params(atomic_param, IntegerKeyKind::Radix);
+
+    //
+    // Latency test: 64 bits, 10 elements
+    //
+    {
+        let num_elements = 10;
+        let bits = 64;
+        let num_blocks = 32;
```
soonum: By looking at this it seems that the two cases 64 bits + 10 elements // 8 bits + 256 elements could be factorized into one for loop.
Unless I'm missing something here.

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,217 @@
...
+        let (cpu_cks, _) = KEY_CACHE.get_from_params(atomic_param, IntegerKeyKind::Radix);
+        let sks = CudaServerKey::new(&cpu_cks, &streams);
+
+        //
+        // Latency test: 64 bits, 10 elements
+        //
+        {
+            let num_elements = 10;
+            let bits = 64;
+            let num_blocks = 32;
```
soonum: Idem for factorization.

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,217 @@
...
+    //
+    // Latency test: 64 bits, 10 elements
+    //
+    {
+        let num_elements = 10;
+        let bits = 64;
+        let num_blocks = 32;
+
+        let bench_id =
+            format!("{bench_name}::{param_name}::latency_{bits}bit_{num_elements}elements");
```
soonum: Is throughput is planned to be implemented at some point ?
If yes, please remove `latency_` from the naming since we add `::throughput::` in the name to differentiate the two cases.

tfhe-benchmark/Cargo.toml
soonum: It depends on how and when you intend to run these benchmarks.
More broadly, if a target is added, we need a make recipe to be able to trigger it easily.

tfhe-benchmark/benches/integer/vector_find.rs
agnesLeroy: It's not! :wink:

tfhe-benchmark/benches/integer/vector_find.rs
agnesLeroy: I don't know if we'll have a throughput measurement for this at any point :thinking: let's remove `latency_` from the name anyway to match other benchmarks' names.

tfhe-benchmark/Cargo.toml
agnesLeroy: It should just be executed alongside integer benches, really I don't think we want a specific recipe here.

tfhe-benchmark/benches/integer/vector_find.rs
soonum: Ok so wee need to add `#[cfg(not(feature = "gpu"))]` for the CPU case

tfhe-benchmark/Cargo.toml
soonum: Got it, in this case, we need to assess the duration of this benchmarks and add them to the adequate bench group in integer, I would say `default` and maybe `fast_default` ?

agnesLeroy: It's not too long so could be both. The second case takes ~200ms.

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,109 @@
...
+                black_box(cpu_sks.match_value_parallelized(&ct_input, &match_values));
+            })
+        });
+        group.finish();
+
+        write_to_json::<u64, _>(
+            &bench_id,
+            atomic_param,
+            param.name(),
+            "match_value_latency",
```
soonum: Can you remove the `_latency` suffix ?

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,109 @@
...
+                    black_box(sks.match_value(&d_ct_input, &match_values, &streams));
+                })
+            });
+            group.finish();
+
+            write_to_json::<u64, _>(
+                &bench_id,
+                atomic_param,
+                param.name(),
+                "match_value_latency",
```
soonum: Same here.

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,109 @@
...
+pub fn match_value(c: &mut Criterion) {
+    let bench_name = "integer::match_value";
+
+    let param = BENCH_PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+    let atomic_param: AtomicPatternParameters = param.into();
+    let param_name = param.name();
+
+    let (cpu_cks, cpu_sks) = KEY_CACHE.get_from_params(atomic_param, IntegerKeyKind::Radix);
+
+    let scenarios = vec![(64, 10, 32), (8, 256, 4)];
```
soonum: If the cases are the same between CPU and GPU, why not factorize that at the top of the file ?
You can look at what's been done for [ZK benchmarks](https://github.com/zama-ai/tfhe-rs/blob/main/tfhe-benchmark/benches/integer/zk_pke.rs#L34)

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let cks = RadixClientKey::from((cpu_cks.clone(), num_blocks));
+
+        let mapping_data: Vec<(u64, u64)> = (0..num_elements as u64).map(|i| (i, i)).collect();
+        let match_values = MatchValues::new(mapping_data).expect("Invalid match values");
+
+        let input_val = 1u64;
+        let ct_input = cks.encrypt(input_val);
+
+        let mut group = c.benchmark_group(&bench_id);
```
soonum: I believe that the group should be instantiated at the top of the function, right after `let bench_name = ...` . Otherwise, you may end up with veeeery long group name. You can look how group are handled in [integer benchmarks](https://github.com/zama-ai/tfhe-rs/blob/main/tfhe-benchmark/benches/integer/bench.rs#L46)

tfhe-benchmark/benches/integer/vector_find.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            let mapping_data: Vec<(u64, u64)> = (0..num_elements as u64).map(|i| (i, i)).collect();
+            let match_values = MatchValues::new(mapping_data).expect("Invalid match values");
+
+            let input_val = 1u64;
+            let ct_input = cks.encrypt(input_val);
+            let d_ct_input =
+                CudaUnsignedRadixCiphertext::from_radix_ciphertext(&ct_input, &streams);
+
+            let mut group = c.benchmark_group(&bench_id);
```
soonum: Same remark regarding the group.

tfhe-benchmark/benches/integer/vector_find.rs
soonum: Not blocking though.

soonum (approved):
Nice, thanks for the work ! 👍

PR #3103, opened 2025-12-10
chore: bump dyn-stack to 0.13

tmontaigu:
Notable changes:
- StackReq methods no longer returns Result<StackReq, SizeOverflow> instead, StackReq contains the invalid state. Now, its when we create a PodBuffer that we can check/catch if the size req is invalid by catching errors when calling `PodBuffer::try_new`. Its also possible to manually check that `stack_req != StackReq::OVERFLOW`

- GlobalaPodBuffer is now PodBuffer

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3103)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ rust-version = "1.91.1"
 [workspace.dependencies]
 aligned-vec = { version = "0.6", default-features = false }
 bytemuck = "1.24"
-dyn-stack = { version = "0.11", default-features = false }
+dyn-stack = { version = "0.13", default-features = false }
 itertools = "0.14"
 num-complex = "0.4"
 pulp = { version = "0.22", default-features = false }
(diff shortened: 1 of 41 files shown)


IceTDrinker:
what impact does this have ? the fact all the unwraps are gone are checked how now ?

tmontaigu:
> Notable changes:
> 
>     * StackReq methods no longer returns Result<StackReq, SizeOverflow> instead, StackReq contains the invalid state. Now, its when we create a PodBuffer that we can check/catch if the size req is invalid by catching errors when calling `PodBuffer::try_new`. Its also possible to manually check that `stack_req != StackReq::OVERFLOW`
> 

So now its checked in PodBuffer::new (panic) or PodBuffer::try_new (Result)

IceTDrinker:
and a few cargo builds too

PR #3115, opened 2025-12-11
refactor(core): simplify PlanMap

mayeul-zama:
entries of the `HashMap` are always initialized so we don't need them to be wrapped in `OnceLock`

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115)

### DIFF
--- a/tfhe/src/core_crypto/fft_impl/fft128/math/fft/mod.rs
+++ b/tfhe/src/core_crypto/fft_impl/fft128/math/fft/mod.rs
@@ -8,35 +8,19 @@ use dyn_stack::{PodStack, StackReq};
 use std::sync::{Arc, OnceLock};
 use tfhe_fft::fft128::{f128, Plan};
 
-#[derive(Clone)]
-pub(crate) struct PlanWrapper(Plan);
-impl core::ops::Deref for PlanWrapper {
-    type Target = Plan;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
-impl core::fmt::Debug for PlanWrapper {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        "[?]".fmt(f)
-    }
-}
-
 /// Negacyclic Fast Fourier Transform. See [`Fft128View`] for transform functions.
 ///
 /// This structure contains the twisting factors as well as the
 /// FFT plan needed for the negacyclic convolution over the reals.
 #[derive(Clone, Debug)]
 pub struct Fft128 {
-    plan: Arc<PlanWrapper>,
+    plan: Arc<Plan>,
 }
 
 /// View type for [`Fft128`].
 #[derive(Clone, Copy, Debug)]
 pub struct Fft128View<'a> {
-    pub(crate) plan: &'a PlanWrapper,
+    pub(crate) plan: &'a Plan,
 }
 
 impl Fft128 {
@@ -45,7 +29,7 @@ impl Fft128 {
     }
 }
 
-type PlanMap = GenericPlanMap<PolynomialSize, PlanWrapper>;
+type PlanMap = GenericPlanMap<PolynomialSize, Plan>;
 
 pub(crate) static PLANS: OnceLock<PlanMap> = OnceLock::new();
 fn plans() -> &'static PlanMap {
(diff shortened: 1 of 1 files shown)


mayeul-zama:
The 2 commits should be reviewed separately

We may only keep the first one

IceTDrinker (commented):
just as an FYI, if we go ahead with this, I would prefer it for Q1 26 (because it's a change at the bottom of the stack with potential weird bugs due to threading) so not for 1.5

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115#-:-OgD0GOJ0kNGnk_Z6rE_:b-ml3v9j)* status: 0 of 5 files reviewed, all discussions resolved

IceTDrinker (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @mayeul-zama).

___
*[`tfhe/src/core_crypto/commons/plan.rs` line 12 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115#-Ok8CTtp4-_3_YFcxRL3:-Ok8CTtp4-_3_YFcxRL4:b6wvaj2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/a5a27c75ef064f20dbfedfa4ffb8014860ee1593/tfhe/src/core_crypto/commons/plan.rs#L12)):*
> ```XML
>     }
> 
>     pub fn get_or_init(&self, key: Key, new_value: impl Fn(Key) -> Value) -> Arc<Value> {
> ```

I think we need a set function that overwrites the previous value

for example with the manual FFT stuff with `setup_custom_fft_plan` one may want to test several FFT algorithms but given how this function is written you cannot replace a plan once it has been setup

this is perfectly fine to keep for the normal way TFHE-rs operates given we don't change FFT plans once they are setup automatically but for the manual approach it would likely be a bug

mayeul-zama (commented):
> @mayeul-zama made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @IceTDrinker).

___
*[`tfhe/src/core_crypto/commons/plan.rs` line 12 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115#-Ok8CTtp4-_3_YFcxRL3:-OkD5vlaDdVLq8K8-F3Y:bve9mbz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/a5a27c75ef064f20dbfedfa4ffb8014860ee1593/tfhe/src/core_crypto/commons/plan.rs#L12)):*

I think we need a set function that overwrites the previous value

for example with the manual FFT stuff with `setup_custom_fft_plan` one may want to test several FFT algorithms but given how this function is written you cannot replace a plan once it has been setup

this is perfectly fine to keep for the normal way TFHE-rs operates given we don't change FFT plans once they are setup automatically but for the manual approach it would likely be a bug
</blockquote></details>

I think `setup_custom_fft_plan` was still working as intended
But I still extracted the `GenericPlanMap::set` function as it's cleaner

mayeul-zama (commented):
> @mayeul-zama made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @IceTDrinker).

___
*[`tfhe/src/core_crypto/commons/plan.rs` line 12 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3115#-Ok8CTtp4-_3_YFcxRL3:-OkDKi6e6-DCT70ITsHf:byx1szo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/a5a27c75ef064f20dbfedfa4ffb8014860ee1593/tfhe/src/core_crypto/commons/plan.rs#L12)):*

if you called it twice in a row with the same PolySize the code was re-using the already present value from the HashMap if I'm not mistaken, you can check the previous revision on reviewable
</blockquote></details>

I don't think so
Code was ` Entry::Occupied(mut occupied_entry) => *occupied_entry.get_mut() = plan,`

PR #3127, opened 2025-12-17
fix(gpu): Using tbc for classical 64 bits pbs on H100

enzodimaria:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -347,27 +347,25 @@ uint64_t scratch_cuda_programmable_bootstrap_64(
 
   auto max_shared_memory = cuda_get_max_shared_memory(gpu_index);
 #if (CUDA_ARCH >= 900)
-  if (has_support_to_cuda_programmable_bootstrap_tbc<uint64_t>(
```
guillermo-oyarzun, line 350: I think what you have to change is the function has_support_to_cuda_programmable_bootstrap_tbc and remove there the constraint of the number of samples. Because we should still check if the arch supports tbc.

enzodimaria: Ok, I'm on it!

PR #3129, opened 2025-12-18
fix(gpu): fix cpu memory leak in expand and rerand

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
@@ -47,7 +47,7 @@ template <typename Torus> struct compact_lwe_list {
 
 template <typename Torus> struct flattened_compact_lwe_lists {
   Torus *d_ptr;
-  Torus **d_ptr_to_compact_list;
+  Torus **ptr_array_to_d_compact_list;
   const uint32_t *h_num_lwes_per_compact_list;
   uint32_t num_compact_lists;
   uint32_t lwe_dimension;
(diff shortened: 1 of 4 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/rerand.cu
```diff
@@ -88,9 +88,6 @@ void cuda_rerand_64(
           " in the interval [256..16384].");
     break;
   }
-
```
andrei-stoian-zama, line 91: could you explain why this isn't needed anymore? synchronization is done on the rust side ?

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -79,6 +79,11 @@ template <typename Torus> struct flattened_compact_lwe_lists {
                             lwe_dimension,
                             h_num_lwes_per_compact_list[compact_list_index]);
   }
+
+  void release(CudaStreams streams) {
+    cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
+    free(d_ptr_to_compact_list);
```
andrei-stoian-zama: `d_ptr_to_compact_list` is a CPU pointer.. having a `d_` prefix is maybe misleading

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -79,6 +79,11 @@ template <typename Torus> struct flattened_compact_lwe_lists {
                             lwe_dimension,
                             h_num_lwes_per_compact_list[compact_list_index]);
   }
+
```
andrei-stoian-zama, line 82:
```
    d_ptr_to_compact_list =
        static_cast<Torus **>(malloc(num_compact_lists * sizeof(Torus **)));
```

annoying nitpicking here, but it should be `sizeof(Torus*)`.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
andrei-stoian-zama: just a small warning: the constructor doesn't take `streams` but it would have been nice if it did. only the destructor takes streams and it can lead to mistakes where the wrong streams are passed.

backends/tfhe-cuda-backend/cuda/src/integer/rerand.cu
agnesLeroy: The new release function does the synchronization and is called at the end of the host function. I wrote comments in the two cases involved. We were actually missing a synchronization in expand.

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
agnesLeroy: I agree the name isn't ideal, I changed it to `ptr_array_to_d_compact_list`.
The constructor doesn't need streams. The release function only takes some because I thought it would be safer to synchronize in that release function, vs calling synchronize then release. But I could also do the latter.

backends/tfhe-cuda-backend/cuda/src/integer/rerand.cu
agnesLeroy: I removed sync from release and kept it only in the cleanup function in the end

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
agnesLeroy: Now the release function  doesn't take streams and rerand & expand only call synchronize in their release functions.

PR #3138, opened 2025-12-22
fix(gpu): return to 64 regs in multi-bit pbs

guillermo-oyarzun:

The registers grew with when the lwe_chunk_size changed to be 64 bit, this change was necessary to avoid overflow when calculating the multi-bit temporal array size. In the pbs we know the chunk size won't overflow in 32-bit cause that would mean having a lwe dimension too large. So i casted it back to uint32_t, and that led to having 64 regs again.

### Check-list:


bbarbakadze (approved):
Thank you @guillermo-oyarzun .

PR #3143, opened 2025-12-29
fix(hpu): erc20 whitepaper bench

pgardratzama:
Full story is:
- @agnesLeroy updated the whitepaper ERC20 implementation in bench to match fhevm description but trivial encryption is not supported by HPU so when running ERC20 bench on HPU it fails
- I tried to replace if_then_else(cond, then, trivial(0)) by a if_then_zero which uses HPU IOp IF_THEN_ZERO and discovered that performance was better than ERC_20 IOp which was very strange
- @Bapt-Roux discovered ERC_20 IOp was incorrectly linked to ERC_20_SIMD and was not used kogge-stone for Add/Sub which explained poor performance
- @Bapt-Roux updated HLAPI if_then_zero for CPU and I tried to fix bench so that it could stay the same for GPU
- performance of ERC20 whitepaper and HPU_optimized moved from 31/34 TPS to 41/42 TPS on a single FPGA

I have run the CPU version on our HPU machines but not the GPU version.
So we need to check that it still works on GPU.
We also need to discuss if adding if_then_zero in HLAPI is ok?

### DIFF
--- a/tfhe-benchmark/benches/high_level_api/erc20.rs
+++ b/tfhe-benchmark/benches/high_level_api/erc20.rs
@@ -34,12 +34,12 @@ where
 {
     let has_enough_funds = (from_amount).ge(amount);
     let amount_to_transfer = {
-        #[cfg(feature = "gpu")]
+        #[cfg(not(feature = "hpu"))]
         {
             let zero_amount = FheType::encrypt_trivial(0u64);
             has_enough_funds.select(amount, &zero_amount)
         }
-        #[cfg(not(feature = "gpu"))]
+        #[cfg(feature = "hpu")]
         {
             has_enough_funds.if_then_zero(amount)
         }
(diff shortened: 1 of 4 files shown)


tfhe/src/high_level_api/booleans/base.rs
```diff
@@ -552,6 +552,66 @@ where
...
+where
+    Id: FheUintId,
+{
+    /// Conditional selection.
+    ///
+    /// The output value returned depends on the value of `self`.
+    ///
+    /// - if `self` is true, the output will have the value of `ct_then`
+    /// - if `self` is false, the output will be an encryption of 0
+    fn if_then_zero(&self, ct_then: &FheUint<Id>) -> FheUint<Id> {
```
IceTDrinker, line 565: could use tests (even if it uses existing primitives)

pgardratzama: a test has been added for CPU & HPU... and it works both on CPU & FPGA!!!

PR #3157, opened 2026-01-08
fix(js): unsafe coop bench was overwritting mt one

nsarlin-zama:

### PR content/description
In the workflow we run the multithread bench, then the unsafe coop. The same json is updated and then sent to grafana. Since the events have the same name they were overwritten.


IceTDrinker (approved):
Thanks !

PR #3178, opened 2026-01-15
chore(ci): update toolchain and some tools

soonum:
Here's the update to the latest Rust nightly toolchain.
Also we have updates for:

* Zizmor `1.16.2` -> `1.20.0`
* Typos   `1.39.0` -> `1.42.0

All new findings resulting for these updates are also fixed.

### DIFF
--- a/mockups/tfhe-hpu-mockup/src/modules/params.rs
+++ b/mockups/tfhe-hpu-mockup/src/modules/params.rs
@@ -46,10 +46,11 @@ impl MockupOptions {
         }
     }
     pub fn report_trace(&self, iop: hpu_asm::AsmIOpcode) -> Option<File> {
-        if self.report_out.is_some() && self.report_trace {
-            let report_out = &self.report_out.as_ref().unwrap();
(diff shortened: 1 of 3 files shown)


mockups/tfhe-hpu-mockup/src/modules/params.rs
```diff
@@ -46,10 +46,14 @@ impl MockupOptions {
...
-            let iop_file = format!("{report_out}/{iop}.json");
-            Some(Self::open_wr_file(&iop_file))
+        if self.report_trace {
+            match self.report_out {
+                Some(ref report_out) => {
+                    let iop_file = format!("{report_out}/{iop}.json");
+                    Some(Self::open_wr_file(&iop_file))
+                }
+                None => None,
+            }
```
nsarlin-zama: Nit, but I think here this could simply be:
```rust
            self.report_out.map(|report_out| {
                let iop_file = format!("{report_out}/{iop}.json");
                Self::open_wr_file(&iop_file)
            })
```

PR #3186, opened 2026-01-20
fix(gpu): destroy temporary events in scatter/gather

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
@@ -158,13 +158,11 @@ __global__ void realign_with_indexes(Torus *d_vector,
 /// The output indexing is always the trivial one
 /// num_inputs: total num of lwe in src
 template <typename Torus>
-void multi_gpu_scatter_lwe_async(CudaStreams streams,
-                                 const std::vector<Torus *> &dest,
-                                 Torus const *src, Torus const *d_src_indexes,
-                                 bool is_trivial_index,
-                                 std::vector<Torus *> &aligned_vec,
-                                 uint32_t max_active_gpu_count,
-                                 uint32_t num_inputs, uint32_t lwe_size) {
+void multi_gpu_scatter_lwe_async(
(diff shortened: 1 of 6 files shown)


backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cuh
```diff
@@ -252,6 +254,7 @@ void multi_gpu_gather_lwe_async(CudaStreams streams, Torus *dest,
       cuda_event_record(temp_event2, streams.stream(0), streams.gpu_index(0));
       cuda_stream_wait_event(streams.stream(i), temp_event2,
                              streams.gpu_index(i));
+      cuda_event_destroy(temp_event2, streams.gpu_index(0));
```
guillermo-oyarzun: We need to do something more tricky here, because the cuda_event_destroy internally just calls cudaEventDestroy, which doesn't check if the event has finished, it just destroys it, if that happens we will lose the correct waiting order. We probably need an external structure that allocates a couple or a list of events that are reused after.

andrei-stoian-zama: isn't the above `cuda_stream_wait_event` sufficient ?

guillermo-oyarzun: i think it might lead to undefined behaviour, cause with the cuda_stream_wait_event, we do the stream to wait for that event but the host code is not blocked only the stream work, and right after that we destroy the event (this call is not enqueued in the stream, is called directly from the host. So, my understanding is that the event would be destroyed and the stream when arrives to the wait, does not have the restriction to wait anything. I am running multi-gpu tests to see if this is true.

guillermo-oyarzun: I modified the commit, now i create a pool of events in the lut that can dynamically grow and is destroyed during release of the lut. @agnesLeroy  and @andrei-stoian-zama can you check if i missed something?

PR #3190, opened 2026-01-23
Backport gpu fixes for create_on_same_gpus and expand

agnesLeroy:

### Check-list:

### DIFF
--- a/tfhe/src/integer/gpu/zk/mod.rs
+++ b/tfhe/src/integer/gpu/zk/mod.rs
@@ -128,7 +128,7 @@ impl<'de> serde::Deserialize<'de> for CudaProvenCompactCiphertextList {
     }
 }
 
-#[cfg(feature = "zk-pok")]
+#[cfg(all(feature = "zk-pok", feature = "gpu"))]
 #[cfg(test)]
 mod tests {
     // Test utils for tests here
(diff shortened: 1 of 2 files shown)


tfhe/src/integer/gpu/zk/mod.rs
```diff
@@ -493,4 +493,100 @@ mod tests {
...
+            expander.get_kind_of(6).is_none(),
+            "Index 6 should return None (beyond data item count)"
+        );
+        assert!(
+            expander.get_kind_of(65).is_none(),
+            "Index 65 should return None (beyond data item count)"
+        );
+
+        // Verify we can actually retrieve the values
+        let _bool_val: crate::FheBool = expander.get(0).unwrap().unwrap();
```
IceTDrinker, line 585: suggestion: decrypt to check they are the initial values, same for all types

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
IceTDrinker, line 127: I don't remember the patterns for the GPU backend, this feels like it could be handled with a vec having the right capacity to avoid reallocs, should avoid forgetting deletes elsewher ?

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
```diff
@@ -114,11 +114,13 @@ struct CudaStreams {
   // streams on the same GPU
   void create_on_same_gpus(const CudaStreams &other) {
     PANIC_IF_FALSE(_streams == nullptr,
```
IceTDrinker, line 116: I'm not sure why _streams being nullptr is an issue, it looks like it gets overwritten anyways line 127, also might be a memory leak ?

I'm unclear on the responsibilities of the backend/rust library for managing CudaStreamsFFI and their lifetimes/allocations

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
IceTDrinker, line 130: I think these two lines leak the previous this->_streams and this->_gpu_indexes

tfhe/src/integer/gpu/zk/mod.rs
agnesLeroy: @pdroalves it's a good idea, we should open a small PR to add this check next week

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
agnesLeroy: Using std::vector for us is dangerous because we need to make sure to synchronize in the right places (for example a resize will require to first synchronize). This is a pain to handle, and the only use case we have for it is this CudaStreams struct basically. We discussed it extensively in the team already but arrived to the conclusion that we don't want to do that.

agnesLeroy: this->_streams and this->gpu_indexes will be released when the release function for this CudaStreams is called.

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
agnesLeroy: We only allow "cloning" a set of streams into an empty CudaStreams, not into an already initialized CudaStreams.

IceTDrinker: ah ok yes, makes sense then !

backends/tfhe-cuda-backend/cuda/include/helper_multi_gpu.h
IceTDrinker: it was linked to my misunderstanding of the clone only allowed on unitialized streams, maybe the function could be called "initialize_on_same_gpus" or something like that

agnesLeroy: We settled on `create_on_same_gpus`, which is quite similar, I would rather not change it now.

IceTDrinker: yep no problem, just a thought from reading the code

tfhe/src/integer/gpu/zk/mod.rs
pdroalves: Thanks @IceTDrinker !
I added that suggestion in https://github.com/zama-ai/tfhe-rs/pull/3214.

pdroalves: Thanks @IceTDrinker !
I added that suggestion in https://github.com/zama-ai/tfhe-rs/pull/3214.

PR #3191, opened 2026-01-23
chore(ci): add toml file format checker

soonum:
Taplo is a CLI tool meant to validate and format any TOML file. In the continuous integration pipeline, only the format checker is used.


Makefile
```diff
@@ -171,6 +171,12 @@ install_cargo_dylint:
 install_cargo_audit:
 	cargo install --locked cargo-audit
 
+.PHONY: install_taplo # Check Cargo.toml format
+install_taplo:
+	@taplo --version > /dev/null 2>&1 || \
+	cargo install --locked taplo-cli || \
```
IceTDrinker: question (non blocking): do we need a locked version ?

soonum: Everything we install with cargo is `--locked`.

IceTDrinker: sorry meant a sepcific version, the lock only specifies that we use the lock file, not which version it is

soonum: Oh you're right

soonum: Script is added to lock version.

PR #3212, opened 2026-01-27
chore(ci): ensure no credentials are leaked in action log

soonum:
When using direct git command, credentials are exposed in the console logs. Despite the fact GitHub is redacting its secrets, adding --quiet flag ensures that, even if this redaction feature is flawed, we don't leak secrets in the action log.

### DIFF
--- a/.github/workflows/sync_on_push.yml
+++ b/.github/workflows/sync_on_push.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           echo ">>> Cloning source repo..."
           git lfs install
-          git clone "https://${USERNAME}:${TOKEN}@github.com/${SOURCE_REPO}.git" ./tfhe-rs --origin source && cd ./tfhe-rs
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_on_push.yml
```diff
@@ -30,8 +30,8 @@ jobs:
         run: |
           echo ">>> Cloning source repo..."
           git lfs install
-          git clone "https://${USERNAME}:${TOKEN}@github.com/${SOURCE_REPO}.git" ./tfhe-rs --origin source && cd ./tfhe-rs
-          git remote add destination "https://${USERNAME}:${TOKEN}@github.com/${DEST_REPO}.git"
+          git clone "https://${USERNAME}:${TOKEN}@github.com/${SOURCE_REPO}.git" ./tfhe-rs --origin source && cd ./tfhe-rs --quiet
```
IceTDrinker: --quiet is not applied to the git command

let's do

git command --quiet

to make sure we are good

PR #3214, opened 2026-01-27
fix(gpu): fix an invalid access in expand when the number of LWEs is odd

pdroalves:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
@@ -15,6 +15,70 @@
 #include "zk/zk_utilities.h"
 #include <functional>
 
+/*
+ * =============================================================================
+ * GPU Expand Algorithm: Overview
+ * =============================================================================
+ *
+ * The expand algorithm transforms compact ciphertexts into standard LWE
+ * ciphertexts. Compact ciphertexts save space by sharing a single mask across
+ * multiple encrypted messages.
+ *
+ * -----------------------------------------------------------------------------
+ * INPUT STRUCTURE (lwe_flattened_compact_array_in)
+ * -----------------------------------------------------------------------------
+ *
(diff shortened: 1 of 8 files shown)


tfhe/src/integer/gpu/zk/mod.rs
```diff
@@ -582,11 +588,25 @@ mod tests {
...
+        let bool_val: crate::FheBool = expander.get(0).unwrap().unwrap();
+        let u8_val_1: crate::FheUint8 = expander.get(1).unwrap().unwrap();
+        let u8_val_2: crate::FheUint8 = expander.get(2).unwrap().unwrap();
+        let u16_val: crate::FheUint16 = expander.get(3).unwrap().unwrap();
+        let u32_val: crate::FheUint32 = expander.get(4).unwrap().unwrap();
+        let u64_val: crate::FheUint64 = expander.get(5).unwrap().unwrap();
+
+        // Check if we retrieve the original values after decryption
+        let decrypted_m0: bool = bool_val.decrypt(&client_key);
+        assert_eq!(decrypted_m0, m0, "{} {}", decrypted_m0, m0);
```
IceTDrinker: assert_eq already prints the vaues if they differ, no need for the added message after the first tow args I would think

pdroalves: Indeed!
Thanks.

backends/tfhe-cuda-backend/cuda/src/zk/zk.cuh
```diff
@@ -15,6 +15,70 @@
 #include "zk/zk_utilities.h"
 #include <functional>
 
+/*
+ * =============================================================================
+ * GPU Expand Algorithm: Overview
```
pdroalves, line 20: @agnesLeroy Not sure this is the ideal PR for this change, but I added these diagrams and explanations (in zk_utilities.h too) about data structure and how the algorithm works.

PR #3215, opened 2026-01-27
Tm/hlapi gpu decomp uses expand

tmontaigu:
This PR has 6 commits, 3 that refactor the code, 1 that add a feature, 2 very small fixes

The refactors consists in making the decompression of hlapi::CompressedServerKey to CPU/GPU ServerKey (aka compute ready) a two step process: expand, then convert

The feature builds on the refactor to easily add the decompression of CompressedXofKeySet onto a GPU

closes https://github.com/zama-ai/tfhe-rs-internal/issues/1271

Some follow up cleanups could be done in a different PR:

- rename `decompress_with_pre_seed_generator` to `expand_with_pre_seed_generator` in xof_key_set/internal (optional)
- move the shortint specific 'Expanded' types (ShortintExpandedBootstrapKey, etc) into the shortint module

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3215)


tfhe/src/high_level_api/keys/expanded.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    pub(crate) fn polynomial_size(&self) -> PolynomialSize {
+        match self {
+            Self::Classic { bsk, .. } => bsk.polynomial_size(),
+            Self::MultiBit { bsk, .. } => bsk.polynomial_size(),
+        }
+    }
+}
+
+impl<ModSwitchScalar> ShortintExpandedBootstrappingKey<u64, ModSwitchScalar>
```
agnesLeroy, line 67: Why do you need "Shortint" prefixed types? For the keyswitching key we have a KeyswitchingKeyMaterial feature gated by "gpu" maybe it would be better to follow that pattern?

tmontaigu: That key is for GPU and CPU, and in shortint the bootstrapping keys are `ShortintBootstrappingKey` and `ShortintCompressedBootstrappingKey`

tfhe/src/high_level_api/xof_key_set/test.rs
```diff
@@ -1,132 +1,124 @@
...
-
-    assert_eq!(cks.tag(), compressed_key_set.compressed_public_key.tag());
-    assert_eq!(cks.tag(), &tag);
-    test_xof_key_set(&compressed_key_set, config, &cks);
+#[cfg(feature = "gpu")]
+mod gpu {
+    use super::*;
+
+    #[test]
+    fn test_xof_key_set_multibit_group_4() {
```
agnesLeroy: We would need to test both CPU parameters and GPU multi-bit parameters, as production will first happen with classical PBS.

tfhe/src/high_level_api/keys/expanded.rs
agnesLeroy: Ah ok I see, then indeed moving these shortint types to shortint would probably be good.

tfhe/src/high_level_api/xof_key_set/test.rs
agnesLeroy: Also shouldn't we cover the small PKE and big PKE cases also in the GPU case?

tmontaigu: We can cover the other PKE case

For the CPU parameters used at first, Are they exactly the same as CPU but we just use them for GPU instead of CPU ?

tfhe/src/high_level_api/keys/expanded.rs
tmontaigu: Yep I plan to do that in the next PR

tfhe/src/high_level_api/xof_key_set/test.rs
agnesLeroy: Yes :+1:

IceTDrinker:
this will need a backport when ready

tfhe/src/high_level_api/xof_key_set/test.rs
tmontaigu: Done.

IceTDrinker (commented):
Last nit, let me know if you want to go ahead as is, in that case I approve

> @IceTDrinker reviewed 6 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3215#-:-OkYa9_34yefLy1Cn-xZ:bmm84k8)* status: all files reviewed, 5 unresolved discussions (waiting on @agnesLeroy, @mayeul-zama, @nsarlin-zama, @pdroalves, and @tmontaigu).

___
*[`tfhe/src/high_level_api/xof_key_set/internal.rs` line 691 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3215#-Ok9HqyB7m9be0miYZmA:-OkY_Kxh6ttGJ15A6uT9:bawg96i) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f52eb16581e7b4b9cec77235b17f9779c1bf9a58/tfhe/src/high_level_api/xof_key_set/internal.rs#L691)):*

I removed to be consistent with its equivalent IntegerServerKey and IntegerCompressedServerKey that do not store the tag. The tag is stored by the ServerKey struct that wraps these Integer*ServerKey
</blockquote></details>

as long as it's not going to cause issues, the reason I'm wondering is because down the line we may want to store Expanded keys maybe ?
___
*[`tfhe/src/integer/gpu/list_compression/server_keys.rs` line 410 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3215#-OkY_gaAB7i4_8DVwSqd:-OkY_gaAB7i4_8DVwSqe:b61bej8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f21e7a1e5aa792c72ffac774df32ec8828923b9b/tfhe/src/integer/gpu/list_compression/server_keys.rs#L410)):*
> ```XML
>     ///
>     /// The glwe_dimensionn polynomial_sizen, message_modulus, carry_modulus, ciphertext_modulus
>     /// should be of the output parameters this key decompresses to
> ```

not sure this is correct then about the parameters, let's maybe not keep the new comment and let the GPU team figure out which is which and update on their end ?

PR #3216, opened 2026-01-27
fix(gpu): add __syncthreads and threadIdx condition for sample_extract_body in all pbs versions

bbarbakadze:
### PR content/description
In some versions of PBS there is no synchronization after `add_to_torus`  that can cause race condition for `sample_extract_body`, this pr adds synchronizations and thread id check for `sample_extract_body` since its modifying only one coefficient would be good if only one thread is responsible to process it.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_amortized.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_amortized.cuh
@@ -211,6 +211,8 @@ __global__ void device_programmable_bootstrap_amortized(
   // For the mask it's more complicated
   sample_extract_mask<Torus, params>(block_lwe_array_out, accumulator,
                                      glwe_dimension);
+
+  // No need to sync here, it is already synchronized after add_to_torus
   sample_extract_body<Torus, params>(block_lwe_array_out, accumulator,
                                      glwe_dimension);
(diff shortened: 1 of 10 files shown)


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_amortized.cuh
```diff
@@ -211,8 +211,11 @@ __global__ void device_programmable_bootstrap_amortized(
   // For the mask it's more complicated
   sample_extract_mask<Torus, params>(block_lwe_array_out, accumulator,
                                      glwe_dimension);
-  sample_extract_body<Torus, params>(block_lwe_array_out, accumulator,
-                                     glwe_dimension);
+
+  if (threadIdx.x == 0) {
```
guillermo-oyarzun: in sample extract body, do we always just take 1? if so, we could add this if inside the function and keep things a bit cleaner

agnesLeroy: Also maybe a comment explaining why we don't need to syncthreads in the places where it's not done would be good, you added it in one place but it would be good to have it in the other places as well if possible.

guillermo-oyarzun (approved):

PR #3220, opened 2026-01-28
ci(gpu): avoid running gpu noise in integer tests

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/scripts/test_filtering.py
+++ b/scripts/test_filtering.py
@@ -74,6 +74,12 @@ parser.add_argument(
     action="store_true",
     help="Do not run tests with big parameters set (e.g. 3bits message with 3 bits carry) for GPU",
 )
+parser.add_argument(
+    "--all-but-noise",
+    dest="all_but_noise",
+    action="store_true",
+    help="Run all tests except noise tests",
+)
 
 # block PBS are too slow for high params
 # mul_crt_4_4 is extremely flaky (~80% failure)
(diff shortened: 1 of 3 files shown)


scripts/test_filtering.py
```diff
@@ -175,7 +190,7 @@ def filter_integer_tests(input_args):
 
     # Do not run noise check tests by default as they can be very slow
     # they will be run e.g. nightly or on demand
-    filter =  filter_expression.append(f"not test(/^integer::gpu::server_key::radix::tests_noise_distribution::.*::test_gpu_noise_check.*/)")
+    filter_expression.append(f"not test(/^integer::gpu::server_key::radix::tests_noise_distribution::.*::test_gpu_noise_check.*/)")
```
IceTDrinker: isn't that enough ?

IceTDrinker (changes requested):
Stopping this PR for now as it seems the fix is only 2 lines

scripts/test_filtering.py
guillermo-oyarzun: nope, the problem is that for the test_integer_gpu we don't use the filtering at all, and if we want to use it, i need a filter that mimics the current behavior of test_integer_gpu, that's why i need to add the other flag

IceTDrinker: can it use the current filtering script with the two line fix then ?

guillermo-oyarzun: nope, the thing is that the current filter for integer is too restrictive, it runs only 256 tests, without the filter we were running 512 tests there. So the change in line 126 is to avoid the extra restrictions for the integer test, to keep the same amount of testing.
Line 193 already existed before, i just removed the filter variable that was never used.

IceTDrinker: one last thing could you check the filters remain equivalent for the CPU ?

it's hard to tell just from the code change

Listing the tests that would run with old filter/new filter should allow to distinguish, thanks !

IceTDrinker (approved):
Thanks, approving to unblock, looks ok to me!

scripts/test_filtering.py
soonum: Is @IceTDrinker request is fulfilled ?

IceTDrinker: Yes, CPU test does not use the flag

soonum (approved):
Nice I like better this consistency.
How long it takes to run noise tests on CPU @IceTDrinker ?

IceTDrinker:
Way too long to have in regular CI

PR #3222, opened 2026-01-28
refactor(gpu): improve f128 precision, bit-consistency, and performance

bbarbakadze:
### PR content/description
Refactor f128 and fft128 to prevent possible precision losses, keep bit-consistency and improve performance
### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/fft128/f128.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/fft128/f128.cuh
@@ -68,9 +68,15 @@ struct alignas(16) f128 {
     auto t = two_sum(a.lo, b.lo);
 
     double hi = s.hi;
+#ifdef __CUDA_ARCH__
+    double lo = __dadd_rn(s.lo, t.hi);
+    hi = __dadd_rn(hi, lo);
+    lo = __dsub_rn(lo, __dsub_rn(hi, s.hi));
+#else
     double lo = s.lo + t.hi;
     hi = hi + lo;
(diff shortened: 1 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/fft128/f128.cuh
```diff
@@ -68,9 +68,15 @@ struct alignas(16) f128 {
     auto t = two_sum(a.lo, b.lo);
 
     double hi = s.hi;
+#ifdef __CUDA_ARCH__
+    double lo = __dadd_rn(s.lo, t.hi);
+    hi = __dadd_rn(hi, lo);
+    lo = __dsub_rn(lo, __dsub_rn(hi, s.hi));
+#else
```
agnesLeroy, line 75: We always compile for Cuda code no, so what's in the else doesn't end up being used?

bbarbakadze: In case we need f128 arithmetics on CPU too,  or maybe for debugging. so like this it will work on both CPU and GPU

agnesLeroy: Makes sense

agnesLeroy: Actually it's nice to have the equivalent code without intrinsics it makes the code easier to read as well.

guillermo-oyarzun (approved):

PR #3229, opened 2026-01-29
feat(gpu): add zk-cuda-backend

pdroalves:
closes: None

### Check-list:


.gitignore
```diff
@@ -25,6 +25,9 @@ dieharder_run.log
 
 # Cuda local build
 backends/tfhe-cuda-backend/cuda/cmake-build-debug/
+backends/tfhe-cuda-backend/cuda/build/
```
agnesLeroy: I don't think we need to add the two build directories, these are local (cmake-build-debug maybe?)

_typos.toml
```diff
@@ -21,6 +21,7 @@ extend-exclude = [
     "backends/tfhe-cuda-backend/cuda/src/fft128/twiddles.cu",
     "backends/tfhe-cuda-backend/cuda/src/fft/twiddles.cu",
     "backends/tfhe-hpu-backend/config_store/**/*.link_summary",
+    "backends/zk-cuda-backend/cuda/build/**",
```
agnesLeroy: I don't think this is necessary, this is a local folder no?

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -33,12 +33,32 @@ fn bigint_to_le_bytes(x: [u64; 7]) -> [u8; 7 * 8] {
...
 
-mod g1 {
-    use tfhe_versionable::Versionize;
-
-    use crate::serialization::{InvalidSerializedAffineError, SerializableG1Affine};
+/// Extract normal-form limbs from an Fq2 element (for CUDA backend).
+/// Uses into_bigint() to convert from Montgomery to normal form.
+#[cfg(feature = "experimental-gpu")]
+#[inline]
+fn fq2_to_normal_limbs(fq2: &crate::curve_446::Fq2) -> ([u64; 7], [u64; 7]) {
```
nsarlin: I guess the 7 here and below could be made a constant

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -311,21 +636,219 @@ mod g2 {
...
+            use ark_ec::AffineRepr;
+            use ark_ff::Zero;
+            if value.inner.is_zero() {
+                return zk_cuda_backend::G2Affine::infinity();
+            }
+            // Use xy() to get normalized coordinates (not Montgomery form)
+            // zk-cuda-backend expects normal form and will convert to Montgomery itself
+            let zero = <crate::curve_446::Fq2 as Zero>::zero();
+            let xy = value.inner.xy().unwrap_or((zero, zero));
+            let x_c0_limbs = {
```
nsarlin: It looks like this could call `fq2_to_normal_limbs(xy.0)` and `fq2_to_normal_limbs(xy.1)`

_typos.toml
pdroalves: Without this I had it complaining about things in that folder.

agnesLeroy: You could remove the folder locally maybe?

tfhe/Cargo.toml
```diff
@@ -102,6 +102,7 @@ integer = ["shortint", "dep:strum"]
 strings = ["integer"]
 internal-keycache = ["dep:fs2"]
 gpu = ["dep:tfhe-cuda-backend", "shortint"]
+experimental-gpu-zk = ["gpu", "zk-pok", "tfhe-zk-pok/experimental-gpu"]
```
nsarlin-zama: small nitpick, this feature is `experimental-gpu-...` and the one below is `gpu-experimental-...`, it might be worth unifying this

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -33,12 +33,32 @@ fn bigint_to_le_bytes(x: [u64; 7]) -> [u8; 7 * 8] {
...
+    let c1_bigint = fq2.c1.into_bigint();
+
+    let mut c0_limbs = [0u64; 7];
+    let mut c1_limbs = [0u64; 7];
+    for (i, limb) in c0_bigint.as_ref().iter().take(7).enumerate() {
+        c0_limbs[i] = *limb;
+    }
+    for (i, limb) in c1_bigint.as_ref().iter().take(7).enumerate() {
+        c1_limbs[i] = *limb;
+    }
```
nsarlin-zama: Since bigint is a wrapper for `[u65; 7]`, here you can simply do:

```suggestion
    let c0_bigint = fq2.c0.into_bigint();
    let c1_bigint = fq2.c1.into_bigint();

    let c0_limbs = c0_bigint.0;
    let c1_limbs = c1_bigint.0;
```
You can do this in other places below.

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -78,21 +98,190 @@ mod g1 {
...
+                if is_infinity {
+                    return G1::ZERO;
+                }
+
+                let x = Fq::from_sign_and_limbs(true, &normalized.X());
+                let y = Fq::from_sign_and_limbs(true, &normalized.Y());
+                let z = Fq::from_sign_and_limbs(true, &z_limbs);
+                G1 {
+                    inner: G1Projective::new_unchecked(x, y, z),
+                }
```
nsarlin-zama: I believe this could simply call `multi_mul_scalar_gpu(bases, scalars, None)` to reduce code duplication.
And this can also be done for G2

tfhe-zk-pok/src/curve_api/bls12_446.rs
```diff
@@ -311,21 +636,219 @@ mod g2 {
...
+                limbs
+            };
+            let y_c1_limbs = {
+                let bigint = xy.1.c1.into_bigint();
+                let mut limbs = [0u64; 7];
+                for (i, limb) in bigint.as_ref().iter().take(7).enumerate() {
+                    limbs[i] = *limb;
+                }
+                limbs
+            };
```
nsarlin-zama: Here you could call the function above:

```suggestion
            let xy = value.inner.xy().unwrap_or((zero, zero));
            let (x_c0_limbs, x_c1_limbs) = fq2_to_normal_limbs(&xy.0);
            let (y_c0_limbs, y_c1_limbs) = fq2_to_normal_limbs(&xy.1);
```

tfhe-benchmark/benches/zk/msm.rs
```diff
@@ -0,0 +1,438 @@
...
+                    });
+                });
+            }
+            BenchmarkType::Throughput => {
+                // Throughput mode disabled - too slow
+                // Benchmarks are skipped when this mode is selected (see early return above)
+                // This branch is unreachable due to early return, but kept for code visibility
+                // To re-enable: uncomment the code below and remove the early return
+                /*
+                use criterion::Throughput;
```
nsarlin-zama: I believe we avoid having commented code in the code base, since it won't be tested or even compiled. I think it's better to add it if/when we need it someday.

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
```diff
@@ -26,6 +26,37 @@ fn bit_iter(x: u64, nbits: u32) -> impl Iterator<Item = bool> {
...
+    let num_gpus = unsafe { cuda_get_number_of_gpus() };
+    assert!(num_gpus > 0, "No GPU available");
+    num_gpus as u32
+}
+
+/// GPU selection for MSM operations.
+/// Returns `Some(gpu_index)` to run on a specific GPU, or `None` for default (GPU 0).
+/// Uses rayon thread index to distribute MSM operations across available GPUs.
+#[inline]
+fn select_gpu_for_msm() -> Option<u32> {
```
nsarlin-zama: maybe this logic could be moved to bls12_446.rs, inside `multi_mul_scalar` ? That way gpu specific impl does not leak here and this file can be left unmodified.
It looks like `multi_mul_scalar_gpu` is always called with the result of this function anyways.

_typos.toml
pdroalves: Maybe I've done something wrong with zk-cuda-backend compilation logic. It needs to create the build/ folder and store compilation files in there, which triggers in typos. Let me go back to this and see if I can do better.

pdroalves: Ok, it's easy. `build.rs` was set to write the compilation files there instead of reading the default output destination (which I believe is inside `target/`). Fixed.

tfhe/Cargo.toml
pdroalves: Actually all below are `gpu-*`. @agnesLeroy should I rename this to `gpu-experimental-zk`?

tfhe-zk-pok/src/curve_api/bls12_446.rs
pdroalves: Cool! I thought that by doing that I would be copying a pointer for the input, not the content. That's why I avoided it. I will refactor.

pdroalves: Done, check it now.

tfhe-zk-pok/src/curve_api/bls12_446.rs
pdroalves: Totally, I forgot to change this after introducing `multi_mul_scalar_with_gpu`.

tfhe-benchmark/benches/zk/msm.rs
pdroalves: Actually I need to reactivate throughput benchmarks. Probably I should do that (and uncomment this) before merging.

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
pdroalves: I don't think that's a good idea. Right now, yes, it is clear that this logic should be inside `multi_mul_scalar_with_gpu`. However the next step will be to have more operations running on the GPU, and all methods on the same "computation flow" need to be on the same device. WHen that happens I will need to compute the target device outside that method.

tfhe/Cargo.toml
agnesLeroy: Ah I had forgotten about that other experimental gpu feature. Ok let's have all names be gpu-experimental-something then to avoid having to rename the `gpu-experimental-multi-arch` 👍

tfhe-benchmark/benches/zk/msm.rs
agnesLeroy: As you want: there was an OOM error in the throughput bench if I remember correctly. That could be a separate PR for later, we could merge this one first without it.

.gitignore
agnesLeroy: I still see backends/tfhe-cuda-backend/cuda/build/ here

tfhe-benchmark/benches/zk/msm.rs
pdroalves: Done. Now we have throughput benchmarks working. cc @agnesLeroy

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
nsarlin-zama: Ah yes understood ! What worries me a bit is that when reading the code for prove/verify it looks like the msm is always done on gpu, whereas it actually depends on feature flags. 
And when you do this next step you will have to duplicate prove/verify anyways?

pdroalves: Ah yes, you have a point. I don't think we should be modifying the code in `verify_impl` / `prove_impl` so deeply.

backends/zk-cuda-backend/LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause Clear License
+
+Copyright © 2025 ZAMA.
```
IceTDrinker: 2026

backends/zk-cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+keywords = ["fully", "homomorphic", "encryption", "fhe", "cryptography"]
+
+[lib]
+name = "zk_cuda_backend"
+crate-type = ["rlib"]
+
+[dependencies]
+ark-ec = "0.5.0"
+ark-ff = "0.5.0"
+ark-serialize = "0.5.0"
```
IceTDrinker: @pdroalves we likely need to put those in the workspace deps to have them synced between the GPU and CPU

I'm not sure I saw ark-serialize being used in the GPU rust code though ?

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
pdroalves: I see two options:

1) Revert to the generic `multi_mul_scalar` call (like on main) and handle the GPU acceleration inside that function behind a compile-time feature flag. This keeps the call sites simple, avoids implying the GPU path is always used, and generally makes the code easier to read.
2) Branch in `verify_impl` and `prove_impl` using `#[cfg(feature = "gpu-experimental-zk")]` to select CPU vs GPU explicitly. This adds some complexity to those functions, but it sets up the structure we’ll need later. Once I implement pairing (and scalar multiplication) and start fusing these operations, we’ll need explicit branching like this anyway.

What do you think?

backends/zk-cuda-backend/src/conversions.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Helper to extract limbs from an arkworks field element
+fn field_to_limbs<F: PrimeField>(f: &F) -> [u64; 7] {
+    let bigint = f.into_bigint();
+    let mut limbs = [0u64; 7];
+    for (i, limb) in bigint.as_ref().iter().take(7).enumerate() {
+        limbs[i] = *limb;
+    }
+    limbs
```
IceTDrinker: limbs.as_mut().copy_from_slice(bigint.as_ref()) might be enough ?

if the size is an issue could restrict by doing

limbs.as_mut().copy_from_slice(bigint.as_ref()[..7])

tfhe-benchmark/benches/integer/zk_pke.rs
```diff
@@ -34,7 +34,7 @@ impl ProofConfig {
 fn default_proof_config() -> Vec<ProofConfig> {
     vec![
         ProofConfig::new(64, &[64]),
-        ProofConfig::new(2048, &[64, 4 * 64, 2048]),
+        ProofConfig::new(2048, &[64, 4 * 64, 10 * 640, 2048]),
```
IceTDrinker: not sure about that 10 * 640 new config ? what's the intended use/effect for the benchmarks ?

tfhe-benchmark/benches/integer/zk_pke.rs
```diff
@@ -65,8 +65,26 @@ fn zk_throughput_num_elements() -> u64 {
     }
 }
 
+/// Compute the number of elements for GPU ZK throughput benchmarks.
+/// Values are tuned to avoid OOM on H100 GPUs while still saturating the GPU.
+/// Memory usage scales with both CRS size and bits being proven.
+#[cfg(all(feature = "gpu", feature = "zk-pok"))]
+fn gpu_zk_throughput_elements(crs_size: usize, bits: usize) -> u64 {
```
IceTDrinker: could be put in the cuda module directly to avoid having to repeat the cfg attributes

tfhe/Cargo.toml
pdroalves: Done.

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
nsarlin-zama: If you think that you will need to add this kind of branching later in prove_impl/verify_impl, I think option 2 is better, then !

backends/zk-cuda-backend/build.rs
```diff
@@ -0,0 +1,223 @@
...
+            "/usr/local/cuda/lib64",
+            "/usr/lib/x86_64-linux-gnu",
+            "/opt/cuda/lib64",
+        ];
+        for path in COMMON_PATHS {
+            if PathBuf::from(path).exists() {
+                println!("cargo:rustc-link-search=native={}", path);
+            }
+        }
+    }
```
IceTDrinker: isn't there something in the existing cuda backend that already manages part of that ?

tfhe-zk-pok/benches/utils.rs
```diff
@@ -344,6 +344,7 @@ impl PkeTestcase {
 }
 
 /// Compact key params used with pkev2
+#[allow(dead_code)]
```
IceTDrinker: I'm surprised this is needed

tfhe-zk-pok/src/curve_446/mod.rs
```diff
@@ -60,6 +60,14 @@ impl Fp2Config for Fq2Config {
     }
 }
 
+/// Create an Fq2 from sign and limb arrays for both components
+pub fn fq2_from_fq_sign_and_limbs(sign: bool, c0_limbs: &[u64; 7], c1_limbs: &[u64; 7]) -> Fq2 {
```
IceTDrinker: is that a GPU specific function ?

tfhe-zk-pok/Cargo.toml
```diff
@@ -24,9 +24,13 @@ serde = { workspace = true, features = ["default", "derive"] }
 zeroize = "1.7.0"
 num-bigint = "0.4.5"
 tfhe-versionable = { version = "0.7.0", path = "../utils/tfhe-versionable" }
+zk-cuda-backend = { path = "../backends/zk-cuda-backend", optional = true }
+tfhe-cuda-backend = { path = "../backends/tfhe-cuda-backend", optional = true }
```
IceTDrinker: those will need a version to avoid problems when publishing to crates.io

Makefile
```diff
@@ -1513,6 +1554,10 @@ bench_signed_integer_multi_bit_gpu: install_rs_check_toolchain
 
 .PHONY: bench_integer_zk # Run benchmarks for integer encryption with ZK proofs
 bench_integer_zk: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" __TFHE_RS_BENCH_TYPE=$(BENCH_TYPE) \
+	cargo $(CARGO_RS_CHECK_TOOLCHAIN) bench \
+	--bench zk-msm \
+	--features=zk-pok -p tfhe-benchmark --profile release_lto_off --
```
IceTDrinker: you already have the zk bench target above ?

also releast lto off on CPU will be bad 

edit: use release

Makefile
```diff
@@ -1451,26 +1478,40 @@ bench_integer_compression_128b_gpu: install_rs_check_toolchain
 	--bench	glwe_packing_compression_128b-integer-bench \
 	--features=integer,internal-keycache,gpu,pbs-stats -p tfhe-benchmark --
 
+.PHONY: bench_msm_zk
+bench_msm_zk: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" __TFHE_RS_BENCH_TYPE=$(BENCH_TYPE) \
+	cargo $(CARGO_RS_CHECK_TOOLCHAIN) bench \
+	--bench zk-msm \
+	--features=zk-pok -p tfhe-benchmark --profile release_lto_off --
```
IceTDrinker: release profile for CPU

tfhe-zk-pok/src/curve_api.rs
```diff
@@ -114,6 +114,16 @@ pub trait CurveGroupOps<Zp>:
...
+    /// Multi-scalar multiplication with optional GPU device selection.
+    /// If `gpu_index` is `None`, defaults to GPU 0 (or CPU if GPU is not available).
+    fn multi_mul_scalar_with_gpu(
+        bases: &[Self::Affine],
+        scalars: &[Zp],
+        _gpu_index: Option<u32>,
+    ) -> Self {
+        // Default implementation ignores gpu_index and calls the regular method
+        Self::multi_mul_scalar(bases, scalars)
+    }
```
IceTDrinker: where is this function called in practice ?

it feels weird to have functions that are only for the GPU on the generic trait

you could have extension traits for the GPU to group the various methods

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
```diff
@@ -2182,19 +2245,21 @@ fn pairing_check_two_steps<G: Curve>(
...
             lhs3 = Some(pairing(
                 C_R,
-                G::G2::multi_mul_scalar(
+                G::G2::multi_mul_scalar_with_gpu(
                     &g_hat_list[n - 128..n],
                     &(0..128)
                         .rev()
                         .map(|j| delta_r * phi[j] + delta_dec * xi[j])
                         .collect::<Box<[_]>>(),
+                    select_gpu_for_msm(),
```
IceTDrinker: re my comment for GPU : you don't need the function it seems given the GPU specific implementation can call the select_gpu_for_msm itself directly ?

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    let x = Fq::from_sign_and_limbs(true, &x_limbs);
+    let y = Fq::from_sign_and_limbs(true, &y_limbs);
+
+    use crate::curve_446::g1::G1Projective;
+    let one = MontFp!("1");
+    let proj = G1Projective::new_unchecked(x, y, one);
+    let inner = <G1Projective as CurveGroup>::into_affine(proj);
+
+    // SAFETY: G1Affine is repr(transparent)
+    unsafe { core::mem::transmute(inner) }
```
IceTDrinker: does it really need a transmute here ?

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    use crate::curve_446::g2::G2Projective;
+    let one = MontFp!("1");
+    let zero = MontFp!("0");
+    let one_fq2 = Fq2::new(one, zero);
+    let proj = G2Projective::new_unchecked(x, y, one_fq2);
+    let inner = <G2Projective as CurveGroup>::into_affine(proj);
+
+    // SAFETY: G2Affine is repr(transparent)
+    unsafe { core::mem::transmute(inner) }
```
IceTDrinker: same question, I would think you could take the data and rewrap it in the cuda type ?

backends/zk-cuda-backend/src/conversions.rs
nsarlin-zama: or simply `bigint.0` as discussed above

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
nsarlin-zama: @IceTDrinker I think it could be interesting to have your opinion here since this overlaps some of your comments

nsarlin-zama: Also note that the wasm unsafe-coop mode will likely also require some cfg specific quirks in prove_impl...

backends/zk-cuda-backend/src/conversions.rs
IceTDrinker: might have missed the other discussion, you are right

pdroalves: Yep, I replaced by `bigint.0`. As I explained, I thought this would be copying a pointer and not the content. C mindset :-P

tfhe-benchmark/benches/integer/zk_pke.rs
pdroalves: None. At some point I was trying to bench this for a very big instance. I will rollback.

backends/zk-cuda-backend/Cargo.toml
pdroalves: Yeah, we can remove ark-serialize and cc. They are not used anymore, thanks.

I added `ark-ec` and `ark-ff` to workspace dependencies. I probably should have `zk-cuda-backend` as a workspace member as well, right?

backends/zk-cuda-backend/build.rs
pdroalves: Yes, we should be compiling it using the exactly same script. As is this is not a good practice. I just copied tfhe's `build.rs` here to make sure everything is done in the same way.

tfhe-zk-pok/src/curve_446/mod.rs
pdroalves: Yes. Should be behind "gpu-experiment" and not here. I will move it..

Makefile
pdroalves: Yes, but I refactored the bench target a few iteration ago so MSM has its own. Forgot this one.

pdroalves: About the profile, we use release_lto_off in maybe all GPU targets. Is that wrong?

tfhe-zk-pok/Cargo.toml
pdroalves: Should I simply hardcode the current versions? `0.13.0` for `tfhe-cuda-backend` and `0.1.0` for `zk-cuda-backend`?

tfhe-zk-pok/src/proofs/pke_v2/mod.rs
pdroalves: Nicolas mentioned the same. Please check our discussion: https://github.com/zama-ai/tfhe-rs/pull/3229#discussion_r2754157988

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: I can replace it by a constructor, `::from_inner(inner)`. Would you prefer it?

pdroalves: I use this to convert from arkworks. From what I read this is cheaper than a constructor.

nsarlin-zama: Since this is repr(transparent), G1Affine does not really exist in the compiled binary, so you should be able to wrap it for free without unsafe.

backends/zk-cuda-backend/src/conversions.rs
nsarlin-zama: To give you a bit of context, this is because [T; N] implements Copy if T implements Copy: https://doc.rust-lang.org/std/primitive.array.html#impl-Copy-for-%5BT;+N%5D
This basically means that there is no pointer inside and the compiler will be able to automatically call `memcpy` on the content.

tfhe-zk-pok/src/gpu/mod.rs
nsarlin-zama: If you want to check you can use [cargo asm](https://github.com/gnzlbg/cargo-asm):
```
cargo asm --features experimental-gpu -p tfhe-zk-pok --lib --release g1_affine_from_zk_cuda
```
It looks like there is no difference in the resulting assembly between safe and unsafe

backends/zk-cuda-backend/src/c_wrapper.cu
```diff
@@ -0,0 +1,402 @@
...
+        cuda_memcpy_with_size_tracking_async_gpu_to_gpu(d_points_converted, d_points, n * sizeof(G2Affine), stream, gpu_index, true);
+        convert_g2_points_to_montgomery(stream, gpu_index, d_points_converted, n);
+        auto err = cudaGetLastError();
+        PANIC_IF_FALSE(err == cudaSuccess, "G2 MSM error: Montgomery conversion failed: %s", cudaGetErrorString(err));
+        points_to_use = d_points_converted;
+    }
+    
+    point_msm_async_g2(stream, gpu_index, d_result, points_to_use, d_scalars, d_scratch, n);
+
+    auto err = cudaPeekAtLastError();
```
andrei-stoian-zama: maybe `  check_cuda_error(cudaGetLastError());` ?

backends/zk-cuda-backend/src/c_wrapper.cu
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    cuda_synchronize_stream(stream, gpu_index);
+}
+
+// Managed MSM wrapper for G1 (handles memory management internally)
+// This allocates device memory, copies data, runs MSM, and copies result back
+void g1_msm_managed_wrapper(
+    G1Projective* result,
+    cudaStream_t stream,
```
andrei-stoian-zama, line 173: I suggest streams should always be argument 0

backends/zk-cuda-backend/src/c_wrapper.cu
```diff
@@ -0,0 +1,402 @@
...
+    const auto threadsPerBlock = get_msm_threads_per_block<G1Affine>(n);
+    const auto num_blocks = (n + threadsPerBlock - 1) / threadsPerBlock;
+    size_t scratch_elems = 0;
+    PANIC_IF_FALSE(!__builtin_mul_overflow((size_t)(num_blocks + 1),
+                                          (size_t)MSM_G1_BUCKET_COUNT,
+                                          &scratch_elems),
+                   "G1 MSM error: scratch element count overflow (num_blocks=%u)",
+                   num_blocks);
+    size_t scratch_size = 0;
+    PANIC_IF_FALSE(!__builtin_mul_overflow(scratch_elems, sizeof(G1Projective),
```
andrei-stoian-zama: I'm not very confortable with this use of PANIC_IF_ELSE. 

It's true that this panic is always enabled. But if, for whatever reason, in the future, the macro might be disabled by `#define PANIC_WITH_ELSE(...) ();` the code inside (the call to __builtin_mul_overflow) isn't compiled anymore. 

I prefer the `  check_cuda_error(cudaGetLastError());` approach. it's always called separately from the actual function that might produce an error. 

In our case we'd have:

`if !(__builtin__(...)) PANIC("G1 MSM...)`

backends/zk-cuda-backend/src/c_wrapper.cu
```diff
@@ -0,0 +1,402 @@
...
+    size_t points_bytes = 0;
+    PANIC_IF_FALSE(!__builtin_mul_overflow((size_t)n, sizeof(G1Affine), &points_bytes),
+                   "G1 MSM error: points byte size overflow (n=%u)", n);
+    size_t scalars_bytes = 0;
+    PANIC_IF_FALSE(!__builtin_mul_overflow((size_t)n, sizeof(Scalar), &scalars_bytes),
+                   "G1 MSM error: scalars byte size overflow (n=%u)", n);
+    
+    auto* d_points = static_cast<G1Affine*>(cuda_malloc_with_size_tracking_async(points_bytes, stream, gpu_index, size_tracker_ref, true));
+    auto* d_scalars = static_cast<Scalar*>(cuda_malloc_with_size_tracking_async(scalars_bytes, stream, gpu_index, size_tracker_ref, true));
+    auto* d_result = static_cast<G1Projective*>(cuda_malloc_with_size_tracking_async(sizeof(G1Projective), stream, gpu_index, size_tracker_ref, true));
```
andrei-stoian-zama, line 230: should we try, maybe in a later PR, to split into scratch / compute / release ?

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: Wonderful. I will replace it by `G1Affine{inner}`.

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: Replaced by `G2Affine{inner}`

tfhe-zk-pok/src/gpu/mod.rs
nsarlin-zama: and I guess now you can also remove the static asserts ! :)

backends/zk-cuda-backend/src/c_wrapper.cu
```diff
@@ -0,0 +1,402 @@
...
+        cuda_drop_with_size_tracking_async(d_points_converted, stream, gpu_index, true);
+    }
+
+    cuda_synchronize_stream(stream, gpu_index);
+}
+
+// Unmanaged MSM wrapper for G2 (assumes all data is already on device)
+// If points_in_montgomery is false, a temporary copy will be made and converted.
+// For best performance, provide points already in Montgomery form to avoid allocation overhead.
+void g2_msm_unmanaged_wrapper(
```
andrei-stoian-zama, line 121: we don't need to expose this to rust no? 

and thus we can avoid the `cudaStream_t` in the bindings as well?

backends/zk-cuda-backend/src/c_wrapper.cu
pdroalves: For sure. That's the end goal.  I was thinking on doing that later, when they really justify the change. I will add a comment "TODO: " mentioning that so I don't forget (eventually I would remember anyway, but...). That's going to be a non-trivial task and I want to close this PR ASAP.

backends/zk-cuda-backend/src/c_wrapper.cu
pdroalves: Eventually we will be using only `_unmanaged_`, so it's not bad to already expose it.

pdroalves:
@andrei-stoian-zama I fixed those, thanks.

About the duplicated symbols, I think it's because zk-cuda-backend depends on device.cu. I couldn't find a way to compile without that.

backends/zk-cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+version = "0.1.0"
+edition = "2021"
+authors = ["Zama team"]
+license = "BSD-3-Clause-Clear"
+description = "Cuda implementation of TFHE-rs' ZK primitives."
+homepage = "https://www.zama.ai/"
+documentation = "https://docs.zama.ai/tfhe-rs"
+repository = "https://github.com/zama-ai/tfhe-rs"
+readme = "README.md"
+keywords = ["fully", "homomorphic", "encryption", "fhe", "cryptography"]
```
IceTDrinker, line 13: you can add

`rust-version.workspace = true`

to the Cargo.toml to make sure the minimum rust version is consistent with the rest of the workspace

backends/zk-cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[build-dependencies]
+cmake = { version = "0.1" }
+pkg-config = { version = "0.3" }
+bindgen = "0.71"
+
+[dependencies]
+ark-ec.workspace = true
+ark-ff.workspace = true
+tfhe-cuda-backend = { path = "../tfhe-cuda-backend" }
```
IceTDrinker: forgot to ask : why is the cuda backend a dependency of the zk code ? feels like it should not be required to use FHE primitives in the ZK ?

.github/workflows/gpu_zk_tests.yml
```diff
@@ -0,0 +1,174 @@
+# Compile and test zk-cuda-backend
```
IceTDrinker, line 1: @soonum if you can check that the workflow looks reasonable to you

backends/zk-cuda-backend/src/tests/basic.rs
```diff
@@ -0,0 +1,137 @@
...
+#[test]
+fn test_g1_affine_creation() {
+    // Create a G1 affine point at infinity
+    let g1_inf = G1Affine::infinity();
+    assert!(g1_inf.is_infinity());
+
+    // Create a G1 affine point from coordinates
+    let x = [0x1234567890abcdef, 0, 0, 0, 0, 0, 0];
+    let y = [0xfedcba0987654321, 0, 0, 0, 0, 0, 0];
+    let g1_point = G1Affine::new(x, y, false);
```
IceTDrinker: I'm no expert on elliptic curves, given the points are constructed from manual coordinates, are we sure those are in fact on the curve ?

Same remark for all tests

backends/zk-cuda-backend/src/tests/basic.rs
```diff
@@ -0,0 +1,137 @@
...
+
+#[test]
+fn test_scalar_multi_limb() {
+    // Create a scalar with multiple limbs: 2^64 + 1
+    let scalar = Scalar::new([1u64, 1u64, 0u64, 0u64, 0u64]);
+    let limbs = scalar.limbs();
+    assert_eq!(limbs[0], 1);
+    assert_eq!(limbs[1], 1);
+    assert_eq!(limbs[2], 0);
+}
```
IceTDrinker: why are the 3rd and 4th limb not checked ?

backends/zk-cuda-backend/src/conversions.rs
```diff
@@ -0,0 +1,92 @@
...
+    G2Projective as ZkG2Projective,
+};
+use ark_ec::AffineRepr;
+use ark_ff::{PrimeField, Zero};
+
+/// Type alias for Fp2 field element represented as limbs (c0, c1)
+pub type Fp2Limbs = ([u64; 7], [u64; 7]);
+
+/// Type alias for G2 projective point coordinates (X, Y, Z) as Fp2 limbs
+pub type G2ProjectiveLimbs = (Fp2Limbs, Fp2Limbs, Fp2Limbs);
```
IceTDrinker: if you want strict typing (to avoid mistaking a tuple for another) 

you can define a tuple struct like

pub struct G2ProjectiveLimbs (Fp2Limbs, Fp2Limbs, Fp2Limbs);

same for the Fp2Limbs above

it's not strictly necessary, just that a type alias has less type check guarantees than a struct

backends/zk-cuda-backend/src/conversions.rs
IceTDrinker: I maintain you should be able to use copy_from_slice since BigInteger is AsRef<[u64]>, if not then I'd be curious to know what the compiler error is

the bigin.as_ref is returning &[u64], so you should be able to copy it with

limbs.as_mut_slice().copy_from_slice(bigin.as_ref());

no need for the loop I would think

backends/zk-cuda-backend/src/ffi.rs
```diff
@@ -0,0 +1,35 @@
...
+//! by bindgen from wrapper.h during the build process. Do not edit them manually.
+//! See build.rs for the bindgen configuration.
+
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+#![allow(dead_code)]
+#![allow(clippy::all)]
+
+include!("bindings.rs");
```
IceTDrinker: can't bindings be a mod declared in lib.rs rather ?
like
lib.rs
```rust
pub mod bindings;
```
including files is consideredt to be very hackish and seems like it's not required here ?
then you can do
```
pub use crate::bindings::*;
```
if needed ?

backends/zk-cuda-backend/src/ffi.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl Fp {
+    /// Create a new Fp from limbs (in normal form - must convert to Montgomery for computation)
+    pub fn new(limbs: [u64; 7]) -> Self {
+        Self { limb: limbs }
+    }
+
+    /// Create a new Fp from a BigInt-like structure (in normal form)
+    /// This is a convenience method for compatibility with tfhe-zk-pok's BigInt<7>
+    pub fn from_bigint<const N: usize>(bigint: &[u64; N]) -> Self {
```
IceTDrinker: should this throw an error if N > 7 ?

backends/zk-cuda-backend/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+//! // Convert back to affine
+//! let g1_affine_again = g1_proj.to_affine();
+//! ```
+
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+// Force linking of tfhe-cuda-backend which provides device utilities (cuda_malloc, cuda_set_device,
+// etc.)
+extern crate tfhe_cuda_backend;
```
IceTDrinker, line 49: is there going to be a problem long term with linking several things together ?

also is the full CUDA backend compiled as a result ?

not required in the short term I guess but having a common crate with the cuda bindings might be something you would want to do

backends/zk-cuda-backend/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+pub mod conversions;
+pub mod ffi;
+pub mod types;
+
+#[cfg(test)]
+mod tests;
+
+pub use types::*;
+// Re-export Fp from ffi module (it's the Fq equivalent - 7-limb field element)
+pub use ffi::Fp;
```
IceTDrinker: you could rather do pub use bindings::Fp

given Fp is defined in the bindings, not in the ffi module

what's the purpose of the ffi module ?

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+        Self { inner: point }
+    }
+
+    /// Check if this point is at infinity
+    #[inline]
+    pub fn is_infinity(&self) -> bool {
+        // SAFETY: `self.inner` is a valid G2Point with repr(C) layout that remains
+        // valid for the duration of this call. The FFI function only reads from it.
+        unsafe { crate::ffi::g2_is_infinity_wrapper(&self.inner) }
+    }
```
IceTDrinker: how is the infinity status checked in CUDA ?

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+        if stream.is_null() {
+            return Err("GPU MSM: stream pointer is null".to_string());
+        }
+        if points.len() > u32::MAX as usize {
+            return Err(format!(
+                "GPU MSM: input length too large for u32: {}",
+                points.len()
+            ));
+        }
+        let n = points.len() as u32;
```
IceTDrinker: could do

let n: u32 = poitns.len().try_into().map_err(format!(
                "GPU MSM: input length too large for u32: {}",
                points.len()
            ))?;

and remove the check of points len above

see https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+        // For projective points, we compare all coordinates directly.
+        // Note: This is coordinate equality, not geometric equality (which would require
+        // checking if they represent the same point via cross-multiplication).
+        self.inner.X.limb == other.inner.X.limb
+            && self.inner.Y.limb == other.inner.Y.limb
+            && self.inner.Z.limb == other.inner.Z.limb
+    }
+}
+
+impl Eq for G1Projective {}
```
IceTDrinker: use derive Macros

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+
+        Self::new(result)
+    }
+}
+
+impl fmt::Debug for Scalar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Scalar({:?})", self.limbs())
+    }
+}
```
IceTDrinker: use derive macro

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+
+impl Eq for Scalar {}
+
+impl Default for Scalar {
+    fn default() -> Self {
+        Self::from_u64(0)
+    }
+}
```
IceTDrinker: use derive macros

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+        let affine = self.to_affine();
+        write!(f, "{}", affine)
+    }
+}
+
+impl Default for G2Projective {
+    fn default() -> Self {
+        Self::infinity()
+    }
+}
```
IceTDrinker: these cannot be generated by derive macros so fine to keep as is

unless infinity is full of 0s in which case the default impl would already fill with 0s normally

backends/zk-cuda-backend/src/types.rs
```diff
@@ -0,0 +1,927 @@
...
+        self.inner.X.c0.limb == other.inner.X.c0.limb
+            && self.inner.X.c1.limb == other.inner.X.c1.limb
+            && self.inner.Y.c0.limb == other.inner.Y.c0.limb
+            && self.inner.Y.c1.limb == other.inner.Y.c1.limb
+            && self.inner.Z.c0.limb == other.inner.Z.c0.limb
+            && self.inner.Z.c1.limb == other.inner.Z.c1.limb
+    }
+}
+
+impl Eq for G2Projective {}
```
IceTDrinker: use derive macros

backends/zk-cuda-backend/build.rs
```diff
@@ -0,0 +1,140 @@
...
+                // Allow the core types needed for FFI
+                .allowlist_type("G1Point")
+                .allowlist_type("G2Point")
+                .allowlist_type("G1ProjectivePoint")
+                .allowlist_type("G2ProjectivePoint")
+                .allowlist_type("Fp")
+                .allowlist_type("Fp2")
+                .allowlist_type("Scalar")
+                .allowlist_type("BigInt")
+                .allowlist_type("cudaStream_t")
```
IceTDrinker, line 119: just wondering why you preferred to be explicit here for types ? if you add new type you will have to export them here as well

I came across https://github.com/jorenham/xsf-rust which had a nice build.rs file IIRC, if you need some examples it was well made for the configuration part

backends/zk-cuda-backend/Cargo.toml
pdroalves: I don't want to duplicate code, and the zk backend depends on device.cu and the cuda stream structure defined in tfhe backend. I thought would be a good practice to make sure the tfhe backend is really there.

IceTDrinker: if it's only a small number of functions being shared could be worth having a more limited binding that's shared between both projects, since here I guess you are recompiling all the backend code (even the FHE ops)

it's fine as a first step, but probably not in the long run

backends/zk-cuda-backend/src/tests/basic.rs
pdroalves: There are checks internally that will not execute the MSM otherwise.

Anyway I think I will remove these tests. They were there initially to make sure the rust binding worked and I would not break them, but with the current integration they are not needed anymore. We don't have anything like that in the tfhe backend.

backends/zk-cuda-backend/Cargo.toml
pdroalves: Yes, compilation here is definitely not being efficient.
We've discussed reorganizing the backends to have shared code on a common place, maybe even making use of C++'s namespace. I don't know if this level of change in the tfhe backend is something we want to do know, wdyt @agnesLeroy ?

backends/zk-cuda-backend/src/ffi.rs
pdroalves: Indeed! That's what we use in the tfhe backend. I will refactor it.

backends/zk-cuda-backend/src/ffi.rs
pdroalves: Could be good indeed. I will add two compile time asserts to make sure we that that FP_LIMBS is 7 and ZP_LIMBS is 5.

backends/zk-cuda-backend/src/lib.rs
pdroalves: Same problem as the other comment. We need both benchmarks when tfhe or tfhe-zk-pok are compiled since the entire memory allocation and streams logic lie in tfhe-cuda-backend. At some point we should refactor that.

Cargo.toml
```diff
@@ -26,6 +27,8 @@ rust-version = "1.91.1"
 
 [workspace.dependencies]
 aligned-vec = { version = "0.6", default-features = false }
+ark-ec = "0.5.0"
+ark-ff = "0.5.0"
```
agnesLeroy, line 31: What are these two new dependencies?

backends/zk-cuda-backend/src/types.rs
pdroalves: If the point is in affine representation we check a boolean. If projective, we check if Z=0.

`P + (-P)` will set one of those attributes to the infinity state, as well as the point constructor.

backends/zk-cuda-backend/src/types.rs
pdroalves: Of course. I always forget about derive macros. Ty!

backends/zk-cuda-backend/src/types.rs
pdroalves: Didn't know about the Default impl. I will remove it then since infinity in projective repr is just zeros.

backends/zk-cuda-backend/Cargo.toml
agnesLeroy: I would leave it for later I think, we already have many things ongoing right now

backends/zk-cuda-backend/build.rs
pdroalves: In the tfhe backend we use `allowlist_type` to expose `PBS_TYPE` and `SHIFT_OR_ROTATE_TYPE`, I just followed the pattern here. Maybe we should change both @agnesLeroy ?

Cargo.toml
pdroalves: I think it was after a comment by @IceTDrinker . We use those two dependencies in zk-cuda-backend and tfhe-zk-pok. I understand that by adding them here we can assert both crates are being compiled with the same version (when we set `.workspace = true` in their cargos).

pdroalves: We actually have the same problematic pattern with cmake, pkg-config, and bindgen and tfhe-cuda-backend. I'm adding them to the root Cargo.toml.

backends/zk-cuda-backend/src/conversions.rs
nsarlin-zama: it can simply be replaced by:
```suggestion
    bigint.0
```

backends/zk-cuda-backend/src/types.rs
IceTDrinker: you can add a test to verify that the output of the default call is indeed infinity this way you can trust the default implem instead of writing it manually

IceTDrinker: Essentially integer values default to 0 normally, all types can define a default value : https://doc.rust-lang.org/std/default/trait.Default.html

backends/zk-cuda-backend/src/lib.rs
pdroalves: We have it in the tfhe backend, but looking again we don't really need it here. I'm removing it.

backends/zk-cuda-backend/build.rs
IceTDrinker: not saying it's bad thing, just that there may be a way to make thigns easier for you

agnesLeroy: At some point yes, but it's not a very big issue I would say, we have more urgent things ongoing. We can open an issue in the backlog for that though to not forget it.

backends/zk-cuda-backend/src/conversions.rs
pdroalves: @nsarlin-zama f is a generic BigInt, but here I'm converting specially to a FP element (with 7 limbs). Probably It works if I use `bigint.as_ref()`.

```
        Compiling zk-cuda-backend v0.1.0 (/home/zama/Code/zk-gpu-accelerated/public_gpu_zk/backends/zk-cuda-backend)                                                                          
     error[E0609]: no field `0` on type `<F as PrimeField>::BigInt`
       --> backends/zk-cuda-backend/src/conversions.rs:40:23                                                                                                                                  
        |                                                                                                                                                                                     
     40 |     Fp { limb: bigint.0 }
        |                       ^ unknown field                                                                                                                                               
                                                                                                                   
     For more information about this error, try `rustc --explain E0609`.
     error: could not compile `zk-cuda-backend` (lib) due to 1 previous error
```

Cargo.toml
IceTDrinker: correct on the workspace = true, allows to have the same versions

for tfhe-cuda-backend you should add a version in your Cargo.toml that corresponds to the version in the repo

backends/zk-cuda-backend/src/conversions.rs
nsarlin-zama: ah yes indeed sorry, I missed that this one was a trait

backends/zk-cuda-backend/build.rs
pdroalves: The issue is open https://github.com/zama-ai/tfhe-rs-internal/issues/1325

backends/zk-cuda-backend/src/types.rs
pdroalves: @IceTDrinker I like that. Is this what you have in mind? https://github.com/zama-ai/tfhe-rs/blob/eafe386086a2f300c19418eec71653eb93ac5499/backends/zk-cuda-backend/src/types.rs#L872

.github/workflows/gpu_zk_tests.yml
```diff
@@ -0,0 +1,174 @@
...
+  PULL_REQUEST_MD_LINK: ""
+  CHECKOUT_TOKEN: ${{ secrets.REPO_CHECKOUT_TOKEN || secrets.GITHUB_TOKEN }}
+  # Secrets will be available only to zama-ai organization members
+  SECRETS_AVAILABLE: ${{ secrets.JOB_SECRET != '' }}
+  EXTERNAL_CONTRIBUTION_RUNNER: "gpu_ubuntu-22.04"
+
+on:
+  # Allows you to run this workflow manually from the Actions tab as an alternative.
+  workflow_dispatch:
+  pull_request:
```
soonum, line 25: Do we want to run it on every push ? 
I know there is a `should-run` check, but if this is a long test suite then we should consider changing the trigger to react on approval labeling.

.github/workflows/gpu_zk_tests.yml
```diff
@@ -0,0 +1,174 @@
...
+        env:
+          PR_BASE_URL: ${{ vars.PR_BASE_URL }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+
+      - name: Send message
+        if: env.SECRETS_AVAILABLE == 'true'
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661
+        env:
+          SLACK_COLOR: ${{ needs.cuda-tests-linux.result }}
+          SLACK_MESSAGE: "Base GPU tests finished with status: ${{ needs.cuda-tests-linux.result }}. (${{ env.PULL_REQUEST_MD_LINK }}[action run](${{ env.ACTION_RUN_URL }}))"
```
soonum: `Base GPU tests ...` doesn't seem right to me, maybe some like `ZK GPU tests ...`

soonum (commented):
Looks go to me, but just a clarification needed about workflow trigger in Pull-request.

.github/workflows/gpu_zk_tests.yml
pdroalves: We want to run it only when `zk-cuda-backend` or `tfhe-zk-pok` code changes. 

This is a [4-minute test](https://github.com/zama-ai/tfhe-rs/actions/runs/21834765900/job/63002908407?pr=3229). When the rust integration comes this workflow would also integrate a few tests in `tfhe::integer::gpu::zk` that could increase that time to around 10 minutes. 

Do you think this workflow should be different?

.github/workflows/gpu_zk_tests.yml
pdroalves: Totally. This workflow file is based on gpu fast tests. I will change to "ZK".

.github/workflows/gpu_zk_tests.yml
soonum: No, in this case we're good to go.

backends/zk-cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[build-dependencies]
+cmake.workspace = true
+pkg-config.workspace = true
+bindgen.workspace = true
+
+[dependencies]
+ark-ec.workspace = true
+ark-ff.workspace = true
+tfhe-cuda-backend = { path = "../tfhe-cuda-backend" }
```
IceTDrinker: needs the version of the current backend here otherwise we cannot publish the zk-cuda-backend crate and we will have to do it to be able to publish the ZK crate which will depend on it

backends/zk-cuda-backend/src/bindings.rs
```diff
@@ -0,0 +1,223 @@
...
+    ["Offset of field: BigInt::limb"][::std::mem::offset_of!(BigInt, limb) - 0usize];
+};
+pub type Scalar = BigInt;
+#[repr(C)]
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
+pub struct G1Point {
+    pub x: Fp,
+    pub y: Fp,
+    pub infinity: bool,
+}
```
IceTDrinker, line 51: Did you verify that the default value is ok for you ?

boolean will be false and the Fp will have default initialized u64s so 0 as well

same for other types

backends/zk-cuda-backend/src/types/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /// Create a new Fp from a BigInt-like structure (in normal form)
+    /// This is a convenience method for compatibility with tfhe-zk-pok's BigInt<7>
+    pub fn from_bigint<const N: usize>(bigint: &[u64; N]) -> Self {
+        const { assert!(N == 7, "from_bigint requires exactly FP_LIMBS (7) limbs") };
+        let mut limbs = [0u64; 7];
+        limbs.copy_from_slice(bigint.as_slice());
+        Self { limb: limbs }
+    }
```
IceTDrinker: if you force N == 7 then you can have

bigint: &[u64; 7] directly and drop the N from the function definition

backends/zk-cuda-backend/src/types/scalar.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn new(limbs: [u64; 5]) -> Self {
+        Self {
+            inner: ScalarFFI { limb: limbs },
+        }
+    }
+
+    /// Create a new Scalar from a BigInt-like structure
+    /// This is a convenience method for compatibility with tfhe-zk-pok's BigInt<5>
+    pub fn from_bigint<const N: usize>(bigint: &[u64; N]) -> Self {
+        const { assert!(N == 5, "from_bigint requires exactly ZP_LIMBS (5) limbs") };
```
IceTDrinker: if you force N == 5 then you can do:

bigint: &[u64; 5]

and get rid of N in the definition of the function

backends/zk-cuda-backend/src/types/scalar.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Get the limbs of the scalar
+    #[inline]
+    pub fn limbs(&self) -> [u64; 5] {
+        self.inner.limb
+    }
+
+    /// Get the inner FFI type (for internal use)
+    #[inline]
+    #[allow(dead_code)]
```
IceTDrinker: allow dead code necessary ?

backends/zk-cuda-backend/src/types/scalar.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub(crate) fn inner(&self) -> &ScalarFFI {
+        &self.inner
+    }
+
+    /// Get a mutable reference to the inner FFI type (for internal use)
+    #[inline]
+    #[allow(dead_code)]
```
IceTDrinker, line 72: allow dead code necessary ?

backends/zk-cuda-backend/src/types/scalar.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        Self::new(result)
+    }
+}
+
+impl fmt::Debug for Scalar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Scalar({:?})", self.limbs())
+    }
+}
```
IceTDrinker: derive Debug directly ?

backends/zk-cuda-backend/Cargo.toml
pdroalves: Should I just hardcode "0.13.0"?

IceTDrinker: yes the way it works is you say which version you need (the one from the repo) and where to find it with a path so that it can be found during development

backends/zk-cuda-backend/src/bindings.rs
pdroalves: Yes for all types but G1Point/G2Point. These are affine and we should be using infinity `true`. I'm removing the default from them. The other types in binding.rs are fine.

backends/zk-cuda-backend/src/types/g1.rs
```diff
@@ -30,8 +22,7 @@ impl G1Affine {
 
     /// Create the point at infinity
     pub fn infinity() -> Self {
-        let mut point = zero_g1_point();
-        point.infinity = true;
+        let mut point = G1Point::default();
```
IceTDrinker, line 25: question on readability : would it not be better to have G1Point::infinity() instead ? it would convey the intent better

also maybe it's a bad habit of mine but I tend to not use Default much because it's never clear what that means (at least to me)

so in short : I would remove Default implementations on any type you don't use the default for, and have speciliazed constructors (like infinity here) for the constructors you need or even have an associated constant

example : https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2d1e868286ee266742302edbf8562d36

if this is making this PR drag on too much can be done in a subsequent one

I did not realize what you were using the default for initially, otherwise I would have tried to recommend the constants right away

IceTDrinker (commented):
the management of FP and scalar size with consts is a nice improvement, adding safety comments is a good practice in rust

Only thing once the default thing changed was questioning whether you really needed default in the first place and would maybe benefit from having consts defined for the types otherwise you are good to squash I believe

backends/zk-cuda-backend/src/types/g1.rs
pdroalves: Yes, I really like that idea. It would make it clear to any reader that we are initializing the point as the point at infinity.

I don’t think we are initializing points like this outside of the few doctests we have in the Rust bindings, but we should have an API that is easy to read and has transparent behavior.

pdroalves: Done. The commit is again on top to facilitate review. We are not using that concept anywhere, so there are not many places to change.

I think I will keep the Default derive for the cases when we may want to define them and initialize later. We don't have (or want) to define them as infinity in this case. Now we also have `infinity()` for the cases when we really want them as infinity.

I also added some doctests to make sure we are not breaking this in the future.

IceTDrinker (commented):
Good to squash and I'll approve the resulting PR, thanks !

PR #3235, opened 2026-01-30
doc: add svg tables to zero-knowledge benchmarks

soonum:
This removes the embedded GSheet.
These SVGs display more operations and inputs that reflect real-use cases.
Throughput is available only for server-side computation as it's meaningless to perform multiple proof in parallele on the client-side.


IceTDrinker (approved):
🔥 let's backport this beauty!

PR #3243, opened 2026-02-02
chore(gpu): reduce hl test time in CI

agnesLeroy:

### PR content/description

I disabled uniformity test for OPRF from HL API tests for L40 targets as it was too long, rewrote a bit the test to have  better perf for it as well.

### Check-list:

### DIFF
--- a/tfhe/src/high_level_api/integers/oprf.rs
+++ b/tfhe/src/high_level_api/integers/oprf.rs
@@ -563,7 +563,6 @@ fn mod_pow_2(exponent: u64, modulus: u64) -> u64 {
 
 #[cfg(test)]
 mod test {
-
     use super::*;
     use crate::integer::server_key::radix_parallel::tests_unsigned::test_oprf::{
         oprf_density_function, p_value_upper_bound_oprf_almost_uniformity_from_values,
@@ -571,13 +570,18 @@ mod test {
     };
     use crate::prelude::FheDecrypt;
     use crate::shortint::oprf::test::test_uniformity;
-    #[cfg(feature = "gpu")]
-    use crate::shortint::parameters::PARAM_GPU_MULTI_BIT_GROUP_4_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
     use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS32_PBS_TUNIFORM_2M128;
-    use crate::{generate_keys, set_server_key, ClientKey, ConfigBuilder, FheUint8, Seed};
+    use crate::{generate_keys, set_server_key, ConfigBuilder, FheUint8, Seed};
     use num_bigint::BigUint;
     use rand::{thread_rng, Rng};
     use rayon::iter::{IntoParallelIterator, ParallelIterator};
+
+    const BASE_SAMPLE_COUNT: usize = 10_000;
+    const P_VALUE_LIMIT: f64 = 0.001;
+    // [0.7, 0.1] for `max_distance` chosen to have `num_input_random_bits` be [2, 4]
+    // for any of the listed `excluded_upper_bound`
+    const PRF_TEST_CASES: [(u64, f64, [u64; 4]); 2] =
+        [(2, 0.7, [3, 5, 6, 7]), (4, 0.1, [3, 5, 6, 7])];
 
     // Helper: The "Oracle" implementation using BigInt
     // This is slow but mathematically guaranteed to be correct.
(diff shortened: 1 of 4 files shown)


tfhe/src/high_level_api/integers/oprf.rs
```diff
@@ -786,6 +805,62 @@ mod test {
...
+        //                    let img = FheUint8::generate_oblivious_pseudo_random_custom_range(
+        //                        Seed(rng.gen::<u128>()),
+        //                        &range,
+        //                        Some(max_distance),
+        //                    );
+        //                    img.decrypt(cks)
+        //                })
+        //                .collect::<Vec<u64>>()
+        //        })
+        //        .collect();
```
IceTDrinker: comment to remove ?

tfhe/src/high_level_api/integers/oprf.rs
```diff
@@ -773,6 +781,17 @@ mod test {
...
+        #[cfg(feature = "gpu")]
+        let compressed_sks = CompressedServerKey::new(cks);
+        #[cfg(feature = "gpu")]
+        let sks_vec = (0..num_gpus)
+            .map(|i| compressed_sks.decompress_to_specific_gpu(GpuIndex::new(i as u32)))
+            .collect::<Vec<_>>();
+
+        #[cfg(not(feature = "gpu"))]
         let real_values: Vec<u64> = (0..sample_count)
             .into_par_iter()
```
IceTDrinker: given this could we maybe have two different tests to have the logic be easier to follow for each and you can disable the one you don't want on GPU ?

agnesLeroy: Indeed :+1:

tfhe/src/high_level_api/integers/oprf.rs
```diff
@@ -575,9 +577,17 @@ mod test {
...
+    use crate::{CompressedServerKey, GpuIndex};
     use num_bigint::BigUint;
     use rand::{thread_rng, Rng};
+    #[cfg(feature = "gpu")]
+    use rayon::iter::IndexedParallelIterator;
     use rayon::iter::{IntoParallelIterator, ParallelIterator};
+    #[cfg(feature = "gpu")]
+    use rayon::prelude::{IntoParallelRefIterator, ParallelSlice};
+    #[cfg(feature = "gpu")]
+    use rayon::ThreadPoolBuilder;
```
IceTDrinker: additional suggestion (sorry) you can import those directly in the gpu specific function to make import easier to update later on, same for the core gpu import

reading the file there is a gpu test module, maybe move evertyhing GPU related at the bottom in the gpu module ? and if need be make some of the functions required available to the GPU module ? if this proves too annoying the gpu test module can be moved inside this test module to have access to the required functions

that last suggestion is maybe not the best idea but I did not realize there was already a gpu test module

tfhe/src/high_level_api/integers/oprf.rs
IceTDrinker: feels like all this logic is shared between the two test functions, sorry again, but we could maybe have the FHE PRF data generation be backend specific but the verification code be shared ?

IceTDrinker (commented):
a few comments again since I realized there is already a dedicated gpu test module

tfhe/src/high_level_api/integers/oprf.rs
agnesLeroy: Done. I had to change quite a lot the file organization but I think it's better now. Let me know if you think this could be improved further.

tfhe/src/high_level_api/integers/oprf.rs
agnesLeroy: Done

tfhe/src/high_level_api/integers/oprf.rs
```diff
@@ -700,20 +703,30 @@ mod test {
...
+            (
+                EXPECTED_NUM_INPUT_RANDOM_BITS.0,
+                MAX_DISTANCE.0,
+                EXCLUDED_UPPER_BOUNDS,
+            ),
+            (
+                EXPECTED_NUM_INPUT_RANDOM_BITS.1,
+                MAX_DISTANCE.1,
+                EXCLUDED_UPPER_BOUNDS,
+            ),
```
IceTDrinker: the original array could have been made a constant ? also the comment above does not make much sense now

not strictly necessary to merge, but would be more tidy I think (since the test cases are shared between CPU and GPU)

so PRF_TEST_CASES: [(u64, f64, [u64; 4]); 2]: ...

and then use those in both tests

tfhe/src/high_level_api/integers/oprf.rs
```diff
@@ -833,48 +824,145 @@ mod test {
...
+                (
+                    EXPECTED_NUM_INPUT_RANDOM_BITS.0,
+                    MAX_DISTANCE.0,
+                    EXCLUDED_UPPER_BOUNDS,
+                ),
+                (
+                    EXPECTED_NUM_INPUT_RANDOM_BITS.1,
+                    MAX_DISTANCE.1,
+                    EXCLUDED_UPPER_BOUNDS,
+                ),
```
IceTDrinker: same remark as CPU could use the same test cases and comment could be removed and placed with the test cases (if the new type looks weird, then a light struct with named field can be ok like

PrfTestCase {expected_num_input_random_bits: u64, max_distance:f64, ...}

and instantiated in an array of length 2

[PrfTestCase ; 2]

PR #3244, opened 2026-02-02
fix(gpu): fix noise level in match value

agnesLeroy:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
@@ -7,7 +7,8 @@
 #include <functional>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -7,7 +7,8 @@
 #include <functional>
 #include <vector>
 
-const uint32_t MAX_STREAMS_FOR_VECTOR_FIND = 10;
+// If we use more than 5 streams the result is incorrect
+const uint32_t MAX_STREAMS_FOR_VECTOR_FIND = 5;
```
agnesLeroy, line 11: The problem is that if we set a number of streams to more than 5, since in the host function we accumulate host_addition over the number of streams we exceed max noise level = 5.

PR #3245, opened 2026-02-02
chore(gpu): Encapsulate LUT index generation

andrei-stoian-zama:
Encapsulate LUT index generation. Several cases are present:

- LUT indexes generated in int_radix_lut h_lut_indexes CPU array, then copied to GPU
- LUT indexes generated in a temporary CPU buffer then copied to GPU -> needs sync before free
- LUT indexes generated in a temporary CPU and a copied to a temporary GPU buffer -> need sync

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -23,7 +23,7 @@
 /// when all blocks should use the same LUT.
 constexpr std::nullptr_t LUT_0_FOR_ALL_BLOCKS = nullptr;
 
-/// Global function to generate LUT indexes, validate them, and copy to any GPU
+/// Generate LUT indexes with a generator, validate them, and copy to any GPU
 /// buffer.
 ///
 /// @tparam Torus          Integer type for indexes
@@ -34,14 +34,13 @@ constexpr std::nullptr_t LUT_0_FOR_ALL_BLOCKS = nullptr;
 /// @param d_lut_indexes        Destination GPU buffer for indexes
 /// @param num_indexes          Number of indexes to generate
 /// @param num_luts             Number of LUTs (for validation)
-/// @param h_buffer             CPU buffer to use for staging (caller must sync
-/// before freeing)
-/// @param allocate_gpu_memory  Whether to track GPU memory allocation
+/// @param h_buffer             CPU buffer to use for staging. The caller that
+/// created this buffer must sync before freeing)
 template <typename Torus, typename IndexGenerator>
 void generate_lut_indexes(CudaStreams streams, IndexGenerator generator,
                           Torus *d_lut_indexes, uint32_t num_indexes,
                           uint32_t num_luts, Torus *h_buffer,
-                          bool allocate_gpu_memory) {
+                          bool gpu_memory_allocated) {
   GPU_ASSERT(h_buffer != nullptr, "h_buffer must be provided");
 
   // Initialize with sentinel value to detect uninitialized entries
@@ -65,7 +64,7 @@ void generate_lut_indexes(CudaStreams streams, IndexGenerator generator,
 
   cuda_memcpy_with_size_tracking_async_to_gpu(
       d_lut_indexes, h_buffer, num_indexes * sizeof(Torus), streams.stream(0),
-      streams.gpu_index(0), allocate_gpu_memory);
+      streams.gpu_index(0), gpu_memory_allocated);
 }
 
 class NoiseLevel {
@@ -871,15 +870,23 @@ private:
 
   /// Sets all LUT indexes to a constant value on both CPU and GPU.
   /// Does not broadcast - caller must call broadcast_lut if needed.
-  void set_lut_indexes_to_constant(const CudaStreams &streams,
-                                   InputTorus value) {
-    for (uint32_t i = 0; i < num_blocks; i++) {
-      h_lut_indexes[i] = value;
-    }
-    if (gpu_memory_allocated)
+  void set_lut_indexes_to_constant_async(const CudaStreams &streams,
+                                         InputTorus value) {
+    GPU_ASSERT(
+        value < num_luts,
+        "Constant LUT index out of bounds: value = %llu >= num_luts (%u)",
+        (unsigned long long)value, num_luts);
+
+    if (gpu_memory_allocated) {
+      // Keep the CPU buffer equal to the GPU one
+      for (uint32_t i = 0; i < num_blocks; i++) {
+        h_lut_indexes[i] = value;
+      }
+
       cuda_set_value_async<InputTorus>(streams.stream(0), streams.gpu_index(0),
                                        get_lut_indexes(0, 0), value,
                                        num_blocks);
+    }
   }
 
 public:
@@ -894,10 +901,16 @@ public:
   set_lut_indexes_and_broadcast_from_gpu(CudaStreams streams,
                                          InputTorus const *d_src_lut_indexes,
                                          uint32_t num_indexes) {
-    cuda_memcpy_async_gpu_to_gpu(get_lut_indexes(0, 0), d_src_lut_indexes,
-                                 num_indexes * sizeof(InputTorus),
-                                 streams.stream(0), streams.gpu_index(0));
+    cuda_memcpy_with_size_tracking_async_gpu_to_gpu(
+        get_lut_indexes(0, 0), d_src_lut_indexes,
+        num_indexes * sizeof(InputTorus), streams.stream(0),
+        streams.gpu_index(0), gpu_memory_allocated);
     broadcast_lut(streams, false);
+
+    // In some cases, num_indexes can be less than the number of blocks
+    // that the LUT was created for but broadcast_lut sets
+    // last_broadcast_num_radix_blocks = num_blocks
+    last_broadcast_num_radix_blocks = num_indexes;
   }
 
   /// Sets all LUT indexes to a constant value and broadcasts to all active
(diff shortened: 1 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -163,15 +165,10 @@ __host__ void are_all_comparisons_block_true(
           streams, lwe_array_out, accumulator, bsks, ksks, lut, 1);
       // Reset max_value_lut_indexes before returning, otherwise if the lut is
       // reused the lut indexes will be wrong
-      memset(is_max_value_lut->h_lut_indexes, 0,
```
andrei-stoian-zama, line 166: Changed to use `set_lut_indexes_and_broadcast_constant` which sets both CPU and GPU index buffers to the given constant

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -782,6 +840,77 @@ struct int_radix_lut_custom_input_output {
...
+  ///
+  /// @param streams              CUDA streams for async operations
+  /// @param d_src_lut_indexes    Source GPU buffer containing precomputed
+  /// indexes
+  /// @param num_indexes          Number of indexes to copy
+  void
+  set_lut_indexes_and_broadcast_from_gpu(CudaStreams streams,
+                                         InputTorus const *d_src_lut_indexes,
+                                         uint32_t num_indexes) {
+    cuda_memcpy_async_gpu_to_gpu(get_lut_indexes(0, 0), d_src_lut_indexes,
```
agnesLeroy: This has to use `with_size_tracking` and only do the copy if gpu_memory_allocated=true

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -782,6 +840,77 @@ struct int_radix_lut_custom_input_output {
...
+  }
+
+  /// Sets all LUT indexes to a constant value on both CPU and GPU.
+  /// Does not broadcast - caller must call broadcast_lut if needed.
+  void set_lut_indexes_to_constant(const CudaStreams &streams,
+                                   InputTorus value) {
+    for (uint32_t i = 0; i < num_blocks; i++) {
+      h_lut_indexes[i] = value;
+    }
+    if (gpu_memory_allocated)
```
agnesLeroy: The if could be placed above, to avoid setting the h_lut_indexes if not necessary

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -879,41 +1011,59 @@ struct int_radix_lut_custom_input_output {
...
+        "");
+
+    // Generate LUT indexes based on index_generator
+    if constexpr (!std::is_same_v<IndexGenerator, std::nullptr_t>) {
+      // Custom index generator provided
+      InputTorus *index_buffer =
+          (h_index_buffer != nullptr) ? h_index_buffer : h_lut_indexes;
+      GPU_ASSERT(index_buffer != nullptr,
+                 "No buffer provided and h_lut_indexes is null");
+      generate_lut_indexes<InputTorus>(
```
agnesLeroy: Couldn't we have this function private in int_radix_lut_custom_input_output and remove gpu_memory_allocated?

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -782,6 +840,77 @@ struct int_radix_lut_custom_input_output {
...
+public:
+  /// Copies precomputed LUT indexes from a GPU buffer to this LUT's GPU index
+  /// buffer.
+  ///
+  /// @param streams              CUDA streams for async operations
+  /// @param d_src_lut_indexes    Source GPU buffer containing precomputed
+  /// indexes
+  /// @param num_indexes          Number of indexes to copy
+  void
+  set_lut_indexes_and_broadcast_from_gpu(CudaStreams streams,
```
andrei-stoian-zama, line 901: `num_indexes` here can be less than the `num_blocks` that the LUT was created on, or broadcast to.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -782,6 +840,77 @@ struct int_radix_lut_custom_input_output {
...
+                                         InputTorus const *d_src_lut_indexes,
+                                         uint32_t num_indexes) {
+    cuda_memcpy_async_gpu_to_gpu(get_lut_indexes(0, 0), d_src_lut_indexes,
+                                 num_indexes * sizeof(InputTorus),
+                                 streams.stream(0), streams.gpu_index(0));
+    broadcast_lut(streams, false);
```
andrei-stoian-zama, line 908: maybe we neeed `last_broadcast_num_blocks = num_indexes`

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -879,41 +1011,59 @@ struct int_radix_lut_custom_input_output {
...
-        "LUT Generation with pre-allocated CPU buffer only supports generation "
-        "with encoding and expects gpu_memory_allocated==True ");
+        h_lut_value_buffers.empty() || (use_encoding && gpu_memory_allocated),
+        "LUT Generation with pre-allocated CPU buffer only supports "
+        "generation with encoding and expects gpu_memory_allocated==True "
+        "");
+
+    // Generate LUT indexes based on index_generator
+    if constexpr (!std::is_same_v<IndexGenerator, std::nullptr_t>) {
+      // Custom index generator provided
```
andrei-stoian-zama: why not `set_lut_indexes`

andrei-stoian-zama: why not check against the alias `LUT_0_FOR_ALL_BLOCKS` instead of  `std::nullptr_t`

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -879,41 +1011,59 @@ struct int_radix_lut_custom_input_output {
...
-            streams.stream(0), streams.gpu_index(0), get_lut(0, lut_indexes[i]),
-            get_degree(lut_indexes[i]), params.message_modulus,
+            streams.stream(0), streams.gpu_index(0), get_lut(0, lut_ids[i]),
+            get_degree(lut_ids[i]), params.message_modulus,
             params.carry_modulus, params.glwe_dimension, params.polynomial_size,
-            f[i], gpu_memory_allocated);
+            lut_value_generator[i], gpu_memory_allocated);
       }
     }
     broadcast_lut(streams);
```
andrei-stoian-zama, line 1075: broadcast synchronizes stream 0

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: Yes indeed

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
agnesLeroy: No it doesn't, it relies on events

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: ok yes! it should add more protection ! 

currently it doesn't crash since `set_lut_indexes_and_broadcast_from_gpu` is called in operation compute code - thus when gpu memory was allocated, never in scratch where gpu memory might not be allocated to compute GPU memory requirements.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: no, because we sometimes use it to generate indexes to a temporary GPU index buffer that is swapped into the LUT during operation computation.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: sorry, you are right and thus I think some synchronize are missing

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: can't be done.. the compiler doesn't like it

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: no changes to synchronization needed as discussed. When a caller passes a CPU buffer to `generate_and_broadcast_lut` they are responsible for synch+free. `generate_and_broadcast_lut` only synchronizes for device accumulator creation.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
andrei-stoian-zama: I had forgotten to refactor the creation of pre-computed GPU index buffers. They are in div_rem, you can now check out the diff for that file.

e.g.

```
      generate_lut_indexes<Torus>(streams, index_generator,
                                  first_indexes_for_overflow_sub_gpu_0[nb - 1],
                                  nb, 2 * group_size + 1, h_lut_indexes,
```

agnesLeroy: Aaah indeed I had forgotten that we do this in the division. Ok so let's keep things as they are :+1:

PR #3247, opened 2026-02-03
feat(gpu): quick wins in pbs128 flavors

guillermo-oyarzun:

### PR content/description
This PR improves significantly performance of pbs128 classical and multi-bit flavors.
The changes are basically two:
- fft128 change the ternary operators so compiler doesn't spill registers
- use Degree instead of amortizedDegree (512 threads instead of 256)

Additionally i had to add the logic of verify cg support for multi-bit pbs128 (With the old logic we were not actually reading the occupancy of pbs128 but from pbs64).

My measurements show around 2x improvements
latency:
classical: 122ms -> 60 ms (next PR will return to old behavior 80ms->40ms)
multi-bit: 40ms -> 22.9ms

throughput:
classical: 55ops/s -> 130ops/s
multi-bit: 155ops/s -> 330 ops/s

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
@@ -76,7 +76,11 @@ __device__ void negacyclic_forward_fft_f128(double *dt_re_hi, double *dt_re_lo,
     for (Index i = 0; i < BUTTERFLY_DEPTH; i++) {
       Index rank = tid & thread_mask;
       bool u_stays_in_register = rank < lane_mask;
-      F128x2_TO_F64x4(((u_stays_in_register) ? v[i] : u[i]), tid);
+      if (u_stays_in_register) {
+        F128x2_TO_F64x4(v[i], tid);
+      } else {
+        F128x2_TO_F64x4(u[i], tid);
+      }
       tid = tid + STRIDE;
     }
     __syncthreads();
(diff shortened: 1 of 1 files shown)


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit_128.cuh
```diff
@@ -1093,4 +1093,115 @@ __host__ uint64_t scratch_cg_multi_bit_programmable_bootstrap_128(
...
+// group constraints
+template <typename Torus>
+__host__ bool
+supports_cooperative_groups_on_multibit_programmable_bootstrap_128(
+    int glwe_dimension, int polynomial_size, int level_count, int num_samples,
+    uint32_t max_shared_memory) {
+  switch (polynomial_size) {
+  case 256:
+    return verify_cuda_programmable_bootstrap_cg_multi_bit_grid_size_128<
+        Torus, AmortizedDegree<256>>(glwe_dimension, level_count, num_samples,
```
agnesLeroy: I believe it should be Degree everywhere instead of AmortizedDegree to be consistent with the way we call the function?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit_128.cuh
```diff
@@ -1093,4 +1093,115 @@ __host__ uint64_t scratch_cg_multi_bit_programmable_bootstrap_128(
...
+    return verify_cuda_programmable_bootstrap_cg_multi_bit_grid_size_128<
+        Torus, AmortizedDegree<4096>>(glwe_dimension, level_count, num_samples,
+                                      max_shared_memory);
+  case 8192:
+    return verify_cuda_programmable_bootstrap_cg_multi_bit_grid_size_128<
+        Torus, AmortizedDegree<8192>>(glwe_dimension, level_count, num_samples,
+                                      max_shared_memory);
+  case 16384:
+    return verify_cuda_programmable_bootstrap_cg_multi_bit_grid_size_128<
+        Torus, AmortizedDegree<16384>>(glwe_dimension, level_count, num_samples,
```
agnesLeroy: We only support up to N=4096 for pbs128, the higher polynomial sizes should be removed from here and the panic message updated

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_multibit_128.cuh
agnesLeroy: Ah ok I see you only changed to Degree for N=2048. Maybe it's better to change for all values of supported N to be less confusing?

guillermo-oyarzun: yup i can do that

agnesLeroy:
Thank you very much!

PR #3249, opened 2026-02-04
Am/feat/improve compact public key perf

IceTDrinker:
should be good to review, normally it's well organized in several commits

- first commit changes the primitive to expand to use cache friendly monomial muls
- second commits makes it easy to check if some NTT plans are avaible wrt the required AVX512 intrinsics, inline always required for good perf
- harden polymul tests in NTT by having more loops
- the meat of the PR : use the native u64 x binary NTT for a new primitive optimized for the case we encounter when encrypting things with a CompactPublicKey + tests of the primitive agains the unmodified previous code, added a dedicated primitive for rerand encryptions (since there are some questions around the NTT use for all CPK encryptions), the rerand primitive is tested against the pre-existing implem
- to avoid code duplication the impl for the CPK encryption is genericized with a const generic allowing to easily choose the optimized primitive or not, all existing encryption primitives conserve the same code path, only the rerand primitive makes use of the optimized reverse negacyclic convolution algorithm

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249)

### DIFF
--- a/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
+++ b/tfhe/src/core_crypto/algorithms/lwe_encryption.rs
@@ -2291,6 +2291,13 @@ fn encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl<
     mask_noise_distribution: MaskDistribution,
     body_noise_distribution: NoiseDistribution,
     noise_generator: &mut NoiseRandomGenerator<EncryptionGen>,
+    // For calls made to slice_semi_reverse_negacyclic_convolution_impl in this function, rhs is
+    // guaranteed to be binary
+    slice_semi_reverse_negacyclic_convolution_impl: fn(
+        output: &mut [Scalar],
+        lhs: &[Scalar],
+        rhs: &[Scalar],
+    ),
 ) -> CompactPublicKeyRandomVectors<Scalar>
 where
     Scalar: Encryptable<MaskDistribution, NoiseDistribution> + RandomGenerable<UniformBinary>,
@@ -2370,14 +2377,14 @@ where
                 // create a temp buffer to compute the full convolution
                 let mut pk_body_convolved = vec![Scalar::ZERO; max_ciphertext_per_bin];
 
-                slice_semi_reverse_negacyclic_convolution(
+                slice_semi_reverse_negacyclic_convolution_impl(
                     output_mask.as_mut(),
                     pk_mask.as_ref(),
                     binary_random_slice,
                 );
 
                 // Fill the temp buffer with b convolved with r
-                slice_semi_reverse_negacyclic_convolution(
+                slice_semi_reverse_negacyclic_convolution_impl(
                     pk_body_convolved.as_mut_slice(),
                     pk_body.as_ref(),
                     binary_random_slice,
@@ -2517,6 +2524,8 @@ pub fn encrypt_lwe_compact_ciphertext_list_with_compact_public_key<
     OutputCont: ContainerMut<Element = Scalar>,
     EncryptionGen: ByteRandomGenerator,
 {
+    // Conservative for now this implementation does not use the optimized implementation of the
+    // negacyclic convolution
     let _ = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl(
         lwe_compact_public_key,
         output,
@@ -2524,6 +2533,7 @@ pub fn encrypt_lwe_compact_ciphertext_list_with_compact_public_key<
         mask_noise_distribution,
         body_noise_distribution,
         noise_generator,
+        slice_semi_reverse_negacyclic_convolution,
     );
 }
 
@@ -2712,6 +2722,8 @@ where
             .collect::<Vec<_>>(),
     );
 
+    // Conservative for now this implementation does not use the optimized implementation of the
+    // negacyclic convolution
     let CompactPublicKeyRandomVectors {
         binary_random_vector,
         mask_noise,
@@ -2723,6 +2735,7 @@ where
         mask_noise_distribution,
         body_noise_distribution,
         noise_generator,
+        slice_semi_reverse_negacyclic_convolution,
     );
 
     Ok(crs.prove(
@@ -3209,6 +3222,45 @@ where
     ))
 }
 
+pub mod re_randomization {
+    use super::*;
+    pub fn rerand_encrypt_lwe_compact_ciphertext_list_with_compact_public_key<
+        Scalar,
+        MaskDistribution,
+        NoiseDistribution,
+        KeyCont,
+        InputCont,
+        OutputCont,
+        EncryptionGen,
+    >(
+        lwe_compact_public_key: &LweCompactPublicKey<KeyCont>,
+        output: &mut LweCompactCiphertextList<OutputCont>,
+        encoded: &PlaintextList<InputCont>,
+        mask_noise_distribution: MaskDistribution,
+        body_noise_distribution: NoiseDistribution,
+        noise_generator: &mut NoiseRandomGenerator<EncryptionGen>,
+    ) where
+        Scalar: Encryptable<MaskDistribution, NoiseDistribution> + RandomGenerable<UniformBinary>,
+        MaskDistribution: Distribution,
+        NoiseDistribution: Distribution,
+        KeyCont: Container<Element = Scalar>,
+        InputCont: Container<Element = Scalar>,
+        OutputCont: ContainerMut<Element = Scalar>,
+        EncryptionGen: ByteRandomGenerator,
+    {
+        // Rerand uses the optimized implementation of the negacyclic convolution
+        let _ = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl(
+            lwe_compact_public_key,
+            output,
+            encoded,
+            mask_noise_distribution,
+            body_noise_distribution,
+            noise_generator,
+            slice_binary_semi_reverse_negacyclic_convolution,
+        );
+    }
+}
+
 #[cfg(test)]
 mod test {
     use crate::core_crypto::commons::generators::DeterministicSeeder;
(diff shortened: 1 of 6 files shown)


tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -2517,7 +2533,17 @@ pub fn encrypt_lwe_compact_ciphertext_list_with_compact_public_key<
     OutputCont: ContainerMut<Element = Scalar>,
     EncryptionGen: ByteRandomGenerator,
 {
-    let _ = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl(
+    let _ = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl::<
+        false, /* We don't use the optimized variant for encryption by default because it can
```
IceTDrinker: comment is wrong and from another iteration of the PR

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -2716,7 +2742,17 @@ where
         binary_random_vector,
         mask_noise,
         body_noise,
-    } = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl(
+    } = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl::<
+        false, /* We don't use the optimized variant for encryption by default because it can
+                * lead to timing attacks */
```
IceTDrinker: same here comment is wrong, no timing attack on NTT normally

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
```diff
@@ -3209,6 +3245,53 @@ where
...
+        MaskDistribution: Distribution,
+        NoiseDistribution: Distribution,
+        KeyCont: Container<Element = Scalar>,
+        InputCont: Container<Element = Scalar>,
+        OutputCont: ContainerMut<Element = Scalar>,
+        EncryptionGen: ByteRandomGenerator,
+    {
+        let _ = encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl::<
+            true, /* We use the optimized variant for rerand, timing attacks are not relevant in
+                   * this case */
```
IceTDrinker: comment is wrong no timing attack

tfhe/src/core_crypto/algorithms/lwe_encryption.rs
nsarlin-zama: should fix? If the comment is wrong, does this mean that we can use the optimized version everywhere?

IceTDrinker: we preferred to limit the usage of the NTT primitive for now

IceTDrinker (commented):
> @IceTDrinker made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249)* status: 0 of 10 files reviewed, 7 unresolved discussions (waiting on @agnesLeroy, @mayeul-zama, @nsarlin-zama, and @tmontaigu).

___
*[`tfhe/src/core_crypto/algorithms/lwe_encryption.rs` line 2280 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlC3Mqa41tw7mei5fxi:-OlC9Mf02HZChyOYjFI2:b-8qn98w) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/lwe_encryption.rs#L2280)):*

Instead of a boolean const generic, which is a bit hard to read at call site, what you can do here is to have a `ConvFn: Fn(&mut [Scalar], &[Scalar], &[Scalar])` generic parameter, and give the convolution function as parameter:

```rust
encrypt_lwe_compact_ciphertext_list_with_compact_public_key_impl(
   ..., 
   slice_binary_semi_reverse_negacyclic_convolution
);
```
Should give the same monomorphisation but a bit more readable at call site
</blockquote></details>

fair, I was having doubts, I'll have to rebench though to make sure
___
*[`tfhe/src/core_crypto/algorithms/lwe_encryption.rs` line 3624 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlByntl8mUwpkGLDC6m:-OlC9XQrBYhTaQ_J3xDs:b-1emiox) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/lwe_encryption.rs#L3624)):*

In this test I guess we assume that `encrypt_lwe_compact_ciphertext_list_with_compact_public_key` is correct, so is it really useful to do this second block of encrypt then decrypt? 
Unless we compare the ciphertexts, we will just get the already known inputs as a result.

IMO the test could only do `rerand_encrypt_lwe_compact_ciphertext_list_with_compact_public_key`, decrypt and compare the results with the generated inputs
</blockquote></details>

I would think so, you could have non bit exact results but still decrypt correctly, arguably difficult to find such an example but this way we know the bits are the same
___
*[`tfhe/src/core_crypto/algorithms/slice_algorithms.rs` line 678 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlC6z2lBHlNbXLEAjAb:-OlC9kec5ac9jqeC0z62:b-gywboh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/slice_algorithms.rs#L678)):*

copy paste error, should be slice_binary_semi_reverse_negacyclic_convolution
</blockquote></details>

indeed thanks
___
*[`tfhe/src/core_crypto/algorithms/slice_algorithms.rs` line 688 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlC7AJb7saoPskeqpa6:-OlC9lsj8YgyCJksDBDr:bjmulmt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/slice_algorithms.rs#L688)):*

In tfhe-ntt::native_binary64::Plan32::negacyclic_polymul, it looks like only rhs is assumed to be binary, so we should assert this I guess?
</blockquote></details>

might be a mix-up in the way I wrote the code (since I had another iteration of this code at some point), I'll recheck, where do you see the assumption about rhs for the NTT ?

tmontaigu:
*[`tfhe/src/core_crypto/algorithms/slice_algorithms.rs` line 709 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlBjZ_JDn7AamBbjw1G:-OlBjZ_JDn7AamBbjw1H:b-3o84iw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/slice_algorithms.rs#L709)):*
> ```XML
>         // Runtime check
>         // 32 and 32768 are limit sizes valid for the native64 NTT from tfhe_ntt
>         if poly_size.0 >= 32 && poly_size.0 <= 32768 {
> ```

Are these constants avaiblable in tfhe_ntt ?

mayeul-zama (commented):
> @mayeul-zama reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249)* status: all files reviewed, 9 unresolved discussions (waiting on @agnesLeroy, @IceTDrinker, @nsarlin-zama, and @tmontaigu).

___
*[`tfhe/src/core_crypto/algorithms/slice_algorithms.rs` line 757 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-OlGTePw7au2iiPIoO65:-OlLJ_nc6Cf_5yFiseSU:b-elbp93) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/73b68538afaf804a17d279c7c0acbc29b7237a48/tfhe/src/core_crypto/algorithms/slice_algorithms.rs#L757)):*

no the Makefile tests both


```
test_core_crypto:
	RUSTFLAGS="$(RUSTFLAGS)" cargo test --profile $(CARGO_PROFILE) \
		--no-default-features --features=experimental,zk-pok -p tfhe -- core_crypto::
	@if [[ "$(AVX512_SUPPORT)" == "ON" ]]; then \
		RUSTFLAGS="$(RUSTFLAGS)" cargo test --profile $(CARGO_PROFILE) \
			--features=experimental,zk-pok -p tfhe -- core_crypto::; \
	fi
```

and we call it with 

```
      - name: Run core tests
        if: needs.should-run.outputs.core_crypto_test == 'true'
        run: |
          AVX512_SUPPORT=ON make test_core_crypto
```

as avx512 is a default feature we should have both paths tested
</blockquote></details>

Great, thanks for the explanation!

nsarlin-zama (approved):
:lgtm:, thanks!

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-:-OlafrEf-aS0hv3KqzyO:bpqxd5f)* status: 9 of 10 files reviewed, 4 unresolved discussions (waiting on @agnesLeroy, @IceTDrinker, @mayeul-zama, and @tmontaigu).

___
*[`tfhe/src/core_crypto/algorithms/lwe_encryption.rs` line 3532 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3249#-Olaf6Cm3mrpHtI2I1bQ:-OlafiYADPNd8XQiwHYN:b-2vy1gi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/fb1bced63b33fcf07a3a8a077d51d0e6d6013e04/tfhe/src/core_crypto/algorithms/lwe_encryption.rs#L3532)):*

@nsarlin-zama actually I can't clone the EncryptionRandomGenerator, and given it needs a seeder (on purpose) it's hard to replicate the same RNG state, so I would keep it as is
</blockquote></details>

erf too bad, so I'm ok with keeping like that!

PR #3275, opened 2026-02-06
feat(gpu): integrate zk-cuda-backend with tfhe-zk-pok

pdroalves:
Wire the zk-cuda-backend into tfhe-zk-pok and tfhe.

### GPU acceleration module (`tfhe-zk-pok/src/gpu/`)

New `gpu` module gated behind the `gpu-experimental` feature flag, containing:

- **Type conversions** between `tfhe-zk-pok` (arkworks) and `zk-cuda-backend` types: `G1Affine`, `G2Affine`, `Zp`/`Scalar`, `Fq`/`Fp`, `Fq2`/`Fp2`. Includes compile-time `size_of`/`align_of` assertions to guarantee transmute safety between wrapper types and inner arkworks types.
- **GPU MSM helpers** (`g1_msm_gpu`, `g2_msm_gpu`): create a CUDA stream, dispatch MSM to the backend, convert results from Montgomery form back to arkworks projective coordinates. Multi-GPU support via `select_gpu_for_msm()` which distributes work across GPUs by rayon thread index.

### GPU prove/verify for PKE v1 and v2

GPU variants of prove and verify are structured as submodules of each proof module, mirroring the CPU API:

| CPU | GPU |
|-----|-----|
| `pke::prove(...)` | `pke::gpu::prove(...)` |
| `pke::verify(...)` | `pke::gpu::verify(...)` |
| `pke_v2::prove(...)` | `pke_v2::gpu::prove(...)` |
| `pke_v2::verify(...)` | `pke_v2::gpu::verify(...)` |

- `proofs/pke/gpu.rs`: Duplicates v1 `prove_impl` logic, replacing every `multi_mul_scalar` with `g1_msm_gpu`/`g2_msm_gpu`. Verification delegates to the CPU verifier since v1 verify has no MSM calls.
- `proofs/pke_v2/gpu.rs`: Duplicates v2 `prove_impl` and `verify_impl`, replacing MSM sites. Also provides GPU variants of `pairing_check_two_steps` and `pairing_check_batched` for verification.

### Visibility changes in `proofs/`

Several items that were previously `pub(super)` or private are now `pub(crate)` so the GPU submodules can access them:
- `proofs/mod.rs`: `OneBased` inner field, `assert_pke_proof_preconditions`, `decode_q`, `compute_r1`, `compute_r2`, `Sid::to_le_bytes`, `SidBytes`, `run_in_pool`, test utilities (`PkeTestParameters`, `PkeTestcase`, etc.)
- `proofs/pke/mod.rs`: `bit_iter`, `compute_a_theta`, `PublicCommit`/`PrivateCommit` fields, test constants
- `proofs/pke_v2/mod.rs`: `bit_iter`, `compute_a_theta`, `compute_crs_params`, `inf_norm_bound_to_euclidean_squared`, `ComputeLoadProofFields::to_le_bytes`, `GeneratedScalars`, `EvaluationPoints`, `PublicCommit`/`PrivateCommit` fields, test constants
- `proofs/pke_v2/hashes.rs`: `RHash`, `PhiHash`, `XiHash`, `YHash`, `THash`, `ThetaHash`, `OmegaHash`, `DeltaHash`, `ZHash` and their chained `gen_*` methods

### Tests

- `gpu/tests/zk_cuda_backend.rs`: Low-level integration tests for the GPU backend — MSM correctness for G1/G2, type round-trip conversions, multi-GPU dispatch.
- `gpu/tests/prove_verify_stress.rs`: Exhaustive GPU-vs-CPU equivalence tests for both v1 and v2. Each test iterates over 3 CRS variants x 32 invalid-witness combinations x 2 compute loads (v2 also sweeps both pairing modes), asserting byte-identical serialized proofs and matching accept/reject decisions.

### Benchmarks

- `tfhe-zk-pok/benches/pke_v1.rs` and `pke_v2.rs`: GPU benchmark groups added behind `gpu-experimental`, covering prove and verify for both protocol versions.
- `tfhe-benchmark/benches/zk/msm.rs`: Standalone MSM benchmarks (CPU and GPU) for G1 and G2 at various sizes, useful for profiling the MSM kernel in isolation.
- `tfhe-benchmark/benches/integer/zk_pke.rs`: GPU throughput benchmark tuning — replaces hardcoded element counts with a `gpu_zk_throughput_elements` function sized per CRS/bit configuration to avoid OOM. Creates per-GPU server keys and streams for proper multi-GPU scaling.

### Build system and CI

- `tfhe-zk-pok/Cargo.toml`: New `gpu-experimental` feature depending on `zk-cuda-backend` and `tfhe-cuda-backend`. `ark-ec`/`ark-ff` switched to workspace dependencies. Added `itertools` dependency.
- `tfhe/Cargo.toml`: New `gpu-experimental-zk` feature forwarding to `tfhe-zk-pok/gpu-experimental`.
- `tfhe-benchmark/Cargo.toml`: New `gpu-experimental-zk` feature; `zk-pok` now brings in `dep:tfhe-zk-pok` directly. New `zk-msm` bench target.
- `Makefile`: New targets `test_zk_pok_gpu`, `test_integer_zk_gpu`, `test_zk_cuda`, `bench_msm_zk`, `bench_msm_zk_gpu`, `bench_tfhe_zk_pok_gpu`. Existing GPU clippy/check/build targets updated to include `gpu-experimental-zk`. `bench_integer_zk_gpu` switched from `release_lto_off` to `release` profile. Minor trailing-comma fixes in `bench_integer_aes*_gpu`.
- `.github/workflows/`: `benchmark_gpu.yml` adds `tfhe_zk_pok` and `msm_zk` commands. `benchmark_cpu.yml` adds `msm_zk`. `gpu_zk_tests.yml` broadens file change triggers and runs `test_zk_pok_gpu` + `test_integer_zk_gpu`.
- `.gitignore`: Ignores `backends/tfhe-cuda-backend/cuda/build/`.

### New Makefile targets

- `test_zk_cuda_backend`: Builds the zk-cuda-backend C++ tests via CMake and runs both the C++ and Rust test suites.
- `test_zk_pok_gpu`: Runs `tfhe-zk-pok` tests with the `gpu-experimental` feature, filtering to GPU tests.
- `test_integer_zk_gpu`: Runs tfhe integer ZK tests with expand accelerated on the GPU.
- `test_integer_zk_experimental_gpu`: Same as above but accelerating the entire verify/proof circuit using `zk-cuda-backend`.
- `bench_msm_zk`: Runs CPU MSM benchmarks for G1 and G2 at various sizes.
- `bench_msm_zk_gpu`: Runs GPU MSM benchmarks for G1 and G2 at various sizes.
- `bench_tfhe_zk_pok_gpu`: Runs `tfhe-zk-pok` prove/verify benchmarks with GPU acceleration.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1336

### Check-list:

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275)

### DIFF
--- a/backends/zk-cuda-backend/cuda/tests_and_benchmarks/benchmarks/CMakeLists.txt
+++ b/backends/zk-cuda-backend/cuda/tests_and_benchmarks/benchmarks/CMakeLists.txt
@@ -23,7 +23,8 @@ set(ZK_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../src)
 set(ZK_PRIMITIVES_DIR ${ZK_SRC_DIR}/primitives)
 
 # Build device library from tfhe-cuda-backend
-add_library(tfhe_device_bench STATIC ${TFHE_CUDA_BACKEND_DIR}/src/device.cu)
+add_library(tfhe_device_bench STATIC ${TFHE_CUDA_BACKEND_DIR}/src/device.cu
+                                     ${TFHE_CUDA_BACKEND_DIR}/src/utils/helper_profile.cu)
 set_target_properties(
   tfhe_device_bench
   PROPERTIES CUDA_SEPARABLE_COMPILATION ON
--- a/backends/zk-cuda-backend/cuda/tests_and_benchmarks/tests/CMakeLists.txt
+++ b/backends/zk-cuda-backend/cuda/tests_and_benchmarks/tests/CMakeLists.txt
@@ -20,7 +20,8 @@ set(ZK_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../src)
 set(ZK_PRIMITIVES_DIR ${ZK_SRC_DIR}/primitives)
 
 # Build device library from tfhe-cuda-backend
-add_library(tfhe_device STATIC ${TFHE_CUDA_BACKEND_DIR}/src/device.cu)
+add_library(tfhe_device STATIC ${TFHE_CUDA_BACKEND_DIR}/src/device.cu
+                               ${TFHE_CUDA_BACKEND_DIR}/src/utils/helper_profile.cu)
 set_target_properties(
   tfhe_device
   PROPERTIES CUDA_SEPARABLE_COMPILATION ON
--- a/tfhe-benchmark/benches/integer/zk_pke.rs
+++ b/tfhe-benchmark/benches/integer/zk_pke.rs
@@ -981,7 +981,7 @@ mod cuda {
 
                         let shortint_params: PBSParameters = *param_fhe;
 
-                        write_to_json::<u64, _>(
+                        write_to_json_unchecked::<u64, _>(
                             &bench_id,
                             shortint_params,
                             param_name,
--- a/tfhe-benchmark/benches/zk/msm.rs
+++ b/tfhe-benchmark/benches/zk/msm.rs
@@ -19,9 +19,8 @@
 //! cargo bench --package tfhe-benchmark --bench zk-msm --features gpu-experimental-zk
 //! ```
 
-use benchmark::utilities::{
-    get_bench_type, write_to_json, BenchmarkType, CryptoParametersRecord, OperatorType,
-};
+use benchmark::utilities::{write_to_json_unchecked, CryptoParametersRecord, OperatorType};
+use benchmark_spec::{get_bench_type, BenchmarkType};
 use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion, Throughput};
 use rand::rngs::StdRng;
 use rand::SeedableRng;
@@ -195,7 +194,7 @@ fn bench_cpu_msm<T: MsmBenchGroup>(c: &mut Criterion) {
 
         // MSM benchmarks are curve operations, use minimal parameters
         let params: CryptoParametersRecord<u64> = CryptoParametersRecord::default();
-        write_to_json(
+        write_to_json_unchecked(
             &bench_id,
             params,
             T::DISPLAY_NAME,
@@ -270,7 +269,7 @@ fn bench_gpu_msm<T: MsmBenchGroup>(c: &mut Criterion) {
         }
 
         let params: CryptoParametersRecord<u64> = CryptoParametersRecord::default();
-        write_to_json(
+        write_to_json_unchecked(
             &bench_id,
             params,
             format!("{}_CUDA", T::DISPLAY_NAME),


Makefile
```diff
@@ -1210,12 +1210,31 @@ test_tfhe_csprng_big_endian: install_cargo_cross
...
 	RUSTFLAGS="$(RUSTFLAGS)" cargo test --profile $(CARGO_PROFILE) \
 		-p tfhe-zk-pok --features experimental
 
+.PHONY: test_zk_pok_gpu # Run tfhe-zk-pok GPU-accelerated tests
+test_zk_pok_gpu:
+	RUSTFLAGS="$(RUSTFLAGS)" cargo test --profile $(CARGO_PROFILE) \
+		-p tfhe-zk-pok --features experimental,gpu-experimental -- gpu
+
+.PHONY: test_integer_zk_gpu # Run tfhe-zk-pok tests
+test_integer_zk_gpu: install_rs_check_toolchain
```
andrei-stoian-zama: this target and the above one do not use the `gpu-experimental` feature. what do these targets test ? old code ?

.github/workflows/gpu_zk_tests.yml
```diff
@@ -126,6 +131,9 @@ jobs:
       - name: Run zk-cuda-backend integration tests
         run: |
           make test_zk_cuda_backend
+          make test_zk_pok_gpu
```
andrei-stoian-zama: how much CI time is added ?

Makefile
```diff
@@ -774,7 +774,7 @@ test_zk_cuda_backend:
 
 
 .PHONY: test_gpu # Run the tests of the core_crypto module including experimental on the gpu backend
-test_gpu: test_core_crypto_gpu test_integer_gpu test_cuda_backend
+test_gpu: test_core_crypto_gpu test_integer_gpu test_cuda_backend test_zk_cuda_backend
```
andrei-stoian-zama, line 794: should we run the zk tests on approved ? only when the zk backend changes ?

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+/// GPU-accelerated multi-scalar multiplication for G1. The `gpu_index` parameter
+/// selects which GPU device to use; `None` defaults to GPU 0.
+///
+/// # Panics
+///
+/// - If `gpu_index` is `Some(i)` where `i >= number of available GPUs`.
+/// - If `bases` and `scalars` have different lengths (checked inside the backend).
+/// - If the GPU MSM call fails.
+#[must_use]
+pub fn g1_msm_gpu(bases: &[G1Affine], scalars: &[Zp], gpu_index: Option<u32>) -> G1 {
```
andrei-stoian-zama: I don't see a good reason to use Optional gpu_index.

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+/// GPU-accelerated multi-scalar multiplication for G2. The `gpu_index` parameter
+/// selects which GPU device to use; `None` defaults to GPU 0.
+///
+/// # Panics
+///
+/// - If `gpu_index` is `Some(i)` where `i >= number of available GPUs`.
+/// - If `bases` and `scalars` have different lengths (checked inside the backend).
+/// - If the GPU MSM call fails.
+#[must_use]
+pub fn g2_msm_gpu(bases: &[G2Affine], scalars: &[Zp], gpu_index: Option<u32>) -> G2 {
```
andrei-stoian-zama: again, no need for optional gpu_index

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+            .try_into()
+            .expect("cuda_get_number_of_gpus returned negative value")
+    })
+}
+
+/// Selects a GPU for MSM based on the rayon thread index, distributing work
+/// across all available GPUs. Returns `None` (meaning GPU 0) when only one GPU
+/// is present.
+#[inline]
+pub fn select_gpu_for_msm() -> Option<u32> {
```
andrei-stoian-zama: I don't understand why we have to rely on rayon here. is our multi-gpu logic necessarily related to rayon use ? in the cuda backend we use rayon in benchmarks but we don't assume rayon is needed.

tfhe-zk-pok/src/gpu/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+
+// ---------------------------------------------------------------------------
+// GPU helpers
+// ---------------------------------------------------------------------------
+
+/// Returns the number of available GPUs. Panics if no GPU is found.
+///
+/// The result is cached after the first call since GPU count cannot change
+/// during execution.
+pub(crate) fn get_num_gpus() -> u32 {
```
andrei-stoian-zama, line 30: can't we have a common file with the tfhe backend and use the function we used traditionally ?

tfhe-zk-pok/src/gpu/pke.rs
andrei-stoian-zama, line 1: some comments: 
-     // FIXME: div_round -> anything to do for it ? remove it ? 
- is there a reference for the proof - a link to some PDF ? CPU doesn't have one either..

tfhe-zk-pok/src/gpu/pke_v2.rs
andrei-stoian-zama, line 1: why do we use the `run_in_pool` mechanism ? CPU uses it, is it the best for us ?

Makefile
pdroalves: Yes, `test_integer_zk_gpu` executes `verify_and_expand` with GPU-accelerated expand (but CPU verify).

`test_zk_pok_gpu` has `gpu-experimental` so the target name actually should be `test_zk_pok_experimental_gpu`, as `test_integer_zk_experimental_gpu`.

.github/workflows/gpu_zk_tests.yml
pdroalves: 15 minutes. https://github.com/zama-ai/tfhe-rs/actions/runs/21573070408

Makefile
pdroalves: Given it's a 15min overhead, I would run them every time some zk-related thing changes.

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: That was useful when we were sharing implementation code with the CPU. Not anymore. I will remove the optional modifier.

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: I will remove this one either.

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: Quick answer: yes, out multi-gpu logic depends on rayon. We link one GPU per rayon thread.

Longer answer: The original version of this PR had `verify_impl` and `prove_impl` shared between CPU and GPU. This version splits those, so we now have our own. This opens the possibility to rewrite and reorganize those methods freely, but that's not what I'm doing in this PR. It has the minimum necessary changes to integrated the zk backend.

That's what I meant that I want to work next.

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: We could, just need to think where.

We can't use the version on `tfhe/` otherwise it would create a circular dependency with `tfhe-zk-pok`. We could maybe convert the current method we have in `tfhe/src/core_crypto/gpu/mod.rs`  to `tfhe-cuda-backend` bindings as a safe rust wrapper. I suppose that's feasible.

tfhe-zk-pok/src/gpu/pke.rs
pdroalves: Yes, we do: https://www.notion.so/zamaai/ZK-resources-d803e74fff9d4f0bbe1fa504679a2864

tfhe-zk-pok/src/gpu/pke_v2.rs
pdroalves: Possibly not. As before, this integration is done with the minimum code change to `verify_impl` and `proof_impl`. I plan to approach this type of thing in a future PR.

Makefile
andrei-stoian-zama: ok, could you add a workflow filter for that ?

tfhe-zk-pok/src/gpu/mod.rs
andrei-stoian-zama: makes sense! can you create an issue detailing the future work please ?

tfhe-zk-pok/src/gpu/mod.rs
andrei-stoian-zama: should it be done in this PR or a new one ?

tfhe-zk-pok/src/gpu/pke.rs
andrei-stoian-zama: if there is a publicly accessible reference we should add, if possible, comments in the code such as. see "Fancy ZK paper" by  Joye et al. 2007 - algorithm A. If there isn't a publicly available paper then nothing to be done, we'll handle docs internally.

Makefile
pdroalves: That's already what's happening.

In the `should-run` job the current file patterns are:

```                                                                                      
- tfhe/Cargo.toml                                                                                                                                                                         
- tfhe/build.rs                                                                                                                                                                           
- backends/zk-cuda-backend/**                                                                                                                                                             
- tfhe/src/integer/gpu/zk/**                                                                                                                                                              
- tfhe-zk-pok/**                                               
- tfhe/docs/**/**.md
- .github/workflows/gpu_zk_tests.yml
 - ci/slab.toml
```

For completeness I will add a few more items to that list:

```
- tfhe/src/zk/**
- backends/tfhe-cuda-backend/**
- tfhe/src/core_crypto/gpu/** (tfhe-zk-pok depends on CudaStreams)
- tfhe/src/integer/gpu/** (ZK depends on ciphertext::compact_list and key_switching_key, we want to catch anything that could break there)
- tfhe/src/shortint/parameters/** (in case something changes in the parameters)
```

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: I would do it in a new PR. We already have to work on the common compilation artifacts between both CUDA backends. I think this would be part of that. I added it to the issue we have about that: https://github.com/zama-ai/tfhe-rs-internal/issues/1325

tfhe-zk-pok/src/gpu/mod.rs
pdroalves: Done: https://github.com/zama-ai/tfhe-rs-internal/issues/1358

tfhe-zk-pok/src/gpu/pke.rs
pdroalves: Not sure if we have a public reference for that. @nsarlin-zama do you?

tfhe-zk-pok/src/gpu/pke_v2.rs
nsarlin-zama: the run_in_pool has been introduced since a benchmark showed that it's actually faster with less cpu threads available (likely caused by the way msm are parallelized in arkworks). 
I don't think it will stay true for the gpu impl.

tfhe-zk-pok/src/gpu/pke.rs
nsarlin-zama: The paper for zkv1 is in the zk doc: https://docs.zama.org/tfhe-rs/fhe-computation/advanced-features/zk-pok

For zkv2 it's this one: https://eprint.iacr.org/2026/033. 
I forgot to add it in the doc once it got published on eprint, I will do it

tfhe-zk-pok/src/gpu/pke_v2.rs
andrei-stoian-zama: It's fine to use this if the multi-threading needs to be handled in the backend. If we decide it's the caller's responsibility to multi-thread then we should remove it. But maybe in a later PR?

tfhe-zk-pok/src/gpu/pke.rs
pdroalves: @andrei-stoian-zama Do you want me to add any of these links as comment to the code?

tfhe-zk-pok/src/gpu/pke_v2.rs
pdroalves: Yes, I think a next PR is better.

IceTDrinker (commented):
Had some time to read the bench stuff, nothing major, just some questions on my end

> @IceTDrinker partially reviewed 9 files and all commit messages, made 5 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-:-Oo-YUuW-FhbesAXm93J:b-tgv196)* status: all files reviewed, 44 unresolved discussions (waiting on agnesLeroy, andrei-stoian-zama, guillermo-oyarzun, nsarlin-zama, pdroalves, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 493 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-VQCn5trMQebz-A3O:-Oo-VQCn5trMQebz-A3P:b-o8xhtn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L493)):*
> ```XML
>     /// Empirically measured with ZkComputeLoad::Verify on 8xH100-NVLink (80GB) GPUs.
>     /// Tuned to avoid OOM — expansion allocates server keys,
>     /// key switching material, and PBS buffers per element on GPU.
> ```

per element ? :monocle_face: 

it's decompressed per GPU once rather I hope ?
___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 811 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-VkwD1eIQT0QJ-bem:-Oo-VkwD1eIQT0QJ-ben:b-5vvnhp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L811)):*
> ```XML
>                                                                .unwrap();
>                                                        });
>                                                    }, BatchSize::PerIteration);
> ```

curious about that change ? not a criterion pro, so wondering if it could be useful for the CPU benchmarks elsewhere
___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 898 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-W6RV-a2SLqMuxDg5:-Oo-W6RV-a2SLqMuxDg6:b-7h97ni) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L898)):*
> ```XML
>             let compact_private_key = CompactPrivateKey::new(*param_pke);
>             let pk = CompactPublicKey::new(&compact_private_key);
>             // Kept for consistency
> ```

why ?
___
*[`tfhe-benchmark/benches/zk/msm.rs` line 68 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-YHFY1TKKB6p49r_T:-Oo-YHFY1TKKB6p49r_U:bkf9ns7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/zk/msm.rs#L68)):*
> ```XML
> }
> 
> fn bench_cpu_g1_msm(c: &mut Criterion) {
> ```

wondering if you may want to use the same trick you used in the tests here to limit the amount of code in the benchmarks, not critical

pdroalves (commented):
> @pdroalves made 21 comments and resolved 14 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275)* status: all files reviewed, 30 unresolved discussions (waiting on agnesLeroy, andrei-stoian-zama, guillermo-oyarzun, IceTDrinker, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`Makefile` line 1238 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwsw5tB4n5Jqr0-C3L:-Oo0RRBLEt4xXqG0dTvw:b-78fayc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/Makefile#L1238)):*

mismatched toolchains installed/used
</blockquote></details>

Not sure I'm following. Should I use the default toolchain (i.e to remove `$(CARGO_RS_BUILD_TOOLCHAIN)`)? Probably I should also be using `$(CARGO_PROFILE)` instead of release.
___
*[`Makefile` line 1243 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwt0VX9A3hhuJsA27j:-Oo0TEjRBoYMixjgbpJc:b-4rpkdu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/Makefile#L1243)):*

probably no need to install the toolchain here, since each sub target should request the right one if they need one
</blockquote></details>

Ok! Removed.
___
*[`backends/zk-cuda-backend/src/types/g1.rs` line 286 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwnyc80JCBePbiL3qa:-Oo0TqVZ24U2MoG5Ddgc:b-3vp7se) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/backends/zk-cuda-backend/src/types/g1.rs#L286)):*

just wondering what the size_tracker is used for ?
</blockquote></details>

For now, nothing. It's there to mirror what we have in the TFHE backend. There the blockchain people will rely on it to check if the instance size fits the GPU memory beforehand. This implementation is not currently used but Agnes asked to keep it wired for the future.
___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 493 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-VQCn5trMQebz-A3O:-Oo10lEJ6B2OGIZQ8f7g:b-yi5vol) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L493)):*

per element ? :monocle_face: 

it's decompressed per GPU once rather I hope ?
</blockquote></details>

The keys are allocated per GPU, not per element. The "per element" refers to the PBS buffers, but that's actually an internal info that doesn't matter at this level. I will simplify the comment to avoid confusion.
___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 811 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-VkwD1eIQT0QJ-bem:-Oo12Ve186X9RMrrD37u:bjqgcd4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L811)):*

curious about that change ? not a criterion pro, so wondering if it could be useful for the CPU benchmarks elsewhere
</blockquote></details>

I was initially having a lot of trouble finding a setup (keys distribution and number of elements) that didn't produce an OOM. `BatchSize::PerIteration` helped with that. I think the overall bench will takes longer to run, but the memory consumption should be smaller.
___
*[`tfhe-benchmark/benches/integer/zk_pke.rs` line 898 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-W6RV-a2SLqMuxDg5:-Oo131G_5bSN9fLAzwBQ:bgptyjh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/integer/zk_pke.rs#L898)):*

why ?
</blockquote></details>

No need actually.  We don't use the ksk here, so we can remove this. That was only an attempt to mirror the GPU one. 
___
*[`tfhe-benchmark/benches/zk/msm.rs` line 68 at r11](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Oo-YHFY1TKKB6p49r_T:-Oo1413g2Lx0_34rX1To:b-jrrdbg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5740706524ea6ffa3e883a42b33cf0b999bdc11b/tfhe-benchmark/benches/zk/msm.rs#L68)):*

wondering if you may want to use the same trick you used in the tests here to limit the amount of code in the benchmarks, not critical
</blockquote></details>

That's actually a good idea and really easy to implement! Will work on it.
___
*[`tfhe-zk-pok/Cargo.toml` line 28 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwoghl27X-WO6QCLsn:-Oo0UyeqEW2i7Fm1VFLi:bgovrrk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/Cargo.toml#L28)):*

do you use the tfhe-cuda-backend directly in the zk code ? (just wondering if it's necessary or the zk-cuda-backend has everything it needs from the tfhe-cuda-backend)
</blockquote></details>

I need it for `cuda_create_stream`, `cuda_destroy_stream`, `cuda_get_number_of_gpus`. We have an open issue to move these methods to a common backend so we don't need to import the entire tfhe-cuda-backend. That's going to be done in a future PR.
___
*[`tfhe-zk-pok/src/gpu/mod.rs` line 23 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwy6Jz-DbW470IeMqn:-Oo0VJQV213AMC66lzST:b-czixf1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/mod.rs#L23)):*

do you still do transmutes ? I recall Nicolas recommending not doing transmutes
</blockquote></details>

No. The comment is old. I will remove it.
___
*[`tfhe-zk-pok/src/gpu/mod.rs` line 56 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-OnwyZ_f9RyoTppTKbUv:-Oo0pBPP9X9DiK8QyojQ:b1wqd3m) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/mod.rs#L56)):*

should this assert ?
</blockquote></details>

I think so. It's safe. `cuda_get_number_of_gpus()` should throw a driver error in case it tries to run on a machine without a compatible device, but it may also just return 0. With this assert be catch the issue before it propagates.
___
*[`tfhe-zk-pok/src/gpu/mod.rs` line 113 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwz9NP5yr4ZkvVBIOL:-Oo0r-ZJArxdyXR8JjxY:b-9r343l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/mod.rs#L113)):*

could maybe use CpuG1Affine, CudaG1Affine as aliases where both exist, and you can also differentiate Ark types with like ArkG1Affine if needed

not sure, but I'm getting a bit confused on what the Zk prefix means (and I think it may not be the same everywhere it's used)
</blockquote></details>

Could be a good idea. I will rename them to `CudaG*Affine`.
___
*[`tfhe-zk-pok/src/gpu/mod.rs` line 192 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onx-1gp-8INz6f6-kNx:-Oo0tcGI0d240iJZRF8T:b-kh52ou) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/mod.rs#L192)):*

Zp from ark, or the tfhe-zk-pok ?
</blockquote></details>

tfhe-zk-pok. I will improve the comment.
___
*[`tfhe-zk-pok/src/gpu/mod.rs` line 237 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onx-KKJDrdaM_kENI2E:-Oo0tfqn69d2RZ2dnTHw:bizxucc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/mod.rs#L237)):*

is creating/destroying streams costly ?

if so down the line you may want to have a way to manage them without creating them just in time every time
</blockquote></details>

Not costly, not free. We surely don't want to create them all the time as we do now. That's the first I want to address after we merge this PR.
___
*[`tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs` line 7 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-OnwpwI62WvHrYmHYQWy:-Oo0u9as6H-2wprQ5i-m:bt07h94) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs#L7)):*

tiny bit confused by the both directions covered only indicating GPU verify when the above talks about both verifiers agreeing

it's a nit, but it seems inconsistent at the moment to me
</blockquote></details>

The comment is incomplete. We test all combinations except `CPU-Prove -> CPU-Verify`. I will fix it.
___
*[`tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs` line 10 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwq8SM1CSmHgM0uWTG:-Oo0uVUaC0iRvsrZtGCj:bfeq2aw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs#L10)):*

only invalid inputs are verified ?
</blockquote></details>

Again a misleading comment. Among the combinations we have the case when all of them are valid. I will fix the comment.
___
*[`tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs` line 155 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwr5eBC2EXTwX3pkkV:-Oo0w1co43lffr1CD8f0:br05mju) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/prove_verify_stress.rs#L155)):*

fiat shamir binding ?

I'm not a zk pro, so not sure what that means
</blockquote></details>

Fiat-Shamir is the algorithm we use to generate challenges deterministically in a NIZK. We hash the metadata together with the encryption parameters, so mismatched metadata would fail the verification although the ciphertext is well formed and the proof is valid. We just would be able to regenerate the challenges we need.
___
*[`tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs` line 174 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-OnwwhVW3Qo16_cSFCN7:-Oo0y9bhBfjpqU24L62S:b-jt7fqi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs#L174)):*

I looked at the test function and this is the first time there is a call to CUDA, so all comments saying things are checked ahead of time are wrong ?
</blockquote></details>

The others look good to me. The only ones that are not precise are the ones about `gpu_index`. I will replace them with `// SAFETY: gpu_index 0 is assumed valid on any CUDA-capable machine.`. Wdyt?
___
*[`tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs` line 204 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwvz_-6isWLRZkNaLq:-Oo0yVxM3iGG-qMpFvHx:b-7ug84z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs#L204)):*

tfhe ?

maybe gpu_result ?
</blockquote></details>

tfhe because it's the GPU result converted back to tfhe-zk-pok. `gpu_result` is less confusing but is taken. I will rename it to `gpu_result_affine`. Ty!
___
*[`tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs` line 213 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-OnwvrnM20Sc-k82vyp5:-Oo0z8cJ9SP2OQErhXJ3:b-pef81r) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs#L213)):*

is it possible to reach infinity in this test in theory ?

if not could be good to assert points are not at infinity ?
</blockquote></details>

Not really. I will replace the conditional by an assert.
___
*[`tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs` line 393 at r10](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-Onwwys43tmYa_n3cwUp:-Oo1-5QS-ES4t5OcC-BA:bno41er) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1cdb8e8ec399dab5ee527e6ef0ffd26dd7aad9c4/tfhe-zk-pok/src/gpu/tests/zk_cuda_backend.rs#L393)):*

tfhe ?
</blockquote></details>

Same thing as before. I will rename to `cpu_`.
___
*[`tfhe-zk-pok/src/gpu/pke_v2.rs` line 247 at r9](https://reviewable.io/reviews/zama-ai/tfhe-rs/3275#-OnrOJ-82NMUmji0GY16:-Oo16p4xBGmMhxT0ku1W:b-cp1lbb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5d287c9e6206e57c5320891b8ab408d9fd27a333/tfhe-zk-pok/src/gpu/pke_v2.rs#L247)):*

using try_from instead of as to convert between i128 and i64
</blockquote></details>

You mean we should use `as` instead of `try_from`?

I believe `ProofSanityCheckMode` is being respected. The bounds checks at lines 138, 262, and 310 are all guarded by `if  sanity_check_mode == ProofSanityCheckMode::Panic`

PR #3281, opened 2026-02-09
chore(bench): allow result parsing of kv store benchmarks

soonum:
Naming has been reworked to stick with the future naming standard.
Call to `write_to_json()` has been added to be able to parse the results and send them to database.

In addition, KV Store benchmarks have been moved to their own file to reduce execution time of unsigned operations benchmark.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3281)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1672,7 +1672,7 @@ bench_web_js_api_unsafe_coop_firefox_ci: setup_venv
 	nvm use $(NODE_VERSION) && \
 	$(MAKE) bench_web_js_api_unsafe_coop_firefox
 
-.PHONY: bench_pi_unsigned # Run benchmarks for integer operations
+.PHONY: bench_hlapi_unsigned # Run benchmarks for integer operations
 bench_hlapi_unsigned: install_rs_check_toolchain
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1672,7 +1672,7 @@ bench_web_js_api_unsafe_coop_firefox_ci: setup_venv
 	nvm use $(NODE_VERSION) && \
 	$(MAKE) bench_web_js_api_unsafe_coop_firefox
 
-.PHONY: bench_hlapi_unsigned # Run benchmarks for integer operations
+.PHONY: bench_pi_unsigned # Run benchmarks for integer operations
```
IceTDrinker: something went wrong for the naming here

tfhe-benchmark/benches/high_level_api/kv_store.rs
```diff
@@ -0,0 +1,205 @@
...
+        let config =
+            ConfigBuilder::with_custom_parameters(BENCH_PARAM_MESSAGE_2_CARRY_2_KS32_PBS).build();
+        let cks = ClientKey::generate(config);
+        let compressed_sks = CompressedServerKey::new(&cks);
+
+        let sks = compressed_sks.decompress();
+        rayon::broadcast(|_| set_server_key(sks.clone()));
+        set_server_key(sks);
+        (cks, tfhe::Device::Cpu)
+    };
```
IceTDrinker, line 177: KVStore only available on CPU for now or is it available elsewhere ?

soonum: Yes, previously it was behind a condition `if backend == TfheBackend::CPU` for all KV Store benchmarks.

PR #3284, opened 2026-02-09
chore: revert backward compatibility change

IceTDrinker:
- this change should not have been needed and poses risks for backward compatibility
- HL CUDA: use dedicated type for the ReRand

This will be backported in 1.5 too

the gist:
- reverted a change on the CPU definition of the Key, moved the GPU primitive to the integer layer instead of shortint
- defined the type for GPU to be able to decorrelate the key types (since the enum is just to encode a key choice)

### DIFF
--- a/tfhe-benchmark/benches/integer/rerand.rs
+++ b/tfhe-benchmark/benches/integer/rerand.rs
@@ -197,10 +197,10 @@ mod cuda {
     use tfhe::integer::ciphertext::ReRandomizationContext;
     use tfhe::integer::gpu::ciphertext::compressed_ciphertext_list::CudaCompressedCiphertextListBuilder;
     use tfhe::integer::gpu::ciphertext::{CudaIntegerRadixCiphertext, CudaUnsignedRadixCiphertext};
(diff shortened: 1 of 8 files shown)


IceTDrinker:
> Looks good to me, sorry that change was merged without someone from the CPU team reviewing I don't know what happened on this one 😞 Usually we open a PR with the integer integration, then another separate one for the HL API so we don't have this kind of problem, I'll tell the team to be careful to do that systematically.

It happens, we caught this one in time, there is another change that makes some old versions of specific structs incompatible, we can live with that (and maybe have a workaround if we deem it necessary)

Again really don't hesitate to check in with us, we are more than happy to help :)

IceTDrinker:
due to the code owners restriction on main it looks like I need an approval from @agnesLeroy, @soonum and @mayeul-zama or @nsarlin-zama

nsarlin-zama (approved):
From a backward perspective, since CudaKeySwitchingKeyMaterial is not Serialize, we know that this change is not breaking unless a user did something really weird.

PR #3294, opened 2026-02-11
fix(gpu): add semgrep to pcc gpu

andrei-stoian-zama:
Makes changes to GPU PCC to ensure synchronous CUDA backend functions call synchronize appropriately.

**Adds semgrep rules to ensure **
- all cleanup functions should not be async : 
  - they don't have have a name ending in _async
  - they calls to `->release()` on memory structures
  - or they call synchronize stream
- all release methods of memory structures synchronize stream 0

**2. A script to check that all scratch_ / cuda_ / cleanup_ are triplets of functions ***
- have consistent naming - I have duplicated scratch / cleanup so all triplets match
- are called in this order -> to void anyone forgetting a cleanup_ call.
- scratch_ and cuda_ functions in such triplets are _async

** 3. Checks that Rust functions are always sync: **
- either they call an `_async` binding and `stream.synchronize()`
- or they call a sync binding

** 4. Checks that synchronized bindings call synchronize in C++**

 ## Removed `streams.synchronize()` in Rust (5 calls)                                                                                                      
                                                                                                                                                            
  | # | File | Line (main) | Function |                                                                                                                     
  |---|------|-------------|----------|                                                                                                                     
  | 1 | `tfhe/src/core_crypto/gpu/algorithms/lwe_keyswitch.rs` | 187 | `cuda_scratch_keyswitch_lwe_ciphertext` |                                            
  | 2 | `tfhe/src/core_crypto/gpu/algorithms/lwe_keyswitch.rs` | 205 | `cleanup_cuda_keyswitch` |                                                           
  | 3 | `tfhe/src/core_crypto/gpu/algorithms/lwe_programmable_bootstrapping.rs` | 137 | `cuda_programmable_bootstrap_lwe_ciphertext` |                      
  | 4 | `tfhe/src/core_crypto/gpu/algorithms/lwe_programmable_bootstrapping.rs` | 265 | `cuda_programmable_bootstrap_128_lwe_ciphertext` |                  
  | 5 | `tfhe/src/core_crypto/gpu/mod.rs` | 868 | `cuda_modulus_switch_ciphertext` |                                                                        
                                                                                                                                                            
  ## Added `cuda_synchronize_stream` in C++ (12 calls)                                                                                                      
                                                                                                                                                            
  | # | File | Line | Function / Context |                                                                                                                  
  |---|------|------|--------------------|                                                                                                                  
  | 1 | `cuda/include/pbs/pbs_utilities.h` | 246 | `pbs_buffer<Torus, CLASSICAL>::release()` |                                                              
  | 2 | `cuda/include/pbs/pbs_utilities.h` | 411 | `pbs_buffer_128<InputTorus, CLASSICAL>::release()` |                                                     
  | 3 | `cuda/include/pbs/pbs_multibit_utilities.h` | 291 | `pbs_buffer<Torus, MULTI_BIT>::release()` |                                                     
  | 4 | `cuda/include/pbs/pbs_multibit_utilities.h` | 435 | `pbs_buffer_128<InputTorus, MULTI_BIT>::release()` |                                            
  | 5 | `cuda/include/trivium/trivium_utilities.h` | 257 | `int_trivium_state_workspaces<Torus>::release()` |                                               
  | 6 | `cuda/src/crypto/keyswitch.cu` | 134 | `cleanup_cuda_keyswitch_gemm_64_64` |                                                                        
  | 7 | `cuda/src/crypto/keyswitch.cu` | 143 | `cleanup_cuda_keyswitch_gemm_64_32` |
  | 8 | `cuda/src/crypto/keyswitch.cu` | 175 | `cleanup_cuda_packing_keyswitch_lwe_list_to_glwe_64` |
  | 9 | `cuda/src/crypto/keyswitch.cu` | 183 | `cleanup_cuda_packing_keyswitch_lwe_list_to_glwe_128` |
  | 10 | `cuda/src/crypto/keyswitch.cuh` | 718 | `cleanup_cuda_keyswitch` (template) |
  | 11 | `cuda/src/linearalgebra/multiplication.cu` | 76 | `cleanup_cuda_wrapping_polynomial_mul_one_to_many_64` |
  | 12 | `cuda/src/pbs/programmable_bootstrap_amortized.cu` | 376 | `cleanup_cuda_programmable_bootstrap_amortized_64` |

  Line numbers for the removed Rust calls reference main (since those lines no longer exist in the branch). Line numbers for the added C++ calls reference
  the current branch.


Scratch functions that don't have cuda ops or cleanup:

```
'scratch_cuda_programmable_bootstrap_32_async' and
'scratch_cuda_programmable_bootstrap_amortized_32_async'
```

---
## Standalone Async cuda Functions

| Old Name | New Name |
|----------|----------|
| `cuda_centered_modulus_switch_64` | `cuda_centered_modulus_switch_64_async` |
| `cuda_convert_lwe_ciphertext_vector_to_cpu_64` | `cuda_convert_lwe_ciphertext_vector_to_cpu_64_async` |
| `cuda_convert_lwe_ciphertext_vector_to_gpu_64` | `cuda_convert_lwe_ciphertext_vector_to_gpu_64_async` |
| `cuda_convert_lwe_multi_bit_programmable_bootstrap_key_128` | `cuda_convert_lwe_multi_bit_programmable_bootstrap_key_128_async` |
| `cuda_convert_lwe_multi_bit_programmable_bootstrap_key_64` | `cuda_convert_lwe_multi_bit_programmable_bootstrap_key_64_async` |
| `cuda_convert_lwe_programmable_bootstrap_key_128` | `cuda_convert_lwe_programmable_bootstrap_key_128_async` |
| `cuda_convert_lwe_programmable_bootstrap_key_32` | `cuda_convert_lwe_programmable_bootstrap_key_32_async` |
| `cuda_convert_lwe_programmable_bootstrap_key_64` | `cuda_convert_lwe_programmable_bootstrap_key_64_async` |
| `cuda_fourier_polynomial_mul` | `cuda_fourier_polynomial_mul_async` |
| `cuda_glwe_sample_extract_128` | `cuda_glwe_sample_extract_128_async` |
| `cuda_glwe_sample_extract_64` | `cuda_glwe_sample_extract_64_async` |
| `cuda_glwe_wrapping_polynomial_mul_one_to_many_64` | `cuda_glwe_wrapping_polynomial_mul_one_to_many_64_async` |
| `cuda_integer_aes_ctr_256_encrypt_64` | `cuda_integer_aes_ctr_256_encrypt_64_async` |
| `cuda_integer_extract_glwe_128` | `cuda_integer_extract_glwe_128_async` |
| `cuda_integer_extract_glwe_64` | `cuda_integer_extract_glwe_64_async` |
| `cuda_integer_reverse_blocks_64_inplace` | `cuda_integer_reverse_blocks_64_inplace_async` |
| `cuda_keyswitch_gemm_lwe_ciphertext_vector_64_32` | `cuda_keyswitch_gemm_64_32_async` |
| `cuda_keyswitch_gemm_lwe_ciphertext_vector_64_64` | `cuda_keyswitch_gemm_64_64_async` |
| `cuda_keyswitch_lwe_ciphertext_vector_64_32` | `cuda_keyswitch_lwe_ciphertext_vector_64_32_async` |
| `cuda_keyswitch_lwe_ciphertext_vector_64_64` | `cuda_keyswitch_lwe_ciphertext_vector_64_64_async` |
| `cuda_modulus_switch_64` | `cuda_modulus_switch_64_async` |
| `cuda_modulus_switch_inplace_64` | `cuda_modulus_switch_inplace_64_async` |
| `cuda_programmable_bootstrap_amortized_lwe_ciphertext_vector_32` | `cuda_programmable_bootstrap_amortized_lwe_ciphertext_vector_32_async` |
| `cuda_programmable_bootstrap_lwe_ciphertext_vector_32` | `cuda_programmable_bootstrap_lwe_ciphertext_vector_32_async` |
| `cuda_small_scalar_multiplication_integer_64_inplace` | `cuda_small_scalar_multiplication_integer_64_inplace_async` |

---

##  Sync (non-async) cuda Functions

All 12 unchanged from main:

| Function |
|----------|
| `cuda_add_lwe_ciphertext_vector_32` |
| `cuda_add_lwe_ciphertext_vector_64` |
| `cuda_add_lwe_ciphertext_vector_plaintext_64` |
| `cuda_add_lwe_ciphertext_vector_plaintext_vector_32` |
| `cuda_add_lwe_ciphertext_vector_plaintext_vector_64` |
| `cuda_bitnot_ciphertext_64` |
| `cuda_mult_lwe_ciphertext_vector_cleartext_vector_32` |
| `cuda_mult_lwe_ciphertext_vector_cleartext_vector_64` |
| `cuda_negate_ciphertext_64` |
| `cuda_negate_lwe_ciphertext_vector_32` |
| `cuda_negate_lwe_ciphertext_vector_64` |
| `cuda_scalar_addition_ciphertext_64_inplace` |

### DIFF
--- a/tfhe/src/core_crypto/gpu/ffi.rs
+++ b/tfhe/src/core_crypto/gpu/ffi.rs
@@ -1,3 +1,4 @@
+#![deny(clippy::cast_possible_truncation)]
 use super::{CudaStreams, PBSMSNoiseReductionType};
 use crate::core_crypto::gpu::lwe_bootstrap_key::CudaModulusSwitchNoiseReductionConfiguration;
 use crate::core_crypto::gpu::vec::CudaVec;
@@ -49,10 +50,10 @@ pub unsafe fn programmable_bootstrap<T: UnsignedInteger>(
         streams.ptr[0],
         streams.gpu_indexes[0].get(),
         std::ptr::addr_of_mut!(pbs_buffer),
-        lwe_dimension.0 as u32,
-        glwe_dimension.0 as u32,
-        polynomial_size.0 as u32,
-        level.0 as u32,
+        u32::try_from(lwe_dimension.0).unwrap(),
+        u32::try_from(glwe_dimension.0).unwrap(),
+        u32::try_from(polynomial_size.0).unwrap(),
+        u32::try_from(level.0).unwrap(),
         num_samples,
         true,
         noise_reduction_type as u32,
@@ -69,11 +70,11 @@ pub unsafe fn programmable_bootstrap<T: UnsignedInteger>(
         lwe_in_indexes.as_c_ptr(0),
         bootstrapping_key.as_c_ptr(0),
         pbs_buffer,
-        lwe_dimension.0 as u32,
-        glwe_dimension.0 as u32,
-        polynomial_size.0 as u32,
-        base_log.0 as u32,
-        level.0 as u32,
+        u32::try_from(lwe_dimension.0).unwrap(),
+        u32::try_from(glwe_dimension.0).unwrap(),
+        u32::try_from(polynomial_size.0).unwrap(),
+        u32::try_from(base_log.0).unwrap(),
+        u32::try_from(level.0).unwrap(),
         num_samples,
         num_many_lut,
         lut_stride,
@@ -106,10 +107,10 @@ pub fn get_programmable_bootstrap_size_on_gpu(
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(pbs_buffer),
-            lwe_dimension.0 as u32,
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            level.0 as u32,
+            u32::try_from(lwe_dimension.0).unwrap(),
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             false,
             noise_reduction_type as u32,
@@ -160,10 +161,10 @@ pub unsafe fn programmable_bootstrap_128<T: UnsignedInteger>(
         streams.ptr[0],
         streams.gpu_indexes[0].get(),
         std::ptr::addr_of_mut!(pbs_buffer),
-        lwe_dimension.0 as u32,
-        glwe_dimension.0 as u32,
-        polynomial_size.0 as u32,
-        level.0 as u32,
+        u32::try_from(lwe_dimension.0).unwrap(),
+        u32::try_from(glwe_dimension.0).unwrap(),
+        u32::try_from(polynomial_size.0).unwrap(),
+        u32::try_from(level.0).unwrap(),
         num_samples,
         true,
         noise_reduction_type as u32,
@@ -177,11 +178,11 @@ pub unsafe fn programmable_bootstrap_128<T: UnsignedInteger>(
         lwe_array_in.as_c_ptr(0),
         bootstrapping_key.as_c_ptr(0),
         pbs_buffer,
-        lwe_dimension.0 as u32,
-        glwe_dimension.0 as u32,
-        polynomial_size.0 as u32,
-        base_log.0 as u32,
-        level.0 as u32,
+        u32::try_from(lwe_dimension.0).unwrap(),
+        u32::try_from(glwe_dimension.0).unwrap(),
+        u32::try_from(polynomial_size.0).unwrap(),
+        u32::try_from(base_log.0).unwrap(),
+        u32::try_from(level.0).unwrap(),
         num_samples,
     );
 
@@ -225,9 +226,9 @@ pub unsafe fn programmable_bootstrap_multi_bit<T: UnsignedInteger, B: Any + Unsi
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(pbs_buffer),
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            level.0 as u32,
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             true,
         );
@@ -241,12 +242,12 @@ pub unsafe fn programmable_bootstrap_multi_bit<T: UnsignedInteger, B: Any + Unsi
             input_indexes.as_c_ptr(0),
             bootstrapping_key.as_c_ptr(0),
             pbs_buffer,
-            lwe_dimension.0 as u32,
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            grouping_factor.0 as u32,
-            base_log.0 as u32,
-            level.0 as u32,
+            u32::try_from(lwe_dimension.0).unwrap(),
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(grouping_factor.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             num_many_lut,
             lut_stride,
@@ -261,9 +262,9 @@ pub unsafe fn programmable_bootstrap_multi_bit<T: UnsignedInteger, B: Any + Unsi
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(pbs_buffer),
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            level.0 as u32,
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             true,
         );
@@ -278,12 +279,12 @@ pub unsafe fn programmable_bootstrap_multi_bit<T: UnsignedInteger, B: Any + Unsi
             input_indexes.as_c_ptr(0),
             bootstrapping_key.as_c_ptr(0),
             pbs_buffer,
-            lwe_dimension.0 as u32,
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            grouping_factor.0 as u32,
-            base_log.0 as u32,
-            level.0 as u32,
+            u32::try_from(lwe_dimension.0).unwrap(),
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(grouping_factor.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             num_many_lut,
             lut_stride,
@@ -312,9 +313,9 @@ pub fn get_programmable_bootstrap_multi_bit_size_on_gpu(
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(pbs_buffer),
-            glwe_dimension.0 as u32,
-            polynomial_size.0 as u32,
-            level.0 as u32,
+            u32::try_from(glwe_dimension.0).unwrap(),
+            u32::try_from(polynomial_size.0).unwrap(),
+            u32::try_from(level.0).unwrap(),
             num_samples,
             false,
         )
@@ -362,8 +363,8 @@ pub unsafe fn keyswitch_gemm<T: UnsignedInteger, KST: UnsignedInteger>(
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(ks_tmp_buffer),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
             num_samples,
             true,
         );
@@ -375,10 +376,10 @@ pub unsafe fn keyswitch_gemm<T: UnsignedInteger, KST: UnsignedInteger>(
             lwe_array_in.as_c_ptr(0),
             lwe_in_indexes.as_c_ptr(0),
             keyswitch_key.as_c_ptr(0),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
-            base_log.0 as u32,
-            l_gadget.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(l_gadget.0).unwrap(),
             num_samples,
             ks_tmp_buffer,
             uses_trivial_indices,
@@ -394,8 +395,8 @@ pub unsafe fn keyswitch_gemm<T: UnsignedInteger, KST: UnsignedInteger>(
             streams.ptr[0],
             streams.gpu_indexes[0].get(),
             std::ptr::addr_of_mut!(ks_tmp_buffer),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
             num_samples,
             true,
         );
@@ -407,10 +408,10 @@ pub unsafe fn keyswitch_gemm<T: UnsignedInteger, KST: UnsignedInteger>(
             lwe_array_in.as_c_ptr(0),
             lwe_in_indexes.as_c_ptr(0),
             keyswitch_key.as_c_ptr(0),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
-            base_log.0 as u32,
-            l_gadget.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(l_gadget.0).unwrap(),
             num_samples,
             ks_tmp_buffer,
             uses_trivial_indices,
@@ -457,10 +458,10 @@ pub unsafe fn keyswitch_async<T: UnsignedInteger, KT: UnsignedInteger>(
             lwe_array_in.as_c_ptr(0),
             lwe_in_indexes.as_c_ptr(0),
             keyswitch_key.as_c_ptr(0),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
-            base_log.0 as u32,
-            l_gadget.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(l_gadget.0).unwrap(),
             num_samples,
         );
     } else if TypeId::of::<KT>() == TypeId::of::<u64>() {
@@ -472,10 +473,10 @@ pub unsafe fn keyswitch_async<T: UnsignedInteger, KT: UnsignedInteger>(
             lwe_array_in.as_c_ptr(0),
             lwe_in_indexes.as_c_ptr(0),
             keyswitch_key.as_c_ptr(0),
-            input_lwe_dimension.0 as u32,
-            output_lwe_dimension.0 as u32,
-            base_log.0 as u32,
-            l_gadget.0 as u32,
+            u32::try_from(input_lwe_dimension.0).unwrap(),
+            u32::try_from(output_lwe_dimension.0).unwrap(),
+            u32::try_from(base_log.0).unwrap(),
+            u32::try_from(l_gadget.0).unwrap(),
             num_samples,
         );
     } else {
@@ -524,10 +525,10 @@ pub unsafe fn packing_keyswitch_list_64<T: UnsignedInteger>(
         streams.ptr[0],
         streams.gpu_indexes[0].get(),
         std::ptr::addr_of_mut!(fp_ks_buffer),
-        input_lwe_dimension.0 as u32,
-        output_glwe_dimension.0 as u32,
-        output_polynomial_size.0 as u32,
-        num_lwes.0 as u32,
+        u32::try_from(input_lwe_dimension.0).unwrap(),
+        u32::try_from(output_glwe_dimension.0).unwrap(),
+        u32::try_from(output_polynomial_size.0).unwrap(),
+        u32::try_from(num_lwes.0).unwrap(),
         true,
     );
     cuda_packing_keyswitch_lwe_list_to_glwe_64_async(
@@ -537,12 +538,12 @@ pub unsafe fn packing_keyswitch_list_64<T: UnsignedInteger>(
         lwe_array_in.as_c_ptr(0),
         packing_keyswitch_key.as_c_ptr(0),
         fp_ks_buffer,
-        input_lwe_dimension.0 as u32,
-        output_glwe_dimension.0 as u32,
-        output_polynomial_size.0 as u32,
-        base_log.0 as u32,
-        l_gadget.0 as u32,
-        num_lwes.0 as u32,
+        u32::try_from(input_lwe_dimension.0).unwrap(),
+        u32::try_from(output_glwe_dimension.0).unwrap(),
+        u32::try_from(output_polynomial_size.0).unwrap(),
+        u32::try_from(base_log.0).unwrap(),
+        u32::try_from(l_gadget.0).unwrap(),
+        u32::try_from(num_lwes.0).unwrap(),
     );
     cleanup_cuda_packing_keyswitch_lwe_list_to_glwe_64(
         streams.ptr[0],
@@ -578,10 +579,10 @@ pub unsafe fn packing_keyswitch_list_128<T: UnsignedInteger>(
         streams.ptr[0],
         streams.gpu_indexes[0].get(),
         std::ptr::addr_of_mut!(fp_ks_buffer),
-        input_lwe_dimension.0 as u32,
-        output_glwe_dimension.0 as u32,
-        output_polynomial_size.0 as u32,
-        num_lwes.0 as u32,
+        u32::try_from(input_lwe_dimension.0).unwrap(),
+        u32::try_from(output_glwe_dimension.0).unwrap(),
+        u32::try_from(output_polynomial_size.0).unwrap(),
+        u32::try_from(num_lwes.0).unwrap(),
         true,
     );
     cuda_packing_keyswitch_lwe_list_to_glwe_128_async(
@@ -591,12 +592,12 @@ pub unsafe fn packing_keyswitch_list_128<T: UnsignedInteger>(
         lwe_array_in.as_c_ptr(0),
         packing_keyswitch_key.as_c_ptr(0),
         fp_ks_buffer,
-        input_lwe_dimension.0 as u32,
-        output_glwe_dimension.0 as u32,
-        output_polynomial_size.0 as u32,
-        base_log.0 as u32,
-        l_gadget.0 as u32,
-        num_lwes.0 as u32,
+        u32::try_from(input_lwe_dimension.0).unwrap(),
+        u32::try_from(output_glwe_dimension.0).unwrap(),
+        u32::try_from(output_polynomial_size.0).unwrap(),
+        u32::try_from(base_log.0).unwrap(),
+        u32::try_from(l_gadget.0).unwrap(),
+        u32::try_from(num_lwes.0).unwrap(),
     );
     cleanup_cuda_packing_keyswitch_lwe_list_to_glwe_128(
         streams.ptr[0],
@@ -627,23 +628,23 @@ pub unsafe fn convert_lwe_programmable_bootstrap_key_async<T: UnsignedInteger>(
             cuda_convert_lwe_programmable_bootstrap_key_128_async(
                 stream_ptr,
                 streams.gpu_indexes[i].get(),
-                dest.as_mut_c_ptr(i as u32),
+                dest.as_mut_c_ptr(u32::try_from(i).unwrap()),
                 src.as_ptr().cast(),
-                input_lwe_dim.0 as u32,
-                glwe_dim.0 as u32,
-                l_gadget.0 as u32,
-                polynomial_size.0 as u32,
+                u32::try_from(input_lwe_dim.0).unwrap(),
+                u32::try_from(glwe_dim.0).unwrap(),
+                u32::try_from(l_gadget.0).unwrap(),
+                u32::try_from(polynomial_size.0).unwrap(),
             );
         } else if size_of::<T>() == 8 {
             cuda_convert_lwe_programmable_bootstrap_key_64_async(
                 stream_ptr,
                 streams.gpu_indexes[i].get(),
-                dest.as_mut_c_ptr(i as u32),
+                dest.as_mut_c_ptr(u32::try_from(i).unwrap()),
                 src.as_ptr().cast(),
-                input_lwe_dim.0 as u32,
-                glwe_dim.0 as u32,
-                l_gadget.0 as u32,
-                polynomial_size.0 as u32,
+                u32::try_from(input_lwe_dim.0).unwrap(),
+                u32::try_from(glwe_dim.0).unwrap(),
+                u32::try_from(l_gadget.0).unwrap(),
+                u32::try_from(polynomial_size.0).unwrap(),
             );
         } else {
             panic!("Unsupported torus size for bsk conversion")
@@ -676,25 +677,25 @@ pub unsafe fn convert_lwe_multi_bit_programmable_bootstrap_key_async<T: Any + Un
             cuda_convert_lwe_multi_bit_programmable_bootstrap_key_128_async(
                 stream_ptr,
                 streams.gpu_indexes[i].get(),
-                dest.as_mut_c_ptr(i as u32),
+                dest.as_mut_c_ptr(u32::try_from(i).unwrap()),
                 src.as_ptr().cast(),
-                input_lwe_dim.0 as u32,
-                glwe_dim.0 as u32,
-                l_gadget.0 as u32,
-                polynomial_size.0 as u32,
-                grouping_factor.0 as u32,
+                u32::try_from(input_lwe_dim.0).unwrap(),
+                u32::try_from(glwe_dim.0).unwrap(),
+                u32::try_from(l_gadget.0).unwrap(),
+                u32::try_from(polynomial_size.0).unwrap(),
+                u32::try_from(grouping_factor.0).unwrap(),
             );
         } else if TypeId::of::<T>() == TypeId::of::<u64>() {
             cuda_convert_lwe_multi_bit_programmable_bootstrap_key_64_async(
                 stream_ptr,
                 streams.gpu_indexes[i].get(),
-                dest.as_mut_c_ptr(i as u32),
+                dest.as_mut_c_ptr(u32::try_from(i).unwrap()),
                 src.as_ptr().cast(),
-                input_lwe_dim.0 as u32,
-                glwe_dim.0 as u32,
-                l_gadget.0 as u32,
-                polynomial_size.0 as u32,
-                grouping_factor.0 as u32,
+                u32::try_from(input_lwe_dim.0).unwrap(),
+                u32::try_from(glwe_dim.0).unwrap(),
+                u32::try_from(l_gadget.0).unwrap(),
+                u32::try_from(polynomial_size.0).unwrap(),
+                u32::try_from(grouping_factor.0).unwrap(),
             );
         } else {
             panic!("Unsupported torus size for bsk conversion")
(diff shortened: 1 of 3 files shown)


.github/workflows/gpu_pcc.yml
```diff
@@ -130,6 +130,13 @@ jobs:
         run: |
           make pcc_gpu
 
+      - name: Run semgrep and lint checks on CUDA code
+        run: |
+          sudo apt-get update && sudo apt-get install -y python3-venv
```
agnesLeroy: You could perform the installation steps in the step named "Intall Cuda" and rename that step into "Install dependencies"

scripts/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
+#!/usr/bin/env python3
```
agnesLeroy, line 1: Does this file have to be located in this .semgrep directory? Could we place it in `scripts` instead and name it `check_cuda_scratch_cleanup.py` instead?

scripts/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+EXPECTED_CUDA_COUNT = 109
+EXPECTED_CLEANUP_COUNT = 66
+
+# Check 2: Rust call-site scanning
+EXPECTED_CHECK2_RUST_FNS = 133
+EXPECTED_CHECK2_ASYNC_CUDA_CALLS = 89
+EXPECTED_CHECK2_SCRATCH_CALLS = 109
+
+# Check 5: Rust async-caller scanning
+EXPECTED_CHECK5_ASYNC_CALLERS = 117
```
agnesLeroy, line 76: This block seems to be duplicated with lines 44-63

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+
+CPP_DIRS = [
+    "backends/tfhe-cuda-backend/cuda/src",
+    "backends/tfhe-cuda-backend/cuda/include",
+]
+
+CPP_EXTENSIONS = {".cu", ".cuh", ".h", ".cpp"}
+
+RUST_CALL_SITES = [
+    "tfhe/src/core_crypto/gpu/mod.rs",
```
agnesLeroy: Ideally we could refactor core_crypto/gpu to have an ffi.rs module that contains all the binding as well.

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+    "tfhe/src/integer/gpu/ffi.rs",
+]
+
+# Maximum number of extra tokens a cuda suffix may have compared to
+# a _common_ scratch/cleanup suffix for them to be considered matching.
+COMMON_MAX_EXTRA_TOKENS = 1
```
agnesLeroy: Shouldn't this be 0? You're trying to allow for some typo somehow?

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+
+    # Check 4
+    print("\nCheck 4: Non-async cuda functions call synchronize in C++...")
+    v, n_checked = check_4_sync_in_non_async_cpp(cuda_set)
+    all_violations.extend(v)
+    print(f"  Scanned {n_checked} non-async cuda functions")
+    print(f"  {'PASS' if not v else f'{len(v)} violation(s)'}")
+
+    # Check 5
+    print("\nCheck 5: Rust _async callers synchronize...")
```
agnesLeroy: It would be good to refactor core crypto bindings to be all sync or call a sync cleanup, so we could remove this check and stop having to call synchronize in core crypto.

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+
+    # Check 1
+    print("\nCheck 1: No cleanup_ function has _async suffix...")
+    v, n_cleanup = check_1_no_async_cleanup(cleanup_set)
+    all_violations.extend(v)
+    print(f"  Scanned {n_cleanup} cleanup functions")
+    print(f"  {'PASS' if not v else f'{len(v)} violation(s)'}")
+
+    # Check 3 (run before check 2 since it's about naming)
+    print("\nCheck 3: Triplet scratch/cuda have _async suffix...")
```
agnesLeroy: Check 3 should be renamed Check 2 and Check 2 -> Check 3 for better consistency

Makefile
```diff
@@ -279,6 +279,14 @@ fmt_js: check_nvm_installed
 	$(MAKE) -C tfhe/web_wasm_parallel_tests fmt && \
 	$(MAKE) -C tfhe/js_on_wasm_tests fmt
 
+.PHONY: semgrep_and_lint_gpu_code # Run semgrep and lint checks on CUDA backend code
+semgrep_and_lint_gpu_code:
```
IceTDrinker: as an FYI this makefile rule does not setup any python environment, so if you want it to execute locally you would have to run it yourself, or you can add your python environment setup as a dependency

scripts/check_scratch_cleanup.py
IceTDrinker: also what's check 2 and check 5 ?

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
```diff
@@ -0,0 +1,933 @@
...
+def extract_functions(source):
+    """Extract (fn_name, line_number, body_text) for each function."""
+    functions = []
+    lines = source.split("\n")
+    i = 0
+    while i < len(lines):
+        match = re.match(
+            r"^(\s*)(pub(\(crate\))?\s+)?(unsafe\s+)?fn\s+(\w+)", lines[i]
+        )
+        if match:
```
IceTDrinker: guessing this particular function is rust focused, if you are looking for code able to parse rust code I would recommend the syn crate https://docs.rs/syn/latest/syn/

for cpp I believe there is a clang lib that allows to do something similar

I haven't read in detail what you are trying to achieve, but if it's a linter/code analysis for rust you can either use syn or dylint https://github.com/trailofbits/dylint

Nicolas has experience writing custom linters and will be much less error prone for code analysis

for c++ clang tidy would be the way to go https://clang.llvm.org/extra/clang-tidy/

The reason I mention them both is that the "custom python script with regex" is not something I've seen scaling and the complexity of the languages make them somewhat irrelevant/impossible to maintain quickly

IceTDrinker (commented):
I don't know semgrep so can't comment on that

for the python script that would lint cpp/rust code I would strongly recommend going for existing tooling to write lints or parse code

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
andrei-stoian-zama: you're definitely right! I felt this whole python custom parser was quite brittle. I'll check out those libraries

Makefile
andrei-stoian-zama: I'll add a check for the python environment and if it doesn't exist the makefile will create it

scripts/check_scratch_cleanup.py
andrei-stoian-zama: I moved it

scripts/check_scratch_cleanup.py
andrei-stoian-zama: I added more comments. it's explained at the top of the file

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
andrei-stoian-zama: I duplicated scratch/cleanup to avoid using the "common" token

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
andrei-stoian-zama: I only need to parse rust code, for the C++ bindings the format is constrained and a regex suffices.

For rust I changed to use `tree-sitters-rust` to find function bodies. then I use regexp to find binding calls. Using the AST to find the calls was quite complex, I think a regex suffices.

backends/tfhe-cuda-backend/cuda/.semgrep/check_scratch_cleanup.py
andrei-stoian-zama: Done, but I'll leave the disabling of the casting clippy check, as done in integer/ffi.rs to a new PR

Makefile
```diff
@@ -282,6 +282,22 @@ fmt_js: check_nvm_installed
...
+	@if [ ! -d .venv ]; then \
+		echo "Creating .venv virtual environment..."; \
+		python3 -m venv .venv; \
+	fi
+	@if ! .venv/bin/python3 -c "import semgrep, tree_sitter, tree_sitter_rust" 2>/dev/null; then \
+		echo "Installing Python dependencies..."; \
+		.venv/bin/pip install semgrep tree-sitter tree-sitter-rust; \
+	fi
```
IceTDrinker: could be a make target "setup_semgrep_venv" and have it done always, setting up a venv is not that costly to be honest

also with a semgrep-requirements.txt file dedicated for semgrep

andrei-stoian-zama: Good point! I changed it to have a separate target to build the venv

.github/workflows/gpu_pcc.yml
andrei-stoian-zama: Actually it can't be done since the Install CUDA step doesn't get executed for PCC, precisely where we need the semgrep step to run

.github/workflows/gpu_pcc.yml
```diff
@@ -130,6 +130,11 @@ jobs:
         run: |
           make pcc_gpu
 
+      - name: Run semgrep and lint checks on CUDA code
+        run: |
+          sudo apt update && sudo apt -y install python3-venv
```
agnesLeroy, line 135: This should move to the "Install" section

backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
```diff
@@ -58,9 +58,14 @@ template <typename Torus> struct int_compression {
                                        gpu_memory_allocated);
     tmp_glwe_array_out = nullptr;
 
-    cleanup_packing_keyswitch_lwe_list_to_glwe(
-        streams.stream(0), streams.gpu_index(0), &fp_ks_buffer,
-        gpu_memory_allocated);
+    if constexpr (sizeof(Torus) == 8)
+      cleanup_cuda_packing_keyswitch_lwe_list_to_glwe_64(
```
agnesLeroy, line 64: Here we should be calling a `release` function instead of a cleanup one, which is supposed to be only exposed to Rust.

agnesLeroy: Ok I see there is no `release`, `only cuda_drop_with_size_tracking`. We could maybe call `cuda_drop_with_size_tracking_async` instead of the cleanup.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cu
```diff
@@ -121,7 +121,7 @@ void cuda_add_and_propagate_single_carry_64_inplace(
       requested_flag, uses_carry);
 }
 
-void cuda_integer_overflowing_sub_64_inplace(
+void cuda_integer_overflowing_sub_64_inplace_async(
```
agnesLeroy, line 124: Now I'm thinking we should remove _64 from all these names, it's not useful to have it in the names for integer ops

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+use std::any::{Any, TypeId};
+use std::ffi::c_void;
+use tfhe_cuda_backend::bindings::*;
+use tfhe_cuda_backend::ffi;
+
+/// Programmable bootstrap on a vector of LWE ciphertexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
```
agnesLeroy: All the "Safety" comments in this file should be modified to
```
/// # Safety
///
/// - The data must not be moved or dropped while being used by the CUDA kernel.
/// - This function assumes exclusive access to the passed data; violating this may lead to
///   undefined behavior.
```

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+            keyswitch_key.as_c_ptr(0),
+            input_lwe_dimension.0 as u32,
+            output_lwe_dimension.0 as u32,
+            base_log.0 as u32,
+            l_gadget.0 as u32,
+            num_samples,
+        );
+    } else {
+        panic!("Unknown LWE KS dtype of size {}B", size_of::<KT>());
+    }
```
agnesLeroy, line 484: Why do we need to keep a keyswitch_async entry point? We could name it keyswitch and call synchronize here instead, no?

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+}
+/// Convert keyswitch key
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn convert_lwe_keyswitch_key_async<T: UnsignedInteger>(
```
agnesLeroy, line 493: Here also maybe we could synchronize in this wrapper function and not name it _async?

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    } else {
+        panic!("Unknown LWE GEMM KS dtype of size {}B", size_of::<KST>());
+    }
+}
+
+/// Keyswitch on a vector of LWE ciphertexts. Better for small batches of LWEs
+/// (up to 128 LWEs on H100, up to 64 on L40, up to 16 on 4090)
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
```
agnesLeroy, line 434: If we keep it _async this Safety section is ok

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+            l_gadget.0 as u32,
+            num_samples,
+        );
+    } else {
+        panic!("Unknown LWE KS dtype of size {}B", size_of::<KT>());
+    }
+}
+/// Convert keyswitch key
+///
+/// # Safety
```
agnesLeroy, line 488: If we keep it _async this Safety section is ok

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+        streams.ptr[0],
+        streams.gpu_indexes[0].get(),
+        std::ptr::addr_of_mut!(fp_ks_buffer),
+        true,
+    );
+}
+
+/// Convert programmable bootstrap key
+///
+/// # Safety
```
agnesLeroy, line 612: If we keep it _async this Safety section is ok

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+                dest.as_mut_c_ptr(i as u32),
+                src.as_ptr().cast(),
+                input_lwe_dim.0 as u32,
+                glwe_dim.0 as u32,
+                l_gadget.0 as u32,
+                polynomial_size.0 as u32,
+            );
+        } else {
+            panic!("Unsupported torus size for bsk conversion")
+        }
```
agnesLeroy, line 651: Same question here, we could have a non _async wrapper and call synchronize here instead.

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+}
+
+/// Convert multi-bit programmable bootstrap key
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn convert_lwe_multi_bit_programmable_bootstrap_key_async<T: Any + UnsignedInteger>(
```
agnesLeroy, line 662: Same remarks for this function

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+        }
+    }
+}
+
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn extract_lwe_samples_from_glwe_ciphertext_list_async<T: UnsignedInteger>(
```
agnesLeroy, line 711: This entry point doesn't need to be async, we should synchronize before returning and remove _async from the name

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+        panic!("Unsupported torus size for glwe sample extraction")
+    }
+}
+
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn cuda_modulus_switch_ciphertext_async<T: UnsignedInteger>(
```
agnesLeroy, line 758: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    streams.synchronize();
+}
+
+/// Addition of a vector of LWE ciphertexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn add_lwe_ciphertext_vector_async<T: UnsignedInteger>(
```
agnesLeroy, line 792: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Assigned addition of a vector of LWE ciphertexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn add_lwe_ciphertext_vector_assign_async<T: UnsignedInteger>(
```
agnesLeroy, line 846: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Addition of a vector of LWE ciphertexts with a vector of plaintexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn add_lwe_ciphertext_vector_plaintext_vector_async<T: UnsignedInteger>(
```
agnesLeroy, line 889: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Addition of a vector of LWE ciphertexts with a plaintext scalar
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn add_lwe_ciphertext_vector_plaintext_scalar_async<T: UnsignedInteger>(
```
agnesLeroy, line 915: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Assigned addition of a vector of LWE ciphertexts with a vector of plaintexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn add_lwe_ciphertext_vector_plaintext_vector_assign_async<T: UnsignedInteger>(
```
agnesLeroy, line 941: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Negation of a vector of LWE ciphertexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn negate_lwe_ciphertext_vector_async<T: UnsignedInteger>(
```
agnesLeroy, line 966: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Assigned negation of a vector of LWE ciphertexts
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn negate_lwe_ciphertext_vector_assign_async<T: UnsignedInteger>(
```
agnesLeroy, line 990: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+    );
+}
+
+/// Multiplication of a vector of LWEs with a vector of cleartexts (assigned)
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+pub unsafe fn mult_lwe_ciphertext_vector_cleartext_vector_assign_async<T: UnsignedInteger>(
```
agnesLeroy, line 1013: Same here

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+}
+
+/// forward fourier transform for complex f128 as integer
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn fourier_transform_forward_as_integer_f128_async<T: UnsignedInteger>(
```
agnesLeroy, line 1066: Same here, no need to have it _async let's make it a sync entry point

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+}
+
+/// forward fourier transform for complex f128 as torus
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn fourier_transform_forward_as_torus_f128_async<T: UnsignedInteger>(
```
agnesLeroy, line 1097: Here as well

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1170 @@
...
+}
+
+/// backward fourier transform for complex f128 as torus
+///
+/// # Safety
+///
+/// [CudaStreams::synchronize] __must__ be called as soon as synchronization is
+/// required
+#[allow(clippy::too_many_arguments)]
+pub unsafe fn fourier_transform_backward_as_torus_f128_async<T: UnsignedInteger>(
```
agnesLeroy, line 1128: Here as well

tfhe/src/integer/gpu/server_key/radix/tests_noise_distribution/utils/noise_simulation.rs
agnesLeroy, line 411: The entry point won't be async anyore so you can delete this streams.synchronize()

backends/tfhe-cuda-backend/cuda/src/integer/integer.cu
agnesLeroy: (We should keep it explicit only for core_crypto)

.github/workflows/gpu_pcc.yml
andrei-stoian-zama: we can't, because the install CUDA step doesn't run in PCC

backends/tfhe-cuda-backend/cuda/src/integer/integer.cu
andrei-stoian-zama: indeed, if we will only support 64 native modulus we can remove this. but maybe in a new PR ?

backends/tfhe-cuda-backend/cuda/include/integer/compression/compression_utilities.h
andrei-stoian-zama: There are a about 15 of such bindings that are called in the C++ backend. they can all be removed by replacing them with calls to the `host_` function with strict typing and by adding a layer of indirection in the cleanup and calling the templated host scratch in the scratch binding.  

I suggest we leave this for a second PR

scripts/gpu-lint-requirements.txt
```diff
@@ -0,0 +1,3 @@
+semgrep
+tree-sitter
+tree-sitter-rust
```
IceTDrinker: maybe specify a major+minor version and ask for semver compatible version, IIRC ^X.Y or something in python

otherwise you could have random breakage in the CI if they break an API

Makefile
```diff
@@ -282,6 +282,22 @@ fmt_js: check_nvm_installed
 	$(MAKE) -C tfhe/web_wasm_parallel_tests fmt && \
 	$(MAKE) -C tfhe/js_on_wasm_tests fmt
 
+.PHONY: semgrep_lint_setup_venv # Create venv and install Python dependencies for GPU lint checks
+semgrep_lint_setup_venv:
+	@if [ ! -d .venv ]; then \
```
IceTDrinker: still think that check is not necessary, also if the venv exists already it might not install the deps (which can happen)

Just saw also that we already "expect" a venv at location venv and not .venv according to the gitignore file, so maybe use the venv dir ?

IceTDrinker (commented):
small nits on requirements and venv management

scripts/gpu-lint-requirements.txt
andrei-stoian-zama: good idea! done

Makefile
andrei-stoian-zama: yes! changed it. though the "requirement already satistifed" prints are quite annoying. but best to re-run the install every time in case dependency versions change

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -0,0 +1,1177 @@
...
+    let noise_reduction_type = ms_noise_reduction_configuration
+        .map_or(PBSMSNoiseReductionType::NoReduction, |_config| {
+            PBSMSNoiseReductionType::Centered
+        });
+
+    scratch_cuda_programmable_bootstrap_64_async(
+        streams.ptr[0],
+        streams.gpu_indexes[0].get(),
+        std::ptr::addr_of_mut!(pbs_buffer),
+        lwe_dimension.0 as u32,
```
agnesLeroy: I had fixed all these `as u32` on main and disabled casting clippy checks on all these functions, could you please do this in this PR? This is a regression compared to the state of main.

andrei-stoian-zama: You're right, by moving the code to the new file I introduced a regression. I updated the core crypto ffi.rs with fixes + forced clippy to check.

andrei-stoian-zama: the changes are in a second commit so it's easier to check it

tfhe/src/core_crypto/gpu/ffi.rs
andrei-stoian-zama: we keep it async for new,  I updated the safety section on functions that were transformed into synchornized ones

tfhe/src/core_crypto/gpu/ffi.rs
```diff
@@ -1,3 +1,4 @@
+#![deny(clippy::cast_possible_truncation)]
```
agnesLeroy, line 1: You can remove this deny from all files where I had added it in core_crypto/gpu/algorithms/*.rs

agnesLeroy: Missing this one 🙂

andrei-stoian-zama: Isn't it better to leave it ? I'm not sure bindings are only called in ffi.rs. Also ideally it would be disabled in more places no ?

we still call bindings in these files which need the clippy stricter checking:

```
 File: tfhe/src/core_crypto/gpu/algorithms/glwe_linear_algebra.rs                                                                              
  Bindings called: scratch_cuda_wrapping_polynomial_mul_one_to_many_64_async (×2), cuda_wrapping_polynomial_mul_one_to_many_64_async,           
    cuda_glwe_wrapping_polynomial_mul_one_to_many_64_async, cleanup_cuda_wrapping_polynomial_mul_one_to_many_64 (×2)                            
  ────────────────────────────────────────                                                                                                      
  File: tfhe/src/core_crypto/gpu/algorithms/lwe_keyswitch.rs                                                                                    
  Bindings called: cuda_closest_representable_64_async                                                                                          
  ````
  
  and test these files call bindings, but since it's only tests they can live without stricter clippy checks
  ```
  ────────────────────────────────────────                                                                                                      
  File: tfhe/src/core_crypto/gpu/algorithms/test/modulus_switch.rs                                                                              
  Bindings called: cuda_modulus_switch_64_async, cuda_centered_modulus_switch_64_async (×2)                                                     
  ────────────────────────────────────────                                                                                                      
  File: tfhe/src/integer/gpu/server_key/radix/tests_noise_distribution/utils/noise_simulation.rs
  Bindings called: cuda_centered_modulus_switch_64_async         
```

agnesLeroy: Hmm it's a bit strange to not have all bindings in ffi.rs, we should probably move all of them there

agnesLeroy: Why didn't you move all of them? Is it because otherwise the file organization that we had doesn't make sense anymore?

agnesLeroy: Actually if we move all bindings to the ffi.rs file, it means we would remove the files that existed before altogether, we lose the file structure of core crypto that we had up till now. To me it's ok, but maybe @pdroalves has another point of view.

backends/tfhe-cuda-backend/cuda/src/aes/aes.cu
```diff
@@ -21,22 +21,52 @@ uint64_t scratch_cuda_integer_aes_encrypt_64(
...
+    PBS_MS_REDUCTION_T noise_reduction_type, uint32_t num_aes_inputs,
+    uint32_t sbox_parallelism) {
+
+  int_radix_params params(pbs_type, glwe_dimension, polynomial_size,
+                          glwe_dimension * polynomial_size, lwe_dimension,
+                          ks_level, ks_base_log, pbs_level, pbs_base_log,
+                          grouping_factor, message_modulus, carry_modulus,
+                          noise_reduction_type);
+
+  return scratch_cuda_integer_aes_encrypt<uint64_t>(
```
pdroalves, line 39: Shouldn't it be `scratch_cuda_integer_aes_encrypt_async`?

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cu
```diff
@@ -105,32 +105,51 @@ uint64_t scratch_packing_keyswitch_lwe_list_to_glwe_64(
       glwe_dimension, polynomial_size, num_lwes, allocate_gpu_memory);
 }
 
-uint64_t scratch_cuda_keyswitch_gemm_64(void *stream, uint32_t gpu_index,
-                                        void **ks_tmp_buffer,
-                                        uint32_t lwe_dimension_in,
-                                        uint32_t lwe_dimension_out,
-                                        uint32_t num_lwes,
-                                        bool allocate_gpu_memory) {
+uint64_t scratch_cuda_keyswitch_gemm_64_64_async(
```
pdroalves, line 108: `_64_64`?

pdroalves: I see many cases like this in this file. Is that correct?

pdroalves: Maybe used to define input and output types, but I just want to check.

andrei-stoian-zama: the first _64 is the input modulus size, the second one is the output modulus size. the First one will be removed in all functions in a 2nd PR

backends/tfhe-cuda-backend/cuda/src/aes/aes.cu
andrei-stoian-zama: Internal functions to the backend do not necessarily have the _async prefix.

tfhe/src/core_crypto/gpu/ffi.rs
pdroalves: I'm also not totally comfortable with having bindings spread through multiple files, but I think the file structure is important so we can easily find things per context.

IceTDrinker: You can have an FFI file and re-exports in the different files

tfhe/src/integer/gpu/server_key/radix/tests_noise_distribution/utils/noise_simulation.rs
pdroalves: @andrei-stoian-zama Did you see this one?

andrei-stoian-zama: Yes. In this PR we don't ensure we have non-duplicate synchronize.

e.g. 

```
fn a(streams) {
  cuda_op_binding_async();
  streams.synchronize()
}

fn b(streams) { 
  a(streams);
  streams.synchronize()
}
```

this is not detected in this PR. It may be done in a second PR, but this PR already implements many other changes and checks

PR #3296, opened 2026-02-11
chore: add wasm_par_mq crate for unsafe coop wasm parallelism

nsarlin-zama:
this is a first step to implement a proper solution for this issue: https://github.com/zama-ai/tfhe-rs-internal/issues/1232

#### Context
For more context, see the linked issue. The wasm encryption and zk proof currently require specific server side headers to be parallelized ([COOP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Opener-Policy) / [COEP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy)). The issue is that these headers put some restriction on the contexts in which the crate can be used.

This is caused by [wasm-bindgen-rayon](https://github.com/RReverser/wasm-bindgen-rayon?tab=readme-ov-file#setting-up), more specifically by the use of [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer). The goal of this PR is to introduce a way to have wasm parallelism without SharedArrayBuffer.

This pr introduces a new crate, `wasm-par-mq`, that provides a parallel api without SharedArrayBuffer. This pr is only for this crate, the actual usage in `tfhe-zk-pok` will come in a following pr.

#### How it works
The main idea is that we can still use web workers, but are limited in the way we are allowed to communicate between them and the main thread, since without SharedArrayBuffer they cannot share memory. But this is just a limitation about how workers can communicate, not about what they can do.

To circumvent this, the solution used here is to communicate with [message queues](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage). The task input is split into chunks, and each chunk is serialized and sent to an available worker.

#### Sync mode
By default, the API is async since the main thread cannot be blocked in js. This means that code using this crate must be async too. The crate also supports a sync mode, that is a bit more complex. The trick is to execute the user code in a parent worker, called the `SyncExecutor`, that is allowed to block. To block until the compute workers are done, we use a [Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) called the `Coordinator`. We do a blocking request from the SyncExecutor to the Coordinator that only returns when all the work has completed. Since it adds a level of indirection, the code that is sent to the SyncExecutor should ideally be as large as possible, with small inputs, to avoid costly back and forth.

For more information, see the README of the new crate: https://github.com/zama-ai/tfhe-rs/tree/ns/chore/wasm_par_mq/utils/wasm_par_mq
You can see an example of it being used in a slightly realistic situation here: https://github.com/zama-ai/tfhe-rs/tree/ns/chore/wasm_par_mq/utils/wasm_par_mq/examples/msm/src

### AI usage
AI (claude code) was used to:
- help me design this and answer my questions about wasm/js/browser world
- add tests, doc and example

### Check-list:

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296)

### DIFF
--- a/.github/workflows/aws_tfhe_wasm_tests.yml
+++ b/.github/workflows/aws_tfhe_wasm_tests.yml
@@ -117,6 +117,8 @@ jobs:
         run: |
           make install_chrome_browser
           make install_chrome_web_driver
+          make install_firefox_browser
+          make install_firefox_web_driver
 
       - name: Run fmt checks
         run: |
@@ -129,6 +131,11 @@ jobs:
       - name: Run parallel wasm tests
         run: |
           make test_web_js_api_parallel_chrome_ci
+
+      - name: Run wasm_par_mq tests
+        run: |
+          make test_wasm_par_mq_chrome_ci
+          make test_wasm_par_mq_firefox_ci
 
       - name: Run x86_64/wasm zk compatibility tests
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ venv/
 web-test-runner/
 node_modules/
 package-lock.json
+utils/wasm-par-mq/examples/*/pkg/
 
 # Python .env
 .env
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ members = [
     "utils/tfhe-backward-compat-data",
     "utils/tfhe-backward-compat-data/crates/add_new_version",
     "utils/param_dedup",
+    "utils/wasm-par-mq",
+    "utils/wasm-par-mq/web_tests",
+    "utils/wasm-par-mq/examples/msm",
     "tests",
     "mockups/tfhe-hpu-mockup",
     "apps/test-vectors",
@@ -46,6 +49,9 @@ bincode = "=1.3.3"
 cmake = "0.1"
 pkg-config = "0.3"
 clap = { version = "4.5", features = ["derive"] }
+js-sys = "0.3"
+serde-wasm-bindgen = "0.6.5"
+paste = "1.0.15"
 
 [profile.bench]
 lto = "fat"
--- a/Makefile
+++ b/Makefile
@@ -534,6 +534,11 @@ clippy_param_dedup: install_rs_check_toolchain
 	RUSTFLAGS="$(RUSTFLAGS)" cargo "$(CARGO_RS_CHECK_TOOLCHAIN)" clippy --all-targets \
 		-p param_dedup -- --no-deps -D warnings
 
+.PHONY: clippy_wasm_par_mq # Run clippy lints on wasm-par-mq and its examples
+clippy_wasm_par_mq: install_rs_check_toolchain
+	RUSTFLAGS="$(RUSTFLAGS)" cargo "$(CARGO_RS_CHECK_TOOLCHAIN)" clippy --all-targets --all-features \
+		-p wasm-par-mq -p wasm-par-mq-web-tests -p wasm-par-mq-example-msm -- --no-deps -D warnings
+
 .PHONY: clippy_backward_compat_data # Run clippy lints on tfhe-backward-compat-data
 clippy_backward_compat_data: install_rs_check_toolchain # the toolchain is selected with toolchain.toml
 	@# Some old crates are x86 specific, only run in that case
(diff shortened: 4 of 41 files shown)


nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296)* status: 22 of 41 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, soonum, SouchonTheo, and tmontaigu).

___
*[`utils/wasm-par-mq/README.md` line 183 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OlvlpC05GL0p0IfTsWC:-Om95dmKBwspM_cHFWte:b2nign6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e7ad1132b2cfe6bd17912669db9a2e66cc05ba31/utils/wasm-par-mq/README.md#L183)):*

so the above example is called from the main thread ?
</blockquote></details>

yes
___
*[`utils/wasm-par-mq/src/pool.rs` line 434 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OmEAPTEEwNSLP3cxx_q:-OmEDjr4800kEMlapq0y:bj4olhz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e7ad1132b2cfe6bd17912669db9a2e66cc05ba31/utils/wasm-par-mq/src/pool.rs#L434)):*

I saw a bunch of `Clone` requirements in the ParallelSlice trait, maybe because of this line,

Isn't it possible to avoid this clone by making the chunks being `Vec<&[I]>` since they are going to be serialized into Vec<u8> anyway
</blockquote></details>

ah yes excellent! Fixed. Actually I don't even need to collect

IceTDrinker (commented):
sure I'll try to do that

> @IceTDrinker made 4 comments and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-:-OmFI2ln-OecfUl4Vw_h:b-m6zzu0)* status: 35 of 41 files reviewed, 6 unresolved discussions (waiting on nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`utils/wasm-par-mq/src/pool.rs` line 32 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OmEOXjhBnaBGaqFzCMZ:-OmFHeNX5Wqv_yhisKed:b-brrlbr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e1e1e0c1da7517e67f1ee78bed18956093a39284/utils/wasm-par-mq/src/pool.rs#L32)):*

for the csp strict ? Actually I need to check this throughout the whole crate, I forgot about this constraint
</blockquote></details>

no mainly to avoid casting stuff

since wasm-bindgen has accepted the patch asking for global object is less of a problem
___
*[`utils/wasm-par-mq/src/pool.rs` line 98 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OmELs1pCaYwtDf71WPh:-OmFHmfk2KDSyaiv9i7p:b-2podae) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e1e1e0c1da7517e67f1ee78bed18956093a39284/utils/wasm-par-mq/src/pool.rs#L98)):*

len is usize, but usize is 32b in wasm
</blockquote></details>

yes, so the try into can never fail I think ?
___
*[`utils/wasm-par-mq/src/worker.rs` line 21 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OmEKjyC8nXdiM-8GOla:-OmFHt392zqeBiM2YKdr:b-4x7r4j) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e1e1e0c1da7517e67f1ee78bed18956093a39284/utils/wasm-par-mq/src/worker.rs#L21)):*

yes I will try to remove all the js_sys::global calls in the crate
</blockquote></details>

might not be absolutely necessary, since wasm-bindgen accepted the patch, but I believe they say it is good practice

IceTDrinker (commented):
> @IceTDrinker reviewed 5 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296)* status: all files reviewed, 18 unresolved discussions (waiting on nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`utils/wasm-par-mq/src/pool.rs` line 447 at r9](https://reviewable.io/reviews/zama-ai/tfhe-rs/3296#-OmTcitc2cTn6axz_dFY:-OmUWL6t018NC-UrctiZ:b-c716xf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/8f54f80596c3a077a259ba8ee96b81494aba8602/utils/wasm-par-mq/src/pool.rs#L447)):*

could be with_capacity(data.len()) maybe ? unsure
</blockquote></details>

think you update one but not this one

PR #3297, opened 2026-02-11
feat(gpu): implement specialized pbs for any architecture

guillermo-oyarzun:

### PR content/description
Specialized version of the classical pbs with 2_2 params for almost any architecture, only restriction is that shared memory in the device must be at least 80KB
### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/pbs/programmable_bootstrap_testing.h
+++ b/backends/tfhe-cuda-backend/cuda/include/pbs/programmable_bootstrap_testing.h
@@ -3,6 +3,12 @@
 
 #include "pbs_enums.h"
 #include <stdint.h>
+
+template <typename Torus>
+bool specialized_2_2_params_checker(uint32_t polynomial_size,
+                                    uint32_t glwe_dimension,
+                                    uint32_t level_count,
+                                    uint32_t max_shared_memory);
 uint64_t scratch_cuda_programmable_bootstrap_tbc_generic_64_async(
     void *stream, uint32_t gpu_index, int8_t **buffer, uint32_t lwe_dimension,
     uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t level_count,
@@ -25,6 +31,46 @@ void cuda_programmable_bootstrap_tbc_64_generic_async(
     uint32_t num_samples, uint32_t num_many_lut, uint32_t lut_stride);
 
 void cuda_programmable_bootstrap_tbc_64_2_2_async(
+    void *stream, uint32_t gpu_index, void *lwe_array_out,
+    void const *lwe_output_indexes, void const *lut_vector,
+    void const *lut_vector_indexes, void const *lwe_array_in,
+    void const *lwe_input_indexes, void const *bootstrapping_key,
+    int8_t *buffer, uint32_t lwe_dimension, uint32_t glwe_dimension,
+    uint32_t polynomial_size, uint32_t base_log, uint32_t level_count,
+    uint32_t num_samples, uint32_t num_many_lut, uint32_t lut_stride);
+
+// The cuda bootstrap key is converted from torus to fourier domain using
+// the automatic logic that checks if the parameters match the specialized
+// 2_2_params. The layout is different in the specialized because the
+// input/output data of the FFT lives in registers, and if the layout is not
+// changed, global memory reads of the bsk are not coalesced. In the backend
+// tests, we want to be able to use other pbs flavors that need the shared
+// memory FFT layouts, so we provide the following two functions to bypass the
+// check and force the desired layout.
+
+// i) Force standard (non-specialized) FFT layout for the BSK, this is needed
+// when using step1/step2 or cg pbs on a H100 GPU.
+void cuda_convert_lwe_programmable_bootstrap_key_standard_64_async(
+    void *stream, uint32_t gpu_index, void *dest, void const *src,
+    uint32_t input_lwe_dim, uint32_t glwe_dim, uint32_t level_count,
+    uint32_t polynomial_size);
+
+// Force specialized 2_2_params FFT layout for the BSK. This is needed when we
+// want to run the specialized non tbc flavor on a GPU that is not high-end (bad
+// FP64 ratio). On those GPUs, the automatic fft layout is the shared memory
+// one, and that is why we need to force the specialized one.
+void cuda_convert_lwe_programmable_bootstrap_key_specialized_2_2_64_async(
+    void *stream, uint32_t gpu_index, void *dest, void const *src,
+    uint32_t input_lwe_dim, uint32_t glwe_dim, uint32_t level_count,
+    uint32_t polynomial_size);
+
+uint64_t scratch_cuda_programmable_bootstrap_specialized_2_2_64_async(
+    void *stream, uint32_t gpu_index, int8_t **buffer, uint32_t lwe_dimension,
+    uint32_t glwe_dimension, uint32_t polynomial_size, uint32_t level_count,
+    uint32_t input_lwe_ciphertext_count, bool allocate_gpu_memory,
+    PBS_MS_REDUCTION_T noise_reduction_type);
+
+void cuda_programmable_bootstrap_specialized_2_2_64_async(
     void *stream, uint32_t gpu_index, void *lwe_array_out,
     void const *lwe_output_indexes, void const *lut_vector,
     void const *lut_vector_indexes, void const *lwe_array_in,
--- a/backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
@@ -188,7 +188,7 @@ __device__ void NSMFFT_direct_2_2_params(double2 *A, double2 *fft_out,
   }
 
   Index twiddle_shift = 1;
-  for (Index l = LOG2_DEGREE - 1; l >= 5; --l) {
+  for (Index l = LOG2_DEGREE - 1; l > 5; --l) {
     Index lane_mask = 1 << (l - 1);
     Index thread_mask = (1 << l) - 1;
     twiddle_shift <<= 1;
@@ -221,8 +221,8 @@ __device__ void NSMFFT_direct_2_2_params(double2 *A, double2 *fft_out,
       tid = tid + STRIDE;
     }
   }
-
-  for (Index l = 4; l >= 1; --l) {
+  __syncthreads();
+  for (Index l = 5; l >= 1; --l) {
     Index lane_mask = 1 << (l - 1);
     Index thread_mask = (1 << l) - 1;
     twiddle_shift <<= 1;
@@ -425,7 +425,7 @@ __device__ void NSMFFT_inverse_2_2_params(double2 *A, double2 *buffer_regs,
   }
 
   Index twiddle_shift = DEGREE;
-  for (Index l = 1; l <= 4; ++l) {
+  for (Index l = 1; l <= 5; ++l) {
     Index lane_mask = 1 << (l - 1);
     Index thread_mask = (1 << l) - 1;
     tid = threadIdx.x;
@@ -459,7 +459,7 @@ __device__ void NSMFFT_inverse_2_2_params(double2 *A, double2 *buffer_regs,
     }
   }
 
-  for (Index l = 5; l <= LOG2_DEGREE - 1; ++l) {
+  for (Index l = 6; l <= LOG2_DEGREE - 1; ++l) {
     Index lane_mask = 1 << (l - 1);
     Index thread_mask = (1 << l) - 1;
     tid = threadIdx.x;
@@ -467,7 +467,7 @@ __device__ void NSMFFT_inverse_2_2_params(double2 *A, double2 *buffer_regs,
 
     // at this point registers are ready for the  butterfly
     tid = threadIdx.x;
-    __syncthreads();
+
 #pragma unroll
     for (Index i = 0; i < BUTTERFLY_DEPTH; ++i) {
       w = (u[i] - v[i]);
@@ -495,6 +495,7 @@ __device__ void NSMFFT_inverse_2_2_params(double2 *A, double2 *buffer_regs,
 
       tid = tid + STRIDE;
     }
+    __syncthreads();
   }
 
 // last iteration
@@ -511,7 +512,6 @@ __device__ void NSMFFT_inverse_2_2_params(double2 *A, double2 *buffer_regs,
 /*
  * global batch fft
  * does fft in half size
- * unrolling half size fft result in half size + 1 elements
  * this function must be called with actual degree
  * function takes as input already compressed input
  */
@@ -536,6 +536,46 @@ __global__ void batch_NSMFFT(double2 *d_input, double2 *d_output,
 #pragma unroll
   for (int i = 0; i < params::opt / 2; i++) {
     d_output[blockIdx.x * (params::degree / 2) + tid] = fft[tid];
+    tid = tid + params::degree / params::opt;
+  }
+}
+
+/*
+ * global batch fft
+ * does fft in half size
+ * this function must be called with actual degree
+ * function takes as input already compressed input
+ * this function is optimized for 2_2 params, it uses shared memory for
+ * twiddles, and for the fft exchanges between threads. Only 32768 bytes are
+ * needed, and we know in the 2_2 params cases we will always have enough shared
+ * memory.
+ */
+template <class params>
+__global__ void batch_NSMFFT_classical_specialized(double2 *d_input,
+                                                   double2 *d_output) {
+  extern __shared__ double2 sharedMemoryFFT[];
+  // For specialized we will always have enough shared memory
+  double2 *fft = sharedMemoryFFT;
+  int tid = threadIdx.x;
+
+  double2 *shared_twiddles = fft + params::degree / 2;
+
+  double2 fft_regs[params::opt / 2];
+#pragma unroll
+  for (int i = 0; i < params::opt / 2; i++) {
+    shared_twiddles[tid] = negtwiddles[tid];
+    fft_regs[i] = d_input[blockIdx.x * (params::degree / 2) + tid];
+    tid = tid + params::degree / params::opt;
+  }
+  __syncthreads();
+
+  NSMFFT_direct_2_2_params<HalfDegree<params>>(fft, fft_regs, shared_twiddles);
+  __syncthreads();
+
+  tid = threadIdx.x;
+#pragma unroll
+  for (int i = 0; i < params::opt / 2; i++) {
+    d_output[blockIdx.x * (params::degree / 2) + tid] = fft_regs[i];
     tid = tid + params::degree / params::opt;
   }
 }
--- a/backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
@@ -129,7 +129,8 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
                                                  uint32_t gpu_index,
                                                  double2 *dest, ST const *src,
                                                  uint32_t polynomial_size,
-                                                 uint32_t total_polynomials) {
+                                                 uint32_t total_polynomials,
+                                                 bool use_specialized_fft_2_2) {
   cuda_set_device(gpu_index);
   int shared_memory_size = safe_mul_sizeof<double>(polynomial_size);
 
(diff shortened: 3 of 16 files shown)


backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -271,7 +269,7 @@ uint64_t scratch_cuda_programmable_bootstrap(
         glwe_dimension, polynomial_size, level_count,
         input_lwe_ciphertext_count, allocate_gpu_memory, noise_reduction_type);
   case 2048:
-    return scratch_programmable_bootstrap<Torus, AmortizedDegree<2048>>(
+    return scratch_programmable_bootstrap<Torus, Degree<2048>>(
```
agnesLeroy, line 351: Here you change AmortizedDegree to Degree -> the specialised case gets better but not the "default" one, is that so? It would be worth a comment at least to justify the choice.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -6,6 +6,15 @@
 #include <cuda_runtime.h>
 #endif
 
+// This macro is needed because in debug mode the compiler doesn't apply all
+// optimizations
+//  and the register count is higher, which can lead to launch bounds conflicts.
+#ifdef __CUDACC_DEBUG__
+#define SPECIALIZED_2_2_PARAMS_LAUNCH_BOUNDS
+#else
+#define SPECIALIZED_2_2_PARAMS_LAUNCH_BOUNDS __launch_bounds__(1024)
```
agnesLeroy, line 15: Why in the debug case launch bounds 1024 doesn't work?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -17,6 +26,183 @@
...
+              &next_lwe_array_out[lwe_output_indexes[blockIdx.x] *
+                                      (glwe_dimension * polynomial_size + 1) +
+                                  threadIdx.y * polynomial_size];
+
+          sample_extract_mask<Torus, params>(next_block_lwe_array_out,
+                                             accumulator, 1, i * lut_stride);
+        }
+      }
+    } else if (threadIdx.y == glwe_dimension) {
+      sample_extract_body<Torus, params>(block_lwe_array_out, accumulator, 0);
```
agnesLeroy, line 205: Needs a __syncthreads before?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -305,6 +325,29 @@ __device__ void add_to_torus_2_2_params_using_shared(double2 *m_values,
...
 }
 
+/**
+ * In case of 2_2_classical PBS, this method should accumulate the result.
+ * and the result is stored in shared memory
+ */
+template <typename Torus, class params>
+__device__ void add_to_torus_2_2_params_using_regs(double2 *m_values,
+                                                   Torus *result) {
+  // int tid = threadIdx.x;
```
agnesLeroy: Stale comment

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
```diff
@@ -188,7 +188,7 @@ __device__ void NSMFFT_direct_2_2_params(double2 *A, double2 *fft_out,
   }
 
   Index twiddle_shift = 1;
-  for (Index l = LOG2_DEGREE - 1; l >= 5; --l) {
+  for (Index l = LOG2_DEGREE - 1; l > 5; --l) {
```
agnesLeroy, line 191: How come you can change the index of the layer from which we change logic? Is this correct?

agnesLeroy (commented):
Some initial remarks:
- base_log doesn't have to be part of the criteria to select the pbs flavor, does it? I think it's fine if it's only N,l,k
- have you checked the effect of FFT changes on the TBC specialized performance?
- about `mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_1block`, if the previous entry point is deleted we could just name it `mul_ggsw_glwe_in_fourier_domain_2_2_params_classical` as it was before? SHouldn't we place it in polynomial_math.cuh?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
guillermo-oyarzun: right i can add a comment, probably we should try and change that to Degree in all the cases possible, but that would be other PR (and not sure how to test it since we dont try those configurations often)

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: I found this by mistake because i compiled in debug. So the problem is that with 1024 threads we can only have 64 regs per thread due to hw limitations. The thing is that in debug not all optimizations are done, and currently that kernel is that the limit, for some reason in debug doesn't spill the registers so it create versions with 67-70 that fail due to hw.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: the sync is in line 161

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
agnesLeroy: Very strange, but well, we discover new behaviors of the compiler every day :sweat_smile:

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
agnesLeroy: Ah indeed, I had missed it. In that case you could add a comment like Beka did to just say syncthreads is already done above and that's why it's not needed here

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
guillermo-oyarzun: yup it is correct, actually we can do 5 levels because that is 2⁵ that is a warp. I didn't do it last time cause registers were growing much, but now seems to be fine, we get rid of 2 synchronization with this (and we need to add one)

guillermo-oyarzun:
> > * First i had it without the base_log, but if for some reason we have other set of parameters that only differ in base_log, the condition will be true, but the results won't be good. We could fix this but adding base_log as a template instead hardcoding it like now, it would be more resilient, and probably we could do it for the other specialized versions too.
> 
> What do you mean won't be good? Performance is affected or correctness?

i meant correctness cause the base_log is hardcoded in the kernel, that's why i was suggesting using a template for it

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: i can place the sync in the same position than Beka did ( i had done this changes before he introduced the fix)

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: dark magic indeed

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
agnesLeroy: Ah yes would be better then :+1:

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -128,7 +128,8 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
                                                  uint32_t gpu_index,
                                                  double2 *dest, ST const *src,
                                                  uint32_t polynomial_size,
-                                                 uint32_t total_polynomials) {
+                                                 uint32_t total_polynomials,
+                                                 bool use_specialized_fft_2_2) {
```
agnesLeroy, line 133: It would maybe be good to sync with @bbarbakadze to see how this would fit with what he's been doing on the test PR.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -203,15 +203,13 @@ __device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params(
...
-template <typename G, class params, uint32_t polynomial_size,
-          uint32_t glwe_dimension, uint32_t level_count>
-__device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical(
-    double2 *fft, double2 *fft_regs, double2 *buffer_regs,
-    const double2 *__restrict__ bootstrapping_key, int iteration, G &group,
+// We need a different version for the default specialized 2_2 params because
+// in that case we don't need to use dsm and then no cluster sync
+template <class params, uint32_t polynomial_size, uint32_t glwe_dimension,
+          uint32_t level_count>
+__device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_1block(
```
agnesLeroy: Maybe the `_1block` suffix isn't necessary?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
```diff
@@ -482,7 +482,7 @@ void cuda_programmable_bootstrap_lwe_ciphertext_vector(
         num_many_lut, lut_stride);
     break;
   case 2048:
-    host_programmable_bootstrap<Torus, AmortizedDegree<2048>>(
+    host_programmable_bootstrap<Torus, Degree<2048>>(
```
agnesLeroy, line 584: Would be good to have the same comment here 
```
// For polynomial size 2048, we know Degree is better than AmortizedDegree,
//  other sizes require further evaluation.
```

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
agnesLeroy: Or `no_tbc` instead?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -17,6 +26,184 @@
...
+
+// Check if specialized 2_2_params kernel can be used
+template <typename Torus>
+__host__ bool
+supports_specialized_2_2_params(uint32_t polynomial_size,
+                                uint32_t glwe_dimension, uint32_t level_count,
+                                uint32_t base_log, uint32_t max_shared_memory) {
+  uint64_t required_shared_memory =
+      get_buffer_size_full_sm_programmable_bootstrap_specialized_2_2_params<
+          Torus>(polynomial_size);
```
agnesLeroy, line 46: Here in case compute capability is more recent than 8.0 we should return false?

agnesLeroy (commented):
Thank you very much @guillermo-oyarzun, I still have some questions.
I think it's ok to keep the base logs as they are now, but I don't see a condition to activate the 1 block PBS only for A100 or previous architectures.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cu
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -17,6 +26,192 @@
...
+// Check if specialized 2_2_params kernel can be used
+template <typename Torus>
+__host__ bool
+supports_specialized_2_2_params(uint32_t gpu_index, uint32_t polynomial_size,
+                                uint32_t glwe_dimension, uint32_t level_count,
+                                uint32_t max_shared_memory) {
+  // Check compute capability - only support cc80
+  // because of the FP64 ratio.
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, gpu_index);
```
agnesLeroy: Missing check_cuda_error().

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -17,6 +26,192 @@
...
+  if (deviceProp.major != 8 || deviceProp.minor != 0) {
```
agnesLeroy: Here you activate it only for A100. Maybe it's ok, otherwise the condition could be `deviceProp.major <= 8`?

guillermo-oyarzun: i thought it was only A100 we wanted it, or do we want to keep it in cc70 too?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
guillermo-oyarzun: i will comment it with him

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
agnesLeroy: It's probably ok to have it only activated for A100, let's keep it this way. A comment specifying this is an A100-specific implementation would be good though.

guillermo-oyarzun: i will add cc7 too, i am not sure if we want to risk to use a really really old gpu like 3.0 and use it.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -429,11 +429,9 @@ uint64_t get_buffer_size_programmable_bootstrap_cg(
 }
 
 template <typename Torus>
-bool has_support_to_cuda_programmable_bootstrap_cg(uint32_t glwe_dimension,
-                                                   uint32_t polynomial_size,
-                                                   uint32_t level_count,
-                                                   uint32_t num_samples,
-                                                   uint32_t max_shared_memory);
+bool has_support_to_cuda_programmable_bootstrap_cg(
+    uint32_t gpu_index, uint32_t glwe_dimension, uint32_t polynomial_size,
```
agnesLeroy: I think we can avoid passing gpu_index as argument and only check device property of GPU 0, wdyt?

guillermo-oyarzun: at first i had it like that, my only concern was that i could be missing some case in which the gpu 0 is not used somehow, but i guess that is not possible, right?

agnesLeroy: There is always a GPU with index 0 on the machine, if not it means there is not GPU at all. Then I've never seen a config with several GPUs different from each other  so I don't think it's worth trying to cover that case. For shared memory size we check only GPU 0 for example already.

backends/tfhe-cuda-backend/cuda/include/pbs/programmable_bootstrap_testing.h
```diff
@@ -33,6 +39,33 @@ void cuda_programmable_bootstrap_tbc_64_2_2_async(
     uint32_t polynomial_size, uint32_t base_log, uint32_t level_count,
     uint32_t num_samples, uint32_t num_many_lut, uint32_t lut_stride);
 
+// Force standard (non-specialized) FFT layout for the BSK
```
bbarbakadze: Can you please add a a bit more comments here, its a bit hard to understand with one look what is it about.

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
bbarbakadze, line 191: So like this it is fine for all flavors in terms of registers?

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
```diff
@@ -540,6 +541,44 @@ __global__ void batch_NSMFFT(double2 *d_input, double2 *d_output,
...
+  __syncthreads();
+
+  tid = threadIdx.x;
+#pragma unroll
+  for (int i = 0; i < params::opt / 2; i++) {
+    d_output[blockIdx.x * (params::degree / 2) + tid] = fft_regs[i];
+    tid = tid + params::degree / params::opt;
+  }
+}
+
```
bbarbakadze, line 582: I know you replicated this function from existing `batch_NSMFFT` but I think:

1. Comment is wrong do we really unroll it to size + 1 elements?
2. Do we really need buffer here if it fits in shared memory we can do fft in shared, but if it does not fit cant we use `d_ouptut` as a buffer?

Maybe we do this refactor in different pr.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -493,21 +699,171 @@ __host__ void execute_step_two(
...
+    use_specialized = specialized_2_2_params_checker<uint64_t>(
+        polynomial_size, glwe_dimension, level_count,
+        cuda_get_max_shared_memory(gpu_index));
+    PANIC_IF_FALSE(
+        use_specialized,
+        "Cuda error (classical PBS): forced specialized 2_2 requires "
+        "(N=2048, level_count=1, glwe_dimension=1, sufficient shared memory).");
+  } else {
+    // In auto mode we also check the compute capability.
+    use_specialized = supports_specialized_2_2_params<Torus>(
```
bbarbakadze, line 736: Why do we have different check here, should not it be the same check as when we force it to be specialized?

backends/tfhe-cuda-backend/cuda/tests_and_benchmarks/tests/test_classical_pbs.cpp
```diff
@@ -252,12 +267,22 @@ TEST_P(ClassicalProgrammableBootstrapTestPrimitives_u64, classical_tbc_2_2) {
...
-        pbs_base_log == 23)) {
+  if (!supports_classical_specialized_2_2()) {
     GTEST_SKIP()
-        << "TBC specialized 2_2 requires N=2048, glwe=1, level=1, base_log=23.";
+        << "TBC classical specialized 2_2 PBS is not supported on this "
+           "params.";
   }
-
+  if (!(pbs_base_log == 23)) {
+    GTEST_SKIP() << "Specialized 2_2 requires base_log == 23.";
```
bbarbakadze, line 276: Do we still need this check, base_log now is template right?

backends/tfhe-cuda-backend/cuda/tests_and_benchmarks/tests/test_classical_pbs.cpp
```diff
@@ -281,6 +306,47 @@ TEST_P(ClassicalProgrammableBootstrapTestPrimitives_u64, classical_tbc_2_2) {
...
+TEST_P(ClassicalProgrammableBootstrapTestPrimitives_u64,
+       classical_specialized_2_2) {
+  if (!supports_classical_specialized_2_2()) {
+    GTEST_SKIP() << "Specialized 2_2 classical PBS is not supported on this "
+                    "params.";
+  }
+  if (!(pbs_base_log >= 21 && pbs_base_log <= 25)) {
+    GTEST_SKIP() << "Specialized 2_2 requires base_log in [21, 25].";
+  }
+  // We need to change the layout of the bsk to match the specialized kernel.
```
bbarbakadze, line 318: This check is different from 
`if (!(pbs_base_log == 23)) {
    GTEST_SKIP() << "Specialized 2_2 requires base_log == 23.";`

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
bbarbakadze: You don't even use buffer in this function so it can be removed

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
guillermo-oyarzun: yup, all specialized versions are kept to 64 regs max

backends/tfhe-cuda-backend/cuda/src/fft/bnsmfft.cuh
guillermo-oyarzun: you are right since this is specialized the main assumption is that it will always fit in shared memory (we only need 32768bytes, so i eliminated the use of buffer but forgot to remove it from the args.  And yes the comment is wrong about the unrolling.

backends/tfhe-cuda-backend/cuda/tests_and_benchmarks/tests/test_classical_pbs.cpp
guillermo-oyarzun: we will always need that check, because even if it is template we will not compile each option of the template.  So i added the template only in the new specialized non-tbc version for now, so this condition is still equivalent to what it was before.  My idea is to leave the templarization of classical and multibit for other PR, to avoid increasing the scope of this one.

backends/tfhe-cuda-backend/cuda/tests_and_benchmarks/tests/test_classical_pbs.cpp
guillermo-oyarzun: This version is actually templarized, and so far i only compile the kernel with ranges [21,25], that's why the condition is different, my idea is to extend the template idea in another PR.

backends/tfhe-cuda-backend/cuda/include/pbs/programmable_bootstrap_testing.h
guillermo-oyarzun: sure

PR #3302, opened 2026-02-12
fix(gpu): widen uint32 products to size_t in allocation size computations

pdroalves:
There is a common pattern in the CUDA backend to compute allocation sizes using uint32_t * uint32_t products that overflow before being widened to 64-bit by * sizeof(T). We were aware of a few samples in rerand scratch function but using Claude I was able to find others. This PR fixes all of them.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1326

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/checked_arithmetic.h
+++ b/backends/tfhe-cuda-backend/cuda/include/checked_arithmetic.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include <cstddef>
+#include <cstdio>
+
+#include "device.h"
+
+// Variadic checked multiplication of size_t values.
+// Folds left-to-right using __builtin_mul_overflow, returning true on overflow.
+// On overflow the value written to *out is unspecified.
+template <typename... Args>
+inline bool checked_mul(size_t *out, size_t first, Args... rest) {
+  size_t result = first;
+  for (size_t value : {static_cast<size_t>(rest)...}) {
+    if (__builtin_mul_overflow(result, value, &result))
+      return true;
+  }
+  *out = result;
+  return false;
+}
+
+// Variadic safe multiplication: computes the product and panics on overflow.
+template <typename... Args> inline size_t safe_mul(size_t first, Args... rest) {
+  size_t result;
+  bool overflow = checked_mul(&result, first, rest...);
+  PANIC_IF_FALSE(!overflow, "multiplication overflow wraps size_t");
+  return result;
+}
+
+// Variadic safe multiplication with an appended sizeof(T) factor.
+// Computes (args... * sizeof(T)) with overflow checking.
+template <typename T, typename... Args>
+inline size_t safe_mul_sizeof(Args... args) {
+  return safe_mul(args..., sizeof(T));
+}
--- a/backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
@@ -1,4 +1,5 @@
 #pragma once
+#include "checked_arithmetic.h"
 #include "cmux.h"
 #include "integer_utilities.h"
 
@@ -39,8 +40,8 @@ template <typename Torus> struct int_are_all_block_true_buffer {
         max_chunks, params.big_lwe_dimension, size_tracker,
         allocate_gpu_memory);
 
-    preallocated_h_lut = (Torus *)malloc(
-        (params.glwe_dimension + 1) * params.polynomial_size * sizeof(Torus));
+    preallocated_h_lut = (Torus *)malloc(safe_mul_sizeof<Torus>(
+        params.glwe_dimension + 1, params.polynomial_size));
 
     is_max_value = new int_radix_lut<Torus>(streams, params, 2, max_chunks,
                                             allocate_gpu_memory, size_tracker);
@@ -214,8 +215,8 @@ template <typename Torus> struct int_tree_sign_reduction_buffer {
     tree_last_leaf_lut = new int_radix_lut<Torus>(
         streams, params, 1, 1, allocate_gpu_memory, size_tracker);
 
-    preallocated_h_lut = (Torus *)malloc(
-        (params.glwe_dimension + 1) * params.polynomial_size * sizeof(Torus));
+    preallocated_h_lut = (Torus *)malloc(safe_mul_sizeof<Torus>(
+        params.glwe_dimension + 1, params.polynomial_size));
 
     tree_last_leaf_scalar_lut = new int_radix_lut<Torus>(
         streams, params, 1, 1, allocate_gpu_memory, size_tracker);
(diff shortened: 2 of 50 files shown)


backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
```diff
@@ -39,8 +39,9 @@ template <typename Torus> struct int_are_all_block_true_buffer {
         max_chunks, params.big_lwe_dimension, size_tracker,
         allocate_gpu_memory);
 
-    preallocated_h_lut = (Torus *)malloc(
-        (params.glwe_dimension + 1) * params.polynomial_size * sizeof(Torus));
+    preallocated_h_lut =
```
agnesLeroy: To be extra safe it would be good to have some utility functions like
```
#pragma once

#include <stddef.h>
#include <stdbool.h>
#include <stdint.h>

/* ---------- 2-operand checked multiply ---------- */

static inline bool
checked_mul_size_t(size_t a, size_t b, size_t *out)
{
#if defined(__GNUC__) || defined(__clang__)
    return !__builtin_mul_overflow(a, b, out);
#else
    if (a != 0 && b > SIZE_MAX / a)
        return false;

    *out = a * b;
    return true;
#endif
}

/* ---------- 3-operand checked multiply ---------- */

static inline bool
checked_mul3_size_t(size_t a, size_t b, size_t c, size_t *out)
{
    size_t tmp;

#if defined(__GNUC__) || defined(__clang__)
    if (__builtin_mul_overflow(a, b, &tmp))
        return false;

    if (__builtin_mul_overflow(tmp, c, out))
        return false;

    return true;
#else
    if (!checked_mul_size_t(a, b, &tmp))
        return false;

    return checked_mul_size_t(tmp, c, out);
#endif
}
```
and then 
```
size_t allocation_size;

if (!checked_mul3_size_t(
        (size_t)num_lwes,
        (size_t)params.small_lwe_dimension + 1,
        sizeof(Torus),
        &allocation_size)) {
   PANIC("Cuda error: overflow in buffer size computation");
}

tmp_ksed_zero_lwes = (Torus *)cuda_malloc_with_size_tracking_async(
    allocation_size,
    streams.stream(0),
    streams.gpu_index(0),
    size_tracker,
    allocate_gpu_memory);
```

pdroalves: I like that. Extra careful and easy to do nowadays.

pdroalves: Done. I added a method for save computing the allocation size and asked claude to find other unsafe allocations we were doing. It found many.

pdroalves: Check and see if this is what you had in mind.

backends/tfhe-cuda-backend/cuda/include/checked_arithmetic.h
```diff
@@ -0,0 +1,87 @@
...
+#include <cstddef>
+#include <cstdio>
+
+#include "device.h"
+
+// Checked multiplication of two size_t values.
+// Returns true if the product overflows, false otherwise.
+// On overflow the value written to *out is unspecified.
+inline bool checked_mul_size_t(size_t a, size_t b, size_t *out) {
+#if defined(__GNUC__) || defined(__clang__)
```
pdroalves: Do we need this?

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
agnesLeroy: Yes looks good, thanks!

backends/tfhe-cuda-backend/cuda/include/checked_arithmetic.h
agnesLeroy: No we don't :+1:

backends/tfhe-cuda-backend/cuda/include/zk/zk_utilities.h
```diff
@@ -301,13 +302,15 @@ template <typename Torus> struct zk_expand_mem {
...
-        num_lwes * (casting_params.big_lwe_dimension + 1) * sizeof(Torus),
+        safe_alloc_size_2<Torus>(num_lwes,
+                                 casting_params.big_lwe_dimension + 1),
         streams.stream(0), streams.gpu_index(0), size_tracker,
         allocate_gpu_memory);
 
     tmp_ksed_small_to_big_expanded_lwes =
         (Torus *)cuda_malloc_with_size_tracking_async(
-            num_lwes * (casting_params.big_lwe_dimension + 1) * sizeof(Torus),
+            safe_alloc_size_2<Torus>(num_lwes,
```
agnesLeroy: If we only use this on the host side we could have a single entry point for the safe_alloc_size, with variable argument numbers. But we could also want to use it in device code in which case what you did here would be better (we could just have 1 name for the 3 functions though to reduce a bit naming complexity).

agnesLeroy: But looking at the PBS kernels I don't see where we would need to use this there, so I guess having it only on host is enough and we don't really care about inlining in this case.

pdroalves: Of course! I originally wrote that as a struct, that's why I didn't even consider function overloading. I'm refactoring to have a single `safe_alloc_size`.

pdroalves: Done! 

Now we have `safe_mul` and `safe_mul_sizeof<T>` which will safely compute the sizes. The first computes `a * b * ...` and the second `a * b * ... * sizeof<T>`.

backends/tfhe-cuda-backend/cuda/include/checked_arithmetic.h
```diff
@@ -0,0 +1,80 @@
...
+
+// Checked multiplication of four size_t values (a * b * c * d).
+// Returns true on overflow at any intermediate step.
+inline bool checked_mul4_size_t(size_t a, size_t b, size_t c, size_t d,
+                                size_t *out) {
+  size_t abc;
+  if (checked_mul3_size_t(a, b, c, &abc))
+    return true;
+  return checked_mul_size_t(abc, d, out);
+}
```
agnesLeroy: Maybe something like this for checked_mul?

```
#include <cstddef>

template <typename... Args>
inline bool checked_mul(size_t* out, size_t first, Args... rest) {
    size_t result = first;

    for (size_t value : {static_cast<size_t>(rest)...}) {
        if (__builtin_mul_overflow(result, value, &result))
            return true;
    }

    *out = result;
    return false;
}
```
instead of having one entry point per number of args?

pdroalves: Cool, I didn't know we had this in C++. It should work also for `safe_mul`.

PR #3303, opened 2026-02-12
fix(gpu): rework compression to fix the GLWE addresses when lwe_per_glwe != polynomial_size

pdroalves:
The original implementation doesn't support the case when a GLWE is not full and `lwe_per_glwe < polynomial`.

```
  GLWE 0                                              GLWE 1
  ├──────────── 2048 elements ────────────┤           ├──────────── 2048 elements ────────────┤
  ┌───────────────┬────────┬──────────────┐           ┌───────────────┬────────┬──────────────┐
  │   mask0       │ body0  │   garbage    │           │   mask1       │ body1  │   garbage    │
  │  (1024 elems) │ (128)  │   (896)     │           │  (1024 elems) │ (128)  │   (896)     │
  └───────────────┴────────┴──────────────┘           └───────────────┴────────┴──────────────┘
   offset 0        1024     1152           2048        2048           3072     3200           4096
```
 This PR implement strides in compression and decompression to adjust the indexes on that case.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1315

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/ciphertext.h
+++ b/backends/tfhe-cuda-backend/cuda/include/ciphertext.h
@@ -19,7 +19,9 @@ void cuda_convert_lwe_ciphertext_vector_to_cpu_64(void *stream,
 void cuda_glwe_sample_extract_64(void *stream, uint32_t gpu_index,
                                  void *lwe_array_out, void const *glwe_array_in,
                                  uint32_t const *nth_array, uint32_t num_nths,
-                                 uint32_t lwe_per_glwe, uint32_t glwe_dimension,
+                                 uint32_t num_lwes_to_extract_per_glwe,
+                                 uint32_t num_lwes_stored_per_glwe,
+                                 uint32_t glwe_dimension,
                                  uint32_t polynomial_size);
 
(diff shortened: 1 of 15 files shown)


backends/tfhe-cuda-backend/cuda/include/ciphertext.h
```diff
@@ -19,6 +19,7 @@ void cuda_convert_lwe_ciphertext_vector_to_cpu_64(void *stream,
 void cuda_glwe_sample_extract_64(void *stream, uint32_t gpu_index,
                                  void *lwe_array_out, void const *glwe_array_in,
                                  uint32_t const *nth_array, uint32_t num_nths,
+                                 uint32_t num_nths_per_glwe,
```
agnesLeroy: I think naming is a bit confusing, how about some more explicit naming:
- num_nth_per_glwe: num_lwes_to_extract_per_glwe
- lwe_per_glwe -> num_lwes_stored_per_glwe
What do you think?

pdroalves: I like it! No reason to have variables with mysterious names.

PR #3430, opened 2026-03-26
feat: add bitonic_shuffle

tmontaigu:
Bitonic shuffles allows to shuffle a vec of homomorphic integers by using a bitonic sort using random generated keys

closes https://github.com/zama-ai/tfhe-rs-internal/issues/1357

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)

### DIFF
--- a/tfhe/src/shortint/oprf.rs
+++ b/tfhe/src/shortint/oprf.rs
@@ -628,9 +628,38 @@ impl<C: Container<Element = c64> + Sync> GenericOprfServerKey<C> {
         random_bits_count: u64,
         target_sks: &ServerKey,
     ) -> Vec<Ciphertext> {
+        self.inner
+            .generate_pseudo_random_bits(
+                seed,
+                &[random_bits_count],
+                target_sks.message_modulus.0.ilog2() as u64,
+                target_sks,
+            )
+            .pop()
+            .unwrap()
+    }
+
+    /// Uniformly generates a batch of independent encrypted random bit-chunks
+    /// from a single seed in one call.
+    ///
+    /// For each entry `bc` in `random_bits_counts`, returns a `Vec<Ciphertext>`
+    /// of `ceil(bc / random_bits_per_block)` blocks, where `random_bits_per_block`
+    /// is the message-modulus log2. Within each chunk, every ciphertext encrypts a
+    /// value in `[0, 2^random_bits_per_block[` except for the last one, which may
+    /// hold fewer random bits if `bc` does not divide evenly.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if `self` is not compatible with `target_sks`
+    pub fn generate_oblivious_pseudo_random_bits_chunks(
+        &self,
+        seed: impl OprfSeed,
+        random_bits_counts: &[u64],
+        target_sks: &ServerKey,
+    ) -> Vec<Vec<Ciphertext>> {
         self.inner.generate_pseudo_random_bits(
             seed,
-            random_bits_count,
+            random_bits_counts,
             target_sks.message_modulus.0.ilog2() as u64,
             target_sks,
         )
@@ -738,6 +767,25 @@ pub(crate) struct PrfSeededModulusSwitched {
     log_modulus: CiphertextModulusLog,
 }
 
+impl PrfSeededModulusSwitched {
+    fn zero(lwe_size: LweSize, log_modulus: CiphertextModulusLog) -> Self {
+        Self {
+            mask: vec![0usize; lwe_size.to_lwe_dimension().0],
+            body: 0,
+            log_modulus,
+        }
+    }
+
+    fn fill_mask_with_random(&mut self, rng: &mut RandomGenerator<DefaultRandomGenerator>) {
+        for mask_element in &mut self.mask {
+            *mask_element = rng.random_from_distribution_custom_mod::<u32, _>(
+                Uniform,
+                CiphertextModulus::new(1u128 << self.log_modulus.0),
+            ) as usize;
+        }
+    }
+}
+
 impl ModulusSwitchedLweCiphertext<usize> for PrfSeededModulusSwitched {
     fn log_modulus(&self) -> CiphertextModulusLog {
         self.log_modulus
@@ -819,71 +867,81 @@ impl MultiBitModulusSwitchedLweCiphertext for PrfMultiBitSeededModulusSwitched {
 // XOF-based seeded modulus switch (current implementation, uses OprfSeed)
 // ============================================================================
 
-/// Takes as input the `bit_count` to be generated (e.g. 21),
-/// the `random_bits_per_blocks` (e.g. 2) and output as many
-/// [PrfSeededModulusSwitched] as needed (ceil(bit_count/random_bits_per_blocks))
+/// Takes as input the `bit_chunks` to be generated (e.g [15, 15])
+/// the `random_bits_per_blocks` (e.g. 2).
 ///
-/// Also returns how many random bits the last block shall have.
-/// e.g 1 in the case of 21 bits over blocks of 2 bits
+/// Outputs a flat array of all [PrfSeededModulusSwitched] needed
+///
+/// Also returns the position of blocks that should be generated to contain
+/// less than `random_bits_per_block`, along with how many bits they should contain
 pub(crate) fn create_random_from_seed_modulus_switched(
     seed: impl OprfSeed,
     lwe_size: LweSize,
     polynomial_size: PolynomialSize,
-    bit_count: u64,
+    bit_chunks: &[u64],
     random_bits_per_block: u64,
-) -> (Vec<PrfSeededModulusSwitched>, u64) {
-    if bit_count == 0 || random_bits_per_block == 0 {
-        return (Vec::new(), 0);
+) -> Vec<(PrfSeededModulusSwitched, u64)> {
+    if random_bits_per_block == 0 {
+        return Vec::new();
     }
 
-    let num_blocks = bit_count.div_ceil(random_bits_per_block);
-
-    let mut last_block_bits = bit_count % random_bits_per_block;
-    if last_block_bits == 0 {
-        last_block_bits = random_bits_per_block;
-    }
-
+    // Init the hasher
     let bytes = seed.into_bytes();
     let mut hasher = sha3::Sha3_256::default();
     hasher.update(b"TFHE_PRF");
     hasher.update(bytes.as_ref());
-    if last_block_bits == random_bits_per_block {
-        // All blocks contain the same number of random bits
-        hasher.update(num_blocks.to_le_bytes());
-        hasher.update(random_bits_per_block.to_le_bytes());
-    } else {
-        let head_count = num_blocks - 1;
-        if head_count != 0 {
-            hasher.update(head_count.to_le_bytes());
-            hasher.update(random_bits_per_block.to_le_bytes());
+
+    let total_blocks: u64 = bit_chunks
+        .iter()
+        .map(|&bits| bits.div_ceil(random_bits_per_block))
+        .sum();
+    let mut seededs = Vec::with_capacity(total_blocks as usize);
+    for &bit_count in bit_chunks {
+        if bit_count == 0 {
+            continue;
         }
-        hasher.update(1u64.to_le_bytes());
-        hasher.update(last_block_bits.to_le_bytes());
+
+        let (num_full_blocks, last_block_bits) = (
+            bit_count / random_bits_per_block,
+            bit_count % random_bits_per_block,
+        );
+
+        if num_full_blocks != 0 {
+            hasher.update(num_full_blocks.to_le_bytes());
+            hasher.update(random_bits_per_block.to_le_bytes());
+
+            for _ in 0..num_full_blocks {
+                seededs.push((
+                    PrfSeededModulusSwitched::zero(
+                        lwe_size,
+                        polynomial_size.to_blind_rotation_input_modulus_log(),
+                    ),
+                    random_bits_per_block,
+                ));
+            }
+        }
+        if last_block_bits != 0 {
+            hasher.update(1u64.to_le_bytes());
+            hasher.update(last_block_bits.to_le_bytes());
+
+            seededs.push((
+                PrfSeededModulusSwitched::zero(
+                    lwe_size,
+                    polynomial_size.to_blind_rotation_input_modulus_log(),
+                ),
+                last_block_bits,
+            ));
+        }
     }
 
     let seed = hasher.finalize();
-
     let mut xof =
         RandomGenerator::<DefaultRandomGenerator>::new(XofSeed::new(seed.to_vec(), *b"PRF_INIT"));
+    for (seeded, _) in &mut seededs {
+        seeded.fill_mask_with_random(&mut xof);
+    }
 
-    let seeded = (0..num_blocks)
-        .map(|_| {
-            let mut mask = vec![0usize; lwe_size.to_lwe_dimension().0];
-            for mask_element in &mut mask {
-                *mask_element = xof.random_from_distribution_custom_mod::<u32, _>(
-                    Uniform,
-                    CiphertextModulus::new(2 * polynomial_size.0 as u128),
-                ) as usize;
-            }
-            PrfSeededModulusSwitched {
-                mask,
-                body: 0,
-                log_modulus: polynomial_size.to_blind_rotation_input_modulus_log(),
-            }
-        })
-        .collect::<Vec<_>>();
-
-    (seeded, last_block_bits)
+    seededs
 }
 
 // ============================================================================
@@ -965,28 +1023,32 @@ impl<C: Container<Element = c64> + Sync> OprfBootstrappingKey<C> {
             random_bits_count, target_sks.message_modulus.0
         );
 
-        let mut blocks = self.generate_pseudo_random_bits(
+        let mut chunks = self.generate_pseudo_random_bits(
             seed,
-            random_bits_count,
+            &[random_bits_count],
             random_bits_count, // This means we will have one block
             target_sks,
         );
+        assert_eq!(chunks.len(), 1);
+        let mut blocks = chunks.pop().unwrap();
         assert_eq!(blocks.len(), 1);
         blocks.pop().unwrap()
     }
 
-    /// Generates `bit_count` random bits split over multiple blocks.
+    /// Generates random bits split over multiple blocks, grouped per input chunk.
     ///
-    /// Each ciphertext encrypts a value in `[0, 2^random_bits_per_block[`
-    /// except for the last one which may have fewer random bits:
-    /// `bit_count=3, random_bits_per_block=2` -> first_block: 2 bits, last block: 1 bit
+    /// For each entry `bc` in `bit_chunks`, produces a `Vec<Ciphertext>` of
+    /// `ceil(bc / random_bits_per_block)` blocks. Each ciphertext encrypts a value
+    /// in `[0, 2^random_bits_per_block[` except for the last block of each chunk
+    /// which may have fewer random bits:
+    /// `bc=3, random_bits_per_block=2` -> first_block: 2 bits, last block: 1 bit.
     pub(crate) fn generate_pseudo_random_bits(
         &self,
         seed: impl OprfSeed,
-        bit_count: u64,
+        bit_chunks: &[u64],
         random_bits_per_block: u64,
         target_sks: &ServerKey,
-    ) -> Vec<Ciphertext> {
+    ) -> Vec<Vec<Ciphertext>> {
         let ks_key = match &target_sks.atomic_pattern {
             AtomicPatternServerKey::Standard(std) => {
                 self.assert_compatible_with_target_bsk(&std.bootstrapping_key);
@@ -1016,46 +1078,33 @@ impl<C: Container<Element = c64> + Sync> OprfBootstrappingKey<C> {
         let polynomial_size = self.polynomial_size();
         let in_lwe_size = self.input_lwe_dimension().to_lwe_size();
 
-        let (seeded_cts, last_block_bits) = create_random_from_seed_modulus_switched(
+        let seeded_cts = create_random_from_seed_modulus_switched(
             seed,
             in_lwe_size,
             polynomial_size,
-            bit_count,
+            bit_chunks,
             random_bits_per_block,
         );
 
         let ciphertext_modulus = target_sks.ciphertext_modulus;
+        let max_bits = seeded_cts.iter().map(|(_, b)| *b).max().unwrap_or(0);
+        let luts = (1..=max_bits)
+            .map(|bit| {
+                generate_oprf_lut(
+                    bit,
+                    bits_in_one_block,
+                    polynomial_size,
+                    self.glwe_size(),
+                    ciphertext_modulus,
+                )
+            })
+            .collect::<Vec<_>>();
 
-        let regular_lut = generate_oprf_lut(
-            random_bits_per_block,
-            bits_in_one_block,
-            polynomial_size,
-            self.glwe_size(),
-            ciphertext_modulus,
-        );
-        let last_block_lut = if last_block_bits == random_bits_per_block {
-            None
-        } else {
-            Some(generate_oprf_lut(
-                last_block_bits,
-                bits_in_one_block,
-                polynomial_size,
-                self.glwe_size(),
-                ciphertext_modulus,
-            ))
-        };
-
-        let num_blocks = seeded_cts.len();
-        seeded_cts
+        let flat: Vec<Ciphertext> = seeded_cts
             .into_par_iter()
-            .enumerate()
-            .map(|(i, seeded)| {
+            .map(|(seeded, num_bits)| {
                 let (LookupTableOwned { mut acc, degree }, post_pbs_constant) =
-                    if i == num_blocks - 1 && last_block_lut.is_some() {
-                        last_block_lut.clone().unwrap()
-                    } else {
-                        regular_lut.clone()
-                    };
+                    luts[num_bits as usize - 1].clone();
 
                 match self {
                     Self::Classic { bsk, .. } => {
@@ -1114,6 +1163,15 @@ impl<C: Container<Element = c64> + Sync> OprfBootstrappingKey<C> {
                     target_sks.carry_modulus,
                     target_sks.atomic_pattern.kind(),
                 )
+            })
+            .collect();
+
+        let mut iter = flat.into_iter();
+        bit_chunks
+            .iter()
+            .map(|&bc| {
+                let n = bc.div_ceil(random_bits_per_block) as usize;
+                iter.by_ref().take(n).collect()
             })
             .collect()
     }
@@ -1291,12 +1349,12 @@ pub(crate) mod test {
             seed,
             lwe_size,
             params.polynomial_size(),
-            params.message_modulus().0.ilog2() as u64,
+            &[params.message_modulus().0.ilog2() as u64],
             params.message_modulus().0.ilog2() as u64,
         )
-        .0
         .pop()
-        .unwrap();
+        .unwrap()
+        .0;
 
         assert!(seeded.mask.iter().all(|v| *v < input_p as usize));
 
--- a/tfhe/src/shortint/oprf.rs
+++ b/tfhe/src/shortint/oprf.rs
@@ -906,6 +906,9 @@ pub(crate) fn create_random_from_seed_modulus_switched(
             bit_count % random_bits_per_block,
         );
 
+        let num_blocks = num_full_blocks + u64::from(last_block_bits != 0);
+        hasher.update(num_blocks.to_le_bytes());
+
         if num_full_blocks != 0 {
             hasher.update(num_full_blocks.to_le_bytes());
             hasher.update(random_bits_per_block.to_le_bytes());
--- a/tfhe-benchmark/Cargo.toml
+++ b/tfhe-benchmark/Cargo.toml
@@ -99,6 +99,12 @@ harness = false
 required-features = ["integer", "internal-keycache"]
 
 [[bench]]
+name = "hlapi-bitonic-shuffle"
+path = "benches/high_level_api/bitonic_shuffle.rs"
+harness = false
+required-features = ["integer", "internal-keycache"]
+
+[[bench]]
 name = "hlapi-dex"
 path = "benches/high_level_api/dex.rs"
 harness = false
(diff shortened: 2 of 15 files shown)


IceTDrinker (commented):
first batch of comments

> @IceTDrinker reviewed 10 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-:-Orn1UZo8OUz7nUfJQ55:b-gh7jz2)* status: 10 of 13 files reviewed, 8 unresolved discussions (waiting on nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 26 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmggZh3vbtpdsju6Iw:-OrmggZiDThJ5-jCkVxF:b-e0koh4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L26)):*
> ```XML
>     data: Vec<T>,
>     key_size: BitonicShuffleKeySize,
>     seeder: &mut S,
> ```

since the process to generate keys uses the PRF, I think we should use the same construction as the PRF in terms of API to take the seed (array of bytes) to match the cryptographic construction we already have

we don't need to support strictly the same types (since I think I recall we manage some backward compatibility) but having a byte array/slice as input is desirable
___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 82 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmhiN_88oGcjt85RXC:-OrmhiN_88oGcjt85RXD:b-b3e6fd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L82)):*
> ```XML
>         let shuffled = bitonic_shuffle(
>             encrypted,
>             BitonicShuffleKeySize::num_blocks(16),
> ```

is it better to ask for a number of blocks or a number of bits here ?

since overall size of the key will change if a number of block is given for different parameters

meaning a parameter update makes the code behavior change wrt collision probability
___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 110 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpikvD9JUOftln809:-OrmpikvD9JUOftln80A:bktg95z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L110)):*
> ```XML
>         let shuffled = bitonic_shuffle(
>             encrypted,
>             BitonicShuffleKeySize::num_blocks(16),
> ```

same remark

do we also test the probability thing ? (haven't seen integer yet)
___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 34 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmryAa1Z1ukmNVZafU:-OrmryAa1Z1ukmNVZafV:b-i92209) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L34)):*
> ```XML
>                     let ascending = (i >> (phase + 1)) & 1 == 0;
>                     comparators.push((i, j, ascending));
>                 }
> ```

would be interested in the rough idea of why the bit tricks work, the xor trick is unclear, and I'm not clear on why ascending is the specific bit in i

is it because below the halfway point it's ascending and beyond it's descending ?

in that case it could be a two stage loop 0..half (do something for first half) loop half..end (do something for second half)
___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 124 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmxcwN83-hfWNZQybk:-OrmxcwN83-hfWNZQybl:b-8qx2pr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L124)):*
> ```XML
>         if key_num_blocks == 0 {
>             return Err(crate::Error::new(
>                 "key_num_blocks must be at least 1".to_string(),
> ```

could be worth having a slightly more explicit message like "failed to find key_num_blocks (key lenght is 0, which cannot be used) for bitonic shuffle of {inputs.len()}, and with key_size: {key_size}"

or something in that spirit
___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 136 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-Ormy3PAAVUHsg2sec1-:-Ormy3PAAVUHsg2sec10:b-mhc74v) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L136)):*
> ```XML
>                 oprf_key.par_generate_oblivious_pseudo_random_unsigned_integer(
>                     seeder.seed(),
>                     key_num_blocks,
> ```

I'm thinking (maybe) to match the PRF construction fully:

array of shortint ciphertexts for all the inputs, which then gets cut down in the respective integers like

[ct 0, .., ct 31, ct32, .., ct 63]

0..=31 => FheUint64
32..=64 => FheUint64

this way there is a single prf call from a single seed 
___
*[`tfhe/src/integer/server_key/radix_parallel/cmux.rs` line 959 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmQyPW6758E8ZJNkux:-OrmQyPW6758E8ZJNkuy:bkr6rj3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88241fee2192be88e29d30b203e2dc8e26f20f1a/tfhe/src/integer/server_key/radix_parallel/cmux.rs#L959)):*
> ```XML
>         // A drawback is that everyblock needs to be cloned to be shifted, as opposed to
>         // only the condition block, but it's an ok drawback
>         assert!(a.blocks().iter().chain(b.blocks().iter()).all(|block| {
> ```

why do we check this here ?
___
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OrmpuyY-kHfSbJaD9Jf:b-dfv4ij) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*
> ```XML
> use tfhe::{bitonic_shuffle, set_server_key, ClientKey, ConfigBuilder, FheUint64, ServerKey};
> 
> fn bitonic_shuffle_bench(c: &mut Criterion, bench_name: &str, cks: &ClientKey) {
> ```

@SouchonTheo is there a simple way maybe to get a bench spec in here ? if so could be worth showing how to do that to @tmontaigu and record/document the process on notion e.g.

IceTDrinker (commented):
> @IceTDrinker reviewed 3 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: all files reviewed, 10 unresolved discussions (waiting on nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 185 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrnPTnQAToNUOYUsgg_:-OrnPTnQAToNUOYUsgga:b-dcm1i7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L185)):*
> ```XML
>                 "all keys must have the same number of blocks".to_string(),
>             ));
>         }
> ```

nit could have a type

BitonicShuffleKeys

to enforce the invariants you need

not sure it's worth it, just a thought
___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 239 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrnQBNC5yjvGiE0tu3-:-OrnQBNC5yjvGiE0tu30:b7hxla9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L239)):*
> ```XML
>         }
> 
>         let padded_n = padded_size(n);
> ```

nit jsut wondering if a function call for a single size is worth it,

next_power_of_2 with a comment works just as well, or even the variable name is ok to indicate the intent here I believe

worst case the bitonic_network could pad and return the padded size if needed, that last point is not strictly necessary

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/cmux.rs` line 959 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmQyPW6758E8ZJNkux:-OrrQBA74uDpLJ9fvAYw:b-y6avqx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88241fee2192be88e29d30b203e2dc8e26f20f1a/tfhe/src/integer/server_key/radix_parallel/cmux.rs#L959)):*

why do we check this here ?
</blockquote></details>

where would you want to check it ? the check is here and not ealier in the function as this precondition does not apply to the `unchecked_if_then_else_parallelized_no_cleanup`

IceTDrinker:
*[`tfhe/src/integer/server_key/radix_parallel/cmux.rs` line 959 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmQyPW6758E8ZJNkux:-OrrQY9l2tlIF1TIUGwt:bf7wqcz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/88241fee2192be88e29d30b203e2dc8e26f20f1a/tfhe/src/integer/server_key/radix_parallel/cmux.rs#L959)):*

where would you want to check it ? the check is here and not ealier in the function as this precondition does not apply to the `unchecked_if_then_else_parallelized_no_cleanup`
</blockquote></details>

I guess the question is more "is it needed ?"

I don't recall similar checks for bivariate-like ops, in practice I'm good with more asserts/checks, worst case this could be an error if the calling functions return results

IceTDrinker:
seems one change is making the backward data fail

tfhe/src/shortint/oprf.rs
```diff
@@ -738,6 +767,25 @@ pub(crate) struct PrfSeededModulusSwitched {
     log_modulus: CiphertextModulusLog,
 }
 
+impl PrfSeededModulusSwitched {
+    fn zero(lwe_size: LweSize, log_modulus: CiphertextModulusLog) -> Self {
+        Self {
+            mask: vec![0usize, lwe_size.to_lwe_dimension().0],
```
tmontaigu: `,` instead of `;`

Image

tmontaigu (commented):
> @tmontaigu made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: 8 of 15 files reviewed, 10 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 26 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmggZh3vbtpdsju6Iw:-OsW-j6XA9bYxpD5b0jq:bvnrpcm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L26)):*

since the process to generate keys uses the PRF, I think we should use the same construction as the PRF in terms of API to take the seed (array of bytes) to match the cryptographic construction we already have

we don't need to support strictly the same types (since I think I recall we manage some backward compatibility) but having a byte array/slice as input is desirable
</blockquote></details>

updated
___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 82 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmhiN_88oGcjt85RXC:-OsW-lanCCBVWkcqTgK0:bqd2idl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L82)):*

is it better to ask for a number of blocks or a number of bits here ?

since overall size of the key will change if a number of block is given for different parameters

meaning a parameter update makes the code behavior change wrt collision probability
</blockquote></details>

Changed to ask for a number of bits
___
*[`tfhe/src/high_level_api/integers/shuffle.rs` line 110 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpikvD9JUOftln809:-OsW-q1rBvsIq-i49pMk:b53n6mw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/high_level_api/integers/shuffle.rs#L110)):*

same remark

do we also test the probability thing ? (haven't seen integer yet)
</blockquote></details>

I added some tests with the collision proba as input
___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 239 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrnQBNC5yjvGiE0tu3-:-OsW01N31_qgtV05PgzD:b-n9tkhd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L239)):*

nit jsut wondering if a function call for a single size is worth it,

next_power_of_2 with a comment works just as well, or even the variable name is ok to indicate the intent here I believe

worst case the bitonic_network could pad and return the padded size if needed, that last point is not strictly necessary
</blockquote></details>

changed

tmontaigu:
*[`tfhe/src/shortint/oprf.rs` line 908 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OsW6ffs9YzjgPknqhf4:-OsW6ffs9YzjgPknqhf5:b-vf2vye) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/20c7dd16e796b96b586a1ebc9f819925463d80f9/tfhe/src/shortint/oprf.rs#L908)):*
> ```XML
>             bit_count % random_bits_per_block,
>         );
> ```

missing hasher.update((num_full_block + u64::from(last_block_bits != 0).to_le_byte());

tfhe/src/shortint/oprf.rs
IceTDrinker: 🤣

SouchonTheo (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: 8 of 15 files reviewed, 11 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OsvLD7zDo7YW8Dkey2k:b-fnr9jr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*

@SouchonTheo is there a simple way maybe to get a bench spec in here ? if so could be worth showing how to do that to @tmontaigu and record/document the process on notion e.g.
</blockquote></details>

yes of course !
the best would be to see how @tmontaigu explore the crate and implement his new bench id
I remain available just ping me when you need a hand !

btw the issue here is that we do not use write_to_json function and I cannot force the user to use it. In fine if the bench would not go on grafana we cannot spot it :/

IceTDrinker:
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OsvNPzt4EtkZtOlabv9:b-vf3ona) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*
<details><summary><i>Previously, SouchonTheo (Theo Souchon) wrote…</i></summary><blockquote>

yes of course !
the best would be to see how @tmontaigu explore the crate and implement his new bench id
I remain available just ping me when you need a hand !

btw the issue here is that we do not use write_to_json function and I cannot force the user to use it. In fine if the bench would not go on grafana we cannot spot it :/
</blockquote></details>

can be done in a follow-up PR so that this one does not get slowed down

IceTDrinker (commented):
> @IceTDrinker reviewed 7 files and all commit messages, made 3 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: all files reviewed, 8 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs` line 148 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszNrVD4F0DOt-vEJsL:-OszNrVD4F0DOt-vEJsM:b-bj8twx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/integer/server_key/radix_parallel/bitonic_shuffle.rs#L148)):*
> ```XML
> 
>         // manual chunking to be able to move the blocks from one vec to another
>         for blocks in block_chunks {
> ```

nit: into_iter + map + collect not ok ?
___
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OszQS-v4punihPhxOpf:b-t3r9fm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*
> ```XML
>     bit_chunks: &[u64],
>     random_bits_per_block: u64,
> ) -> Vec<(PrfSeededModulusSwitched, u64)> {
> ```

GPU call needs updates to manage the per struct bits ?

any reason for preferring the flat representation when some code later has to re-create the chunks ?
___
*[`tfhe/src/shortint/oprf.rs` line 1177 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszR1QL7bFWxp0gZP8R:-OszR1QL7bFWxp0gZP8S:b-ldwzx1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L1177)):*
> ```XML
>             .map(|&bc| {
>                 let n = bc.div_ceil(random_bits_per_block) as usize;
>                 iter.by_ref().take(n).collect()
> ```

better to do this than having a `vec<vec<_>>` for the chunking ?

SouchonTheo (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: all files reviewed, 8 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OszTI4D6NkDKQt0h2zc:b-7tdf7d) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*

can be done in a follow-up PR so that this one does not get slowed down
</blockquote></details>

agree

IceTDrinker (commented):
Just a few questions to be sure I understand the remaining changes

> @IceTDrinker reviewed 2 files and all commit messages, made 6 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-:-OtZLtSR1hJTCxaZaD1X:b207zu1)* status: all files reviewed, 5 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 908 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OsW6ffs9YzjgPknqhf4:-OtZK7i09YYLuvXpOVNN:bbjng9p) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/20c7dd16e796b96b586a1ebc9f819925463d80f9/tfhe/src/shortint/oprf.rs#L908)):*

missing hasher.update((num_full_block + u64::from(last_block_bits != 0).to_le_byte());
</blockquote></details>

ok missed the fact it was right below, thanks
___
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OtZM1gk02elDx2xtm2t:bdu0odh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

GPU call needs updates to manage the per struct bits ?

any reason for preferring the flat representation when some code later has to re-create the chunks ?
</blockquote></details>

Question I think still stands if a block needs less than message_modulus bits ?
___
*[`tfhe/src/shortint/oprf.rs` line 1177 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszR1QL7bFWxp0gZP8R:-OtZKAwhDPb17qdk3x3u:b-ryyzbv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L1177)):*

better to do this than having a `vec<vec<_>>` for the chunking ?
</blockquote></details>

I take this back, the flat output that gets re-chunked is useful for upcoming prf + rerand
___
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OtZKK3hBtk4hBGxRG_V:bxcc622) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*
<details><summary><i>Previously, SouchonTheo (Theo Souchon) wrote…</i></summary><blockquote>

agree
</blockquote></details>

did you guys sync on this ?
___
*[`tfhe/src/integer/server_key/radix_parallel/cmux.rs` line 839 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OtZKksJBS381_JI9kiK:-OtZKksJBS381_JI9kiL:b-llkpsf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/823dbcaa053168e8b6fe8f4cfdb48b9726157feb/tfhe/src/integer/server_key/radix_parallel/cmux.rs#L839)):*
> ```XML
>                     let degree = Degree::new(lhs_block.degree.get().max(rhs_block.degree.get()));
>                     self.key.unchecked_add_assign(lhs_block, rhs_block);
>                     lhs_block.degree = degree;
> ```

just so I'm clear one of the two blocks is 0 right ? just that the degree does not reflect it since we do'nt know which is strictly 0 ?

so the degree is whichever of the two was higher

tmontaigu:
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OtZVL5O1PhgyeELXn4c:b-oxeid2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

Question I think still stands if a block needs less than message_modulus bits ?
</blockquote></details>

the returned types is `Vec<(Prf, u64)>` the u64 indicates how many bits of random are needed for each prf block, so asking for two chunks of 3 random bits in 2_2 wil return `vec![(prf0, 2), (prf1, 1), (prf2, 2), (prf3, 1)]`

For flat/vs pre-chunked I think both have advantages and drawbacks which probably dont impact much

tmontaigu:
*[`tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs` line 8 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OrmpuyY-kHfSbJaD9Je:-OtZZuYL8sO9ylbaCXKH:btt3gaf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4903fe4c6a69566f3de9f6749e46c8e0007d61f4/tfhe-benchmark/benches/high_level_api/bitonic_shuffle.rs#L8)):*

did you guys sync on this ?
</blockquote></details>

Yes, i'll do something in a followup

tmontaigu:
*[`tfhe/src/integer/server_key/radix_parallel/cmux.rs` line 839 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OtZKksJBS381_JI9kiK:-OtZ_2je5zvd8Oyhnndq:b2nign6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/823dbcaa053168e8b6fe8f4cfdb48b9726157feb/tfhe/src/integer/server_key/radix_parallel/cmux.rs#L839)):*

just so I'm clear one of the two blocks is 0 right ? just that the degree does not reflect it since we do'nt know which is strictly 0 ?

so the degree is whichever of the two was higher
</blockquote></details>

yes

IceTDrinker (commented):
> @IceTDrinker made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430)* status: all files reviewed, 2 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OtZmHJ77kJ9JkaY4Aj-:b-7t1wl9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

the returned types is `Vec<(Prf, u64)>` the u64 indicates how many bits of random are needed for each prf block, so asking for two chunks of 3 random bits in 2_2 wil return `vec![(prf0, 2), (prf1, 1), (prf2, 2), (prf3, 1)]`

For flat/vs pre-chunked I think both have advantages and drawbacks which probably dont impact much
</blockquote></details>

sorry I meant : does the GPU handle those incomplete blocks ?

For now it seems _bits is just ignored, so we could have something silently broken ?

IceTDrinker:
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-Ot_9cEgCXPlJvLUQoVu:b434bgj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

sorry I meant : does the GPU handle those incomplete blocks ?

For now it seems _bits is just ignored, so we could have something silently broken ?
</blockquote></details>

so looks like they manage it the same way we used to manage it : i.e. per one radix call, but not necessarily for a chunked approach

tmontaigu:
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OtcV3nCC-VPMwfI0pDy:bs0g19y) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

so looks like they manage it the same way we used to manage it : i.e. per one radix call, but not necessarily for a chunked approach
</blockquote></details>

Yes, to handle the chunked approach, they need to do some modifications, I cannot really make them

IceTDrinker (approved):
Thanks for the feature, worth a check-in with Enzo for GPU to make sure they are aware of the chunking needs



> @IceTDrinker made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-:-OtcYyO69Nw1obnyNMB9:b-jqs6cb)* status: all files reviewed, 1 unresolved discussion (waiting on mayeul-zama, nsarlin-zama, soonum, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-OtcYTAT5Oj3lLTxfutK:b4acirx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

Yes, to handle the chunked approach, they need to do some modifications, I cannot really make them
</blockquote></details>

agreed, did Enzo contact you for the shuffle and this use case ?

with the way we generate keys for the shuffle the chunking is actually doable from a "mega radix" which would contain all blocks since we force full message bits for random in each block

tmontaigu:
*[`tfhe/src/shortint/oprf.rs` line 883 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3430#-OszQS-v4punihPhxOpe:-Otcjpm-2puxzveRXmz4:bjqp0lu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/edf96c53c7006570994e00fa61331130838bf328/tfhe/src/shortint/oprf.rs#L883)):*

agreed, did Enzo contact you for the shuffle and this use case ?

with the way we generate keys for the shuffle the chunking is actually doable from a "mega radix" which would contain all blocks since we force full message bits for random in each block
</blockquote></details>

I touched a few words about it with him about this, i will re-tell him once merged

PR #3448, opened 2026-04-03
feat(gpu): optimize BLS12-446 field arithmetic for MSM performance

bbarbakadze:
### PR content/description

Optimize BLS12-446 field arithmetic for MSM performance

  - Replace 64-bit CIOS Montgomery multiplication with 32-bit MAD chains                                                                                                                                                                                 
    (mad.lo.cc/madc.hi.cc), exploiting native 2x throughput of 32-bit ops
    on NVIDIA GPUs via even/odd accumulator separation                                                                                                                                                                                                   
                                                                                                                                                                                                                                                         
  - Add fp_mont_sqr using a triangular MAD chain (upper triangle computed                                                                                                                                                                                
    once and doubled, diagonal added separately), saving ~40% of the                                                                                                                                                                                     
    multiplications versus treating squaring as a general multiplication                                                                                                                                                                                 
                  
  - Add fp_add_lazy/fp_sub_lazy (and Fp2 variants): skip the final                                                                                                                                                                                       
    conditional subtraction when the result feeds fp_mont_mul, which
    accepts inputs in [0, 2p). Wired into fp2_mont_mul, fp2_mont_square,                                                                                                                                                                                 
    and G1/G2 projective_point_double                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                         
  - Replace all fp_mont_mul(c, a, a) squaring patterns with fp_mont_sqr                                                                                                                                                                                  
    across curve.cu and fp2.cu (point addition, doubling, inversion)  


<!-- Feel free to delete the template if the PR (bumping a version e.g.) does not fit the template -->
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1346

### Check-list:

### DIFF
--- a/backends/zk-cuda-backend/cuda/include/curve.h
+++ b/backends/zk-cuda-backend/cuda/include/curve.h
@@ -17,7 +17,13 @@ __host__ __device__ void fp2_zero(Fp2 &a);
 
 // G1 point: (x, y) coordinates in Fp
 // Curve equation: y^2 = x^3 + b (short Weierstrass form with a = 0)
-struct G1Affine {
+//
+// alignas(sizeof(uint64_t)): The bool infinity field causes the struct to be
+// padded to the largest field alignment (4 bytes in 32-bit limb mode, 8 bytes
+// in 64-bit). Forcing alignment to sizeof(uint64_t) ensures
+// sizeof(G1Affine)==120 in both modes, matching the Rust FFI bindings which
+// are always generated from the 64-bit layout regardless of LIMB_BITS_CONFIG.
+struct alignas(sizeof(uint64_t)) G1Affine {
   Fp x;
   Fp y;
   bool infinity; // true if point at infinity (identity element)
@@ -36,7 +42,9 @@ struct G1Affine {
 
 // G2 point: (x, y) coordinates in Fp2
 // Curve equation: y^2 = x^3 + b' (twisted curve over Fp2)
-struct G2Affine {
+//
+// alignas(sizeof(uint64_t)): same ABI-stability reason as G1Affine above.
+struct alignas(sizeof(uint64_t)) G2Affine {
   Fp2 x;
   Fp2 y;
   bool infinity; // true if point at infinity (identity element)
--- a/backends/zk-cuda-backend/cuda/include/fp.h
+++ b/backends/zk-cuda-backend/cuda/include/fp.h
@@ -17,7 +17,7 @@
 // Supported values: 32, 64.
 // ============================================================================
 #ifndef LIMB_BITS_CONFIG
-#define LIMB_BITS_CONFIG 64
+#define LIMB_BITS_CONFIG 32
 #endif
 
 #if LIMB_BITS_CONFIG == 64
@@ -208,6 +208,17 @@ __host__ __device__ void fp_add(Fp &c, const Fp &a, const Fp &b);
 // MONTGOMERY: Both inputs and output must be in Montgomery form
 __host__ __device__ void fp_sub(Fp &c, const Fp &a, const Fp &b);
 
+// Lazy addition: c = a + b, output in [0, 2p) for inputs in [0, p).
+// Skips the final conditional subtraction of fp_add.
+// Safe as input to fp_mont_mul (CIOS accepts [0, 2p)); NOT safe for final
+// results or as input to fp_sub/fp_neg which require [0, p) inputs.
+__host__ __device__ void fp_add_lazy(Fp &c, const Fp &a, const Fp &b);
+
+// Lazy subtraction: c ≡ a - b (mod p), output in [0, 2p) for inputs in [0, p).
+// Adds p unconditionally, skipping the borrow-select of fp_sub.
+// Same safety concerns as fp_add_lazy.
+__host__ __device__ void fp_sub_lazy(Fp &c, const Fp &a, const Fp &b);
+
 // Multiplication: c = a * b (without reduction)
 // "Raw" means the operation is performed without modular reduction modulo p.
 // The result is stored in double-width (2*FP_LIMBS limbs) and may be >= p.
@@ -223,6 +234,11 @@ __host__ __device__ void fp_mont_reduce(Fp &c, const UNSIGNED_LIMB *a);
 // Montgomery multiplication: c = (a * b * R_INV) mod p
 // Both a and b are in Montgomery form, result is in Montgomery form
 __host__ __device__ void fp_mont_mul(Fp &c, const Fp &a, const Fp &b);
+
+// Montgomery squaring: c = (a^2 * R_INV) mod p
+// Both input and output in Montgomery form.
+// On device uses a triangular MAD chain (fewer multiplications).
+__host__ __device__ void fp_mont_sqr(Fp &c, const Fp &a);
 
 // CONVERSION: Input is normal form, output is Montgomery form
 __host__ __device__ void fp_to_montgomery(Fp &c, const Fp &a);
--- a/backends/zk-cuda-backend/cuda/include/fp2.h
+++ b/backends/zk-cuda-backend/cuda/include/fp2.h
@@ -72,6 +72,11 @@ __host__ __device__ void fp2_add(Fp2 &c, const Fp2 &a, const Fp2 &b);
 // Subtraction: c = a - b
 __host__ __device__ void fp2_sub(Fp2 &c, const Fp2 &a, const Fp2 &b);
 
+// Lazy add/sub: each component output in [0, 2p) for inputs in [0, p).
+// Safe as input to fp2_mont_mul; same contract as fp_add_lazy / fp_sub_lazy.
+__host__ __device__ void fp2_add_lazy(Fp2 &c, const Fp2 &a, const Fp2 &b);
+__host__ __device__ void fp2_sub_lazy(Fp2 &c, const Fp2 &a, const Fp2 &b);
+
 // Multiplication: c = a * b
 // (a0 + a1*i) * (b0 + b1*i) = (a0*b0 - a1*b1) + (a0*b1 + a1*b0)*i
 // NOTE: Assumes inputs are in normal form and converts to/from Montgomery
@@ -84,7 +89,7 @@ __host__ __device__ void fp2_mont_mul(Fp2 &c, const Fp2 &a, const Fp2 &b);
 // Montgomery squaring: c = a^2 (all in Montgomery form)
 // Uses the complex-squaring identity: c0 = (a0+a1)(a0-a1), c1 = 2*a0*a1
 // Only 2 Fp multiplications vs 3 for fp2_mont_mul(c, a, a).
-// NOTE: All inputs and outputs are in Montgomery form (no conversions)
+// NOTE: All inputs and output are in Montgomery form
 __host__ __device__ void fp2_mont_square(Fp2 &c, const Fp2 &a);
 
 // Squaring: c = a^2
--- a/backends/zk-cuda-backend/cuda/src/primitives/fp.cu
+++ b/backends/zk-cuda-backend/cuda/src/primitives/fp.cu
@@ -7,8 +7,6 @@
 #include <cuda_runtime.h>
 
 // For CUDA device code, we use __constant__ memory
-// Constants are hardcoded at compile time (like sppark) to avoid
-// cudaMemcpyToSymbol
 // Note: DEVICE_MODULUS is in normal form (not Montgomery)
 __constant__ const Fp DEVICE_MODULUS = {BLS12_446_MODULUS_LIMBS};
 
@@ -104,7 +102,7 @@ __host__ __device__ ComparisonType fp_cmp(const Fp &a, const Fp &b) {
 
 __host__ __device__ bool fp_is_zero(const Fp &a) {
   // By doing this way we avoid branching
-  uint64_t acc = 0;
+  UNSIGNED_LIMB acc = 0;
   for (int i = 0; i < FP_LIMBS; i++) {
     acc |= a.limb[i];
   }
@@ -114,8 +112,8 @@ __host__ __device__ bool fp_is_zero(const Fp &a) {
 __host__ __device__ bool fp_is_one(const Fp &a) {
   if (a.limb[0] != 1)
     return false;
-  // By doing this way we avoid branching
-  uint64_t acc = 0;
+  // All higher limbs must be zero.
+  UNSIGNED_LIMB acc = 0;
   for (int i = 1; i < FP_LIMBS; i++) {
     acc |= a.limb[i];
   }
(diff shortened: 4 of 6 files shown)


backends/zk-cuda-backend/cuda/include/fp.h
pdroalves: This PR should change this line to 32 by default.

pdroalves (changes requested):

backends/zk-cuda-backend/cuda/include/fp.h
bbarbakadze: Done

backends/zk-cuda-backend/cuda/include/curve.h
```diff
@@ -17,7 +17,12 @@ __host__ __device__ void fp2_zero(Fp2 &a);
 
 // G1 point: (x, y) coordinates in Fp
 // Curve equation: y^2 = x^3 + b (short Weierstrass form with a = 0)
-struct G1Affine {
+//
+// alignas(8): The bool infinity field causes the struct to be padded to the
+// largest field alignment (4 bytes in 32-bit limb mode, 8 bytes in 64-bit).
+// Forcing alignas(8) ensures sizeof(G1Affine)==120 in both modes, matching
+// the Rust FFI bindings which are always generated from the 64-bit layout.
+struct alignas(8) G1Affine {
```
pdroalves: Can you replace this magic number by a function based on `LIMB_BITS_CONFIG`?

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -207,6 +205,40 @@ __host__ __device__ UNSIGNED_LIMB fp_add_raw(Fp &c, const Fp &a, const Fp &b) {
         "l"(b.limb[1]), "l"(b.limb[2]), "l"(b.limb[3]), "l"(b.limb[4]),
         "l"(b.limb[5]), "l"(b.limb[6]));
   return carry_out;
+#elif defined(__CUDA_ARCH__) && LIMB_BITS_CONFIG == 32
+  // 32-bit PTX carry chain: add.cc.u32 sets the hardware carry flag,
+  // addc.cc.u32 propagates it. Eliminates software carry-detect comparisons
+  // across all 14 limbs.
+  // Operand map: %0..%13 = c[0..13], %14 = carry_out,
+  //              %15..%28 = a[0..13], %29..%42 = b[0..13].
+  uint32_t carry_out;
```
pdroalves, line 214: @guillermo-oyarzun do you want to double check this PTX? It seems ok to me.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -248,6 +280,41 @@ __host__ __device__ UNSIGNED_LIMB fp_sub_raw(Fp &c, const Fp &a, const Fp &b) {
...
   // borrow. Normalize to 0/1 for callers that check (borrow != 0) or add it.
   return borrow_out & 1;
+#elif defined(__CUDA_ARCH__) && LIMB_BITS_CONFIG == 32
+  // 32-bit PTX borrow chain: sub.cc.u32 sets the hardware borrow flag,
+  // subc.cc.u32 propagates it across all 14 limbs.
+  // subc.u32 with 0-0-BF gives 0xFFFFFFFF on borrow; normalize to 0/1.
+  // Operand map: %0..%13 = c[0..13], %14 = borrow_out,
+  //              %15..%28 = a[0..13], %29..%42 = b[0..13].
+  uint32_t borrow_out;
+  asm("sub.cc.u32   %0,  %15, %29;\n\t" // c[0]  = a[0]  - b[0],  set BF
```
pdroalves, line 290: @guillermo-oyarzun here too.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +860,411 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
 #endif // LIMB_BITS_CONFIG == 64
 #endif // __CUDA_ARCH__
 
+// 32-bit dual MAD-chain Montgomery multiplication (device path)
```
pdroalves, line 842: Do you have a reference for this MAD-chain multiplication? If so, a link as comment would help.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +860,411 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
...
+  fp_mul_n_32(wide + 2, a32 + 2, a32[0], n - 2);
+  fp_mul_n_32(wtemp, a32 + 1, a32[0], n);
+
+#pragma unroll
+  for (int i = 2; i <= n - 4; i += 2) {
+    fp_mad_row_32(&wide[2 * i], &wtemp[2 * i - 2], &a32[i], a32[i - 1], n - i);
+    fp_qad_row_32(&wtemp[2 * i], &wide[2 * i + 2], &a32[i + 1], a32[i], n - i);
+  }
+
+  asm("mul.lo.u32 %0, %2, %3; mul.hi.u32 %1, %2, %3;"
```
pdroalves: I don't like PTX in the middle of a function like this one. Maybe you could move it to a macro and add comments explaining what it is.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +860,411 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
...
+        : "r"(even[2 * j]), "r"(even[2 * j + 1]));
+
+  // subc.u64 0-0-borrow gives 0xFFFF... when c<p (keep), 0 when c>=p (reduce).
+  // shr.s64 sign-extends to a per-bit selection mask.
+  const uint64_t p0 = DEVICE_MODULUS.limb[0], p1 = DEVICE_MODULUS.limb[1],
+                 p2 = DEVICE_MODULUS.limb[2], p3 = DEVICE_MODULUS.limb[3],
+                 p4 = DEVICE_MODULUS.limb[4], p5 = DEVICE_MODULUS.limb[5],
+                 p6 = DEVICE_MODULUS.limb[6];
+  uint64_t r0, r1, r2, r3, r4, r5, r6, mask64;
+  asm("sub.cc.u64   %0, %8,  %15;\n\t"
```
pdroalves, line 1189: This diff is full of PTX. We need to careful read them and if possible remove them from within functions.

pdroalves (commented):
I added a batch of high-level comments. Let me know when you are done with this PR so I can do a careful review line by line.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
bbarbakadze: will do it

backends/zk-cuda-backend/cuda/include/curve.h
bbarbakadze: So this is actually not dependent on a `LIMB_BITS_CONFIG`, it depends on the layout Rust is using.  So once 64-bit limbs are used, we need the same alignment. still the magic number is replaced with `sizeof`

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
guillermo-oyarzun: yup the ptx looks good!

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
guillermo-oyarzun: same here, it looks good too!

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +838,440 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
...
+
 // CIOS (Coarsely Integrated Operand Scanning) Montgomery multiplication
 // Fuses multiplication and reduction in a single pass for better efficiency.
 // Uses only FP_LIMBS+1 limbs of working space instead of 2*FP_LIMBS.
 // Both a and b are in Montgomery form, result is in Montgomery form.
 __host__ __device__ void fp_mont_mul_cios(Fp &c, const Fp &a, const Fp &b) {
-#if defined(__CUDA_ARCH__) && LIMB_BITS_CONFIG == 64
-  // Device path: fully unrolled PTX with hardware carry flags
-  fp_mont_mul_cios_ptx(c, a, b);
+#ifdef __CUDA_ARCH__
```
guillermo-oyarzun, line 1240: i understand that now we have 2 versions for 32 and 64-bit limbs, can we add a panic in the correct place in case someone attempts to use it with 128-bit?

bbarbakadze: @guillermo-oyarzun you mean if someone tries to set value other than 32 and 64 to `LIMB_BITS_CONFIG`

bbarbakadze: in this case maybe use enum? with two values `32BIT` and `64BIT`

guillermo-oyarzun: yup enum should work, just trying be extra safe because the code shouldn't work with 128-bit, right? we would need to emulate them somehow

bbarbakadze: for now limbs can only be 32 or 64 I will rewrite it with enum, should be better than panic.

bbarbakadze: btw there is already a protection implemented for this inside fp.h line:55

`static_assert(LIMB_BITS == 32 || LIMB_BITS == 64, "LIMB_BITS_CONFIG must be 32 or 64");`

So I guess it is fine to leave it as it is.

backends/zk-cuda-backend/cuda/include/fp.h
```diff
@@ -18,7 +18,7 @@
 // Supported values: 32, 64.
 // ============================================================================
 #ifndef LIMB_BITS_CONFIG
-#define LIMB_BITS_CONFIG 64
+#define LIMB_BITS_CONFIG 32
 #endif
 
 #if LIMB_BITS_CONFIG == 64
```
pdroalves, line 23: We cannot forget to deprecate this. Once we merge this PR we should completely remove the 64-bit mode. Do you agree @bbarbakadze ?

backends/zk-cuda-backend/cuda/include/fp2.h
```diff
@@ -84,7 +89,7 @@ __host__ __device__ void fp2_mont_mul(Fp2 &c, const Fp2 &a, const Fp2 &b);
 // Montgomery squaring: c = a^2 (all in Montgomery form)
 // Uses the complex-squaring identity: c0 = (a0+a1)(a0-a1), c1 = 2*a0*a1
 // Only 2 Fp multiplications vs 3 for fp2_mont_mul(c, a, a).
-// NOTE: All inputs and outputs are in Montgomery form (no conversions)
+// NOTE: All inputs should be in Montgomery form
```
pdroalves: What about outputs?

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +838,440 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
...
+    // madc_n_rshift.
+    asm("mul.lo.u32 %0, %1, %2;" : "=r"(mi) : "r"(even[0]), "r"(M0));
+    fp_madc_n_rshift_32(odd, p + 1, mi, n);
+    fp_cmad_n_32(even, p, mi, n);
+    asm("addc.u32 %0, %0, 0;" : "+r"(odd[n - 1]));
+  }
+}
+
+// Montgomery squaring using CIOS with triangular 32-bit MAD chains.
+// See fp_mont_mul_mad32 for the algorithm reference (Koç et al., 1996).
```
pdroalves, line 1039: Where can I find this reference? Could be good to add the paper name and venue with full author names.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -483,14 +590,32 @@ __host__ __device__ void fp_mont_reduce(Fp &c, const UNSIGNED_LIMB *a) {
...
-    // Add u * p to t, starting at position i
+    // Add u * p to t, starting at position i.
+    // Use uint64_t accumulator in 32-bit mode to avoid carry overflow:
+    // hi + carry1 + carry2 can reach 2^32 which overflows uint32_t.
+#if LIMB_BITS_CONFIG == 32
+    uint64_t carry = 0;
+    for (int j = 0; j < FP_LIMBS; j++) {
+      uint64_t acc =
+          (uint64_t)t[i + j] + (uint64_t)u * (uint64_t)p.limb[j] + carry;
+      t[i + j] = (UNSIGNED_LIMB)acc;
```
pdroalves: We've been slowly trying to avoid this type of cast in new C++ code. In the ZK backend that's a convention that we should be following.

Instead of 

```
a = (type_t) b
```

you should be using

```
a = static_cast<type_t>(b)
```.

File: backends/zk-cuda-backend/cuda/src/primitives/fp.cu [hunk 12]
```diff
@@ -483,14 +590,32 @@ __host__ __device__ void fp_mont_reduce(Fp &c, const UNSIGNED_LIMB *a) {
...
+      carry = acc >> LIMB_BITS;
+    }
+    // Propagate remaining carry (carry ≤ 2^32-1 at this point)
+    int idx = i + FP_LIMBS;
+    while (carry != 0 && idx <= 2 * FP_LIMBS) {
+      uint64_t acc = (uint64_t)t[idx] + carry;
+      t[idx] = (UNSIGNED_LIMB)acc;
```
pdroalves: Here too.

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
pdroalves: I asked code to point other lines changed in this PR that needs to be fixed.

```
● 14 C-style casts in new PR lines, across two functions:
                                                                                                                                                                                              
  fp_mont_reduce (32-bit path), lines 596-611:                                                                                                                                                
                                                                                                                                                                                              
  ┌──────┬──────────────────────────────────────────────────────┐                                                                                                                             
  │ Line │                         Cast                         │                                                                                                                             
  ├──────┼──────────────────────────────────────────────────────┤                                                                                                                             
  │ 600  │ (uint64_t)t[i + j], (uint64_t)u, (uint64_t)p.limb[j] │
  ├──────┼──────────────────────────────────────────────────────┤
  │ 601  │ (UNSIGNED_LIMB)acc                                   │                                                                                                                             
  ├──────┼──────────────────────────────────────────────────────┤
  │ 607  │ (uint64_t)t[idx]                                     │                                                                                                                             
  ├──────┼──────────────────────────────────────────────────────┤   
  │ 608  │ (UNSIGNED_LIMB)acc                                   │
  └──────┴──────────────────────────────────────────────────────┘                                                                                                                             
  
  fp_mont_mul_cios (32-bit Step 1), lines 1263-1273:                                                                                                                                          
                                                                    
  ┌──────┬──────────────────────────────────────────────────────────┐
  │ Line │                           Cast                           │
  ├──────┼──────────────────────────────────────────────────────────┤
  │ 1267 │ (uint64_t)t[j], (uint64_t)a.limb[j], (uint64_t)b.limb[i] │
  ├──────┼──────────────────────────────────────────────────────────┤
  │ 1268 │ (UNSIGNED_LIMB)acc                                       │                                                                                                                         
  ├──────┼──────────────────────────────────────────────────────────┤                                                                                                                         
  │ 1271 │ (uint64_t)t[FP_LIMBS]                                    │                                                                                                                         
  ├──────┼──────────────────────────────────────────────────────────┤                                                                                                                         
  │ 1272 │ (UNSIGNED_LIMB)(sum64 >> LIMB_BITS)                      │
  ├──────┼──────────────────────────────────────────────────────────┤                                                                                                                         
  │ 1273 │ (UNSIGNED_LIMB)sum64                                     │
  └──────┴──────────────────────────────────────────────────────────┘                                                                                                                         
                                                                    
  fp_mont_mul_cios (32-bit Step 2), lines 1298-1310:

  ┌──────┬──────────────────────────────────────────────────┐                                                                                                                                 
  │ Line │                       Cast                       │
  ├──────┼──────────────────────────────────────────────────┤                                                                                                                                 
  │ 1302 │ (uint64_t)t[j], (uint64_t)m, (uint64_t)p.limb[j] │       
  ├──────┼──────────────────────────────────────────────────┤
  │ 1303 │ (UNSIGNED_LIMB)acc                               │
  ├──────┼──────────────────────────────────────────────────┤                                                                                                                                 
  │ 1308 │ (uint64_t)t[FP_LIMBS], (uint64_t)overflow        │
  ├──────┼──────────────────────────────────────────────────┤                                                                                                                                 
  │ 1309 │ (UNSIGNED_LIMB)s64                               │       
  ├──────┼──────────────────────────────────────────────────┤                                                                                                                                 
  │ 1310 │ (UNSIGNED_LIMB)(s64 >> LIMB_BITS)                │
  └──────┴──────────────────────────────────────────────────┘                                                                                                                                 
                                                                    
  All 14 are widening (uint32_t → uint64_t) or truncating (uint64_t → uint32_t) integer casts that should use static_cast<>.  
```

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
```diff
@@ -735,26 +838,440 @@ __device__ __noinline__ void fp_mont_mul_cios_ptx(Fp &c, const Fp &a,
...
+    fp_mul_by_1_row_32(&red_odd[0], &wide[0], p32, M0, false);
+  }
+  // Merge the final red_odd word into wide[0..n-1].
+  fp_cadd_n_32(&wide[0], &red_odd[1], n - 1);
+  FP_ADDC_32(wide[n - 1]);
+
+  // Add reduced lower half into upper half wide[n..2n-1]; the result lives
+  // in wide[n..2n-1] and is in [0, 2p).
+  fp_cadd_n_32(&wide[n], &wide[0], n);
+  FP_CARRY_32(wide[0]); // discard overflow (always 0 for p<2^446)
```
pdroalves, line 1096: Do we need this line?  The comment says the overflow is always 0, is that right?

pdroalves: Ahh I think I get it, you are just consuming the carry right? Maybe the comment should be replaced by `consume the carry flag so CC is clean`

pdroalves (changes requested):
Just a few minor comments and style changes. 

The PR is quite good in my opinion. Is there anything else you need to do here? Otherwise it's a good moment to rebase. We can merge after these changes.

backends/zk-cuda-backend/cuda/include/fp.h
bbarbakadze: I don't see any use of it, so yes we can remove it.

backends/zk-cuda-backend/cuda/include/fp2.h
bbarbakadze: Yes, both should be in a Montgomery form

backends/zk-cuda-backend/cuda/src/primitives/fp.cu
bbarbakadze: I would keep it. should not affect performance, but makes it more generic.

pdroalves (approved):

andrei-stoian-zama (approved):

PR #3460, opened 2026-04-10
Tm/bind contains

tmontaigu:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460)

### DIFF
--- a/tfhe/src/high_level_api/array/mod.rs
+++ b/tfhe/src/high_level_api/array/mod.rs
@@ -24,6 +24,8 @@ pub use traits::{IOwnedArray, Slicing, SlicingMut};
 use crate::array::stride::DynDimensions;
 use crate::core_crypto::prelude::{Numeric, OverflowingAdd, SignedNumeric, UnsignedNumeric};
 use crate::integer::block_decomposition::DecomposableInto;
+#[cfg(feature = "gpu")]
+use crate::integer::gpu::ciphertext::CudaIntegerRadixCiphertext;
 use crate::integer::RadixCiphertext;
 use crate::prelude::{CastFrom, CastInto};
 pub use cpu::{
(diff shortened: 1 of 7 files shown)


IceTDrinker (approved):
It's looking good to me, worth having a second pair of eyes, but approving, thanks !

> @IceTDrinker reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460#-:-OqH0haHA0NjfbF-2drM:bqed7nh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on IceTDrinker and nsarlin-zama).

nsarlin-zama (approved):
looks good to me ! Small api idea but not necessary

> @nsarlin-zama reviewed 4 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460#-:-OqKM9mxEpkSPDPjsoC-:bdnhaz1)* status: all files reviewed, 1 unresolved discussion (waiting on tmontaigu).

___
*[`tfhe/src/high_level_api/integers/signed/base.rs` line 258 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460#-OqKLv5G9ipKSrV73YTv:-OqKLv5G9ipKSrV73YTw:b-rv5peh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b8e7dfa6beb6e9c2c47da9a6fdf4a282098ec86/tfhe/src/high_level_api/integers/signed/base.rs#L258)):*
> ```XML
>     /// assert!(decrypted);
>     /// ```
>     pub fn contains(cts: &[Self], value: &Self) -> FheBool {
> ```

If we want methods directly on the array, I think we can create an extension trait like FheArrayOps and do `impl<T> FheArrayOps for [T] where T: FheIntegerType`
So users can do `array.fhe_contains(val)` (contains will likely be already taken)

tmontaigu:
*[`tfhe/src/high_level_api/integers/signed/base.rs` line 258 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460#-OqKLv5G9ipKSrV73YTv:-OqLdavM0c7LzR2zbxG3:b-q7asfn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b8e7dfa6beb6e9c2c47da9a6fdf4a282098ec86/tfhe/src/high_level_api/integers/signed/base.rs#L258)):*

If we want methods directly on the array, I think we can create an extension trait like FheArrayOps and do `impl<T> FheArrayOps for [T] where T: FheIntegerType`
So users can do `array.fhe_contains(val)` (contains will likely be already taken)
</blockquote></details>

That is indeed a good idea we should keep it in my for the next time we work on array

tmontaigu:
*[`tfhe/src/high_level_api/integers/signed/base.rs` line 258 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3460#-OqKLv5G9ipKSrV73YTv:-OqLe52-CKAppwQcGVtG:b-kdo7od) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b8e7dfa6beb6e9c2c47da9a6fdf4a282098ec86/tfhe/src/high_level_api/integers/signed/base.rs#L258)):*

That is indeed a good idea we should keep it in my for the next time we work on array
</blockquote></details>

I could also maybe solve the naming problem for when in [T] T is a clear, but it might not be as easy to due rules about conflicting imp

PR #3461, opened 2026-04-10
chore(gpu): optimize CI

andrei-stoian-zama:
- integer_long_run_tests scheduled on 4xL40 changed to run weekly instead of nightly
- integer_long_run_tests should only run when relevant files are changed
- Moves H100 tests to "approved" if no changes are made in core crypto
- Only run ZK tests if there are changes to relevant parts since there is no dependency to core_crypto/integer
- Do not trigger workflows that don't run doc tests on doc changes
- Workflows shouldn't run when only other workflows have changed

### DIFF
--- a/.github/workflows/gpu_core_h100_tests.yml
+++ b/.github/workflows/gpu_core_h100_tests.yml
@@ -23,7 +23,7 @@ on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
-    types: [ labeled ]
+    types: [ labeled, opened, synchronize ]
 
 permissions:
   contents: read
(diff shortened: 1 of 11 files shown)


.github/workflows/gpu_core_h100_tests.yml
```diff
@@ -23,7 +23,7 @@ on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
-    types: [ labeled ]
+    types: [ labeled, opened, synchronize ]
```
soonum, line 26: This would trigger the workflow on PR opening and on any commit push. Is that what we want ?

soonum: I think I get it. You want to run this workflow if there is any changes in core_crypto code.

soonum (approved):
Nice ! Let's save some CI time !

soonum:
A nitpick though, could you elaborate a bit more in your commit title/message ? For history purposes.

enzodimaria (approved):
That's great!

PR #3464, opened 2026-04-13
fix(gpu): AES noise fix

enzodimaria:
This PR contains

about **AES/AES256**:
- Noise fix, before 6 additions were performed at multiple places in the process.


andrei-stoian-zama (approved):

bbarbakadze (approved):
Thanks for the fix @enzodimaria

PR #3467, opened 2026-04-13
chore: allow rand audit advisory

nsarlin-zama:

### PR content/description
Fix a cargo audit warning on rand 0.8, caused by an unsoundness triggered by the use of a custom logger. Upgrading rand would be too complex now because of arkworks is still depending on 0.8.
PR to update to 0.9 exists in arkworks but has not been merged yet, and looks stalled: https://github.com/arkworks-rs/std/pull/56


IceTDrinker (approved):
thanks!

PR #3468, opened 2026-04-13
doc(bench): update benchmark results tables

soonum:
Please check if changes looks good to you.
I know that there are some discrepancies regarding GPU PBS benchmarks,  @andrei-stoian-zama could you point me the culprits and the associated values please?


andrei-stoian-zama (changes requested):
I had updated the two GPU pbs ones myself quite recently. there was an issue with the automatic creation of the tables. While for the 2^-128 pfail it looks ok for the 2^-64 one it doesn't. I would prefer you revert at least for the latter one

see my previous change here: https://github.com/zama-ai/tfhe-rs/commit/a7476d0aaae371750152c1d9af1e7d19f5751f68#diff-d7332a221d2063eaadc93fbe57b4e4f8f0671ae6853a054cbf13125cf232599e

soonum:
I reverted both of them @andrei-stoian-zama.

andrei-stoian-zama (approved):

IceTDrinker (approved):
Thanks !

PR #3471, opened 2026-04-14
fix(bench): wrong size in wasm benchmarks

nsarlin-zama:

### PR content/description
since bench name format changed I guess, the size bench results were multiplied by 1000_000 because the "_size" marker was not found in the short name


soonum (approved):
Thanks for the quick fix mate !

PR #3472, opened 2026-04-14
feat(gpu): implement shuffle

enzodimaria:
This PR contains the new feature FHE **bitonic_shuffle**

```
+-----+------------+------------+------------+------------+----------+------------+
|   n | 1 GPU 2⁻²⁰ | 8 GPU 2⁻²⁰ | 1 GPU 2⁻³⁰ | 8 GPU 2⁻³⁰ | CPU 2⁻⁴⁰ | 8 GPU 2⁻⁴⁰ |
+-----+------------+------------+------------+------------+----------+------------+
|  32 |     4.82 s |    0.760 s |     5.46 s |    0.877 s |   7.10 s |    0.954 s |
|  64 |    13.59 s |     1.97 s |    15.48 s |     2.25 s |  17.59 s |     2.48 s |
| 128 |    37.52 s |     5.32 s |    41.64 s |     5.92 s |  43.71 s |     6.61 s |
| 256 |    97.82 s |    13.77 s |   109.30 s |    15.33 s | 109.91 s |    16.98 s |
+-----+------------+------------+------------+------------+----------+------------+
params MB 2_2 GP4
```

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
@@ -1039,6 +1039,24 @@ void cuda_cast_to_signed_64_async(CudaStreamsFFI streams,
 
 void cleanup_cuda_cast_to_signed_64(CudaStreamsFFI streams,
                                     int8_t **mem_ptr_void);
+
+uint64_t scratch_cuda_integer_bitonic_sort_64_async(
+    CudaStreamsFFI streams, int8_t **mem_ptr, uint32_t glwe_dimension,
+    uint32_t polynomial_size, uint32_t big_lwe_dimension,
+    uint32_t small_lwe_dimension, uint32_t ks_level, uint32_t ks_base_log,
+    uint32_t pbs_level, uint32_t pbs_base_log, uint32_t grouping_factor,
+    uint32_t num_radix_blocks, uint32_t num_values, uint32_t message_modulus,
+    uint32_t carry_modulus, PBS_TYPE pbs_type, bool is_signed,
+    bool allocate_gpu_memory, PBS_MS_REDUCTION_T noise_reduction_type);
+
+void cuda_integer_bitonic_sort_64_async(CudaStreamsFFI streams,
+                                        CudaRadixCiphertextFFI **values,
+                                        uint32_t num_values, int8_t *mem_ptr,
+                                        void *const *bsks, void *const *ksks,
+                                        int32_t direction);
+
+void cleanup_cuda_integer_bitonic_sort_64(CudaStreamsFFI streams,
+                                          int8_t **mem_ptr_void);
 } // extern C
 
 #endif // CUDA_INTEGER_H
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer.h
@@ -981,6 +981,40 @@ void cuda_cast_to_signed_64_async(CudaStreamsFFI streams,
 
 void cleanup_cuda_cast_to_signed_64(CudaStreamsFFI streams,
                                     int8_t **mem_ptr_void);
+
+uint64_t scratch_cuda_integer_bitonic_shuffle_64_async(
+    CudaStreamsFFI streams, int8_t **mem_ptr,
+    CudaLweBootstrapKeyParamsFFI bsk_params,
+    CudaLweKeyswitchKeyParamsFFI ksk_params, uint32_t key_num_radix_blocks,
+    uint32_t data_num_radix_blocks, uint32_t num_values,
+    uint32_t message_modulus, uint32_t carry_modulus, bool allocate_gpu_memory,
+    PBS_MS_REDUCTION_T noise_reduction_type);
+
+void cuda_integer_bitonic_shuffle_64_async(CudaStreamsFFI streams,
+                                           CudaRadixCiphertextFFI **keys,
+                                           CudaRadixCiphertextFFI **values,
+                                           uint32_t num_values, int8_t *mem_ptr,
+                                           void *const *bsks,
+                                           void *const *ksks);
+
+void cleanup_cuda_integer_bitonic_shuffle_64(CudaStreamsFFI streams,
+                                             int8_t **mem_ptr_void);
+
+uint64_t scratch_cuda_integer_oprf_bitonic_shuffle_64_async(
+    CudaStreamsFFI streams, int8_t **mem_ptr,
+    CudaLweBootstrapKeyParamsFFI bsk_params,
+    CudaLweKeyswitchKeyParamsFFI ksk_params, uint32_t key_num_radix_blocks,
+    uint32_t data_num_radix_blocks, uint32_t num_values,
+    uint32_t message_modulus, uint32_t carry_modulus, bool allocate_gpu_memory,
+    PBS_MS_REDUCTION_T noise_reduction_type);
+
+void cuda_integer_oprf_bitonic_shuffle_64_async(
+    CudaStreamsFFI streams, CudaRadixCiphertextFFI **values,
+    uint32_t num_values, const void *seeded_lwe_input, int8_t *mem_ptr,
+    void *const *oprf_bsks, void *const *bsks, void *const *ksks);
+
+void cleanup_cuda_integer_oprf_bitonic_shuffle_64(CudaStreamsFFI streams,
+                                                  int8_t **mem_ptr_void);
 } // extern C
 
 #endif // CUDA_INTEGER_H
--- a/backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
@@ -1,0 +1,217 @@
+#pragma once
+#include "checked_arithmetic.h"
+#include "comparison.h"
+#include "integer_utilities.h"
+
+template <typename Torus> struct int_bitonic_sort_buffer {
+  int_radix_params params;
+  uint32_t max_num_pairs;
+  uint32_t num_radix_blocks;
+
+  int_comparison_buffer<Torus> *comparison_mem;
+  CudaRadixCiphertextFFI *comparison_results;
+
+  // Unsigned compare path: K*N packed, K*N/2 diffs, K*N/2 tree scratch (x, y).
+  CudaRadixCiphertextFFI *batch_cmp_packed;
+  CudaRadixCiphertextFFI *batch_cmp_comparisons;
+  CudaRadixCiphertextFFI *batch_cmp_tree_x;
+  CudaRadixCiphertextFFI *batch_cmp_tree_y;
+
+  int_radix_lut<Torus> *batch_identity_lut;
+  int_radix_lut<Torus> *batch_is_non_zero_lut;
+  int_radix_lut<Torus> *batch_inner_tree_leaf_lut;
+  int_radix_lut<Torus> *batch_last_tree_leaf_lut;
+  Torus *preallocated_h_lut;
+
+  // Batched cmux: 4KN = [true-side (2KN)] [false-side (2KN)]; condition
+  // broadcast per block.
+  CudaRadixCiphertextFFI *batch_buffer_in;
+  CudaRadixCiphertextFFI *batch_buffer_out;
+  CudaRadixCiphertextFFI *batch_condition;
+
+  int_radix_lut<Torus> *batch_predicate_lut;
+  int_radix_lut<Torus> *batch_message_extract_lut;
+
+  bool is_signed;
+  bool gpu_memory_allocated;
+
+  int_bitonic_sort_buffer(CudaStreams streams, int_radix_params params,
+                          uint32_t num_radix_blocks, uint32_t num_values,
+                          bool is_signed, bool allocate_gpu_memory,
+                          uint64_t &size_tracker) {
+    this->params = params;
+    this->is_signed = is_signed;
+    this->gpu_memory_allocated = allocate_gpu_memory;
+    this->num_radix_blocks = num_radix_blocks;
+    this->max_num_pairs = num_values / 2;
+
+    uint32_t K = max_num_pairs;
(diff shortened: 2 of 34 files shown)


andrei-stoian-zama (commented):
Should the API work with [GpuFheXYArray](https://docs.rs/tfhe/latest/tfhe/array/type.GpuFheUint32Array.html) ?

enzodimaria:
> Should the API work with [GpuFheXYArray](https://docs.rs/tfhe/latest/tfhe/array/type.GpuFheUint32Array.html) ?

Yes but I have to write the HL api before, I'll do it after the merge of the CPU shuffle :+1:

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
```diff
@@ -0,0 +1,461 @@
...
+#include "oprf.h"
+
+// Per-half-step scratch for the bitonic sort on K = num_values/2 pairs of
+// N-block radix integers: per-pair compare scratch and the 4*K*N batched
+// cmux buffers with their LUTs.
+// Attributes:
+// - params: shared radix parameters
+// - max_num_pairs, num_radix_blocks: K and N
+// - comparison_mem, comparison_results: per-pair compare scratch and K signs
+// - batch_cmp_*, batch_*_lut, preallocated_h_lut: legacy batched compare
```
andrei-stoian-zama: are these fields not actually used ? then remove them ?

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
```diff
@@ -0,0 +1,461 @@
...
+                                  uint32_t data_num_blocks, uint32_t num_values,
+                                  bool data_is_signed, bool allocate_gpu_memory,
+                                  uint64_t &size_tracker) {
+    this->params = params;
+    this->gpu_memory_allocated = allocate_gpu_memory;
+    this->num_values = num_values;
+    this->key_num_blocks = key_num_blocks;
+
+    uint64_t message_bits_per_block =
+        (uint64_t)std::log2(params.message_modulus);
```
andrei-stoian-zama: use log2_int

tfhe/src/integer/gpu/server_key/radix/shuffle.rs
```diff
@@ -0,0 +1,321 @@
...
+        if keys[1..]
+            .iter()
+            .any(|k| k.as_ref().d_blocks.lwe_ciphertext_count().0 != key_num_blocks)
+        {
+            return Err(crate::Error::new(
+                "all keys must have the same number of blocks".to_string(),
+            ));
+        }
+
+        for v in data.iter_mut() {
```
andrei-stoian-zama, line 216: why do we need this? shouldn't the caller ensure that carries are empty ?

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
```diff
@@ -0,0 +1,533 @@
...
+      Torus lsb = x & (message_modulus - 1);
+      return sign_handler_f((msb == IS_EQUAL) ? lsb : msb);
+    };
+  } else {
+    y = x;
+    f = sign_handler_f;
+  }
+
+  auto active = streams.active_gpu_subset(total_blocks, params.pbs_type);
+  last_lut->generate_and_broadcast_lut(active, {0}, {f}, LUT_0_FOR_ALL_BLOCKS,
```
andrei-stoian-zama: generate_and_broadcast_lut should be done in the scratch function

andrei-stoian-zama (changes requested):
Looks good but please remove:
                                                                                                                                                          
Two dead functions:
  - batched_tree_sign_reduction — only called from host_batched_unsigned_comparison                                                                          
  - host_batched_unsigned_comparison — never called (explicitly flagged "Unused" in its own comment)

  Nine dead fields on int_bitonic_sort_buffer, each with alloc + release:
  - batch_cmp_packed — only in host_batched_unsigned_comparison
  - batch_cmp_comparisons — only in host_batched_unsigned_comparison
  - batch_identity_lut — only in host_batched_unsigned_comparison
  - batch_is_non_zero_lut — only in host_batched_unsigned_comparison
  - batch_cmp_tree_x — only in batched_tree_sign_reduction
  - batch_cmp_tree_y — only in batched_tree_sign_reduction
  - batch_inner_tree_leaf_lut — only in batched_tree_sign_reduction
  - batch_last_tree_leaf_lut — only in batched_tree_sign_reduction
  - preallocated_h_lut — only in batched_tree_sign_reduction

andrei-stoian-zama:
For a next PR:

- batch the comparisons (based on initial work in host_batched_unsigned_comparison)
- the cmux implemented here (shuffle.cuh:283) performs a "message extract" step to clean noise. but we might not need to clean the noise since the subsequent comparison first performs a subtraction, then cleans the noise (so it can pack) - only works for message_modulus==carry_modulus
- explore if a single kernel can perform the various copy operations that are done on loops atm

andrei-stoian-zama (commented):
I'd also like to have a better explanation of the algorithm in the code comments. 

I asked Claude to produce pseudo code. Could you confirm this pseudo code is correct and then break it up and copy it into comments next to the code blocks that implement it ?

```

High-Level Pseudocode

  FUNCTION bitonic_shuffle_with_keys(keys[], data[], n):
      # Pad to next power of 2
      padded_n = next_power_of_2(n)
      FOR i IN [n, padded_n):
          keys[i] = MAX_VALUE   # sentinel: always sorts to end
          data[i] = 0

      # Bitonic network
      FOR k = 2, 4, 8, ... while k <= padded_n:
          FOR j = k/2, k/4, ... while j >= 1:
              bitonic_substep(keys, data, padded_n, k, j)

      RETURN keys[0..n], data[0..n]   # drop sentinels


  FUNCTION bitonic_substep(keys[], data[], n, k, j):
      # Step 1: Compare all pairs in parallel (one PBS per block)
      FOR each i where (i XOR j) > i:
          l = i XOR j
          ascending = ((i AND k) == 0)
          sign[i] = FHE_compare(keys[i], keys[l])   # → {INF, EQ, SUP}
     # Step 2: Conditional swap keys (batched CMUX)
      FOR each pair (i, l):
          should_swap = ascending ? (sign == SUP) : (sign == INF)
          (keys[i], keys[l]) = CMUX(should_swap,
                                     (keys[l], keys[i]),   # swapped
                                     (keys[i], keys[l]))   # unchanged

      # Step 3: Same CMUX for data (reuse comparison result)
      FOR each pair (i, l):
          (data[i], data[l]) = CMUX(should_swap,
                                     (data[l], data[i]),
                                     (data[i], data[l]))

  ---
  CMUX (Conditional Multiplexing) — the core primitive

  Since we can't branch on encrypted values, swaps are done via:

  FUNCTION CMUX(condition, true_val, false_val):
     # condition ∈ {INF=0, EQ=1, SUP=2}

      # Bivariate PBS: zero out the losing branch
      out_true  = bivariate_PBS(true_val,  condition,
                      LUT: (b, c) -> if c == SUP then b else 0)
      out_false = bivariate_PBS(false_val, condition,
                      LUT: (b, c) -> if c != SUP then b else 0)

      # Add: exactly one branch is nonzero
      result = HE_add(out_true, out_false)
      result = message_extract(result)    # clean up noise
      RETURN result

```

andrei-stoian-zama:
Could you please also look into refactoring CMUX : could the FheUintXY Cmux call into the batched version with a batch of 1 ? then a single function could be used in shuffle and the single value cmux operation

enzodimaria:
- removing msg extract cannot work, the noise would be equal to 6
- concerning cmux, I'll see in an other pr if we don't lose perfs

backends/tfhe-cuda-backend/cuda/include/integer/integer.h
```diff
@@ -981,6 +981,44 @@ void cuda_cast_to_signed_64_async(CudaStreamsFFI streams,
 
 void cleanup_cuda_cast_to_signed_64(CudaStreamsFFI streams,
                                     int8_t **mem_ptr_void);
+
+uint64_t scratch_cuda_integer_bitonic_shuffle_64_async(
+    CudaStreamsFFI streams, int8_t **mem_ptr, uint32_t glwe_dimension,
+    uint32_t polynomial_size, uint32_t big_lwe_dimension,
+    uint32_t small_lwe_dimension, uint32_t ks_level, uint32_t ks_base_log,
+    uint32_t pbs_level, uint32_t pbs_base_log, uint32_t grouping_factor,
```
guillermo-oyarzun: we can use CudaLweBootstrapKeyParamsFFI and CudaLweKeyswitchKeyParamsFFI

backends/tfhe-cuda-backend/cuda/include/integer/integer.h
```diff
@@ -981,6 +981,44 @@ void cuda_cast_to_signed_64_async(CudaStreamsFFI streams,
...
+                                           void *const *bsks,
+                                           void *const *ksks);
+
+void cleanup_cuda_integer_bitonic_shuffle_64(CudaStreamsFFI streams,
+                                             int8_t **mem_ptr_void);
+
+uint64_t scratch_cuda_integer_oprf_bitonic_shuffle_64_async(
+    CudaStreamsFFI streams, int8_t **mem_ptr, uint32_t glwe_dimension,
+    uint32_t polynomial_size, uint32_t big_lwe_dimension,
+    uint32_t small_lwe_dimension, uint32_t ks_level, uint32_t ks_base_log,
```
guillermo-oyarzun: same here

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
```diff
@@ -0,0 +1,546 @@
...
+    if (k_min < 3)
+      k_min = 3;
+    this->key_num_blocks_padded = k_min;
+    this->needs_pad = needs_pad_n || (key_num_blocks_padded != key_num_blocks);
+
+    this->batched_compare = new int_batched_compare_buffer<Torus>(
+        streams, params, padded_num_values / 2, key_num_blocks_padded,
+        allocate_gpu_memory, size_tracker);
+    this->fused_cmux = new int_fused_cmux_buffer<Torus>(
+        streams, params, padded_num_values / 2, key_num_blocks_padded,
```
guillermo-oyarzun, line 373: what happens with num_values=1, it seems like here i would create luts of size 0? is num_values =1 a case?

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
```diff
@@ -0,0 +1,546 @@
...
+    bool needs_pad_n = (num_values & (num_values - 1)) != 0;
+
+    uint32_t on = 1;
+    while (on < num_values)
+      on <<= 1;
+    this->padded_num_values = needs_pad_n ? on : num_values;
+
+    uint32_t k_min = needs_pad_n ? (key_num_blocks + 1) : key_num_blocks;
+    if (k_min < 3)
+      k_min = 3;
```
guillermo-oyarzun, line 363: where is that 3 coming from?

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
```diff
@@ -0,0 +1,825 @@
...
+ * @param n_blocks_per_row     Number of input blocks per row.
+ * @param factor               Packing scale applied to paired MSB blocks.
+ * @param orphan_factor        Scale applied to the orphan block (1 = copy
+ *                             verbatim; factor = treat as MSB).
+ * @param orphan_body_sentinel Constant injected at the body coefficient of
+ *                             the orphan output (0 or some Torus-encoded value
+ * : clear_value * delta).
+ */
+template <typename Torus>
+__global__ void pack_bivariate_adjacent_blocks_kernel(
```
guillermo-oyarzun, line 31: what is the difference with device_pack_blocks?

andrei-stoian-zama (dismissed):

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
andrei-stoian-zama: pair-packing of adjacent blocks + tree reduction last step

pair-packing only 2 blocks (a single FheUint4) results in a single block in the comparison lwe list. the Tree reduction does not support this, as the last step of the tree reduction makes an assumption about the input being produced by a tree-reduction step. here the comparison result would be fed into the final step of tree reduction directly.

I can add a comment

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
andrei-stoian-zama: it's pretty much the same but: 

pack_bivariate_adjacent_blocks_kernel - input is a matrix of lwes, and each row is packed. if the number of values is odd there can be some custom handling of the last block

`device_pack_blocks` - single list of LWEs, last block is only copied

`device_pack_blocks` is a specific case of `pack_bivariate_adjacent_blocks_kernel` and could be refactored

backends/tfhe-cuda-backend/cuda/include/integer/shuffle_utilities.h
andrei-stoian-zama: we have 

```
assert!(
        num_values >= 2,
        "Bitonic shuffle requires at least 2 values, got {num_values}"
    );
``` on the rust side. 

but it's worth explaining in C++ too. Basically there is a passthrough in rust for num_values = 1 -> no call to ffi in that case. 

the num_values here is the number of values to shuffle

backends/tfhe-cuda-backend/cuda/include/integer/integer.h
andrei-stoian-zama: fixed

backends/tfhe-cuda-backend/cuda/include/integer/integer.h
andrei-stoian-zama: fixed

guillermo-oyarzun (approved):

PR #3481, opened 2026-04-16
feat(gpu): Extended Jacobian EC arithmetics for BLS12-446 to improve MSM

bbarbakadze:
### PR content/description

Implements the extended Jacobian projective type for elliptic curve arithmetics to improve MSM performance

### Check-list:

### DIFF
--- a/backends/zk-cuda-backend/cuda/include/curve.h
+++ b/backends/zk-cuda-backend/cuda/include/curve.h
@@ -17,7 +17,13 @@ __host__ __device__ void fp2_zero(Fp2 &a);
 
 // G1 point: (x, y) coordinates in Fp
 // Curve equation: y^2 = x^3 + b (short Weierstrass form with a = 0)
-struct G1Affine {
+//
+// alignas(sizeof(uint64_t)): The bool infinity field causes the struct to be
+// padded to the largest field alignment (4 bytes in 32-bit limb mode, 8 bytes
+// in 64-bit). Forcing alignment to sizeof(uint64_t) ensures
+// sizeof(G1Affine)==120 in both modes, matching the Rust FFI bindings which
+// are always generated from the 64-bit layout regardless of LIMB_BITS_CONFIG.
+struct alignas(sizeof(uint64_t)) G1Affine {
   Fp x;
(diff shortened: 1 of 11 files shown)


backends/zk-cuda-backend/cuda/include/curve.h
pdroalves, line 26: I remember we were using an alias to the integer type instead of hardcoding `uint64_t`. Didn't that work?

backends/zk-cuda-backend/cuda/include/fp.h
pdroalves, line 21: Can you rebase on main? We are already using 32 bits.

backends/zk-cuda-backend/cuda/src/primitives/xyzz.cu
```diff
@@ -0,0 +1,175 @@
...
+  fp_zero(xyzz.ZZ.c1);
+  fp_one_montgomery(xyzz.ZZZ.c0);
+  fp_zero(xyzz.ZZZ.c1);
+}
+
+__host__ __device__ void xyzz_mixed_add(G2XYZZ &acc, const G2Affine &p) {
+  if (p.infinity)
+    return;
+
+  if (xyzz_is_infinity(acc)) {
```
pdroalves, line 117: We are using two different standards for check infinity here: we just check the attribute in `G*Affine` and we check `xyzz_is_infinity()` for acc. Perhaps we could standardize that by implementing `g*affine_is_infinity()` for the former.

backends/zk-cuda-backend/cuda/include/curve.h
bbarbakadze: this pr does not introduce any `uint64_t`

backends/zk-cuda-backend/cuda/src/primitives/xyzz.cu
bbarbakadze: Good catch g1_is_infinity()/g2_is_infinity() already exist, so I've switched the affine checks in xyzz_mixed_add to use them

pdroalves (approved):

PR #3482, opened 2026-04-17
chore(lint): remove upgrade false positive warning if new variant add…

SouchonTheo:
remove upgrade false positive warning if new variant added

the logic behind is that source can't change but target yes (cf test example)
in that case we need to match with index the old variant names map them with their new possible name
in that case we are guarantee that we avoid false positive
we are also safe because from another linter we know that the naming convention is respected

test were written with the help of AI

### DIFF
--- a/utils/tfhe-backward-compat-checker/src/diff.rs
+++ b/utils/tfhe-backward-compat-checker/src/diff.rs
@@ -1,4 +1,5 @@
 use std::cmp::Ordering;
+use std::collections::HashMap;
 use std::fs;
 use std::path::Path;
 use std::process::ExitCode;
@@ -375,6 +376,15 @@ impl Registry {
     fn diff_enum(old_enum: &EnumSnapshot, new_enum: &EnumSnapshot, entries: &mut Vec<DiffEntry>) {
         let name = &old_enum.enum_name;
 
+        let mut rename_map: HashMap<&str, &str> = HashMap::new();
+        for ov in &old_enum.variants {
+            if let Some(nv) = new_enum.variants.iter().find(|nv| nv.index == ov.index) {
+                if ov.inner_type_def_path != nv.inner_type_def_path {
+                    rename_map.insert(&ov.inner_type_def_path, &nv.inner_type_def_path);
+                }
+            }
+        }
+
         // Variant additions & modifications
         for nv in &new_enum.variants {
             match old_enum.variants.iter().find(|ov| ov.index == nv.index) {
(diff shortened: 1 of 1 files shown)


nsarlin-zama (approved):
Nice and clean fix, thanks a lot!

utils/tfhe-backward-compat-checker/src/diff.rs
```diff
@@ -676,6 +694,110 @@ mod tests {
...
+            "FooVersions",
+            vec![variant(0, "FooV0", "aaa"), variant(1, "Foo", "bbb")],
+            vec![upgrade("FooV0", "Foo", "uuu")],
+        )]);
+        // After: V0(FooV0), V1(FooV1), V2(Foo) with upgrades FooV0 → FooV1, FooV1 → Foo
+        let new = registry(vec![enum_snap(
+            "FooVersions",
+            vec![
+                variant(0, "FooV0", "aaa"),
+                variant(1, "FooV1", "bbb"), // same hash, just renamed
```
IceTDrinker, line 715: I don't see a hash check in the code doing the check, but the test mentions the hash ?

Could you point to the location where the "stable" hash is checked ? or is the naming logic enough and the hash check is already in place ?

IceTDrinker (commented):
Just a small question on my end :)

utils/tfhe-backward-compat-checker/src/diff.rs
SouchonTheo: I'm not sure this answer your question
```
        for nu in &new_enum.upgrades {
            match old_enum.upgrades.iter().find(|ou| {
                let resolved_target = rename_map
                    .get(ou.target_def_path.as_str())
                    .copied()
                    .unwrap_or(ou.target_def_path.as_str());
                ou.source_def_path == nu.source_def_path && resolved_target == nu.target_def_path
            }) {
                None => entries.push(DiffEntry::upgrade_added(name, nu)),
                Some(ou) if ou.body_hash != nu.body_hash => {
                    entries.push(DiffEntry::upgrade_hash_changed(name, ou, nu));
                }
                _ => {}
            }
        }
```
you can see if we match something we compare hash

SouchonTheo: this is line 416 of diff.rs

IceTDrinker: thanks, I think that should ansnwer it, since I was only seeing the naming diff, will take a look 👍

IceTDrinker: yep that answers it :) thanks !

PR #3490, opened 2026-04-21
chore(lint): remove snapshots folder and make it more dynamic and eas…

SouchonTheo:
remove snapshots dir to make it simpler for dev
this lead to a lot of simplification in the ci and also in makefile
we also removed the two sh script which make everything simplier

AI where involve for the checkout depth 1 which was missing in both ci
AI where also involve to find the right shell call in the makefile for example shell mktemp -u and -d. Also the simplification of the macro with run-all-crates was made by AI which is very convenient to simplify the makefile

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490)


.github/workflows/backward_compat_snapshot_consistency.yml
```diff
@@ -1,5 +1,4 @@
-# Generates snapshots from code and diffs against committed base files.
-# Ensures snapshots are up to date on PRs and catches stale ones on main.
+# Checks backward compatibility between the PR head and its base branch.
```
nsarlin-zama, line 1: Maybe now we could have only one single workflow that:
1. generate snapshot for head and base
2. generate report
3. check

(maybe 2 and 3 can be the same step)

That way the snapshots are generated only once

SouchonTheo: Just did it !

utils/tfhe-lints/Makefile
```diff
@@ -57,28 +55,22 @@ tfhe_lints_crate: install_cargo_dylint install_tfhe_lints_toolchain
...
 
-.PHONY: backward_snapshot_base # Generate the base snapshot for all crates
-backward_snapshot_base: install_cargo_dylint install_tfhe_lints_toolchain
-	@echo "Generating base snapshot"
-	$(call run-all-crates,tfhe_lints_snapshot,$(SNAPSHOT_DIR))
+.PHONY: backward_snapshot # Generate snapshots for all crates. Required: SNAPSHOT_OUTPUT_DIR
+backward_snapshot: install_cargo_dylint install_tfhe_lints_toolchain
+	$(call require,SNAPSHOT_OUTPUT_DIR)
+	@mkdir -p $(SNAPSHOT_OUTPUT_DIR)
+	cargo clean
```
nsarlin-zama: if this is needed I wouldn't do this here, since it's a side effect that the user might not expect. Maybe cleaner to do it in the workflow

.github/workflows/backward_compatibility.yml
```diff
@@ -0,0 +1,93 @@
...
+          BASE_REF: ${{ github.event.pull_request.base.sha || github.event.before }}
+        run: make backward_correctness BASE_REF="$BASE_REF"
+
+      - name: Generate diff report
+        if: github.event_name == 'pull_request'
+        id: report
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha }}
+        run: |
+          make backward_report \
```
nsarlin-zama: If I understand correctly, the workflow does:
- backward_correctness
- backward_report

But they both call backward_snapshot, meaning that the snapshot will be done twice. You could do it once I think. For example, backward_snapshot could be a non PHONY make target that generates the snapshot only if the rust files in the crates have been modified

nsarlin-zama (dismissed):

utils/tfhe-lints/Makefile
SouchonTheo: done

.github/workflows/backward_compatibility.yml
SouchonTheo: done

nsarlin-zama (dismissed):
LGTM, thanks!!

nsarlin-zama (approved):

soonum (commented):
Good to go, just a small question.

> @soonum reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490#-:-OrIV-Mx56eYighvBPjm:b-ml0xk1)* status: all files reviewed, 4 unresolved discussions (waiting on nsarlin-zama and SouchonTheo).

___
*[`utils/tfhe-lints/Makefile` line 96 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490#-OrITAIP6EDCwjY01wFA:-OrITAIP6EDCwjY01wFB:b-pee691) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43de0252a52cfa5122d25cbf463db38fd31fc34d/utils/tfhe-lints/Makefile#L96)):*
> ```Makefile
> 
> .PHONY: backward_correctness # Check backward compat of current code against BASE_REF. Required: BASE_REF
> backward_correctness:
> ```

Are this recipe and the following meant to be run locally ? I don't see them in the GitHub workflow.

SouchonTheo (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490)* status: all files reviewed, 4 unresolved discussions (waiting on nsarlin-zama and soonum).

___
*[`utils/tfhe-lints/Makefile` line 96 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490#-OrITAIP6EDCwjY01wFA:-OrNKSiYCdQ-L8ckb9Xa:b-j5e5ty) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43de0252a52cfa5122d25cbf463db38fd31fc34d/utils/tfhe-lints/Makefile#L96)):*
<details><summary><i>Previously, soonum (David Testé) wrote…</i></summary><blockquote>

Are this recipe and the following meant to be run locally ? I don't see them in the GitHub workflow.
</blockquote></details>

yes local rule
in ci we use the one lower to be able to build just one time the snapshot

IceTDrinker:
*[`.github/workflows/backward_compatibility.yml` line 111 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3490#-OrOuI9p-oXSWvqTllDJ:-OrOuI9p-oXSWvqTllDK:bjej1jw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/43de0252a52cfa5122d25cbf463db38fd31fc34d/.github/workflows/backward_compatibility.yml#L111)):*
> ```YAML
> 
>       - name: Fail if check failed
>         if: steps.check.outcome == 'failure'
> ```

I would think (nit) that this should be

steps.check.outcome != 'success'

reason is : if the status is cancelled the status would be green ? unsure

PR #3496, opened 2026-04-22
chore: update typos file filter

IceTDrinker:
- with HPU data file checked out the typos CLI finds typos in essentially binary data
- exclude .hpu files from the checks


soonum (approved):
Let's gooo

PR #3507, opened 2026-04-23
chore: bump rand to 0.8.6 in data generation crate

IceTDrinker:
- 1.6 is done in a separate PR which will use the officially published tag as source for the code, which also updates the lock

this closes the advisories about unsoundness with the log feature (which we do not use)


nsarlin-zama (approved):
thanks!

PR #3517, opened 2026-04-27
Am/chore/misc

IceTDrinker:
First commit is to make it less confusing to reason about the custom fft code which was mixing the notion of base_n used in tfhe-fft to select the base fft algorithm size, which has nothing to do with how we store the plan in that case

Second commit is an unfortunate typo reported by KMS when they used the newly available cleartext_prf for testing purposes


mayeul-zama (approved):
Thanks!

PR #3518, opened 2026-04-28
feat(gpu): expose ILP in standard keyswitch

guillermo-oyarzun:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
@@ -301,7 +301,11 @@ __global__ void keyswitch_zero_output_with_output_indices(
 // in two parts, a constant part is calculated before the loop, and a variable
 // part is calculated inside the loop. This seems to help with the register
 // pressure as well.
-template <typename Torus, typename KSTorus>
+// LevelCount template fully unrolls the level loop.
+// Accumulation step is divided in even and odd to expose more instruction level
+// parallelism. When LevelCount == 0 (default), the original runtime path is
+// used unchanged.
+template <typename Torus, typename KSTorus, int LevelCount = 0>
 __global__ void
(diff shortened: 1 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
```diff
@@ -301,7 +301,11 @@ __global__ void keyswitch_zero_output_with_output_indices(
 // in two parts, a constant part is calculated before the loop, and a variable
 // part is calculated inside the loop. This seems to help with the register
 // pressure as well.
-template <typename Torus, typename KSTorus>
+// LevelCount template fully unrolls the level loop.
+// Accumulation step is divided in even and odd to expose more instruction level
+// parallelism. When LevelCount == 0 (default), the original runtime path is
+// used unchanged.
+template <typename Torus, typename KSTorus, int LevelCount = 0>
```
andrei-stoian-zama, line 308: why do even/odd accumulator help? I would imagine the `decompose` step takes sufficiently many cycles to allow for the previous write to the accumulator to finish so the new write doesn't stall -> was that the problem ?

andrei-stoian-zama (approved):

backends/tfhe-cuda-backend/cuda/src/crypto/keyswitch.cuh
guillermo-oyarzun: if you don't split it the compiler sees a chain of dependancies between the instruction because they accumulate on top of the same register, which is bad because operations involving reading memory can be delayed and math instructions have to wait too. This exposes 2 independent execution paths, so you pay the price of adding more registers and one more add (but is an acceptable price, no changing occupancy), and you give the compiler the chance to fire pairs of operations without having a dependancy between them (only at the final reduction). Ideally we could unroll more, but then we would need more live registers to keep all the intermediate ops, and we would need extra additions too. To be honest, I thought the compiler would be smart enough to see this, but it seems we need to chew its food a little bit.

PR #3531, opened 2026-05-04
fix(gpu): fix vector find parallelization

enzodimaria:
vector_find is now parallelized via _vectorization_ instead of _multi-streams_

Before, we had two bottlenecks:
- a hardcoded cap at 5 CUDA streams
- tons of small pbs launched one by one (one per candidate for match_value, one per ct for contains), stitched together by a series of ct copies

After, the idea is simple: batch everything and let the PBS dispatcher handle the parallelism.
- match_value (ct vs N clears): one many-LUT computes all the block-wise equalities, a small align_with_indexes kernel lays them out in a flat buffer, then an AND-reduce stacks up to max_degree selector blocks per candidate into the carry space before applying one batched pbs. All candidates progress through a few big PBS calls instead of N small are_all_comparisons_block_true loops.
- contains (N cts vs 1 ct): a kernel packs the (input, value) pairs block by block, then one batched bivariate pbs does all the equalities at once
- One-hot aggregation (final step): a tree reduction instead of per-stream naive sums. Each level sums chunk_size rows in a fused addition, then a batched identity PBS refreshes the noise.
- Most copy_radix_ciphertext_slice_async got replaced by small homemade scatter/pack kernels, killing the host-side launch overhead (good gains on small inputs !)

```
+--------------------+------+---------------+---------------+-------------+
| Benchmark Function | Size | Before        | After         | Improvement |
+--------------------+------+---------------+---------------+-------------+
| FheUint64 (1 x H100-SXM)                                                |
+--------------------+------+---------------+---------------+-------------+
| contains           |    5 |         21 ms |         24 ms |       x 0.9 |
|                    |   10 |         39 ms |         37 ms |       x 1.1 |
|                    |   20 |         70 ms |         61 ms |       x 1.1 |
|                    |   30 |        102 ms |         86 ms |       x 1.2 |
|                    |   40 |        133 ms |        112 ms |       x 1.2 |
|                    |   50 |        164 ms |        137 ms |       x 1.2 |
|                    |  500 |       1548 ms |       1327 ms |       x 1.2 |
|                    | 1000 |       3091 ms |       2653 ms |       x 1.2 |
+--------------------+------+---------------+---------------+-------------+
| match_value        |    5 |         21 ms |         21 ms |       x 1.0 |
|                    |   10 |         32 ms |         27 ms |       x 1.2 |
|                    |   20 |         50 ms |         33 ms |       x 1.5 |
|                    |   30 |         71 ms |         41 ms |       x 1.7 |
|                    |   40 |         89 ms |         47 ms |       x 1.9 |
|                    |   50 |        106 ms |         53 ms |       x 2.0 |
|                    |  500 |        931 ms |        386 ms |       x 2.4 |
|                    | 1000 |       1846 ms |        758 ms |       x 2.4 |
+--------------------+------+---------------+---------------+-------------+
| FheUint64 (8 x H100-SXM)                                                |
+--------------------+------+---------------+---------------+-------------+
| contains           |    5 |         21 ms |         17 ms |       x 1.2 |
|                    |   10 |         32 ms |         22 ms |       x 1.5 |
|                    |   20 |         53 ms |         26 ms |       x 2.0 |
|                    |   30 |         74 ms |         31 ms |       x 2.4 |
|                    |   40 |         94 ms |         35 ms |       x 2.7 |
|                    |   50 |        114 ms |         39 ms |       x 2.9 |
|                    |  500 |        997 ms |        283 ms |       x 3.5 |
|                    | 1000 |       1966 ms |        555 ms |       x 3.5 |
+--------------------+------+---------------+---------------+-------------+
| match_value        |    5 |         29 ms |         23 ms |       x 1.3 |
|                    |   10 |         39 ms |         28 ms |       x 1.4 |
|                    |   20 |         59 ms |         31 ms |       x 1.9 |
|                    |   30 |         81 ms |         35 ms |       x 2.3 |
|                    |   40 |        100 ms |         37 ms |       x 2.7 |
|                    |   50 |        119 ms |         38 ms |       x 3.1 |
|                    |  500 |        971 ms |         88 ms |      x 11.0 |
|                    | 1000 |       1906 ms |        146 ms |      x 13.1 |
+--------------------+------+---------------+---------------+-------------+
| FheUint8 (1 x H100-SXM)                                                 |
+--------------------+------+---------------+---------------+-------------+
| contains           |    5 |          9 ms |          6 ms |       x 1.5 |
|                    |   10 |         14 ms |          9 ms |       x 1.6 |
|                    |   20 |         21 ms |         12 ms |       x 1.8 |
|                    |   30 |         30 ms |         16 ms |       x 1.9 |
|                    |   40 |         37 ms |         20 ms |       x 1.9 |
|                    |   50 |         44 ms |         23 ms |       x 1.9 |
|                    |  500 |        367 ms |        191 ms |       x 1.9 |
|                    | 1000 |        722 ms |        373 ms |       x 1.9 |
+--------------------+------+---------------+---------------+-------------+
| match_value        |    5 |         15 ms |         10 ms |       x 1.5 |
|                    |   10 |         21 ms |         14 ms |       x 1.5 |
|                    |   20 |         30 ms |         16 ms |       x 1.9 |
|                    |   30 |         41 ms |         20 ms |       x 2.1 |
|                    |   40 |         50 ms |         23 ms |       x 2.2 |
|                    |   50 |         58 ms |         25 ms |       x 2.3 |
|                    |  500 |        253 ms |         79 ms |       x 3.2 |
|                    | 1000 |        253 ms |         79 ms |       x 3.2 |
+--------------------+------+---------------+---------------+-------------+
```


backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
 #include "integer/radix_ciphertext.cuh"
-#include "integer/scalar_comparison.cuh"
 #include "integer/vector_find.h"
 
+// Gathers `dst[i]` block from `src[map[i]]`, for `i` in `[0, gridDim.x)`.
+// Each CUDA block handles one destination block, threads cooperatively copy
+// `block_size` Torus words.
+template <typename Torus>
+__global__ void
+map_gather_kernel(Torus *__restrict__ dst, const Torus *__restrict__ src,
```
andrei-stoian-zama: how is this different from `align_with_indexes` ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+    dst_ptr[tid] = src_ptr[tid];
+  }
+}
+
+// Scatters one block per (i, j) to disjoint output ciphertexts.
+// Block grid is sized `num_outputs * num_blocks`. Output `i = blockIdx.x /
+// num_blocks` receives `j = blockIdx.x % num_blocks` from
+// `src_batched[src_offsets[blockIdx.x]]`.
+template <typename Torus>
+__global__ void scatter_to_ptr_array_kernel(
```
andrei-stoian-zama: how is this different to `realign_with_indexes` ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+  uint32_t max_items = mem_ptr->max_items;
 
   integer_radix_apply_many_univariate_lookup_table<Torus>(
       streams, mem_ptr->tmp_many_luts_output, lwe_array_in, bsks,
       (Torus *const *)ksks, mem_ptr->comparison_luts, message_modulus,
       mem_ptr->lut_stride);
 
-  mem_ptr->internal_cuda_streams.internal_streams_wait_for_main_stream_0(
-      streams);
+  uint32_t *h_map = mem_ptr->h_map;
```
andrei-stoian-zama: in a comment please explain what h_map contains

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+      h_map[j * num_possible_values + i] =
+          (uint32_t)block_value * num_blocks + j;
+    }
+  }
+  cuda_memcpy_async_to_gpu(mem_ptr->d_map, h_map,
+                           safe_mul_sizeof<uint32_t>(total_blocks),
+                           streams.stream(0), streams.gpu_index(0));
 
-  uint32_t num_streams = mem_ptr->num_streams;
+  uint32_t block_size = mem_ptr->tmp_batched_comparisons->lwe_dimension + 1;
```
andrei-stoian-zama: I don't like `block_size` -> we usually use lwe_size no?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
     const uint64_t *h_decomposed_cleartexts,
     int_equality_selectors_buffer<Torus> *mem_ptr, void *const *bsks,
     Torus *const *ksks) {
 
   uint32_t num_possible_values = mem_ptr->num_possible_values;
   uint32_t message_modulus = mem_ptr->params.message_modulus;
+  uint32_t carry_modulus = mem_ptr->params.carry_modulus;
+  uint32_t max_items = mem_ptr->max_items;
 
   integer_radix_apply_many_univariate_lookup_table<Torus>(
```
andrei-stoian-zama, line 291: please add a comment to explain what this does. what is done to each block of the buffer using many lut ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+        &col_slice, mem_ptr->tmp_batched_comparisons, j * num_possible_values,
+        (j + 1) * num_possible_values);
+
+    host_addition<Torus>(streams.stream(0), streams.gpu_index(0),
+                         mem_ptr->packed_accumulator,
+                         mem_ptr->packed_accumulator, &col_slice,
+                         num_possible_values, message_modulus, carry_modulus);
+    items_in_acc++;
+
+    if (items_in_acc == max_items || j == num_blocks - 1) {
```
andrei-stoian-zama: max_items -> not well named, we usually call this max_degree

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+      integer_radix_apply_univariate_lookup_table<Torus>(
```
andrei-stoian-zama: why don't we batch the PBSs ? they are done sequentially no ? first 15 summed blocks, then the next, etc

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
+    CudaStreams streams, CudaRadixCiphertextFFI *lwe_array_out_list,
+    CudaRadixCiphertextFFI const *inputs, CudaRadixCiphertextFFI const *value,
+    uint32_t num_inputs, uint32_t num_blocks,
+    int_equality_selectors_vs_ciphertext_buffer<Torus> *mem_ptr,
+    void *const *bsks, Torus *const *ksks) {
 
-    CudaStreams current_stream = mem_ptr->internal_cuda_streams[stream_idx];
+  uint32_t message_modulus = mem_ptr->params.message_modulus;
+  uint32_t carry_modulus = mem_ptr->params.carry_modulus;
+  uint32_t match_parallelism = mem_ptr->match_parallelism;
```
andrei-stoian-zama: what is match_parallelism ? seems to be a CPU thing no ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
-    CudaRadixCiphertextFFI *current_tmp_block_comparisons =
-        mem_ptr->tmp_block_comparisons[stream_idx];
-    int_comparison_buffer<Torus> *current_reduction_buffer =
-        mem_ptr->reduction_buffers[stream_idx];
+  for (uint32_t i = 0; i < num_inputs; i += match_parallelism) {
+    uint32_t parallel_chunk = std::min(match_parallelism, num_inputs - i);
 
-    const uint64_t *current_clear_blocks =
-        &h_decomposed_cleartexts[i * num_blocks];
+    for (uint32_t c = 0; c < parallel_chunk; c++) {
```
andrei-stoian-zama: why do we need to start the copies in parallel ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,167 +5,321 @@
...
-      uint64_t block_value = current_clear_blocks[j];
+    CudaRadixCiphertextFFI acc_slice, curr_slice, val_slice;
+    as_radix_ciphertext_slice<Torus>(&acc_slice, mem_ptr->packed_accumulator, 0,
+                                     parallel_chunk);
+    as_radix_ciphertext_slice<Torus>(&curr_slice, mem_ptr->packed_current_block,
+                                     0, parallel_chunk);
+    as_radix_ciphertext_slice<Torus>(&val_slice, mem_ptr->packed_value_block, 0,
+                                     parallel_chunk);
+
+    integer_radix_apply_bivariate_lookup_table<Torus>(
```
andrei-stoian-zama, line 402: why do we treat the first block differently ?

andrei-stoian-zama (changes requested):

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Identical indeed

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: realign_with_indexes output is a "Torus *d_vector" and here, the destinationis a "Torus **dst_ptr_array"

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: :+1:

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Done

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: No it's a parameter which enables to adjust the level of vectorisation of the sub-functions: how many independent inputs we fold into a single batched PBS
For now it's hardcoded to the maximum just in case

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Each successive call are dependant
A pbs computes a partial "and"
`auto f_eq_k = [k](Torus x) -> Torus { return (x == k) ? 1 : 0; };`
then the next iterations add more booleans into that same accumulator. So pbs k+1 reads what PBS k just wrote
But, each call already runs in parallel over the "num_possible_values" selectors, so we are batched on this axis

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Not parallel, copies are done in the same stream
I'll change the name, it's not good

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Just to save an addition :+1: 

  Current version:                                          
  LUT_eq(acc, x_0, y_0)
  for j = 1..n-1:      
      LUT_eq(curr, x_j, y_j)
      acc += curr                                                                                                                                                                                                                                       
                                                            
  If j= 0 were in the loop:
  set_zero(acc)                                                                                                                                                                                                                                         
  for j = 0..n-1:
      LUT_eq(curr, x_j, y_j)                                                                                                                                                                                                                            
      acc += curr

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,163 +5,277 @@
...
       streams, mem_ptr->tmp_many_luts_output, lwe_array_in, bsks,
       (Torus *const *)ksks, mem_ptr->comparison_luts, message_modulus,
       mem_ptr->lut_stride);
 
-  mem_ptr->internal_cuda_streams.internal_streams_wait_for_main_stream_0(
-      streams);
+  // For each (candidate i, block j) pair, store the index inside the
+  // many-LUT output that corresponds to the precomputed equality result
+  // (x_j == h_decomposed_cleartexts[i][j])
+  Torus *h_map = mem_ptr->h_map;
```
andrei-stoian-zama, line 303: what is the max number of LUTs that can be applied with a single PBS with the manylut functionality ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
andrei-stoian-zama: sorry,  I don't really understand why this is a good or necessary approach. can't this be done with a parallel reduction so that PBS k+1 does not depend on PBS k?

andrei-stoian-zama: look at are_all_comparisons_block_true -> doesn't it do the same thing ? it accumulates the chunks into a buffer then runs batch PBS on that buffer.  could you compare your function here to that one ? should we use `are_all_comparisons_block_true` instead ?

tmontaigu (commented):
> @tmontaigu reviewed 3 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3531)* status: 3 of 8 files reviewed, 12 unresolved discussions (waiting on andrei-stoian-zama, enzodimaria, and nsarlin-zama).

___
*[`tfhe/src/integer/gpu/ffi.rs` line 9000 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3531#-Os0y9NsCyG2SS3cdXdF:-Os0y9NsCyG2SS3cdXdG:b-fqoo3o) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/8865e8051ac75443d93cff4c32fde282c668337f/tfhe/src/integer/gpu/ffi.rs#L9000)):*
> ```XML
> 
>     scalar_blocks.truncate(num_blocks_in_ct as usize);
>     scalar_blocks.resize(num_blocks_in_ct as usize, 0u64);
> ```

nit: if you do a resize, the truncate done before is not needed

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,163 +5,277 @@
...
+    as_radix_ciphertext_slice<Torus>(&curr_slice, mem_ptr->packed_current_block,
+                                     0, current_batch_size);
+    as_radix_ciphertext_slice<Torus>(&val_slice, mem_ptr->packed_value_block, 0,
+                                     current_batch_size);
+
+    integer_radix_apply_bivariate_lookup_table<Torus>(
+        streams, &acc_slice, &curr_slice, &val_slice, bsks, ksks,
+        mem_ptr->equality_lut, current_batch_size, message_modulus);
+
+    uint32_t items_in_acc = 1;
```
andrei-stoian-zama: this looks very similar to the code above (line 85). It should be refactored into a single function. Better yet, could you find if the existing `host_aggregate_one_hot_vector` is usable

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,547 @@
...
+template <typename T>
+__global__ void
+pack_chunk_eq_pairs_kernel(T *packed_current, T *packed_value,
+                           T *const *src_ptrs, T const *value_ptr,
+                           uint32_t base_input, uint32_t current_chunk,
+                           uint32_t num_blocks, uint32_t lwe_size) {
+  uint32_t c = blockIdx.x;
+  uint32_t j = blockIdx.y;
+  if (c >= current_chunk)
+    return;
```
bbarbakadze: I think you can remove this check, since `blockIdx.x ` will always be less than `chunk_size`.  with `  dim3 grid(current_chunk, num_blocks, 1);`.   If it is the only use of this kernel, the comment can be added that gridDim.x should be chunk_size

bbarbakadze: If it can be used with different grids than lets keep the guard.

bbarbakadze (dismissed):
Thanks @enzodimaria . Nice work.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Fixed! thanks Beka

bbarbakadze (dismissed):

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -97,21 +108,140 @@ template <typename Torus> struct int_equality_selectors_buffer {
...
+                                    int_radix_params params,
+                                    uint32_t num_inputs, uint32_t num_blocks,
+                                    bool allocate_gpu_memory,
+                                    uint64_t &size_tracker) {
+    this->params = params;
+    this->allocate_gpu_memory = allocate_gpu_memory;
+    this->num_inputs = num_inputs;
+    this->num_blocks = num_blocks;
+
+    this->max_degree = (params.message_modulus * params.carry_modulus - 1) /
```
andrei-stoian-zama: this appears many times in the code base, can it be refactored ? maybe there is a function already that you can call ?

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -97,21 +108,140 @@ template <typename Torus> struct int_equality_selectors_buffer {
...
+
+    auto eq_fn = [](Torus x, Torus y) -> Torus { return (x == y) ? 1 : 0; };
+
+    this->equality_lut = new int_radix_lut<Torus>(
+        streams, params, 1, biv_batch, allocate_gpu_memory, size_tracker);
+    this->equality_lut->generate_and_broadcast_bivariate_lut(
+        streams.active_gpu_subset(biv_batch, params.pbs_type), {0}, {eq_fn},
+        LUT_0_FOR_ALL_BLOCKS);
+
+    this->packed_accumulator = new CudaRadixCiphertextFFI;
```
andrei-stoian-zama: no need to change it, but I still don't understand why we don't allocate CudaRadixCiphertextFFI on the stack (i.e. no pointer)

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -71,21 +63,40 @@ template <typename Torus> struct int_equality_selectors_buffer {
...
+        allocate_gpu_memory);
+
+    this->max_degree = (params.message_modulus * params.carry_modulus - 1) /
+                       (params.message_modulus - 1);
 
-      this->reduction_buffers[j] = new int_comparison_buffer<Torus>(
-          streams, EQ, params, num_blocks, false, allocate_gpu_memory,
-          size_tracker);
+    this->num_luts_needed = std::min(this->max_degree, num_blocks);
+    this->luts_eq = new int_radix_lut<Torus> *[this->num_luts_needed + 1]();
```
andrei-stoian-zama: so luts 0 and 1 are not used and not initialized? at least set the pointers to `nullptr` please.

andrei-stoian-zama: also this is a good setting to use std::vector. see some examples in the code base:

```  std::vector<int_radix_lut<Torus> *> lut_buffers_bivariate;```

this way you get automatic initialization to nullptr. but you should still do it explicitly

File: backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h [hunk 17]
```diff
@@ -97,21 +108,140 @@ template <typename Torus> struct int_equality_selectors_buffer {
...
+    delete this->packed_accumulator;
+
+    for (uint32_t k = 2; k <= this->num_luts_needed; k++) {
+      this->luts_eq[k]->release(streams);
+      delete this->luts_eq[k];
     }
-    delete[] this->tmp_block_comparisons;
+    delete[] this->luts_eq;
+
+    delete[] this->h_map;
```
andrei-stoian-zama: you're deleting the host pointer before the device pointer. this is a synchronization issue: you should always delete device pointer, synchronize then delete the host pointer

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -97,21 +108,140 @@ template <typename Torus> struct int_equality_selectors_buffer {
...
+                                   this->allocate_gpu_memory);
+    delete this->packed_value_block;
+
+    for (uint32_t k = 2; k <= this->num_luts_needed; k++) {
+      this->luts_eq[k]->release(streams);
+      delete this->luts_eq[k];
+    }
+    delete[] this->luts_eq;
+
+    delete[] this->h_input_ptrs;
```
andrei-stoian-zama: same synchronization problem here !

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -341,33 +468,39 @@ template <typename Torus> struct int_aggregate_one_hot_buffer {
     create_zero_radix_ciphertext_async<Torus>(
         streams.stream(0), streams.gpu_index(0), this->carry_ct, num_blocks,
         params.big_lwe_dimension, size_tracker, allocate_gpu_memory);
+
+    this->h_input_ptrs = new Torus *[num_matches];
+    this->d_input_ptrs = (Torus **)cuda_malloc_with_size_tracking_async(
```
andrei-stoian-zama, line 533: this is a Torus**  but allocated as a contiguous block of memory. if you need a vector of pointers why not std::vector<Torus*> and then you cuda_malloc each individual pointer. otherwise keep it a Torus* if you need a single contiguous buffer.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -376,7 +509,9 @@ template <typename Torus> struct int_aggregate_one_hot_buffer {
                                    this->carry_ct, this->allocate_gpu_memory);
     delete this->carry_ct;
 
-    internal_cuda_streams.release(streams);
+    delete[] this->h_input_ptrs;
```
andrei-stoian-zama: synchronization issue again. also deletion of Torus** as Torus*, please decide what you need here

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -583,14 +708,11 @@ template <typename Torus> struct int_unchecked_contains_buffer {
...
-  int_comparison_buffer<Torus> **eq_buffers;
+  int_eq_selectors_cts_vs_ct_buffer<Torus> *eq_selectors_buf;
   int_comparison_buffer<Torus> *reduction_buffer;
 
   CudaRadixCiphertextFFI *packed_selectors;
-
-  CudaStreams active_streams;
-  InternalCudaStreams internal_cuda_streams;
-  uint32_t num_streams;
+  CudaRadixCiphertextFFI *unpacked_selectors;
```
andrei-stoian-zama: I think you should try not using a pointer for the CudaRadixCiphertextFFI when you add new ones

andrei-stoian-zama (changes requested):
And some more from Claude:

```
HIGH
                                                                                                                                                     
  host_compute_eq_selectors_cts_vs_ct — no bounds check on caller-supplied num_inputs (vector_find.cuh ~505)

  for (uint32_t k = 0; k < num_inputs; k++) {                                                                                                        
      mem_ptr->h_input_ptrs[k] = (Torus *)inputs[k].ptr;                                                                                             
  }                                                                                                                                                  
                                                                                                                                                     
  h_input_ptrs is heap-allocated at scratch time with a fixed num_inputs. The function accepts a separate num_inputs argument and uses it directly to
   write into that array — no assertion that it matches the allocation size. If the caller passes a larger value, this silently overwrites adjacent  
  heap memory.

  Fix: Add assert(num_inputs <= mem_ptr->num_inputs) or PANIC_IF_FALSE(...) at function entry.

```

```
  Dead GPU allocation: final_aggregated_vector never used (vector_find.h ~456)

  In int_aggregate_one_hot_buffer, this->final_aggregated_vector is allocated with create_zero_radix_ciphertext_async and freed in release(), but
  host_aggregate_one_hot_vector never touches it — it creates a local slice view of src_buf instead. GPU memory is wasted on every scratch
  allocation.
```

```
  uint32_t overflow before size calculations (vector_find.h ~338, ~418)

  uint32_t total_lut_blocks = num_lut_accumulators * num_possible_values;
  // passed to create_zero_radix_ciphertext_async as:
  total_lut_blocks * max_luts_per_call    // still uint32_t, can overflow silently
  num_matches * num_blocks                // same

  The multiplication happens in 32-bit arithmetic before reaching safe_mul_sizeof. For large input sets this silently wraps and causes
  under-allocation.

  Fix: Widen at least one operand: (uint64_t)total_lut_blocks * max_luts_per_call.

```

```
  const dropped silently on h_clear_val (vector_find.cuh ~890, ~1093, ~1282)

  // h_clear_val is const uint64_t *, cast to (Torus *) which is non-const
  set_trivial_radix_ciphertext_async<Torus>(..., (Torus *)h_clear_val, ...);

  This is an implicit const_cast. If set_trivial_radix_ciphertext_async's signature ever changes to write through that pointer, this becomes UB
  silently.

  Fix: Use const_cast<Torus *>(h_clear_val) with a comment, or change the parameter type to const Torus *.

This is about the `new CudaRadixCiphertextFFI` usage instead of allocation on the stack that I pointed out too, but "RAII" is a fancier term for it 
```

```
  No RAII: partial-constructor leaks (all buffer structs)

  All constructors perform multiple allocations without RAII. If any allocation panics or throws mid-constructor, all previously allocated buffers
  leak since the destructor is not called for partially-constructed objects. This is a pre-existing pattern, but the diff adds many more such
  constructors.
```

```
  Missing bounds assertion on luts_eq[items_in_acc] (vector_find.cuh ~424)

  The indexing is safe given current loop arithmetic, but the invariant is non-obvious. A single assert(items_in_acc <= num_luts_needed) would
  prevent silent OOB if the loop logic is ever touched.
```

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
enzodimaria: There's no function that does this but I added it in integer_utilities.h:
```
  // Maximum number of consecutive unbootstrapped additions
  uint32_t max_degree() const {
    return (message_modulus * carry_modulus - 1) / (message_modulus - 1);
  }
```
It will be useful :+1:

enzodimaria:
Everything is fixed ! @andrei-stoian-zama

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,306 @@
...
+// │ j=1: [In_0[1], In_1[1], ...] │ <--> │ j=1: [Val[1],  Val[1],  ...] │
+// ├──────────────────────────────┤      ├──────────────────────────────┤
+// │ ...                          │      │ ...                          │
+// └──────────────────────────────┘      └──────────────────────────────┘
+//
+// Pair (packed_current[k], packed_value[k]) feeds one slot of a batched
+// bivariate PBS that computes the block-wise equality.
+//
+// Used to avoid to loop with host_equality_check
+template <typename T>
```
andrei-stoian-zama, line 74: the comment is not clear. I think it should be re-written entirely. Maybe keep the diagram if you manage to make the text explain it well. 

can you explain in the comment what are the inputs and outputs of this function ? there are 4 radix-ct ptr params.. one is value and one is packed_value -> I'm guessing packed_value is related to value, but what about inputs and packed_current ? 

and for packing I'm expecting the same packing as for bivariate PBS, but I don't see the `v0 * 4 + v1` which is done in host_pack_bivariate_blocks..

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,306 @@
...
+      dst_list[i].degrees[j] = src->degrees[off];
+      dst_list[i].noise_levels[j] = src->noise_levels[off];
+    }
+  }
+}
+
+// and-reduce each row of an num_inputs * num_blocks selector matrix into 1 bit
+// per row
+template <typename Torus>
+__host__ void host_and_reduce_selector_matrix(
```
andrei-stoian-zama, line 224: how is this different from `host_aggregate_one_hot_vector` ? it seems to be the same thing only with a different lut and taking a flat input instead of ptrTable.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
```diff
@@ -301,84 +419,70 @@ template <typename Torus> struct int_aggregate_one_hot_buffer {
...
-          streams.stream(0), streams.gpu_index(0),
-          this->partial_aggregated_vectors[i], num_blocks,
-          params.big_lwe_dimension, size_tracker, allocate_gpu_memory);
-
-      create_zero_radix_ciphertext_async<Torus>(
-          streams.stream(0), streams.gpu_index(0),
-          this->partial_temp_vectors[i], num_blocks, params.big_lwe_dimension,
-          size_tracker, allocate_gpu_memory);
-    }
+    create_zero_radix_ciphertext_async<Torus>(
```
andrei-stoian-zama, line 514: I don't think CudaRadixCiphertextFFI is the right struct to use. CudaRadixCiphertextFFI should have a number of blocks (lwes) that corresponds to some radix-precision: eg. 32b / 64b.

It should not be used as a generic "list of lwes". for that we have CudaLweCiphertextListFFI which should be extended with the necessary functionality. this mirrors the rust-side classes

andrei-stoian-zama: there are many instances...

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,306 @@
...
+      PANIC_IF_FALSE(items_in_acc <= num_luts_needed,
+                     "items_in_acc out of luts_eq bounds");
+      integer_radix_apply_univariate_lookup_table<Torus>(
+          streams, &acc_slice, &acc_slice, bsks, ksks, luts_eq[items_in_acc],
+          num_columns);
+    }
+  }
+}
+
+// 1 ciphertext is compared to N clear values
```
andrei-stoian-zama: please add a clearer comment: explain why/how the output is packed

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,306 @@
...
+  check_cuda_error(cudaGetLastError());
 
-  for (uint32_t i = 0; i < num_possible_values; i++) {
+  for (uint32_t b = 0; b < total_blocks; b++) {
+    mem_ptr->tmp_batched_comparisons.degrees[b] = 1;
+    mem_ptr->tmp_batched_comparisons.noise_levels[b] = NoiseLevel::NOMINAL;
+  }
 
-    uint32_t stream_idx = i % num_streams;
+  host_and_reduce_selector_matrix<Torus>(
```
andrei-stoian-zama, line 337: the and reduction is on rows or columns ? please add a comment explaining the memory layout of packed_accumulator: it's a 2d matrix of how many rows/columns ? the output of the and reduction is a vector of LWE or single LWE ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,306 @@
...
+  for (uint32_t k = 0; k < num_inputs; k++) {
+    h_input_ptrs[k] = (Torus *)inputs[k].ptr;
+  }
+  cuda_memcpy_async_to_gpu(mem_ptr->d_input_ptrs, mem_ptr->h_input_ptrs,
+                           safe_mul_sizeof<Torus *>(num_inputs),
+                           streams.stream(0), streams.gpu_index(0));
 
-      uint32_t input_start = (uint32_t)block_value * num_blocks + j;
+  for (uint32_t base = 0; base < num_inputs; base += chunk_size) {
+    uint32_t current_chunk = std::min(chunk_size, num_inputs - base);
```
andrei-stoian-zama: `current_chunk_size` would have been more clear

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -470,31 +624,12 @@ host_unchecked_contains(CudaStreams streams, CudaRadixCiphertextFFI *output,
...
-
-    host_equality_check<Torus>(current_stream, &current_selector_block,
-                               input_ct, value, mem_ptr->eq_buffers[stream_idx],
-                               bsks, ksks, num_blocks);
-  }
-
-  mem_ptr->internal_cuda_streams.main_stream_0_wait_for_internal_streams(
-      streams);
+  host_compute_eq_selectors_cts_vs_ct<Torus>(
+      streams, mem_ptr->unpacked_selectors.data(), inputs, value, num_inputs,
```
andrei-stoian-zama: there's no much point in using a std::vector if we convert it to the raw pointer at the first opportunity ? can't we pass the std::vector here by reference and use proper indexing in the std::vector instead of pointer arithmetics ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -659,25 +768,25 @@ __host__ void host_unchecked_index_in_clears(
...
+      &mem_ptr->final_index_buf->packed_selectors, num_clears,
       mem_ptr->final_index_buf->h_indices, packed_len,
       mem_ptr->final_index_buf->possible_results_buf, bsks, ksks);
 
   host_aggregate_one_hot_vector<Torus>(
-      streams, index_ct, mem_ptr->final_index_buf->possible_results_ct_list,
-      num_clears, packed_len, mem_ptr->final_index_buf->aggregate_buf, bsks,
-      ksks);
+      streams, index_ct,
+      mem_ptr->final_index_buf->possible_results_ct_list.data(), num_clears,
```
andrei-stoian-zama: again please pass the vector instead of .data().

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -87,69 +314,63 @@ uint64_t scratch_cuda_compute_equality_selectors(
 template <typename Torus>
 __host__ void host_create_possible_results(
```
andrei-stoian-zama, line 444: please add a comment to explain this function

andrei-stoian-zama (changes requested):
Let's discuss. I think there are a few things that still need improving, starting with comments.

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
andrei-stoian-zama: you're now passing the `mem_ptr->luts_eq.data()` -> can't we pass the std::vector reference instead ?

backends/tfhe-cuda-backend/cuda/include/integer/vector_find.h
enzodimaria: CudaLweCiphertextListFFI couldn't be used here because there's no information about noise and degree in it
Plus, CudaRadixCiphertextFFI is often used to hold many lwes

enzodimaria:
PR soon updated @andrei-stoian-zama 
But I can't apply your third comment

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Already written, it's done in ffi.rs (there is a comment)

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: there's already a comment in the content of host_unchecked_match_value
do you want me to move it the head of host_create_possible_results ?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Indeed!

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Done :+1:

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: host_aggregate_one_hot_vector is based on a tree reduction.

host_and_reduce_selector_matrix does a simple reduction. For example, if a Radix ciphertext is composed of 9 blocks (rows), we compare it against N candidates (columns) in parallel. Assuming a max_degree of 5 it will do:

    ret = 0; // ret is a vector of N boolean elements

    // Pass 1: Accumulate up to max_degree blocks
    ret += row_0 + ... + row_4; 
    ret = (ret == 5); // LUT checks if all 5 blocks matched (0 or 1)

    // Pass 2: Accumulate up to (max_degree - 1) new blocks
    // (the previous 'ret' consumes 1 degree of noise)
    ret += row_5 + ... + row_8; 
    ret = (ret == 5); // LUT checks if (previous_match + 4 new blocks) is still a match etc...

The output is a vector of single-block boolean ciphertexts (one boolean per candidate).

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: I don't know host_pack_bivariate_blocks, we just prepare two contiguous flat buffers (packed_current and packed_value) that can be fed into a single batched bivariate PBS later, avoiding a loop over host_equality_check

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: output is a 1D vector of num_columns single-block LWEs
and each single-block LWE contains a boolean (1 or 0) indicating if the full multi-block equality holds for that specific column

enzodimaria: you can read comments in the match_value function to understand better

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
andrei-stoian-zama: got it. I think it could be done with tree reduction in a new PR

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+    dst_ptr[tid] = src_ptr[tid];
+  }
+}
+
+// Dispatches blocks from a flat batched src to multiple output ciphertexts.
+// used to replace multiples copies
 template <typename Torus>
-__host__ void host_compute_equality_selectors(
-    CudaStreams streams, CudaRadixCiphertextFFI *lwe_array_out_list,
+__host__ void host_scatter_to_ptr_array(
```
andrei-stoian-zama, line 173:
```
/* 
Copies radix-ciphertexts from a contiguous buffer of radix-ciphertexts to a list of independantly allocated radix-ciphertexts. It performs the inverse operation of `align_with_indexes`. It allows re-ordering the input radix-ciphertexts through the offsets argument.
 A single CUDA kernel performs the copy.

Args:
- stream, gpu_index: GPU index and its associated stream that performs the copy. This is a single GPU function
- dst_list: the CPU-side list of independantly allocated CudaRadixCiphertextFFI. 
- d_dst_ptrs: GPU copy of the CPU-side pointer list to the CudaRadixCiphertextFFI LWE arrays. 
- src: a CudaRadixCiphertextFFI containing n_output * num_blocks blocks that will be scattered to the destination CudaRadixCiphertextFFI. 
- d_src_offsets: output radix-ciphertext k is set to src[offsets[k]]. This allows re-ordering.
- h_src_offsets: cpu copy of the offsets (indexes), used for noise tracking
- num_outputs: number of input and output radix-ciphertexts
- num_blocks: number of blocks in each radix-ciphertext
*/
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+                     "items_in_acc out of luts_eq bounds");
+      integer_radix_apply_univariate_lookup_table<Torus>(
+          streams, &acc_slice, &acc_slice, bsks, ksks, luts_eq[items_in_acc],
+          num_columns);
+    }
+  }
+}
+
+// 1 ciphertext is compared to N clear values
+template <typename Torus>
```
andrei-stoian-zama, line 270:
```
/* 
Computes an indicator (boolean) vector of LWEs, representing whether the input radix-ciphertext is equal to any of a list of scalars.

Args:
- streams: the set of streams that can be used for this operation, supporting multi-GPU
- lwe_array_out_packed: a flat list of LWEs containing 0 or 1 depending on 
- lwe_array_in: input ciphertext that is matched with each clear value
- num_blocks: blocks in the ciphertext and in the decomposed clear values
- h_decomposed_cleartexts: CPU flat array of the decomposed clear values
- bksks, ksks: keys
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
-    int_equality_selectors_buffer<Torus> *mem_ptr, void *const *bsks,
+    int_eq_selectors_ct_vs_clears_buffer<Torus> *mem_ptr, void *const *bsks,
     Torus *const *ksks) {
 
   uint32_t num_possible_values = mem_ptr->num_possible_values;
   uint32_t message_modulus = mem_ptr->params.message_modulus;
+  uint32_t carry_modulus = mem_ptr->params.carry_modulus;
+  uint32_t max_degree = mem_ptr->max_degree;
 
+  // For every input block, precompute all possible equality results using a
```
andrei-stoian-zama: Replace this with:

// For every block in the input radix ciphertext `lwe_array_in`, precompute indicators of equality to all possible values of the block (0..2**(message_modulus+carry_modulus)-1). 
// Using only num_blocks multi-variate PBS an indicator matrix is produced where columns are one-hot vectors indicating the active value in the input block
// The indicator matrix can be used to assess equality for ALL the clear values to be matched

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+      streams, &mem_ptr->tmp_many_luts_output, lwe_array_in, bsks,
       (Torus *const *)ksks, mem_ptr->comparison_luts, message_modulus,
       mem_ptr->lut_stride);
 
-  mem_ptr->internal_cuda_streams.internal_streams_wait_for_main_stream_0(
-      streams);
+  // For each (candidate i, block j) pair, store the index inside the
+  // many-LUT output that corresponds to the precomputed equality result
+  // (x_j == h_decomposed_cleartexts[i][j])
+  Torus *h_map = mem_ptr->h_map;
```
andrei-stoian-zama, line 303: // Equality between the input radix-ciphertext and each decomposed clear value is determined by inspecting the indicator matrix cells corresponding to the decomposed clear value
// Eg. for clear value = 20, the decomposed clear value with 2_2 params is [1, 1, 0]. Thus we extract the cells [indicator_matrix[0,1], indicator_matrix[1,1], indicator_matrix[2,0]]. If all these cells are 1 then the input radix-ciphertext matched this clear value.

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+      h_map[j * num_possible_values + i] = (Torus)block_value * num_blocks + j;
+    }
+  }
+  cuda_memcpy_async_to_gpu(mem_ptr->d_map, h_map,
+                           safe_mul_sizeof<Torus>(total_blocks),
+                           streams.stream(0), streams.gpu_index(0));
 
-  uint32_t num_streams = mem_ptr->num_streams;
+  uint32_t lwe_size = mem_ptr->tmp_batched_comparisons.lwe_dimension + 1;
+  align_with_indexes<Torus><<<total_blocks, 256, 0, streams.stream(0)>>>(
```
andrei-stoian-zama, line 323:
```
// Extract the indicator booleans for each clear value to be matched and groups them in a flat contiguous 2d array of LWE  in order, so that each column contains the booleans corresponding to a single clear value. The 2d array is of size num_blocks (rows) x num_clear_values (cols)
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+      (Torus *)mem_ptr->tmp_batched_comparisons.ptr,
+      (Torus *)mem_ptr->tmp_many_luts_output.ptr, mem_ptr->d_map, lwe_size);
+  check_cuda_error(cudaGetLastError());
 
-  for (uint32_t i = 0; i < num_possible_values; i++) {
+  for (uint32_t b = 0; b < total_blocks; b++) {
```
andrei-stoian-zama, line 330: // Reset the noise levels since this buffer can be reused and noise is checked at the start of PBS

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+    mem_ptr->tmp_batched_comparisons.degrees[b] = 1;
+    mem_ptr->tmp_batched_comparisons.noise_levels[b] = NoiseLevel::NOMINAL;
+  }
 
-    uint32_t stream_idx = i % num_streams;
+  host_and_reduce_selector_matrix<Torus>(
```
andrei-stoian-zama, line 337: // And-reduce by rows the indicator 2d array. This produces a 1d vector of booleans, each indicating whether the input matches one of the clear values

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+      streams, &mem_ptr->packed_accumulator, &mem_ptr->tmp_batched_comparisons,
+      mem_ptr->luts_eq.data(), mem_ptr->num_luts_needed, num_possible_values,
+      num_blocks, max_degree, message_modulus, carry_modulus, bsks, ksks);
 
-    CudaStreams current_stream = mem_ptr->internal_cuda_streams[stream_idx];
+  copy_radix_ciphertext_slice_async<Torus>(
```
andrei-stoian-zama, line 343: // Place the output booleans into the output LWE list

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+      num_possible_values);
+}
 
-    CudaRadixCiphertextFFI *current_tmp_block_comparisons =
-        mem_ptr->tmp_block_comparisons[stream_idx];
-    int_comparison_buffer<Torus> *current_reduction_buffer =
-        mem_ptr->reduction_buffers[stream_idx];
+// N ciphertexts are compared to 1 encrypted value
+template <typename Torus>
+__host__ void host_compute_eq_selectors_cts_vs_ct(
```
andrei-stoian-zama, line 362: /* 
Computes an indicator (boolean) vector of LWEs, representing whether the input list of radix-ciphertext is equal to the target value that is searched for.

Args:
- streams: the set of streams that can be used for this operation, supporting multi-GPU
- lwe_array_out_list: boolean list, the i-th cell indicating whether value matches the i-th input
- inputs: list of radix-ciphertext to be searched
- value: radix-ciphertext that is searched for in the list
- num_inputs: number of values in the search list of radix-ciphertext
- num_blocks: number of blocks in the radix-ciphertexts used
- bksks, ksks: keys
*/

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+  Torus **h_input_ptrs = mem_ptr->h_input_ptrs;
+  for (uint32_t k = 0; k < num_inputs; k++) {
+    h_input_ptrs[k] = (Torus *)inputs[k].ptr;
+  }
+  cuda_memcpy_async_to_gpu(mem_ptr->d_input_ptrs, mem_ptr->h_input_ptrs,
+                           safe_mul_sizeof<Torus *>(num_inputs),
+                           streams.stream(0), streams.gpu_index(0));
 
-      uint32_t input_start = (uint32_t)block_value * num_blocks + j;
+  for (uint32_t base = 0; base < num_inputs; base += chunk_size) {
```
andrei-stoian-zama, line 386: why do we need chunking here? only to avoid a too high memory allocation without any benefit to latency (batch sufficiently big)?

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+    uint32_t current_chunk_size = std::min(chunk_size, num_inputs - base);
 
-      copy_radix_ciphertext_slice_async<Torus>(
-          current_stream.stream(0), current_stream.gpu_index(0),
-          current_tmp_block_comparisons, j, j + 1,
-          mem_ptr->tmp_many_luts_output, input_start, input_start + 1);
-    }
+    host_pack_chunk_eq_pairs<Torus>(
```
andrei-stoian-zama, line 393: I don't understand why you need to pack the "value" every time. packed_value_block  shouldn't change between for loop iterations..

In a new PR this could be done simply as a new univariate PBS function that calls bivariate packing for N lhs and 1 rhs radix-ct to pack.

Something in between the bivariate packing in cmux which does `host_pack_bivariate_blocks_with_single_block` (single block rhs) and the usual `host_pack_bivariate_blocks` which packs N radix-ct lhs  with N radix-ct rhs. 

Anyway, please add:
```
// Cache the chunk of the input radix-ciphertext list a temporary buffer. Also duplicate the searched-for value radix-ct blocks in a manner in which the bivariate PBS can pack pairs of (inputs[i][k], value[k]) for-each i in [0..num_inptus], for-each k in [0..num_blocks]
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+        streams.stream(0), streams.gpu_index(0), &mem_ptr->packed_current_block,
+        &mem_ptr->packed_value_block, mem_ptr->d_input_ptrs, inputs, value,
+        base, current_chunk_size, num_blocks);
 
-    CudaRadixCiphertextFFI *current_output_block = &lwe_array_out_list[i];
+    uint32_t total = current_chunk_size * num_blocks;
+    integer_radix_apply_bivariate_lookup_table<Torus>(
```
andrei-stoian-zama, line 402:
```
// Compute an indicator value for each block of each input radix-ciphertext, indicating whether the block is equal to the corresponding block in the searched for radix-ciphertext value
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+        streams, &mem_ptr->packed_current_block, &mem_ptr->packed_current_block,
+        &mem_ptr->packed_value_block, bsks, ksks, mem_ptr->equality_lut, total,
+        message_modulus);
 
-    host_integer_are_all_comparisons_block_true<Torus>(
-        current_stream, current_output_block, current_tmp_block_comparisons,
-        current_reduction_buffer, bsks, (Torus **)ksks, num_blocks);
-  }
+    host_and_reduce_selector_matrix<Torus>(
```
andrei-stoian-zama, line 411:
```
// The indicator boolean values produced by the LUT are interpreted as a 2d array of num_blocks (rows) x num_inputs (cols). And-reduce the array to determine which inputs match the searched-for value. This produces a 1d vector of booleans
```

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
```diff
@@ -5,79 +5,317 @@
...
+        streams, &mem_ptr->packed_accumulator, &mem_ptr->packed_current_block,
+        mem_ptr->luts_eq.data(), mem_ptr->num_luts_needed, current_chunk_size,
+        num_blocks, max_degree, message_modulus, carry_modulus, bsks, ksks);
 
-  mem_ptr->internal_cuda_streams.main_stream_0_wait_for_internal_streams(
-      streams);
+    for (uint32_t c = 0; c < current_chunk_size; c++) {
```
andrei-stoian-zama, line 418:
```
// Copy the booleans into individual output radix-ciphertexts which have a single block
```

andrei-stoian-zama (changes requested):
Some comment suggestions

backends/tfhe-cuda-backend/cuda/src/integer/vector_find.cuh
enzodimaria: Yes indeed, but it was simpler to implement for me at first and the cost is almost nothing
It can refactored in a next PR

enzodimaria:
PR updated @andrei-stoian-zama

andrei-stoian-zama (approved):
Excellent job ! Looks amazing !Thanks for fixing all the comments.

In a next PR two potential optimizations:
- better batching in the and-reduce (since rows are only 32 blocks wide in 2_2), and-reducing two rows at the same time would be nice
- same for the one-hot summation
- avoiding duplication of value in packing by refactoring the bivariate PBS calls

PR #3532, opened 2026-05-04
fix(shortint): checked_less_or_equal not calling correct function

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1414#issue-4375585668

### DIFF
--- a/tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
+++ b/tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
@@ -1,3 +1,8 @@
+#![allow(
+    clippy::eq_op,
+    reason = "explicitly test that clear and fhe op give the same result in the equality case"
+)]
+
 use crate::shortint::keycache::KEY_CACHE;
 use crate::shortint::parameters::test_params::*;
 use crate::shortint::server_key::tests::parameterized_test::create_parameterized_test;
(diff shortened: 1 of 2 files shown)


tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
```diff
@@ -1044,6 +1050,204 @@ where
...
+
+        let ct_res = sks.checked_greater(&ctxt_0, &ctxt_1).unwrap();
+        let dec_res = cks.decrypt(&ct_res);
+
+        assert_eq!((clear_0 > clear_1) as u64, dec_res);
+    }
+
+    let clear = rng.gen::<u64>() % modulus;
+    let ctxt_0 = cks.encrypt(clear);
+    let ctxt_1 = cks.encrypt(clear);
```
mayeul-zama: The equal case could be in the loop to be more thorough

tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
```diff
@@ -1044,6 +1050,204 @@ where
     }
 }
 
+fn shortint_checked_greater<P>(param: P)
```
mayeul-zama, line 1058: To share some code, all new tests could call a function which takes
- a closure on encrypted inputs
- a closure on clear inputs
and compare the results on inputs a, b with the 3 cases `a=b`, `a<b` and `a>b`.

Btw we currenlty have no strong guarantee that well always get at least one `clear_0 < clear_1` and one `clear_0 > clear_1` cases.

Maybe we could do the same thing for existing test

nsarlin-zama: I copied existing tests. In theory I like your idea but I'm not sure it's worth refactoring all these tests for a small fix like that?

tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
nsarlin-zama: what do you mean? The idea here is to make sure the equal case is tested since it's what triggered the bug

tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
mayeul-zama: As you prefer

tfhe/src/shortint/server_key/tests/parameterized_test_bivariate_pbs_compliant.rs
mayeul-zama: We could test more equality cases with many random cleartext instead of a single one

nsarlin-zama: done!

mayeul-zama (dismissed):
Thanks!

nsarlin-zama:
clippy was complaining about useless `a < a` checks, I added an "allow" because I think that in that case it's better to be explicit about the clear operation that is tested

mayeul-zama (approved):
Thanks!

PR #3534, opened 2026-05-04
fix(integer): avoid panic in compact list expand with incorrect size

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1415#issue-4375593075

In theory this should not be triggerable on inputs that went through conformance, but since the calling function is already returning a Result, as a defense-in-depth it might be preferable to handle the error gracefully here too.


mayeul-zama (approved):
Thanks!

PR #3536, opened 2026-05-05
chore(ci): add checkout token for internal backward flow

SouchonTheo:
add checkout token for internal backward flow


IceTDrinker (approved):
Looks good, thanks !

PR #3546, opened 2026-05-06
fix(gpu): refactor ks params into a ffi struct

andrei-stoian-zama:
Refactors KS params in CUDA bindings into a FFI struct


guillermo-oyarzun (approved):
nice to see how things get reduced little by little

PR #3554, opened 2026-05-11
chore(gpu): run gpu summary benchmarks weekly on 1h100

andrei-stoian-zama:
Also run the summary benchmarks weekly on 1xh100

### DIFF
--- a/.github/workflows/benchmark_gpu_weekly.yml
+++ b/.github/workflows/benchmark_gpu_weekly.yml
@@ -13,6 +13,26 @@ permissions: {}
 # zizmor: ignore[concurrency-limits] only GitHub can trigger this workflow
 
 jobs:
+  run-benchmarks-1-h100-summary:
+    name: benchmark_gpu_weekly/run-benchmarks-1-h100-summary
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark_gpu_weekly.yml
```diff
@@ -13,6 +13,26 @@ permissions: {}
...
+  run-benchmarks-1-h100-summary:
+    name: benchmark_gpu_weekly/run-benchmarks-1-h100-summary
+    if: github.repository == 'zama-ai/tfhe-rs'
+    uses: ./.github/workflows/benchmark_gpu_common.yml
+    with:
+      profile: single-h100
+      hardware_name: n3-H100x1
+      command: summary
+      bench_type: both
+      params_type: classical + multi_bit
```
soonum, line 25: Do you need to run both classical and multi-bit parameters set ?

andrei-stoian-zama: yes, I think so. we currently do it for the 8xgpu

soonum (approved):
Let's goo

PR #3555, opened 2026-05-12
feat: make SystemTime Versionable

dvdplm:
Pretty much what it says in the title. In the KMS we currently have to cast SystemTime to u64 to work around this limitation.


nsarlin-zama (approved):
Thanks a lot!

PR #3557, opened 2026-05-12
Ts/wasm/create mini wasm sdk to lower the weight of the file

SouchonTheo:
# Creation of the client wasm package
this is following this [issue](https://github.com/zama-ai/tfhe-rs-internal/issues/1394)

## Client package

I create a dedicated folder client in js_high_level_api lived
if you activate few feature you will only support what's inside it which is what we want

Also in the makefile/cargo.toml some optimization has been made to reduce the size a little bit more (only Oz in fine)

some extra function have been added from my perspective for example I found it weird to have TfheClientKey needed for TfheCompactPublicKey and not bring all the creation or parsing process for TfheClientKey
This is something open to the discussion of course (not sure it make the package heavier a lot)

we went from 4237274 bytes to 1348195 bytes which is a reduction of almost 70%

## tooling

I also moved the makefile part inside a file under make/... to make it more readable for everyone

I also used it has a reviewer to correct my code before going in PR

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557)

### DIFF
--- a/.github/workflows/make_release_tfhe.yml
+++ b/.github/workflows/make_release_tfhe.yml
@@ -20,6 +20,14 @@ on:
         description: "Push web compat (cross-origin) js package"
         type: boolean
         default: true
+      push_web_client_package:
+        description: "Push web client js package"
+        type: boolean
+        default: true
+      push_web_compat_client_package:
+        description: "Push web compat (cross-origin) client js package"
+        type: boolean
+        default: true
       push_node_package:
         description: "Push node js package"
         type: boolean
@@ -117,6 +125,40 @@ jobs:
 
       - name: Publish web compat (cross-origin) package
         if: ${{ inputs.push_web_compat_package }}
+        uses: JS-DevTools/npm-publish@0fd2f4369c5d6bcfcde6091a7c527d810b9b5c3f
+        with:
+          package: tfhe/pkg/package.json
+          dry-run: ${{ inputs.dry_run }}
+          tag: ${{ env.NPM_TAG }}
+          provenance: true
+
+      - name: Build web client package
+        if: ${{ inputs.push_web_client_package }}
+        run: |
+          rm -rf tfhe/pkg
+
+          make build_web_js_api_parallel_client
+          sed -i 's/"tfhe"/"tfhe-client"/g' tfhe/pkg/package.json
+
+      - name: Publish web client package
+        if: ${{ inputs.push_web_client_package }}
+        uses: JS-DevTools/npm-publish@0fd2f4369c5d6bcfcde6091a7c527d810b9b5c3f
+        with:
+          package: tfhe/pkg/package.json
+          dry-run: ${{ inputs.dry_run }}
+          tag: ${{ env.NPM_TAG }}
+          provenance: true
+
+      - name: Build web compat (cross-origin) client package
+        if: ${{ inputs.push_web_compat_client_package }}
+        run: |
+          rm -rf tfhe/pkg
+
+          make build_web_js_api_client
+          sed -i 's/"tfhe"/"tfhe-compat-client"/g' tfhe/pkg/package.json
+
+      - name: Publish web compat (cross-origin) client package
+        if: ${{ inputs.push_web_compat_client_package }}
         uses: JS-DevTools/npm-publish@0fd2f4369c5d6bcfcde6091a7c527d810b9b5c3f
         with:
           package: tfhe/pkg/package.json
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ BACKWARD_COMPAT_DATA_GEN_VERSION:=$(TFHE_VERSION)
 CORRUPTED_INPUTS_TEST=tests/corrupted_inputs_deserialization
 TEST_VECTORS_DIR=apps/test-vectors
 CURRENT_TFHE_VERSION:=$(shell grep '^version[[:space:]]*=' tfhe/Cargo.toml | cut -d '=' -f 2 | xargs)
-WASM_PACK_VERSION="0.13.1"
+WASM_PACK_VERSION="0.15.0"
 WASM_BINDGEN_VERSION:=$(shell cargo tree --target wasm32-unknown-unknown -e all --prefix none | grep "wasm-bindgen v" | head -n 1 | cut -d 'v' -f2)
 WEB_RUNNER_DIR=web-test-runner
 WEB_SERVER_DIR=tfhe/web_wasm_parallel_tests
@@ -777,39 +777,8 @@ build_c_api_experimental_deterministic_fft: install_rs_check_toolchain
 		--features=boolean-c-api,shortint-c-api,high-level-c-api,zk-pok,experimental-force_fft_algo_dif4 \
 		-p tfhe
 
-.PHONY: build_web_js_api # Build the js API targeting the web browser, in sequential or cross origin parallelism modes.
-build_web_js_api: install_wasm_pack
-	cd tfhe && \
-	RUSTFLAGS="$(WASM_RUSTFLAGS)" wasm-pack build --release --target=web \
-		-- --features=boolean-client-js-wasm-api,shortint-client-js-wasm-api,integer-client-js-wasm-api,zk-pok,extended-types,cross-origin-wasm-api && \
-	find pkg/snippets -type f -iname worker_helpers.js -exec sed -i 's|import("../../..")|import("../../../tfhe.js")|g' {} \;
-	cp utils/wasm-par-mq/js/coordinator.js tfhe/pkg/
-	jq '.files += ["snippets"]' tfhe/pkg/package.json > tmp_pkg.json && mv -f tmp_pkg.json tfhe/pkg/package.json
-
-.PHONY: build_web_js_api_parallel # Build the js API targeting the web browser with parallelism support
-# parallel wasm requires specific build options, see https://github.com/rust-lang/rust/pull/147225
-build_web_js_api_parallel: install_rs_check_toolchain install_wasm_pack install_wasm_bindgen_cli
-	cd tfhe && \
-	rustup component add rust-src --toolchain $(RS_CHECK_TOOLCHAIN) && \
-	RUSTFLAGS="$(WASM_RUSTFLAGS) -C target-feature=+atomics,+bulk-memory \
-		-Clink-arg=--shared-memory \
-		-Clink-arg=--max-memory=1073741824 \
-		-Clink-arg=--import-memory \
-		-Clink-arg=--export=__wasm_init_tls \
-		-Clink-arg=--export=__tls_size \
-		-Clink-arg=--export=__tls_align \
-		-Clink-arg=--export=__tls_base" \
-		rustup run $(RS_CHECK_TOOLCHAIN) wasm-pack build --release --target=web \
-		-- --features=boolean-client-js-wasm-api,shortint-client-js-wasm-api,integer-client-js-wasm-api,parallel-wasm-api,zk-pok,extended-types \
-		-Z build-std=panic_abort,std && \
-	find pkg/snippets -type f -iname workerHelpers.js -exec sed -i "s|const pkg = await import('..\/..\/..');|const pkg = await import('..\/..\/..\/tfhe.js');|" {} \;
-	jq '.files += ["snippets"]' tfhe/pkg/package.json > tmp_pkg.json && mv -f tmp_pkg.json tfhe/pkg/package.json
-
-.PHONY: build_node_js_api # Build the js API targeting nodejs
-build_node_js_api: install_wasm_pack
-	cd tfhe && \
-	RUSTFLAGS="$(WASM_RUSTFLAGS)" wasm-pack build --release --target=nodejs \
-		-- --features=boolean-client-js-wasm-api,shortint-client-js-wasm-api,integer-client-js-wasm-api,zk-pok,extended-types
+# wasm pkg build rules (full + client, web + node, parallel + cross-origin).
+include make/wasm.mk
 
 .PHONY: build_tfhe_csprng # Build tfhe_csprng
 build_tfhe_csprng:
@@ -1397,7 +1366,8 @@ test_zk_wasm_x86_compat_ci: check_nvm_installed
 	$(MAKE) test_zk_wasm_x86_compat
 
 .PHONY: test_zk_wasm_x86_compat # Check compatibility between wasm and x86_64 proofs
-test_zk_wasm_x86_compat: build_node_js_api
+# Uses the client pkg (lighter to compile) — compat is package-agnostic.
+test_zk_wasm_x86_compat: build_node_js_api_client
 	cd tfhe/tests/zk_wasm_x86_test && npm install
 	RUSTFLAGS="$(RUSTFLAGS)" cargo test --profile $(CARGO_PROFILE) \
 		-p tfhe --test zk_wasm_x86_test --features=integer,zk-pok
--- a/make/wasm.mk
+++ b/make/wasm.mk
@@ -1,0 +1,84 @@
+WEB_CLIENT_OUT_DIR ?= pkg
+
+.PHONY: build_web_js_api # Build the js API targeting the web browser, in sequential or cross origin parallelism modes.
+build_web_js_api: install_wasm_pack
+	cd tfhe && \
+	RUSTFLAGS="$(WASM_RUSTFLAGS)" wasm-pack build --release --target=web \
+		-- --features=boolean-client-js-wasm-api,shortint-client-js-wasm-api,integer-client-js-wasm-api,zk-pok,extended-types,cross-origin-wasm-api && \
+	find pkg/snippets -type f -iname worker_helpers.js -exec sed -i 's|import("../../..")|import("../../../tfhe.js")|g' {} \;
+	cp utils/wasm-par-mq/js/coordinator.js tfhe/pkg/
+	jq '.files += ["snippets"]' tfhe/pkg/package.json > tmp_pkg.json && mv -f tmp_pkg.json tfhe/pkg/package.json
+
+.PHONY: build_web_js_api_parallel # Build the js API targeting the web browser with parallelism support
+# parallel wasm requires specific build options, see https://github.com/rust-lang/rust/pull/147225
+build_web_js_api_parallel: install_rs_check_toolchain install_wasm_pack install_wasm_bindgen_cli
+	cd tfhe && \
+	rustup component add rust-src --toolchain $(RS_CHECK_TOOLCHAIN) && \
+	RUSTFLAGS="$(WASM_RUSTFLAGS) -C target-feature=+atomics,+bulk-memory \
+		-Clink-arg=--shared-memory \
+		-Clink-arg=--max-memory=1073741824 \
+		-Clink-arg=--import-memory \
+		-Clink-arg=--export=__wasm_init_tls \
+		-Clink-arg=--export=__tls_size \
+		-Clink-arg=--export=__tls_align \
+		-Clink-arg=--export=__tls_base" \
+		rustup run $(RS_CHECK_TOOLCHAIN) wasm-pack build --release --target=web \
+		-- --features=boolean-client-js-wasm-api,shortint-client-js-wasm-api,integer-client-js-wasm-api,parallel-wasm-api,zk-pok,extended-types \
+		-Z build-std=panic_abort,std && \
+	find pkg/snippets -type f -iname workerHelpers.js -exec sed -i "s|const pkg = await import('..\/..\/..');|const pkg = await import('..\/..\/..\/tfhe.js');|" {} \;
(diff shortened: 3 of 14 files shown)


nsarlin-zama (commented):
There is something I'm not sure I understand about the wasm/x86 compat test:

This test is supposed to:
- serialize a crs in rust
- deserialize in wasm
- generate a proof in wasm with the crs
- deserialize and check the proof in rust

The goal is to check that we don't have compatibility issues between wasm and rust due to different word sizes (32 vs 64b).

The whole flow is called from rust. This is not supposed to be a standalone js test. And there is nothing to bench I guess? All of this is covered by more targeted benches

It should not be impacted by the package type (compat, client,...) So I would not include it in the refacto, I think it's fine as it is. However it could benefit from using the client package if it reduces compilation times.

> @nsarlin-zama partially reviewed 18 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-:-OsebZ7gCV366mbqJ9gH:bkccz4h)* status: 18 of 40 files reviewed, 3 unresolved discussions (waiting on IceTDrinker, soonum, and SouchonTheo).

___
*[`Makefile` line 1343 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWaSln1cODpHSYGPRr:-OsWaSln1cODpHSYGPRs:b-x5kznt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/02d7b43d3727516bc05bdfae0b56d06da4740d61/Makefile#L1343)):*
> ```Makefile
> 
> .PHONY: test_zk_wasm_x86_compat_client_ci
> test_zk_wasm_x86_compat_client_ci: check_nvm_installed
> ```

I'm not sure this is needed, the test above checks that a proof generated in wasm is compatible with x86. This is just meant to be a "hardware" architecture check (32b vs 64b), so it should not be impacted by the package

The "compat" name here is not the same as the "compat" in the package name. This is just "architecture compatibility".

However you can certainly run this test only with the client package if it's faster to build
___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWbVxW6d9QprC6PLGp:-OsWbVxW6d9QprC6PLGq:bgukg5n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/02d7b43d3727516bc05bdfae0b56d06da4740d61/.github/workflows/aws_tfhe_wasm_tests.yml#L152)):*
> ```YAML
>       - name: Run x86_64/wasm zk compatibility tests (client pkg)
>         run: |
>           make test_zk_wasm_x86_compat_client_ci
> ```

same, no need to duplicate this for the client/complete package
___
*[`tfhe/web_wasm_parallel_tests/build.sh` line 63 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWlU0KEeuFEADATiyJ:-OsWlU0KEeuFEADATiyK:b69g6ij) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/23930b8b491bbec6917260cac15caaedcbf72f4a/tfhe/web_wasm_parallel_tests/build.sh#L63)):*
> ```Shell
> cp -r ./pkg-client dist/pkg-client
> 
> # Step 5: copy fixtures (optional, used by x86CompatTest/Bench)
> ```

I'm not sure this should be handled in this script, the wasm/x86 test is rather stand alone

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557)* status: 18 of 40 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, soonum, and SouchonTheo).

___
*[`tfhe/tests/zk_wasm_x86_test.rs` line 35 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OseeId8Erzz3k5Cu47J:-OseeId8Erzz3k5Cu47K:b-7h97ni) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/23930b8b491bbec6917260cac15caaedcbf72f4a/tfhe/tests/zk_wasm_x86_test.rs#L35)):*
> ```XML
> 
>     println!("Generating crs");
>     // TODO: 2048
> ```

why ?

SouchonTheo (commented):
You are right my naming was not good and the need of the test is dabatable

I think we need to bench a little the client package because I tried some flag and feature optimisation and the issue is that we don't have the right function in client to be able to run `compactPublicKeyZeroKnowledgeBench`
So I took the x86 pseudo test to run a little bench

I renamed it `Encrypt + Prove from fixture`

I also removed the double uselss ci

> @SouchonTheo made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-:-OsuVWkU0yFxAiesrtoI:btvsd6q)* status: 13 of 42 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, nsarlin-zama, and soonum).

___
*[`Makefile` line 1343 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWaSln1cODpHSYGPRr:-OsuyO2VCRo9ZH4H3vKS:bsm7ddp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/02d7b43d3727516bc05bdfae0b56d06da4740d61/Makefile#L1343)):*

I'm not sure this is needed, the test above checks that a proof generated in wasm is compatible with x86. This is just meant to be a "hardware" architecture check (32b vs 64b), so it should not be impacted by the package

The "compat" name here is not the same as the "compat" in the package name. This is just "architecture compatibility".

However you can certainly run this test only with the client package if it's faster to build
</blockquote></details>

yes exactly
Now we only run the client one
___
*[`tfhe/web_wasm_parallel_tests/build.sh` line 63 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWlU0KEeuFEADATiyJ:-OsuW6Mz4NcFIUvtyxES:b-e8mqtn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/23930b8b491bbec6917260cac15caaedcbf72f4a/tfhe/web_wasm_parallel_tests/build.sh#L63)):*

I'm not sure this should be handled in this script, the wasm/x86 test is rather stand alone
</blockquote></details>

cf main comment section
___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OsWbVxW6d9QprC6PLGp:-OsuyW2G2to7OkOUQKZ_:b-hfpv0l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/02d7b43d3727516bc05bdfae0b56d06da4740d61/.github/workflows/aws_tfhe_wasm_tests.yml#L152)):*

same, no need to duplicate this for the client/complete package
</blockquote></details>

removed
___
*[`tfhe/tests/zk_wasm_x86_test.rs` line 35 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OseeId8Erzz3k5Cu47J:-OsuTqiI2b2apyMUDvpL:bt6opgx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/23930b8b491bbec6917260cac15caaedcbf72f4a/tfhe/tests/zk_wasm_x86_test.rs#L35)):*

why ?
</blockquote></details>

old todo 
we wanted to test the perf diff from client and full
not needed anymore as we implemented a solution in the browser

nsarlin-zama (commented):
If I understand correctly, you re-use the wasm/x86 compat test because in the client you don't have access to the functions needed to generate the data for the test? But looking at how it's done here, it feels like we are duplicating some tests. It looks when you run `make test_zk_wasm_x86_compat` you already do: 
1. generate crs and key from x86
2. encrypt and prove from wasm
3. load proof and verify in x86 

Then the `Encrypt + Prove from fixture` re-do step 2 alone, but is this needed if we already did 1+2+3? Or the test is doing something different? (maybe I'm missing something)

Maybe we need to re-think a bit what these tests do and how they work. Before we had a full wasm/js test because we could do everything in js, and another wasm/x86 test just for compatibility. But with the light client package it is not possible anymore to do everything in js, so maybe the x86 test is not needed anymore and can be directly merged in the web\_wasm\_parallel\_tests/ with the other tests ?

> @nsarlin-zama reviewed 3 files and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-:-Ot3kt384nkVDwoe8ddP:b-4wracp)* status: 16 of 42 files reviewed, 2 unresolved discussions (waiting on IceTDrinker, soonum, and SouchonTheo).

___
*[`tfhe/web_wasm_parallel_tests/Makefile` line 10 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-Ot3kNbW20Rdl3_dhFZb:-Ot3kNbW20Rdl3_dhFZc:b-h8t4fr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/c4cfdbfc6ef4a490ed7ceaa311a864164cdb73bc/tfhe/web_wasm_parallel_tests/Makefile#L10)):*
> ```Makefile
> 	else \
> 		echo "==> Fixtures missing, generating via 'make test_zk_wasm_x86_compat'..."; \
> 		$(MAKE) -C ../.. test_zk_wasm_x86_compat; \
> ```

Here you are running the full test to generate the data, so it will be a duplicate test if you run the fixture test later I think?

SouchonTheo (commented):
I made a patch around your comment
hope it will be enough to resolve the duplicaiton issue
remain available for further patch

> @SouchonTheo made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-:-Ou7q2dyBGzEMVJ8fxNH:b-tib1nq)* status: 13 of 45 files reviewed, 1 unresolved discussion (waiting on IceTDrinker, nsarlin-zama, and soonum).

___
*[`tfhe/web_wasm_parallel_tests/Makefile` line 10 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-Ot3kNbW20Rdl3_dhFZb:-Ou7qBM16LxqPmNUPfrz:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/28df6005b5fed79ff45a73ffdf220ce796ce08d4/tfhe/web_wasm_parallel_tests/Makefile#L10)):*

Here you are running the full test to generate the data, so it will be a duplicate test if you run the fixture test later I think?
</blockquote></details>

done

IceTDrinker (changes requested):
While I understand the will to refactor the test structure, I think it's not the time and place.

The worker.js prior to the refacto is 778 lines and now it's scattered across a lot of files

The noise added by the various code moves makes it very very hard to:
- follow what's actually important (the re-org of tfhe JS wasm API)
- makes it hard to trust the new setup actually works

My recommendation for this PR (since I have not really reviewed the interesting part due to everything that moved:
- drop anything not strictly related to the js/wasm update
- make it a dumb PR by adding things in the right locations "dumb" copy paste in Makefile for build targets, "dumb" tests additions using the existing system, no HTML update for now by re-using the previous test harness
- then in follow-ups: one PR for the Makefile update, one PR for the test refactor doing one thing at a time (build script, fixtures etc.) with very small commits which all compile to make sure we can review it knowing things still work

I appreciate the will to update things, but "cleaning up everything the reviewers know" will not help us achieve our goal here :)

Focusing on the feature and then proposing small clean ups will allow to achieve both objectives with less cognitive loads and less opportunities for mistakes

> @IceTDrinker partially reviewed 39 files and all commit messages, and made 15 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-:-OuI2wmv3xOtAEZoHuOt:b-bp7263)* status: 38 of 47 files reviewed, 15 unresolved discussions (waiting on nsarlin-zama, soonum, and SouchonTheo).

___
*[`Makefile` line 214 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuHyKZ_3yEzs35XiNUh:-OuHyKZjBNHiVL2g2jLM:b13551a) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/Makefile#L214)):*
> ```Makefile
> 	rm linelint_checksum && \
> 	chmod +x linelint
> ```

let's maybe keep the Makefile in one place and plan a PR to do the refacto

logic is : let's keep the focus on small diffs for the feature PR

then the refacto PR can be only moving stuff around, much easier to review
___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 138 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuHyYl6763iWOLNzc9q:-OuHyYl6763iWOLNzc9r:b-p5dt3c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/.github/workflows/aws_tfhe_wasm_tests.yml#L138)):*
> ```YAML
>       - name: Run parallel wasm tests
>         run: |
>           make test_web_js_api_parallel_chrome_ci
> ```

let's maybe keep the various targets here ? unsure
___
*[`ci/webdriver.py` line 415 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI-Xj04svvmGiS8Vsg:-OuI-Xj04svvmGiS8Vsh:bkmcfyg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/ci/webdriver.py#L415)):*
> ```Python
>         out_path = pathlib.Path(capture_out)
>         out_path.parent.mkdir(parents=True, exist_ok=True)
>         out_path.write_bytes(base64.b64decode(result.pop(capture_key)))
> ```

do we need to pop here ? (removes the data associated to capture_key
___
*[`ci/webdriver.py` line 432 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI-nJ82ElFicKjQAFo:-OuI-nJ9-SOxvpDp_-oL:b-9065h2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/ci/webdriver.py#L432)):*
> ```Python
>     """
>     # Keep only numeric measurements, so non-numeric payloads a test may return
>     # (e.g. fixtureEncryptProveTest's base64 `proof_b64`) never leak into the dump.
> ```

do we care about that ? otherwise can we not have another output box to avoid mixing the two ?
___
*[`ci/webdriver.py` line 577 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI-va47NpOYR8RQV2P:-OuI-va47NpOYR8RQV2Q:b3yphbm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/ci/webdriver.py#L577)):*
> ```Python
> 
>         dump_benchmark_results(benchmark_results, browser_kind)
>     finally:
> ```

no except to log a potential error ?
___
*[`make/wasm.mk` line 1 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuHzl7tDCIrRwG4ZF8j:-OuHzl7tDCIrRwG4ZF8k:br8svl5) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/make/wasm.mk#L1)):*
> ```Makefile
> WEB_CLIENT_OUT_DIR ?= pkg
> ```

let's keep this change separate to make the diff easier to review
___
*[`tfhe/Cargo.toml` line 267 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuHzub80wNt-qkoTTJ9:-OuHzub9CAZKg8UKNRhK:b-ruk3oo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/tfhe/Cargo.toml#L267)):*
> ```TOML
> 
> [package.metadata.wasm-pack.profile.release]
> wasm-opt = ["-Oz", "--enable-simd", "--enable-threads", "--enable-bulk-memory"]
> ```

that's no good since it applies to all web packages ?

also didn't we say Oz was less performant than O ?

I would remove this : keeps the default for the previous packages

the new build can use Oz if the performance is the same as O
___
*[`tfhe/src/js_on_wasm_api/mod.rs` line 71 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI06lJAry7UhEnLKtJ:-OuI06lJAry7UhEnLKtK:b-khzk40) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/tfhe/src/js_on_wasm_api/mod.rs#L71)):*
> ```XML
> 
> #[cfg(feature = "shortint-client-js-wasm-api")]
> pub(crate) fn generic_safe_deserialize_conformant<
> ```

same as the makefile the generic safe serialize stuff is not related to the current PR's functionnality, so let's revert this diff and do a potential follow-up PR to refactor
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/config.rs` line 5 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI0TtvAwB26r61Kzz6:-OuI0TtvAwB26r61Kzz7:buol1wl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/tfhe/src/js_on_wasm_api/js_high_level_api/config.rs#L5)):*
> ```XML
> 
> #[wasm_bindgen]
> pub struct TfheConfigBuilder(pub(crate) crate::high_level_api::ConfigBuilder);
> ```

maybe import it since it's used 3 times ?

or keep the original approach of an alias to the correct module
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/mod.rs` line 14 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI0siiBrTtBmjYXrXY:-OuI0siiBrTtBmjYXrXZ:b-1y7zp2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/tfhe/src/js_on_wasm_api/js_high_level_api/mod.rs#L14)):*
> ```XML
> use super::into_js_error;
> 
> pub(crate) fn catch_panic_result<F, R>(closure: F) -> Result<R, JsError>
> ```

those could be kept here ? since they are used by the whole JS API ? (the catch panics wrappers)
___
*[`tfhe/web_wasm_parallel_tests/build.sh` line 1 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI1BHHEqJTV17GeqJR:-OuI1BHI2mwmeWJUw3Lw:bbpruuu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1578a2aa8ccf1e8a1ecf261b482375bcad1b29f1/tfhe/web_wasm_parallel_tests/build.sh#L1)):*
> ```Shell
> #!/usr/bin/env bash
> ```

what is the role of that script ?
___
*[`tfhe/web_wasm_parallel_tests/jest.config.js` line 1 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI1J4sBHFQpZkzGg1O:-OuI1J4sBHFQpZkzGg1P:b-1qvv7z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/tfhe/web_wasm_parallel_tests/jest.config.js#L1)):*
> ```JavaScript
> const secs = 3600; // 60 Minutes
> ```

why was this removed ?
___
*[`tfhe/web_wasm_parallel_tests/package.json` line 17 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI1RbE5xqKq_msfjMO:-OuI1RbE5xqKq_msfjMP:b-tmb8hg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/tfhe/web_wasm_parallel_tests/package.json#L17)):*
> ```JSON
>   "license": "BSD-3-Clause-Clear",
>   "devDependencies": {
>     "@babel/preset-env": "^7.29.4",
> ```

why removing it ?
___
*[`tfhe/web_wasm_parallel_tests/worker.js` line 1 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3557#-OuI1jil6JehBWDTw8dQ:-OuI1jil6JehBWDTw8dR:b-1qvv7z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2b2f547ba4f7528822023e30d5d39b6523191852/tfhe/web_wasm_parallel_tests/worker.js#L1)):*
> ```JavaScript
> import * as Comlink from "comlink";
> ```

why was this removed ?

PR #3560, opened 2026-05-13
fix: increase margin for common mask noise test

mayeul-zama:


IceTDrinker:
as discussed let's first make sure the secret keys have averegae hamming weight forced to 1/2, most noise tests behaved correctly with that change only

mayeul-zama:
> as discussed let's first make sure the secret keys have averegae hamming weight forced to 1/2, most noise tests behaved correctly with that change only

We use mean noise reduction in the test which makes noise independent from secret key content

IceTDrinker:
> > as discussed let's first make sure the secret keys have averegae hamming weight forced to 1/2, most noise tests behaved correctly with that change only
> 
> We use mean noise reduction in the test which makes noise independent from secret key content

Untrue if the output is the output of the PBS meaning the key encrypting the GGSWs has an influence, the noise at the mod switch is independent from the key perhaps

IceTDrinker (approved):
Thanks !

PR #3565, opened 2026-05-15
chore(hlapi): add has_*_key methods to check if dedicated key exists

tmontaigu:
Closes https://github.com/zama-ai/tfhe-rs-internal/issues/1424


nsarlin-zama (approved):
thanks!

PR #3570, opened 2026-05-18
fix(gpu): move ci to aws

andrei-stoian-zama:
Move all workflows to AWS

### DIFF
--- a/.github/workflows/benchmark_gpu_common.yml
+++ b/.github/workflows/benchmark_gpu_common.yml
@@ -148,7 +148,7 @@ jobs:
       matrix:
         # explicit include-based build matrix, of known valid options
         include:
-          - cuda: "12.8"
+          - cuda: "12.2"
             gcc: 11
     steps:
       - name: Checkout tfhe-rs repo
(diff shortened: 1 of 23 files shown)


.github/workflows/benchmark_gpu_common.yml
```diff
@@ -148,7 +148,7 @@ jobs:
       matrix:
         # explicit include-based build matrix, of known valid options
         include:
-          - cuda: "12.8"
```
guillermo-oyarzun, line 151: why do we need to move to an older version?

andrei-stoian-zama: the cuda version that is set here is a hint about the cuda version installed in the AMI. for AWS our AMI has 12.2. our installation scripts do not re-install cuda in the CI, as it is quite long to do so. 

in the scaleway PR we improve upon this situation (allow multiple cuda versions, allow installation, etc)

soonum (approved):

.github/workflows/benchmark_gpu_common.yml
guillermo-oyarzun: just wondering if that will imply a slightly worse performance. At some point we need need to review the building blocks there to check how does it change with respect to other instances or cuda versions

PR #3573, opened 2026-05-20
chore(ci): fix pcc matrix results handling

soonum:
The usage of 'builds-result' workflow output came from the need to treat instance teardown failures as infra issue not a CI one. Thus if the build was green but the teardown failed, the dev could see the CI pipeline as green.

However this design had a issue. When multiple jobs were running in parallel via the 'runners' matrix, the 'builds-result' output would be set by the job that finished last. So if a quicker job had failed, the 'failure' status would be overwritten by the one that finished successfully a bit later.
    
Since teardown is not an issue anymore, thanks to the usage of runson service, we can rely on the regular job result instead of using workflow outputs as a medium.

### DIFF
--- a/.github/workflows/cargo_build.yml
+++ b/.github/workflows/cargo_build.yml
@@ -141,22 +141,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check all builds success
-        if: needs.parallel-pcc-cpu.outputs.builds-result == 'success' &&
-          needs.pcc-hpu.outputs.builds-result == 'success' &&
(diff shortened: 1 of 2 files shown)


soonum:
The fix is working, this has been tested manually:
* [previous](https://github.com/zama-ai/tfhe-rs/actions/runs/26168030422/job/76981637651?pr=3573) bad behavior
* [now](https://github.com/zama-ai/tfhe-rs/actions/runs/26169409892/job/76986443542?pr=3573) expected behavior

SouchonTheo (approved):

IceTDrinker (approved):
The best fixes are the "simplest" ones, like this PR

PR #3483, opened 2026-04-17
chore(gpu): noise checks on input

andrei-stoian-zama:
- Enables check noise levels on PBS inputs
- Removes bindings for low-level linear algebra which were not used. Most low-level functions are now internal use only:
    - cuda_add_lwe_ciphertext_vector_32
    - cuda_add_lwe_ciphertext_vector_64
    - cuda_add_lwe_ciphertext_vector_inplace_32
    - cuda_add_lwe_ciphertext_vector_plaintext_64
    - cuda_add_lwe_ciphertext_vector_plaintext_vector_32
    - cuda_add_lwe_ciphertext_vector_plaintext_vector_64
    - cuda_mult_lwe_ciphertext_vector_cleartext_vector_32
    - cuda_mult_lwe_ciphertext_vector_cleartext_vector_64
    - cuda_negate_lwe_ciphertext_vector_32
    - cuda_negate_lwe_ciphertext_vector_64
    - host_subtraction_plaintext

- Moves functions from linear algebra that already take CudaRadixCiphertextFFI to integer.cuh 
-  1.  host_negation                                                                                                        

- Converts some `host_` functions to take `CudaRadixCiphertextFFI` : 
  1. host_addition_plaintext                                                                                                              
  2. host_addition_plaintext_scalar    
  4. host_cleartext_vec_multiplication : and adds degree tracking
  
- Adds noise&degree tracking to some functions that didn't have it: functions that were used in `div_rem` and `comparisons`
  1. host_subtraction : and moves to integer.cuh
  2. host_cleartext_multiplication : : and moves to integer.cuh
  4. accumulate_all_blocks : : and moves to integer.cuh

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3483)

PR #3574, opened 2026-05-21
feat(gpu): noise check tests on L40

enzodimaria:
Noise check is done with the `gpu-debug-noise-check` feature, which enables the `CHECK_NOISE_LEVEL` assertions in the CUDA backend.

New workflow on a Hyperstack L40 instance that runs the signed and unsigned integer GPU tests with this feature. It is triggered automatically on every PR that touches GPU code (file-change filter, no label gate), alongside `gpu_fast_tests`. External (forked) PRs fall back to the GitHub-hosted `gpu_ubuntu-22.04` runner.

### DIFF
--- a/scripts/integer-tests.sh
+++ b/scripts/integer-tests.sh
@@ -123,7 +123,7 @@ if [[ "${NO_BIG_PARAMS_GPU}" == TRUE ]]; then
 fi
 
 if [[ "${backend}" == "gpu" ]]; then
-    gpu_feature="gpu"
+    gpu_feature="${GPU_FEATURE:-gpu}"
 fi
 
 CURR_DIR="$(dirname "$0")"
--- a/.github/workflows/gpu_l40_noise_check_tests.yml
+++ b/.github/workflows/gpu_l40_noise_check_tests.yml
@@ -1,0 +1,200 @@
+# Run signed and unsigned GPU integer tests with the gpu-debug-noise-check
+# feature so that the CHECK_NOISE_LEVEL assertions in the CUDA backend are
+# active, without the multi-GPU emulation that requires a multi-GPU host.
+# Runs on a single L40 Hyperstack instance, triggered automatically on every PR
+# that touches GPU code, alongside gpu_fast_tests.
+name: gpu_l40_noise_check_tests
+
+env:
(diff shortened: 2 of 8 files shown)


scripts/integer-tests.sh
```diff
@@ -123,7 +123,7 @@ if [[ "${NO_BIG_PARAMS_GPU}" == TRUE ]]; then
 fi
 
 if [[ "${backend}" == "gpu" ]]; then
-    gpu_feature="gpu"
+    gpu_feature="${GPU_FEATURE:-gpu}"
```
andrei-stoian-zama, line 126: can you explain `-gpu` here in a comment? if backend is "gpu" and if GPU_FEATURE is not set we use "gpu" right ?

.github/workflows/gpu_4090_fake_multi_gpu_tests.yml
```diff
@@ -0,0 +1,71 @@
+# Run signed and unsigned GPU integer tests with the gpu-debug-fake-multi-gpu
+# feature so that the CHECK_NOISE_LEVEL assertions in the CUDA backend are
+# active.
+name: gpu_4090_fake_multi_gpu_tests
```
andrei-stoian-zama: can you remind us why we want to run this on 4090 ? 

we have some other tests on the local 4090 , having two workflow in parallel could be an issue.. why not run this on a cloud L40 ?

andrei-stoian-zama (changes requested):
Thanks! this should help in the future with new HL ops

scripts/integer-tests.sh
enzodimaria: Thanks Andrei for your review !

GPU_FEATURE is a tag for the additionnal feature. Here it's for gpu-debug-fake-multi-gpu
I added it because before we were only using "gpu"

.github/workflows/gpu_4090_fake_multi_gpu_tests.yml
enzodimaria: I thought we wanted to start it on this machine, but I can use a L40 if you prefer

andrei-stoian-zama:
Superseded by #3483

PR #3580, opened 2026-05-22
Am/feat/prf rerand

IceTDrinker:
Preparatory PR to get a PRF + rerand primitive

First a few commits to fix/update a few small things I noticed while writing the PR.

The theme is :
- remove all the PRF APIs which are just wrappers of others at the shortint level to limit how many functions need to be written for the PRF + rerand
- only shortint CPU has a PRF + rerand API for now
- next PR will focus on integer + GPU (and likely HL) to plug everything
- the rerand seeding has been described as follows :

```
So lets get a bit more formal so we are on the same page...
You have a hash function H which outputs a matrix of size v x n, where v is the number of output bits and n is the LWE dimension of the PRF.
Let x be the input to the PRF
Let 2^{k'} be the output size of the PRF, so 0 <= k' <= 5

You should probably do

H(x || k' || v)
and not just

H(x)
```
k' is here to indicate how random bits are set in the output, it can be more complicated than "a set number of bits for all blocks" (though in practice that's the most common case we will be using for Radix)

so it was agreed to use a Run Length Encoding to describe how random bits are distributed in the output

https://en.wikipedia.org/wiki/Run-length_encoding#:~:text=Run%2Dlength%20encoding%20(RLE),than%20as%20the%20original%20run.

essentially it encodes a chunk in the following way:

chunk_block_count || repeat(block_count, bits_in_these_blocks)

so for example if we have the following random bit chunks requests [15, 14, 1] for a block that contains two bits message that we fill the encoding would be

8 || 7 || 2 || 1 || 1 ---- 7 || 7 || 2 ---- 1 || 1 || 1

8 blocks total, 7 blocks with 2 bits, 1 block with 1 bit, then 7 blocks total, 7 blocks with two bits and then 1 block total, 1 block with 1 bit

the message bit count is also hashed, since I made an LLM assisted review (only reviewing with LLM) and it pointed out that there could be a collision between the representation of the RLE if we only have partial blocks e.g.

[1, 1] -> 1 || 1 || 1 ---- 1 || 1 || 1

the representation is the same no matter the message bits

which makes me think I should also probably hash how many bits are in a block which are part of the cleartext space, so in our regular encoding 5 (1 padding bit + 2 bits of carry and 2 bits of message), so maybe hashing 1 (padding), 2 (carry) and 2 (message)

message today being hashed already

feel free to ask questions

I added a test in shortint which should ensure all properties we care about for the PRF + rerand are tested

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)

### DIFF
--- a/tfhe-benchmark/benches/shortint/oprf.rs
+++ b/tfhe-benchmark/benches/shortint/oprf.rs
@@ -21,7 +21,7 @@ fn oprf(c: &mut Criterion) {
 
     bench_group.bench_function(format!("2-bits-oprf::{}", param.name()), |b| {
         b.iter(|| {
-            _ = black_box(oprf_sk.generate_oblivious_pseudo_random(Seed(0), 2, sks));
+            _ = black_box(oprf_sk.generate_oblivious_pseudo_random_bits_chunks(Seed(0), &[2], sks));
         })
     });
 }
--- a/tfhe/src/high_level_api/booleans/oprf.rs
+++ b/tfhe/src/high_level_api/booleans/oprf.rs
@@ -33,10 +33,22 @@ impl FheBool {
             InternalServerKey::Cpu(key) => {
                 let sk = &key.pbs_key().key;
 
-                let ct = key
+                let ct_wrapped = key
                     .oprf_key()
                     .key
-                    .generate_oblivious_pseudo_random(seed, 1, sk);
+                    .generate_oblivious_pseudo_random_bits_chunks(seed, &[1], sk);
+
+                // We have to do the double unwrap, we want to keep as little primitives as possible
+                // for PRF since they also need a rerandomized_variant, so we don't have a single
+                // block primitive for that
+                let ct = ct_wrapped
+                    .into_iter()
+                    .next()
+                    .expect("A single chunk was expected, got 0")
+                    .into_iter()
+                    .next()
+                    .expect("A single ciphertext was expected, got 0");
+
                 (
                     InnerBoolean::Cpu(BooleanBlock::new_unchecked(ct)),
                     key.tag.clone(),
--- a/tfhe/src/high_level_api/booleans/oprf.rs
+++ b/tfhe/src/high_level_api/booleans/oprf.rs
@@ -57,9 +57,16 @@ impl FheBool {
             #[cfg(feature = "gpu")]
             InternalServerKey::Cuda(cuda_key) => {
                 let streams = &cuda_key.streams;
+                // 1 block with 1 bit of data is a boolean
                 let d_ct: CudaUnsignedRadixCiphertext = cuda_key
                     .oprf_key()
-                    .generate_oblivious_pseudo_random(seed, 1, cuda_key.pbs_key(), streams);
+                    .par_generate_oblivious_pseudo_random_unsigned_integer_bounded(
+                        seed,
+                        1,
+                        1,
+                        cuda_key.pbs_key(),
+                        streams,
+                    );
                 (
                     InnerBoolean::Cuda(CudaBooleanBlock::from_cuda_radix_ciphertext(
                         d_ct.ciphertext,
--- a/tfhe/src/high_level_api/integers/mod.rs
+++ b/tfhe/src/high_level_api/integers/mod.rs
@@ -69,7 +69,7 @@ pub trait IntegerId: FheId + 'static {
     fn num_bits() -> usize;
 
     fn num_blocks(message_modulus: MessageModulus) -> usize {
-        Self::num_bits() / message_modulus.0.ilog2() as usize
+        Self::num_bits().div_ceil(message_modulus.0.ilog2() as usize)
     }
 }
 
--- a/tfhe/src/high_level_api/integers/oprf.rs
+++ b/tfhe/src/high_level_api/integers/oprf.rs
@@ -870,7 +870,7 @@ mod test {
     /// Test OPRF with BootstrapKeyswitch (PBS_KS) parameter order.
     ///
     /// This exercises the keyswitch-after-bootstrap code path in
-    /// `OprfBootstrappingKey::generate_pseudo_random_bits`.
+    /// [`crate::shortint::oprf::OprfBootstrappingKey::generate_pseudo_random_bits_chunks`].
     #[test]
     fn test_oprf_with_pbs_ks_params() {
         let config = ConfigBuilder::with_custom_parameters(
--- a/tfhe/src/high_level_api/integers/oprf.rs
+++ b/tfhe/src/high_level_api/integers/oprf.rs
@@ -591,9 +591,12 @@ mod test {
     };
     use crate::prelude::FheDecrypt;
     use crate::shortint::oprf::test::test_uniformity;
-    use crate::shortint::parameters::test_params::TEST_PARAM_MESSAGE_2_CARRY_2_PBS_KS_GAUSSIAN_2M128;
+    use crate::shortint::parameters::test_params::{
+        TEST_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+        TEST_PARAM_MESSAGE_2_CARRY_2_PBS_KS_GAUSSIAN_2M128,
+    };
     use crate::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS32_PBS_TUNIFORM_2M128;
-    use crate::{generate_keys, set_server_key, ConfigBuilder, FheUint8, Seed};
+    use crate::{generate_keys, set_server_key, ConfigBuilder, FheInt8, FheUint8, Seed};
     use num_bigint::BigUint;
     use rand::{thread_rng, Rng};
     use rayon::iter::{IntoParallelIterator, ParallelIterator};
@@ -890,6 +893,32 @@ mod test {
         assert!(result_bounded < (1 << 3));
     }
 
+    #[test]
+    fn test_oprf_bounded_zero() {
+        let config = ConfigBuilder::with_custom_parameters(
+            TEST_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+        )
+        .use_dedicated_oprf_key(true)
+        .build();
+
+        let (client_key, server_key) = generate_keys(config);
+        set_server_key(server_key);
+
+        // Do not use static seed in production
+        let ct_unsigned_bounded = FheUint8::generate_oblivious_pseudo_random_bounded(Seed(0), 0);
+        assert!(ct_unsigned_bounded.is_trivial());
+        let result_unsigned_bounded: u8 = ct_unsigned_bounded.decrypt(&client_key);
+        // PRF with 0 bits is equivalent to modulo 1 meaning only 0 is generated
+        assert_eq!(result_unsigned_bounded, 0);
+
+        // Do not use static seed in production
+        let ct_signed_bounded = FheInt8::generate_oblivious_pseudo_random_bounded(Seed(0), 0);
+        assert!(ct_signed_bounded.is_trivial());
+        let result_signed_bounded: i8 = ct_signed_bounded.decrypt(&client_key);
+        // PRF with 0 bits is equivalent to modulo 1 meaning only 0 is generated
+        assert_eq!(result_signed_bounded, 0);
+    }
+
     #[cfg(feature = "gpu")]
     mod gpu {
         use super::*;
@@ -903,6 +932,29 @@ mod test {
         use rayon::iter::IndexedParallelIterator;
         use rayon::prelude::{IntoParallelRefIterator, ParallelSlice};
         use rayon::ThreadPoolBuilder;
+
+        #[test]
+        fn test_oprf_bounded_zero() {
+            for setup_fn in crate::high_level_api::integers::unsigned::tests::gpu::GPU_SETUP_FN {
+                let client_key = setup_fn();
+
+                // Do not use static seed in production
+                let ct_unsigned_bounded =
+                    FheUint8::generate_oblivious_pseudo_random_bounded(Seed(0), 0);
+                assert!(ct_unsigned_bounded.is_trivial());
+                let result_unsigned_bounded: u8 = ct_unsigned_bounded.decrypt(&client_key);
+                // PRF with 0 bits is equivalent to modulo 1 meaning only 0 is generated
+                assert_eq!(result_unsigned_bounded, 0);
+
+                // Do not use static seed in production
+                let ct_signed_bounded =
+                    FheInt8::generate_oblivious_pseudo_random_bounded(Seed(0), 0);
+                assert!(ct_signed_bounded.is_trivial());
+                let result_signed_bounded: i8 = ct_signed_bounded.decrypt(&client_key);
+                // PRF with 0 bits is equivalent to modulo 1 meaning only 0 is generated
+                assert_eq!(result_signed_bounded, 0);
+            }
+        }
 
         #[test]
         fn test_oprf_gpu() {
--- a/tfhe/src/high_level_api/integers/signed/base.rs
+++ b/tfhe/src/high_level_api/integers/signed/base.rs
@@ -912,6 +912,27 @@ where
         self.ciphertext.on_cpu().decrypt_trivial()
     }
 
+    /// Returns true if the ciphertext is a trivial encryption
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use tfhe::prelude::*;
+    /// use tfhe::{generate_keys, set_server_key, ConfigBuilder, FheInt16};
+    ///
+    /// let (client_key, server_key) = generate_keys(ConfigBuilder::default());
+    /// set_server_key(server_key);
+    ///
+    /// let non_trivial = FheInt16::encrypt(1i16, &client_key);
+    /// assert!(!non_trivial.is_trivial());
+    ///
+    /// let trivial = FheInt16::encrypt_trivial(2i16);
+    /// assert!(trivial.is_trivial());
+    /// ```
+    pub fn is_trivial(&self) -> bool {
+        self.ciphertext.on_cpu().is_trivial()
+    }
+
     /// Reverse the bit of the signed integer
     ///
     /// # Example
--- a/tfhe/src/integer/gpu/server_key/radix/oprf.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/oprf.rs
@@ -177,12 +177,14 @@ where
         streams: &CudaStreams,
     ) -> CudaUnsignedRadixCiphertext {
         let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
-        let range_log_size = message_bits_count * num_blocks;
+        let range_bits_count = message_bits_count * num_blocks;
+        assert!(range_bits_count > 0);
 
         assert!(
-                random_bits_count <= range_log_size,
-                "The range asked for a random value (=[0, 2^{random_bits_count}[) does not fit in the available range [0, 2^{range_log_size}[",
-            );
+            random_bits_count <= range_bits_count,
+            "The range asked for a random value (=[0, 2^{random_bits_count}[) \
+            does not fit in the available range [0, 2^{range_bits_count}[",
+        );
 
         self.generate_oblivious_pseudo_random_bounded_integer(
             seed,
@@ -290,14 +292,16 @@ where
         streams: &CudaStreams,
     ) -> CudaSignedRadixCiphertext {
         let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
-        let range_log_size = message_bits_count * num_blocks;
+        let range_bits_count = message_bits_count * num_blocks;
+        assert!(range_bits_count > 0);
 
-        #[allow(clippy::int_plus_one)]
         {
+            let signed_range_bits_count = range_bits_count.saturating_sub(1);
             assert!(
-                random_bits_count + 1 <= range_log_size,
-                "The range asked for a random value (=[0, 2^{}[) does not fit in the available range [-2^{}, 2^{}[",
-                random_bits_count, range_log_size - 1, range_log_size - 1,
+                random_bits_count <= signed_range_bits_count,
+                "The range asked for a random value (=[0, 2^{random_bits_count}[) \
+                which does not fit in the available range \
+                [-2^{signed_range_bits_count}, 2^{signed_range_bits_count}[",
             );
         }
 
@@ -518,13 +522,15 @@ where
 
         assert!(
             !excluded_upper_bound.is_power_of_two(),
-            "Use the cheaper par_generate_oblivious_pseudo_random_unsigned_integer_bounded function instead"
+            "Use the cheaper par_generate_oblivious_pseudo_random_unsigned_integer_bounded \
+            function instead"
         );
 
         let num_bits_output = num_blocks_output * message_bits_count;
         assert!(
             (excluded_upper_bound as f64) < 2_f64.powi(num_bits_output as i32),
-            "num_blocks_output(={num_blocks_output}) is too small to hold an integer up to excluded_upper_bound(={excluded_upper_bound})"
+            "num_blocks_output(={num_blocks_output}) is too small to hold an integer \
+            up to excluded_upper_bound(={excluded_upper_bound})"
         );
 
         let CudaDynamicKeyswitchingKey::Standard(computing_ks_key) = &target_sks.key_switching_key
--- a/tfhe/src/integer/gpu/server_key/radix/oprf.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/oprf.rs
@@ -176,6 +176,7 @@ where
         target_sks: &CudaServerKey,
         streams: &CudaStreams,
     ) -> CudaUnsignedRadixCiphertext {
+        assert!(target_sks.message_modulus.0.is_power_of_two());
         let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
         let range_bits_count = message_bits_count * num_blocks;
         assert!(range_bits_count > 0);
@@ -291,6 +292,7 @@ where
         target_sks: &CudaServerKey,
         streams: &CudaStreams,
     ) -> CudaSignedRadixCiphertext {
+        assert!(target_sks.message_modulus.0.is_power_of_two());
         let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
         let range_bits_count = message_bits_count * num_blocks;
         assert!(range_bits_count > 0);
@@ -388,9 +390,10 @@ where
         result
     }
 
-    // Core private implementation that calls the OPRF backend.
-    // This function contains the main logic for both bounded and unbounded generation.
-    //
+    /// Core private implementation that calls the OPRF backend.
+    /// This function contains the main logic for both bounded and unbounded generation.
+    ///
+    /// Caller must ensure total_random_bits is non 0 otherwise this function will panic.
     fn generate_multiblocks_oblivious_pseudo_random(
         &self,
         result: &mut CudaRadixCiphertext,
@@ -412,14 +415,18 @@ where
         let polynomial_size = bootstrapping_key.polynomial_size();
         let in_lwe_size = input_lwe_dimension.to_lwe_size();
         let message_bits_count = target_sks.message_modulus.0.ilog2();
+        let carry_bits_count = target_sks.carry_modulus.0.ilog2();
+        let bits_per_block = message_bits_count + carry_bits_count + 1;
 
-        let seeded = create_random_from_seed_modulus_switched(
+        let (seeded, _rle_info) = create_random_from_seed_modulus_switched(
             seed,
             in_lwe_size,
             polynomial_size,
             &[total_random_bits],
-            message_bits_count as u64,
+            message_bits_count.into(),
+            bits_per_block.into(),
         );
+
         let h_seeded_lwe_list: Vec<u64> = seeded
             .into_iter()
             .flat_map(|(seeded, _bits)| {
@@ -471,6 +478,13 @@ where
         }
     }
 
+    /// # Panics
+    ///
+    /// Panics if:
+    /// - `target_sks.message_modulus` is not a power of 2
+    /// - `excluded_upper_bound` is a power of 2 use
+    ///   [`Self::par_generate_oblivious_pseudo_random_unsigned_integer_bounded`] instead
+    /// - `excluded_upper_bound.ilog2() + 1` is greater than the output bit count
     pub fn par_generate_oblivious_pseudo_random_unsigned_custom_range(
         &self,
         seed: impl OprfSeed,
@@ -484,7 +498,13 @@ where
             target_sks.message_modulus.0.is_power_of_two(),
             "Message modulus must be a power of two"
         );
-        let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
+        assert!(
+            target_sks.carry_modulus.0.is_power_of_two(),
+            "Carry modulus must be a power of two"
+        );
+        let message_bits_count: u64 = target_sks.message_modulus.0.ilog2().into();
+        let carry_bits_count: u64 = target_sks.carry_modulus.0.ilog2().into();
+        let bits_per_block = message_bits_count + carry_bits_count + 1;
 
         assert!(
             !excluded_upper_bound.is_power_of_two(),
@@ -528,12 +548,13 @@ where
             .iter_as::<u64>()
             .collect::<Vec<_>>();
 
-        let seeded = create_random_from_seed_modulus_switched(
+        let (seeded, _rle_info) = create_random_from_seed_modulus_switched(
             seed,
             in_lwe_size,
             polynomial_size,
             &[num_input_random_bits],
             message_bits_count,
+            bits_per_block,
         );
 
         let h_seeded_lwe_list: Vec<u64> = seeded
--- a/tfhe/src/integer/gpu/server_key/radix/oprf.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/oprf.rs
@@ -314,40 +314,6 @@ where
         )
     }
 
-    // Generic interface to generate a single-block oblivious pseudo-random integer.
-    // It performs checks specific to single-block capacity.
-    //
-    pub fn generate_oblivious_pseudo_random<T>(
-        &self,
-        seed: impl OprfSeed,
-        random_bits_count: u64,
-        target_sks: &CudaServerKey,
-        streams: &CudaStreams,
-    ) -> T
-    where
-        T: CudaIntegerRadixCiphertext,
-    {
-        assert!(
-            1 << random_bits_count <= target_sks.message_modulus.0,
-            "The range asked for a random value (=[0, 2^{random_bits_count}[) does not fit in the available range [0, {}[",
-            target_sks.message_modulus.0
-        );
-        let carry_bits_count = target_sks.carry_modulus.0.ilog2() as u64;
-        let message_bits_count = target_sks.message_modulus.0.ilog2() as u64;
-        assert!(
-            random_bits_count <= carry_bits_count + message_bits_count,
-            "The number of random bits asked for (={random_bits_count}) is bigger than carry_bits_count (={carry_bits_count}) + message_bits_count(={message_bits_count})",
-        );
-
-        self.generate_oblivious_pseudo_random_bounded_integer(
-            seed,
-            random_bits_count,
-            1,
-            target_sks,
-            streams,
-        )
-    }
-
     // Generic internal implementation for unbounded pseudo-random generation.
     // It calls the core implementation with parameters for the unbounded case.
     //
--- a/tfhe/src/integer/oprf.rs
+++ b/tfhe/src/integer/oprf.rs
@@ -375,11 +375,16 @@ where
         assert!(target_sks.message_modulus().0.is_power_of_two());
         let message_bits_count = target_sks.message_modulus().0.ilog2() as u64;
 
-        let blocks = self.key.generate_oblivious_pseudo_random_bits(
-            seed,
-            num_blocks * message_bits_count,
-            &target_sks.key,
-        );
+        let blocks = self
+            .key
+            .generate_oblivious_pseudo_random_bits_chunks(
+                seed,
+                &[num_blocks * message_bits_count],
+                &target_sks.key,
+            )
+            .into_iter()
+            .next()
+            .expect("Expected a single chunk for the radix being generated, got 0");
 
         T::from(blocks)
     }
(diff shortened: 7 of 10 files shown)


nsarlin-zama:
I haven't look at the code yet, but I'm not sure I understand this:
> the message bit count is also hashed, since I made an LLM assisted review (only reviewing with LLM) and it pointed out that there could be a collision between the representation of the RLE if we only have partial blocks e.g.
> 
> [1, 1] -> 1 || 1 || 1 ---- 1 || 1 || 1

What is the other payload this could be colliding with?

IceTDrinker:
> I haven't look at the code yet, but I'm not sure I understand this:
> 
> > the message bit count is also hashed, since I made an LLM assisted review (only reviewing with LLM) and it pointed out that there could be a collision between the representation of the RLE if we only have partial blocks e.g.
> > [1, 1] -> 1 || 1 || 1 ---- 1 || 1 || 1
> 
> What is the other payload this could be colliding with?

if you have message_bits = 2 or 4, the [1, 1] would be the same if we don't hash message_bits

nsarlin-zama (commented):
some comment regarding the panics/error and the encoding

> @nsarlin-zama reviewed 9 files and made 14 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-:-Ou6kD0A-jPZk34lQ1aO:bfmw1z6)* status: all files reviewed (commit messages unreviewed), 13 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/booleans/oprf.rs` line 47 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_ikD6t8HzrAiVmaJ:-Ou6_ikD6t8HzrAiVmaK:bms6tol) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/high_level_api/booleans/oprf.rs#L47)):*
> ```XML
>                     .into_iter()
>                     .next()
>                     .unwrap()
> ```

Since this is statically infallible, here I would use "expect" with a string that says why it cannot fail such as "exactly one chunk was requested" and "chunk size is exactly one". That way it is easier when reading/refactoring the code to validate that the condition still holds
___
*[`tfhe/src/integer/oprf.rs` line 387 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6b7JV-BOusoMIbQdy:-Ou6b7JV-BOusoMIbQdz:b-o06vap) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/oprf.rs#L387)):*
> ```XML
>             .into_iter()
>             .next()
>             .unwrap();
> ```

would use expect to document why this cannot fail
___
*[`tfhe/src/integer/oprf.rs` line 430 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6buAQ9nz--lS_Doon:-Ou6buAQ9nz--lS_Dooo:bd7zvto) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/oprf.rs#L430)):*
> ```XML
>             .into_iter()
>             .next()
>             .unwrap();
> ```

same, would use expect
___
*[`tfhe/src/integer/gpu/server_key/radix/oprf.rs` line 483 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6cBei2_3XaioQJNlo:-Ou6cBei2_3XaioQJNlp:b-4bn5cm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/gpu/server_key/radix/oprf.rs#L483)):*
> ```XML
> 
>     /// Caller must ensure num_input_random_bits is non 0 otherwise this function will panic.
>     pub fn par_generate_oblivious_pseudo_random_unsigned_custom_range(
> ```

Usually the convention is to use a `# Panics` block in the doc so that it is immediately visible that the function can panic
___
*[`tfhe/src/shortint/oprf.rs` line 627 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6caeO64LkpoT5NISV:-Ou6caeO64LkpoT5NISW:b-xkungc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L627)):*
> ```XML
>     /// * Panics if `bit_chunks` contains a 0
>     /// * Panics if `self` is not compatible with `target_sks`
>     pub fn generate_oblivious_pseudo_random_bits_chunks(
> ```

Maybe this should return an error too to keep the api consistent with the re-randomize version
___
*[`tfhe/src/shortint/oprf.rs` line 662 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_5p-APdWetNhuXiE:-Ou6_5p-APdWetNhuXiF:b1m08it) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L662)):*
> ```XML
>     ///
>     /// * Panics if `self` is not compatible with `target_sks`.
>     /// * Panics if `bit_chunks` contains a `0`.
> ```

not clear why some failure cause panics and other return an error. I would return an error in all cases now that the function is fallible, and let the caller panic if it wants too.

Maybe this is too late to have this at a higher level, but a good practice I think would be to have:
- invalid crypto params: panic (since we know the protocol will never run with invalid params)
- invalid arguments to a function: error (since fhevm might create wrapper to these functions that the end user will call, so a panic might be triggerable by an attacker if fhevm forget to sanitize the inputs)
___
*[`tfhe/src/shortint/oprf.rs` line 873 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6ezJV9omsRfopc-Al:-Ou6ezJV9omsRfopc-Am:b-w0y7k8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L873)):*
> ```XML
> }
> 
> pub(crate) struct RandomBitsRleLeBytes(Vec<u8>);
> ```

naming suggestion: RandomBitsEncoding ?
___
*[`tfhe/src/shortint/oprf.rs` line 925 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6d56d21b5XLzUWkEt:-Ou6d56d21b5XLzUWkEu:b4o6ijs) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L925)):*
> ```XML
>     let seed_bytes = seed.into_bytes();
>     let mut hasher = sha3::Sha3_256::default();
>     hasher.update(b"TFHE_PRF");
> ```

maybe use a constant here?
___
*[`tfhe/src/shortint/oprf.rs` line 943 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gDc4DeaL8fbNmAGl:-Ou6gDc4DeaL8fbNmAGm:b-jix5l7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L943)):*
> ```XML
>     // 1 for full block count
>     // 1 per bits per full block
>     // 1 to encode the block count "1" for the last block
> ```

isn't this redundant with (total_block_count - full_block_count) ? And if it's always 1 it means that it holds no additional information.

I think without collision we could even have:
- 1 u64 for total block count
- 1 u64 for bits per full block
- 1 u64 for bits in the last block. Can be 0 if there is no last block since we know that we can't have a block of 0 bits.

But maybe it is too late to change this if it's breaking
___
*[`tfhe/src/shortint/oprf.rs` line 994 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gYNs4yJCsGHNh_2S:-Ou6gYNs4yJCsGHNh_2T:b-qy74vs) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L994)):*
> ```XML
>     let mut xof = RandomGenerator::<DefaultRandomGenerator>::new(XofSeed::new(
>         xof_seed.to_vec(),
>         *b"PRF_INIT",
> ```

constant ?
___
*[`tfhe/src/shortint/oprf.rs` line 942 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6finF2ejkibf1qPR5:-Ou6finF2ejkibf1qPR6:bwltgwc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L942)):*
> ```XML
>     // 1 for the total block count
>     // 1 for full block count
>     // 1 per bits per full block
> ```

"1 for the number of bits per full block" ?
___
*[`tfhe/src/shortint/ciphertext/re_randomization.rs` line 252 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6jmDa57PGJ9ip3t_z:-Ou6jmDa57PGJ9ip3ta-:b-mb2sf5) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/ciphertext/re_randomization.rs#L252)):*
> ```XML
>         random_bits_rle_bytes: &RandomBitsRleLeBytes,
>     ) -> Self {
>         const PRF_RERAND_DOMAIN_SEPARATOR: [u8; XofSeed::DOMAIN_SEP_LEN] = *b"PRF_RRND";
> ```

not for this pr, but it might be useful to have all these domain separator defined in a single place since it's an info that can be asked later (for example when writing specs)
___
*[`tfhe/src/shortint/oprf.rs` line 898 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6efp28L9sQUJabWUp:-Ou6efp28L9sQUJabWUq:b-1ylvrt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L898)):*
> ```XML
> ///   the various chunks in the same order as `bit_chunks`, with `num_bits` the random bit count
> ///   actually carried by a given block.
> /// - `rle` is the little-endian bytes of `random_bits_per_block` plus the per-chunk  RLE layout, it
> ```

maybe it could be worth explaining the per-chunk layout here? Or on the doc at the type level

IceTDrinker (commented):
> @IceTDrinker made 13 comments and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed (commit messages unreviewed), 7 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/booleans/oprf.rs` line 47 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_ikD6t8HzrAiVmaJ:-Ou87qHk80vFZTiPuanw:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/high_level_api/booleans/oprf.rs#L47)):*

Since this is statically infallible, here I would use "expect" with a string that says why it cannot fail such as "exactly one chunk was requested" and "chunk size is exactly one". That way it is easier when reading/refactoring the code to validate that the condition still holds
</blockquote></details>

done
___
*[`tfhe/src/integer/oprf.rs` line 387 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6b7JV-BOusoMIbQdy:-Ou88Clq0N2A-oVgwpn2:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/oprf.rs#L387)):*

would use expect to document why this cannot fail
</blockquote></details>

done
___
*[`tfhe/src/integer/oprf.rs` line 430 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6buAQ9nz--lS_Doon:-Ou88DQ_6DGMNzdGftGm:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/oprf.rs#L430)):*

same, would use expect
</blockquote></details>

done
___
*[`tfhe/src/integer/gpu/server_key/radix/oprf.rs` line 483 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6cBei2_3XaioQJNlo:-Ou88GG33_-alypBs5B3:bve6l1o) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/integer/gpu/server_key/radix/oprf.rs#L483)):*

Usually the convention is to use a `# Panics` block in the doc so that it is immediately visible that the function can panic
</blockquote></details>

fair, not sure if I want to document all the failure cases though :melting_face: 

ended up doing it, so done
___
*[`tfhe/src/shortint/oprf.rs` line 627 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6caeO64LkpoT5NISV:-Ou8Ae1V5Ovg4SYMBU1H:bopsgqn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L627)):*

Maybe this should return an error too to keep the api consistent with the re-randomize version
</blockquote></details>

I don't see the point of changing the internal APIs when the HL won't change it's signature, I thought about it and it did not seem to make much sense

I think the mistake was returning errors in the rerand when those are actually correctness asserts :/ 
___
*[`tfhe/src/shortint/oprf.rs` line 662 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_5p-APdWetNhuXiE:-Ou8AvtO1SGNxIlu0FCO:bchlnf7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L662)):*

not clear why some failure cause panics and other return an error. I would return an error in all cases now that the function is fallible, and let the caller panic if it wants too.

Maybe this is too late to have this at a higher level, but a good practice I think would be to have:
- invalid crypto params: panic (since we know the protocol will never run with invalid params)
- invalid arguments to a function: error (since fhevm might create wrapper to these functions that the end user will call, so a panic might be triggerable by an attacker if fhevm forget to sanitize the inputs)
</blockquote></details>

bit chunk 0 is a hard error
incompatible target_sks can be a mishandling but still not recoverable


it's the eternal "panic when something is extremely wrong" vs "return an error" debate, KMS is going the way of panics for some stuff IIRC after doing errors for a while
___
*[`tfhe/src/shortint/oprf.rs` line 873 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6ezJV9omsRfopc-Al:-Ou8BBt60AyJ6tgYXykT:bevcwa6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L873)):*

naming suggestion: RandomBitsEncoding ?
</blockquote></details>

the current name is more descriptive of what's inside so this or your name + docstring I think I prefer the current ?
___
*[`tfhe/src/shortint/oprf.rs` line 898 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6efp28L9sQUJabWUp:-Ou8BIfA9dgS3e7S5vvw:b-3pw52p) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L898)):*

maybe it could be worth explaining the per-chunk layout here? Or on the doc at the type level
</blockquote></details>

yeah like this it lacks details, added an example
___
*[`tfhe/src/shortint/oprf.rs` line 925 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6d56d21b5XLzUWkEt:-Ou8BO4A4X4XlKIIWSZ_:bb7oryy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L925)):*

maybe use a constant here?
</blockquote></details>

sure
___
*[`tfhe/src/shortint/oprf.rs` line 942 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6finF2ejkibf1qPR5:-Ou8DTzvBn0jHFIKg6ae:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L942)):*

"1 for the number of bits per full block" ?
</blockquote></details>

done
___
*[`tfhe/src/shortint/oprf.rs` line 943 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gDc4DeaL8fbNmAGl:-Ou8CDedCE1WsNwhJe2M:b4qadkt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L943)):*

isn't this redundant with (total_block_count - full_block_count) ? And if it's always 1 it means that it holds no additional information.

I think without collision we could even have:
- 1 u64 for total block count
- 1 u64 for bits per full block
- 1 u64 for bits in the last block. Can be 0 if there is no last block since we know that we can't have a block of 0 bits.

But maybe it is too late to change this if it's breaking
</blockquote></details>

it can still be changed, it's just that RLE has a certain definition, it seemed reasonable to have something somewhat mainstream

https://en.wikipedia.org/wiki/Run-length_encoding#:~:text=For%20example%2C%20a%20sequence%20of,drawings%2C%20games%2C%20and%20animations.
___
*[`tfhe/src/shortint/oprf.rs` line 994 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gYNs4yJCsGHNh_2S:-Ou8CSN73An1G55OEPy4:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L994)):*

constant ?
</blockquote></details>

done
___
*[`tfhe/src/shortint/ciphertext/re_randomization.rs` line 252 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6jmDa57PGJ9ip3t_z:-Ou8CT9625x8QPFfRu_C:b-gsu8f4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/ciphertext/re_randomization.rs#L252)):*

not for this pr, but it might be useful to have all these domain separator defined in a single place since it's an info that can be asked later (for example when writing specs)
</blockquote></details>

could you open an iisue ?

IceTDrinker:
ah damn forgot the bits per block hashing 🙄

nsarlin-zama (commented):
> @nsarlin-zama made 3 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: 5 of 9 files reviewed, 5 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 662 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_5p-APdWetNhuXiE:-OuBWm9k3DatOXx1gW2z:bikzuo3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L662)):*

bit chunk 0 is a hard error
incompatible target_sks can be a mishandling but still not recoverable


it's the eternal "panic when something is extremely wrong" vs "return an error" debate, KMS is going the way of panics for some stuff IIRC after doing errors for a while
</blockquote></details>

Why do you consider this a hard error here? I see it as a normal error. If bit chunk is 0, this specific call is invalid, but it does not prevent to follow the normal course of operations after it returns.
Hard error for me means that the app (ie coproc) state is broken beyond repair. Here this is just a specific call that is wrong, so it is just a classical error IMO.

To give an example:
- user A (or dapp or whatever) requires prf of 0 bits
- user B requires a prf of N > 0 bits

A does not impact B in any ways, so you should just return a clean error to A and carry on. 
Wrong params is something different: the way the lib is used, if params are incorrect for A they are also incorrect for B, so this is a "hard" error and I can understand the panic.
___
*[`tfhe/src/shortint/oprf.rs` line 873 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6ezJV9omsRfopc-Al:-OuBn0e-7VRd-mLBDlDk:b9dcshl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L873)):*

the current name is more descriptive of what's inside so this or your name + docstring I think I prefer the current ?
</blockquote></details>

Since it is pub(crate) it does not matter, but it kind of expose internal details that would rather be documented. But not really important.
___
*[`tfhe/src/shortint/oprf.rs` line 943 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gDc4DeaL8fbNmAGl:-OuBnZeyA0hyS4c3DPAb:b-xlbps2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L943)):*

it can still be changed, it's just that RLE has a certain definition, it seemed reasonable to have something somewhat mainstream

https://en.wikipedia.org/wiki/Run-length_encoding#:~:text=For%20example%2C%20a%20sequence%20of,drawings%2C%20games%2C%20and%20animations.
</blockquote></details>

Here your encoding has lots of "invalid" states. I guess it does not really matter since it is just hashed but maybe we want to avoid that.

IceTDrinker (commented):
Updated to have a test on 0 bits requested at HL level

other small changes

> @IceTDrinker made 4 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-:-OuHhI-nDpV0aOqaqjlf:b3wsolu)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 662 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6_5p-APdWetNhuXiE:-OuHVwBkCiB7wUs5ZXta:b-3tev9k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L662)):*

Why do you consider this a hard error here? I see it as a normal error. If bit chunk is 0, this specific call is invalid, but it does not prevent to follow the normal course of operations after it returns.
Hard error for me means that the app (ie coproc) state is broken beyond repair. Here this is just a specific call that is wrong, so it is just a classical error IMO.

To give an example:
- user A (or dapp or whatever) requires prf of 0 bits
- user B requires a prf of N > 0 bits

A does not impact B in any ways, so you should just return a clean error to A and carry on. 
Wrong params is something different: the way the lib is used, if params are incorrect for A they are also incorrect for B, so this is a "hard" error and I can understand the panic.
</blockquote></details>

Actually had to manage this to not get errors on existing "correct" behavior

0 random bits returns a trivial 0 at HL/integer level
___
*[`tfhe/src/shortint/oprf.rs` line 873 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6ezJV9omsRfopc-Al:-OuHWIJDBKIuaGgq8Wce:b-6cmunu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L873)):*

Since it is pub(crate) it does not matter, but it kind of expose internal details that would rather be documented. But not really important.
</blockquote></details>

I'll add a docstring on top
___
*[`tfhe/src/shortint/oprf.rs` line 943 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-Ou6gDc4DeaL8fbNmAGl:-OuHWKi-4KQUSL7X74W8:b-he3436) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0c3a5be9ab5b5e13c0ce49cc2c1bce50667361f/tfhe/src/shortint/oprf.rs#L943)):*

Here your encoding has lots of "invalid" states. I guess it does not really matter since it is just hashed but maybe we want to avoid that.
</blockquote></details>

I have tests on the code producing the encoding, and I get what you mean about the encoding

in this instance the fact RLE is "well known" and matches naturally with how the code is written seems to make it a good fit

nsarlin-zama (changes requested):
a few copy paste left, but otherwise looks good!

> @nsarlin-zama reviewed 6 files and all commit messages, made 4 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-:-OuNVvrY8wnyB4caK-hI:bvx50o6)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/integer/gpu/server_key/radix/oprf.rs` line 486 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNNu2qAYH-8PSMM8vL:-OuNNu2qAYH-8PSMM8vM:b-i4mv8i) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/bb17fe2e7e83ba85b54bc59c4be83b8ab20ae646/tfhe/src/integer/gpu/server_key/radix/oprf.rs#L486)):*
> ```XML
>     ///
>     /// Panics if:
>     /// - `num_input_random_bits` is non 0
> ```

not true anymore I guess?
___
*[`tfhe/src/shortint/oprf.rs` line 1493 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNUnrK1IQqcNIfR8Id:-OuNUnrK1IQqcNIfR8Ie:b-h2n3bm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L1493)):*
> ```XML
>         let ciphertext_modulus = CiphertextModulus::new_native();
>         let message_bits: u64 = params.message_modulus().0.ilog2().into();
>         let carry_bits: u64 = params.message_modulus().0.ilog2().into();
> ```

carry_modulus
___
*[`tfhe/src/shortint/oprf.rs` line 1494 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNUtKj5ZxbcP0cpzKa:-OuNUtKj5ZxbcP0cpzKb:b-a2dqbe) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L1494)):*
> ```XML
>         let message_bits: u64 = params.message_modulus().0.ilog2().into();
>         let carry_bits: u64 = params.message_modulus().0.ilog2().into();
>         let bits_per_lock = message_bits + carry_bits + 1;
> ```

per_block

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed, 4 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 985 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNWeh-3DxTNY4HxMKJ:-OuNWeh-3DxTNY4HxMKK:b71sdi0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L985)):*
> ```XML
>     );
> 
>     bit_count_rle.extend(bits_per_block.to_le_bytes());
> ```

Just to be sure. IIUC this will change the prf output before/after, we are ok with that?

IceTDrinker (commented):
> @IceTDrinker made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed, 4 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/integer/gpu/server_key/radix/oprf.rs` line 486 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNNu2qAYH-8PSMM8vL:-OuNYUp04cUVMb72lFo7:bn12xhk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/bb17fe2e7e83ba85b54bc59c4be83b8ab20ae646/tfhe/src/integer/gpu/server_key/radix/oprf.rs#L486)):*

not true anymore I guess?
</blockquote></details>

yes, this feels like a foot gun however, requesting a range but 0 random input bits is stupid...

the value would collapse to 0 I guess
___
*[`tfhe/src/shortint/oprf.rs` line 985 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNWeh-3DxTNY4HxMKJ:-OuNYZOD3-Xp3jrrmgI4:b-45mfqs) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L985)):*

Just to be sure. IIUC this will change the prf output before/after, we are ok with that?
</blockquote></details>

yes since the breaking changes of the RLE encoding are from the bitonic shuffle to be released

This was checked with JB and it seems reasonable to hash the number of bits to have the encoding hashed, I'll just re-confirm but this is expected
___
*[`tfhe/src/shortint/oprf.rs` line 1493 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNUnrK1IQqcNIfR8Id:-OuNYliy2422pcVD9fjK:bd30ifx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L1493)):*

carry_modulus
</blockquote></details>

good catch
___
*[`tfhe/src/shortint/oprf.rs` line 1494 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNUtKj5ZxbcP0cpzKa:-OuNYwLR7QJxRfi-nf8X:bb7p5e3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/08a7ce6a7b2863640093193244b968a1311d8358/tfhe/src/shortint/oprf.rs#L1494)):*

per_block
</blockquote></details>

thx

IceTDrinker:
*[`tfhe/src/integer/gpu/server_key/radix/oprf.rs` line 486 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OuNNu2qAYH-8PSMM8vL:-OuNbkjT9nY8LE6M1yV3:behstqv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/bb17fe2e7e83ba85b54bc59c4be83b8ab20ae646/tfhe/src/integer/gpu/server_key/radix/oprf.rs#L486)):*

yes, this feels like a foot gun however, requesting a range but 0 random input bits is stupid...

the value would collapse to 0 I guess
</blockquote></details>

also the condition was backwards "panics if 0" was correct

mayeul-zama:
*[`tfhe/src/shortint/oprf.rs` line 1359 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvBAEKSEu6RdVTqNVyZ:-OvBAEKSEu6RdVTqNVy_:bjfp8ir) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L1359)):*
> ```XML
> 
> #[cfg(test)]
> pub(crate) mod test {
> ```

Could be moved to a separate file

The non test part of this file should also be split in smaller files IMO

tmontaigu (commented):
> @tmontaigu reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed, 7 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/shortint/oprf.rs` line 783 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJYtsh7boldXcG9I3m:-OvJYtsh7boldXcG9I3n:b-mfaupx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L783)):*
> ```XML
>         // as it can contain the represented modulus.
>         // log_modulus <= 32 so that 2^log_modulus fits in u32 (native modulus included).
>         assert!(self.log_modulus.0 <= 32);
> ```

Could also technically use a usize::BITS >= 32 or self.log_modulus <= usize::BITS, but pragmatically we dont target platforms with usize < 32 bits
___
*[`tfhe/src/shortint/oprf.rs` line 1660 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJc9qG5Iqr0e95y1AO:-OvJc9qG5Iqr0e95y1AP:bhvjbbg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L1660)):*
> ```XML
> 
>                     assert_eq!(blocks.len(), 1);
>                     let blocks = blocks.into_iter().next().unwrap();
> ```

turbo minor suggestion, but `blocks.pop().unwrap()` could be used instead

nsarlin-zama (dismissed):
thanks!

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-:-OvJnmUk-ASj39w5aus9:bd5l6mr)* status: all files reviewed, 7 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, and SouchonTheo).

mayeul-zama:
*[`tfhe/src/shortint/ciphertext/re_randomization.rs` line 252 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJa8rB2ctWmirpl6pu:-OvJa8rC9GsEcSsvS-vZ:bix5lyp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/ciphertext/re_randomization.rs#L252)):*
> ```XML
>         random_bits_rle_bytes: &RandomBitsRleLeBytes,
>     ) -> Self {
>         const PRF_RERAND_DOMAIN_SEPARATOR: [u8; XofSeed::DOMAIN_SEP_LEN] = *b"PRF_RRND";
> ```

Is that part of the spec?

IceTDrinker (commented):
> @IceTDrinker made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed, 7 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/oprf.rs` line 783 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJYtsh7boldXcG9I3m:-OvQANWZDyjz1sF7yVzR:b-osnuan) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L783)):*

Could also technically use a usize::BITS >= 32 or self.log_modulus <= usize::BITS, but pragmatically we dont target platforms with usize < 32 bits
</blockquote></details>

as discussed, TFHE-rs on platforms with less than 64 bits are not technically in what we support for compute, so 16 bits or less for usize is just not something we'll ever want managing
___
*[`tfhe/src/shortint/oprf.rs` line 1359 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvBAEKSEu6RdVTqNVyZ:-OvQB9a70o6HcwUGJO3A:b1g7fr0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L1359)):*

Could be moved to a separate file

The non test part of this file should also be split in smaller files IMO
</blockquote></details>

I can split the tests in the coming PR, I would like to get this merged, splitting the rest, I'm not so sure for now
___
*[`tfhe/src/shortint/oprf.rs` line 1660 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJc9qG5Iqr0e95y1AO:-OvQAqc483K4uVNJKJcs:b-v980ex) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/oprf.rs#L1660)):*

turbo minor suggestion, but `blocks.pop().unwrap()` could be used instead
</blockquote></details>

I'll check godbolt for the sake of it, I'm curious

unclear from godbolt, just saw it was in a test though :rofl: 
___
*[`tfhe/src/shortint/ciphertext/re_randomization.rs` line 252 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580#-OvJa8rB2ctWmirpl6pu:-OvQAx3d3D8B1AW_eZhC:b-3qjgo4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9650706d17b78a798c255366bc7b9c362a74d5de/tfhe/src/shortint/ciphertext/re_randomization.rs#L252)):*

Is that part of the spec?
</blockquote></details>

yes, as in : I chose it and someone said : "Makes sense"

nsarlin-zama (approved):
> @nsarlin-zama resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: 9 of 10 files reviewed, 4 unresolved discussions (waiting on mayeul-zama, soonum, SouchonTheo, and tmontaigu).

tmontaigu (approved):
> @tmontaigu reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3580)* status: all files reviewed, 2 unresolved discussions (waiting on mayeul-zama, soonum, and SouchonTheo).

PR #3469, opened 2026-04-13
chore: make CompressedXofKeySet::decompress take a reference

dvdplm:
Noticed that the `CompressedXofKeySet::decompress` consumes `self`, forcing consumers to clone keys (e.g. in the `kms`).

At a glance it seems like most `decompress` methods in the repo take a reference and I see no reason for this particular one to be different.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3469)

PR #3582, opened 2026-05-27
chore: CompressedXofKeySet is infallible

dvdplm:
Makes `CompressedXofKeySet::decompress` return an `XofKeySet` instead of a `Result`. This mirrors the behaviour of most other `decompress` methods in the library and is consistent with idiomatic Rust.

Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/1380

### DIFF
--- a/tests/backward_compatibility/high_level_api.rs
+++ b/tests/backward_compatibility/high_level_api.rs
@@ -1062,9 +1062,7 @@ fn test_hl_compressed_xof_key_set_test(
         )
     })?;
 
-    let xof_key_set = compressed_xof_key_set
-        .decompress()
-        .map_err(|e| test.failure(format!("Failed to decompress the xof key set: {e}"), format))?;
+    let xof_key_set = compressed_xof_key_set.decompress();
 
     let (pk, server_key) = xof_key_set.into_raw_parts();
 
--- a/tfhe/src/high_level_api/xof_key_set/mod.rs
+++ b/tfhe/src/high_level_api/xof_key_set/mod.rs
@@ -389,7 +389,7 @@ impl CompressedXofKeySet {
     }
 
     /// Decompress the KeySet
-    pub fn decompress(&self) -> crate::Result<XofKeySet> {
+    pub fn decompress(&self) -> XofKeySet {
         let tag = self.compressed_server_key.tag.clone();
         let (mut public_key, expanded_server_key) = self.expand();
         // Server key tag is the source of truth; sync public key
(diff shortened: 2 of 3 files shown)


nsarlin-zama:
thanks for the pr!
However since we are > 1.0, we try to avoid breaking changes, at least in the HL-api. Maybe this could be added as a new function instead, and removed once we hit 2.0 ?

dvdplm:
> However since we are > 1.0, we try to avoid breaking changes, at least in the HL-api. 

Ok, I accept that argument, but still want to present the counter-argument! :)

It goes like this:

- There are 50 `pub fn decompress(` in the repo
- There are 10 `pub fn decompress(` in `src/high_level_api` alone
- Only one return `Result`, and that's `XofKeySet::decompress`
- Given the above, it is not so much about a refactor as it is a bugfix
- The code is fairly new? I think it was introduced [here](https://github.com/zama-ai/tfhe-rs/commit/f52eb16581e7b4b9cec77235b17f9779c1bf9a58#diff-58c9793af88b226dd99e4e0969630f4b3e13c17c667635ca2e4520a00cacaf4dR321), i.e. end of January this year.
- There is "prior art" in my previous PR touching this: https://github.com/zama-ai/tfhe-rs/pull/3469

nsarlin-zama (approved):
Fair points. Fine by me! :)

dvdplm:
> Maybe this could be added as a new function instead, and removed once we hit 2.0 ?

I am not sure that does very much. Right now we do this all over the place: `keyset.decompress().expect("It's actually infallible")`. 

Adding a new method with some odd name wouldn't do much to clean up the situation. Say we call it `decompress_direct()` or something. Now there are two `pub` methods  that do the same job, and when 2.0 rolls around, we have to deprecate `decompress_direct()` and change `decompress()` to return the right type, thus punishing users that opted for using the "correct" method... All in all an even worse mess imo!

nsarlin-zama:
yes, but the issue is that the dependency resolver in Cargo is semver aware, so for downstream users it means that their code will stop building correctly once 1.7 is released. And behavior might even not be consistent depending on if a user has a Cargo.lock, and when it was generated.
So that's why once we release 1.0 we are not supposed to change any pub api. Since the resolver will not automatically bump from 1.x to 2.0 it's ok to break at this point.

But since this api is made for you I guess we can be a bit more pragmatic here :)

dvdplm:
Don't get me wrong here: I really do buy the semver argument and if you opt to close this PR then I will not harbor any hard feelings!

nsarlin-zama:
Basically going from 1.6 to 1.7 signals to users "it's ok to update without touching your code if you don't need new things". Going from 1.6 to 2.0 means "ok, now you will have to take the time to read the release note and change a few method calls"

nsarlin-zama:
It's ok I approved it, you convinced me. :) There are the theoretical rules but sometimes it's ok to break them if we know that the impact downstream is minimal !

PR #3593, opened 2026-05-29
repairing HPU bench + add BW_NOT support to HPU in HLAPI

pgardratzama:
This depends on https://github.com/zama-ai/zhc/pull/119 from ZHC to enable BW_NOT support
also updated price of hpu_x1, x2, x4, x8 in anticipation of multi-HPU backend that is coming soon
also added overflow_add/sub/mul to HPU integer bench


Bapt-Roux (dismissed):
Looks good to me

Makefile
```diff
@@ -1476,11 +1476,11 @@ check_main_readme_links: install_mlc
 
 .PHONY: check_doc_paths_use_dash # Check paths use "-" instead of "_" in docs for gitbook compatibility
 check_doc_paths_use_dash:
-	python3 ./scripts/check_doc_paths_use_dash.py
+	python3.11 ./scripts/check_doc_paths_use_dash.py
 
 .PHONY: check_parameter_export_ok # Checks exported "current" shortint parameter module is correct
 check_parameter_export_ok:
-	python3 ./scripts/check_current_param_export.py
+	python3.11 ./scripts/check_current_param_export.py
```
IceTDrinker: this should not be necessary, you could have pip install -U pip to make sure you have some more recent libs

IceTDrinker: otherwise we can add a target that manages a venv to avoid the headaches

pgardratzama: Agreed, this was a local change I had to make. All I wanted is the change of target name.
Fixing...

Bapt-Roux (approved):

PR #3599, opened 2026-06-02
ref(gpu): Improve comparisons performance

bbarbakadze:

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
@@ -14,23 +14,30 @@
 #include "pbs/programmable_bootstrap_multibit.cuh"
 #include "utils/helper.cuh"
 
-// lwe_dimension + 1 threads
-// todo: This kernel MUST be refactored to a binary reduction
 template <typename Torus>
 __global__ void
 device_accumulate_all_blocks(Torus *output, Torus const *input_block,
                              uint32_t lwe_dimension, uint32_t num_blocks) {
-  int idx = threadIdx.x + blockIdx.x * blockDim.x;
-  if (idx < lwe_dimension + 1) {
-    auto block = &input_block[idx];
+  extern __shared__ Torus smem[];
 
-    Torus sum = block[0];
-    for (int i = 1; i < num_blocks; i++) {
-      sum += block[i * (lwe_dimension + 1)];
-    }
+  int coeff_idx = blockIdx.x * blockDim.x + threadIdx.x;
+  int lwe_size = lwe_dimension + 1;
 
-    output[idx] = sum;
+  Torus val = (coeff_idx < lwe_size && threadIdx.y < num_blocks)
+                  ? input_block[threadIdx.y * lwe_size + coeff_idx]
+                  : 0;
+  smem[threadIdx.y * blockDim.x + threadIdx.x] = val;
+  __syncthreads();
+
+  for (uint32_t s = blockDim.y >> 1; s > 0; s >>= 1) {
+    if (threadIdx.y < s)
+      smem[threadIdx.y * blockDim.x + threadIdx.x] +=
+          smem[(threadIdx.y + s) * blockDim.x + threadIdx.x];
+    __syncthreads();
   }
+
+  if (threadIdx.y == 0 && coeff_idx < lwe_size)
+    output[coeff_idx] = smem[threadIdx.x];
 }
 
 template <typename Torus>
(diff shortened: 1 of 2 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -40,12 +47,20 @@ __host__ void accumulate_all_blocks(cudaStream_t stream, uint32_t gpu_index,
...
+  while (block_dim_y < num_radix_blocks)
+    block_dim_y <<= 1;
+
+  uint32_t block_dim_x = std::max(1u, std::min(512u / block_dim_y, 512u));
+  uint32_t grid_dim_x = CEIL_DIV(lwe_dimension + 1, block_dim_x);
+
+  dim3 block_dim(block_dim_x, block_dim_y);
+  size_t smem_size = block_dim_x * block_dim_y * sizeof(Torus);
+
+  device_accumulate_all_blocks<Torus>
```
andrei-stoian-zama, line 61: noise isn't tracked here it would seem!

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -40,12 +47,20 @@ __host__ void accumulate_all_blocks(cudaStream_t stream, uint32_t gpu_index,
                                     uint32_t num_radix_blocks) {
 
   cuda_set_device(gpu_index);
-  int num_blocks = 0, num_threads = 0;
-  int num_entries = (lwe_dimension + 1);
-  getNumBlocksAndThreads(num_entries, 512, num_blocks, num_threads);
-  // Add all blocks and store in sum
-  device_accumulate_all_blocks<Torus><<<num_blocks, num_threads, 0, stream>>>(
-      output, input, lwe_dimension, num_radix_blocks);
+
```
andrei-stoian-zama, line 50: let's rename this function:

`host_lwe_array_1d_sum`

and move it to linearalgebra/addition.cuh

The naming was chosen to be consistent with a new convention @enzodimaria is implementing in the match_value PR:

`host_lwe[[_flat][_array]]_[1|2]d_[sum|accumulate|aggregate]_[rows|cols|full]`

e.g. 
- `host_lwe_flat_array_2d_sum_rows` sums the rows of a 2d array of LWEs that is contiguous in memory
- `host_lwe_array_1d_aggregate` could combine all elements of a 1d LWE array using some kind of PBS

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -108,18 +123,25 @@ __host__ void are_all_comparisons_block_true(
     auto is_max_value_lut = are_all_block_true_buffer->is_max_value;
     uint32_t chunk_lengths[num_chunks];
     auto begin_remaining_blocks = remaining_blocks;
+    cudaEventRecord(are_all_block_true_buffer->sync_event, streams.stream(0));
```
andrei-stoian-zama, line 126: I understand you want to use streams to optimize this function (parallelize different chunks?)

what about the approach from @enzodimaria in this match_value PR? 

he has a tree-reduction approach: a kernel processes "chunk_size" chunks of blocks in a single pass. that would parallelize the aggregation without streams.

the function here is a generalization of the one in the match_value -> that one only clears noise after chunk_size lwes are summed, here you apply a different lut.

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -108,18 +123,25 @@ __host__ void are_all_comparisons_block_true(
...
       accumulator_ptr += (big_lwe_dimension + 1);
       remaining_blocks -= (chunk_length - 1);
       input_blocks += (big_lwe_dimension + 1) * chunk_length;
     }
+    for (int i = 0; i < num_chunks; i++)
+      cudaStreamWaitEvent(streams.stream(0),
+                          are_all_block_true_buffer->chunk_done_events[i]);
     auto accumulator = are_all_block_true_buffer->tmp_block_accumulated;
 
     // Selects a LUT
```
andrei-stoian-zama, line 147:
```
        is_max_value_lut->generate_and_broadcast_lut(
            active_streams, {1}, {is_equal_to_num_blocks_lut_f}, index_gen,
            true, {are_all_block_true_bu
```

this could be replaced: instead of generating a special lut depending on the last chunk's size we could:

- add `max_chunk_size - last_chunk_size` encryptions of 1 to the last accumulated value. 
- use the lut `value == max_chunk_size` to check at the end.

the padding could be handled as a parameter in the tree-sum function of @enzodimaria

andrei-stoian-zama (changes requested):
Thanks for the PR! 

I think it requires quite a few changes:
- we should try to avoid using sub-streams. I think it's possible with a tree-reduction approach. but if we have to use streams, we should look into the existing StreamsBarrier class and check if it could be used
- I suggest additional refactoring of the `all_comparisons_blocks_true` function, following the example of tree sums in @enzodimaria 's PR on match_value
- The comparisons for the last chunk of the blocks could be done with the same LUT -> reducing complexity and avoiding re-computing lut indexes. Basically `value == n_remaining` can be expressed as `value + (chunk_size - n_remaining) == chunk_size` -> a unique LUT is sufficient . the padding can be incorporated in the summation kernel maybe

PR #3604, opened 2026-06-03
Ns/chore/nightly floating deps

nsarlin-zama:
Adds a nightly workflow run that removes the Cargo.lock and performs clippy checks to verify that everything build correctly.
Since according to our security model, this means running potentially untrusted deps, this workflow is running in a limited environment without access to any secret. The idea is to limit the blast radius as much as possible.

This means:
- only run inside a github instance
- runs only on public repo -> no token needed to clone
- use another workflow with `workflow_run/completed` trigger to post the result on slack.

So we have 2 workflows: one "trusted" workflow with access to the slack token but does not run any code. One "untrusted" workflow that runs untrusted deps but has no access to any token or secret.

Communication from "untrusted" to "trusted" is only done using github controlled metadata. No execution output or artefact from the workflow is ever read.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604)

### DIFF
--- a/.github/workflows/floating_deps.yml
+++ b/.github/workflows/floating_deps.yml
@@ -1,0 +1,40 @@
+# Build and lint tfhe-rs against the latest semver-compatible dependencies.
+#
+# This job runs UNTRUSTED code: it updates dependencies to their latest semver
+# compatible version (ignoring the committed Cargo.lock), so a newly published
+# malicious dependency version could execute in build scripts / proc-macros.
+# It therefore runs on an ephemeral GitHub-hosted runner with no secrets and
+# read-only permissions.
+name: floating_deps
+
+on:
+  workflow_dispatch:
+  schedule:
+    # runs every day at 5am UTC
+    - cron: '0 5 * * *'
+
+env:
+  CARGO_TERM_COLOR: always
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  check-floating-deps:
+    name: floating_deps/check-floating-deps
+    if: github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'schedule' && github.repository == 'zama-ai/tfhe-rs')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # for actions/checkout
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+        with:
+          persist-credentials: 'false'
+
+      - name: Build and lint against up-to-date dependencies
+        run: |
+          make check_floating_deps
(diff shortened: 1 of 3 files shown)


IceTDrinker (commented):
Just one question on permissions granularity, I know we had some weird result on the GPU CI due to some permission quirk

> @IceTDrinker reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604#-:-OuCerGX1SJ9iku5xsft:bo39709)* status: all files reviewed, 1 unresolved discussion (waiting on nsarlin-zama, soonum, and SouchonTheo).

___
*[`.github/workflows/floating_deps.yml` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604#-OuCeNZi5q4kcSBmCQ1y:-OuCeNZj8Q028JldL2Ub:b260pgi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/eddbe7d1c133055d4b21413c454377e641cff602/.github/workflows/floating_deps.yml#L23)):*
> ```YAML
>   cancel-in-progress: true
> 
> permissions:
> ```

no sure about this but should the workflow level permissions be {}

and the job level permissions contents: read ?

I seem to recall both are not equivalent, could be wrong though

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on IceTDrinker, soonum, and SouchonTheo).

___
*[`.github/workflows/floating_deps.yml` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604#-OuCeNZi5q4kcSBmCQ1y:-OuD6XKFDGGDhv_PJp4H:b-eg3t58) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/eddbe7d1c133055d4b21413c454377e641cff602/.github/workflows/floating_deps.yml#L23)):*

no sure about this but should the workflow level permissions be {}

and the job level permissions contents: read ?

I seem to recall both are not equivalent, could be wrong though
</blockquote></details>

from what I understand, you are correct. That way we are "deny by default" and only add permissions on a per job basis.
In that case since there is only one job it's equivalent but I think it's better to keep the pattern, that way it will already be correct if we add more jobs later.

done

IceTDrinker (approved):
Thanks !

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3604#-:-OuD9VsK5zBKgcGFfS2y:b-hay2j5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on soonum and SouchonTheo).

PR #3607, opened 2026-06-03
chore(ci): add dependabot config for rust

nsarlin-zama:
Adds dependabot for rust.

/!\ This is only for non-security updates. Security updates are done through the alert mechanism (also from dependabot).

Updates to the Cargo.lock are done once every month, to reduce the noise. A 7 days cooldown is applied to reduce supply chain risks (might be modified later, especially if/when we add cargo vet).
Updates pr for minor/patch bumps are batched to reduce the review work. Major version updates get their own pr to be treated with special care.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -9,3 +9,24 @@ updates:
       day: "sunday"
     cooldown:
       default-days: 7
+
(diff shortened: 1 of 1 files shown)


soonum (approved):
Nice ! On our way to have up to date dependencies 🎉

PR #3617, opened 2026-06-05
  feat!: change 'overshift' behavior

tmontaigu:
Previously a in a shift, left or right of a
FheUintX << y, FheUintX >> y (and for FheInt too)

A modulo by trunctaion was applied to `y`.
This worked well when X was a power of 2,
as the trunctaion meant we did a real `y % X` modulo, wrapping y to values < X
However, when X was not a power of two then the modulo was `y % (1 << ceil(log2(X)))` which meant y could still hold values >= X

This made the behavior inconsistent between the two cases, and we relied on the user to know about it and handle it.

This commit changes that behavior to now make 'overshift', i.e shifts where y >= X return 0, (or -1 for right shift of negative
                values).
This cost one additional scalar comparison, done in parallel of the shift, and a conditional zeroing which, depending on the algorithm and parameters is absorbed or not.

BREAKIN_CHANGES: not an api break, but behavior break, however, the new
        behavior is likely more intuive for users

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3617)

### DIFF
--- a/tfhe/src/high_level_api/integers/signed/ops.rs
+++ b/tfhe/src/high_level_api/integers/signed/ops.rs
@@ -1894,6 +1894,11 @@ where
 {
     /// Performs the `<<=` operation on [FheInt]
     ///
+    /// # Overshift
+    ///
+    /// If the shift amount is greater than or equal to the number of bits of the type,
+    /// the result is `0`.
+    ///
     /// # Example
     ///
     /// ```rust
@@ -1950,6 +1955,12 @@ where
     Id2: FheUintId,
 {
     /// Performs the `>>=` operation on [FheInt]
+    ///
+    /// # Overshift
+    ///
+    /// If the shift amount is greater than or equal to the number of bits of the type,
+    /// the result is `0` for non-negative inputs, and `-1` for negative inputs
+    /// (sign-extension).
     ///
     /// # Example
     ///
--- a/tfhe/src/high_level_api/integers/signed/scalar_ops.rs
+++ b/tfhe/src/high_level_api/integers/signed/scalar_ops.rs
@@ -680,6 +680,7 @@ macro_rules! define_scalar_rotate_shifts {
     ) => {
 
         generic_integer_impl_scalar_operation!(
+            op_doc: "# Overshift\n\nIf the shift amount is greater than or equal to the number of bits of the type, the result is `0`.",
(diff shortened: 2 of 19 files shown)


guillermo-oyarzun:
Can this change affect some of the algorithms we currently have for other integer ops that might use the shifts? do we need to keep both behaviours?

tmontaigu (commented):
No, none of our algorithm depended on that behavior so no need to keep both

> @tmontaigu made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3617#-:-OvJxfzT17Q-LepZPXw4:bxf8jph)* status: 0 of 15 files reviewed, all discussions resolved (waiting on nsarlin-zama).

nsarlin-zama (commented):
nice implementation, it looks correct and optimal to me !
Just one thing, I think it could be worth documenting the new overshift behavior at the api boundaries ?

> @nsarlin-zama reviewed 7 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3617#-:-OvVAD1_62S8Tm-BodKd:bbvef67)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on mayeul-zama and tmontaigu).

___
*[`tfhe/src/integer/server_key/radix/shift.rs` line 240 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3617#-OvUf5EpDU_Zt_QeuhIE:-OvUf5EpDU_Zt_QeuhIF:b-r5t2p1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e3421ac4422cb58f91653053f073c288f821ba3e/tfhe/src/integer/server_key/radix/shift.rs#L240)):*
> ```XML
>     }
> 
>     /// Computes homomorphically a left shift by a scalar.
> ```

maybe document the behavior in the doc comment?

nsarlin-zama (approved):
thanks!

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3617#-:-OvjR--E3u4DIaG8ALuR:bd5l6mr)* status: 6 of 19 files reviewed, 1 unresolved discussion (waiting on mayeul-zama and tmontaigu).

PR #3623, opened 2026-06-11
feat(gpu): accelerate comparisons on 2_2 params unsigned integers

guillermo-oyarzun:

### PR content/description
The idea is reuse overflowing sub logic as on rust side to perform some of the comparisons, the limit is 2_2 params since carry prop is optimized for it. It reduces the batches of pbs from 6->5, provides about 18% of latency improvement on the comparisons.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
@@ -349,6 +349,17 @@ template <typename Torus> struct int_comparison_buffer {
   int_comparison_eq_buffer<Torus> *eq_buffer;
   int_comparison_diff_buffer<Torus> *diff_buffer;
 
+  /// @brief Borrow-propagation fast path for unsigned GT/GE/LT/LE on 2_2
+  /// params. When enabled, the comparison reuses the overflow (borrow-out) path
+  /// of the overflowing-sub borrow propagation instead of the sign-tree
+  /// reduction.
+  bool use_borrow_fast_path;
+  /// @brief Borrow-propagation memory, allocated only on the fast path.
+  int_borrow_prop_memory<Torus> *diff_borrow_mem;
+  /// @brief Final single-block LUT applying the per-op inversion:
+  ///   f(block) = ((block >> 2) & 1) ^ invert
+  int_radix_lut<Torus> *lut_borrow_flag_cmp;
+
   CudaRadixCiphertextFFI *tmp_block_comparisons;
   CudaRadixCiphertextFFI *tmp_lwe_array_out;
   CudaRadixCiphertextFFI *tmp_trivial_sign_block;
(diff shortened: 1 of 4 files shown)


pdroalves:
@guillermo-oyarzun I suppose you benchmarked the speedup with this PR. Can you add the numbers to the description? Is is 18% for all operations?

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
```diff
@@ -369,15 +380,34 @@ template <typename Torus> struct int_comparison_buffer {
...
     this->is_signed = is_signed;
+    diff_borrow_mem = nullptr;
+    lut_borrow_flag_cmp = nullptr;
+    // When we are in the fast path we can avoid calculating the luts for the
+    // tree reduction.
+    use_borrow_fast_path =
+        allow_borrow_fast_path &&
+        (op == COMPARISON_TYPE::GT || op == COMPARISON_TYPE::GE ||
+         op == COMPARISON_TYPE::LT || op == COMPARISON_TYPE::LE) &&
+        !is_signed && params.message_modulus == 4 && params.carry_modulus == 4;
```
pdroalves, line 403: Since the 2_2 params are "special" (we have special PBSs but also integer algorithms for it), we could encapsulate the check if they are in use. Nothing to be done in this PR, but maybe something to thing about to avoid rewriting the same thing all the time and simplifying our lifes in case that set changes.

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
```diff
@@ -387,72 +417,78 @@ template <typename Torus> struct int_comparison_buffer {
...
+          streams.stream(0), streams.gpu_index(0), tmp_packed_input,
+          2 * num_radix_blocks, params.big_lwe_dimension, size_tracker,
+          allocate_gpu_memory);
 
-    uint32_t total_modulus = params.message_modulus * params.carry_modulus;
-    auto is_zero_f = [total_modulus](Torus x) -> Torus {
-      return (x % total_modulus) == 0;
-    };
+      // Cleaning LUT
+      identity_lut =
```
pdroalves, line 445: Another idea: since we use identity luts so much, maybe we could have a `create_identity_lut()` method.

pdroalves (dismissed):
Great job. I did my best to find real issues, including asking for two different claude routines to perform AI reviews, but couldn't find anything wrong. I added a few general ideas for future thinking, but IMO this is good to go.

guillermo-oyarzun:
> @guillermo-oyarzun I suppose you benchmarked the speedup with this PR. Can you add the numbers to the description? Is is 18% for all operations?

this change affects only le/ge/gt/lt, i ran more sizes
Multi-bit             Classical
8 : Same                  Same
16: 29%                    31%
32: 27%                    23%
64: 22%                    18%
80: 22%                    15%
96: 9%                     Same
128: Same              Same  

thanks to rerunning I found a mistake in carry prop when using 128-bit, basically we were selecting a different algorithm which made us perform an extra batch in the 128-bit case. it is fixed now

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
guillermo-oyarzun: i think is would be a good addition to our common ops, indeed we do the identity many times, maybe we could even have it preloaded in the gpu memory

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
guillermo-oyarzun: We do have such a function, but it is more restrictive than what is needed here. i will rethink about it when we rework the signed version of the comparisons

backends/tfhe-cuda-backend/cuda/include/integer/comparison.h
pdroalves: I created an issue for that: https://github.com/zama-ai/tfhe-rs-internal/issues/1448

pdroalves (approved):

PR #3624, opened 2026-06-12
chore: update shellflags in Makefile to be in POSIX compliant mode

IceTDrinker:
https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

- the -u option was triggering some errors on a bench machine, reviewing the documentation systems likely do not expect a -u in default make shellflags, so let's use a default that's advertised

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 SHELL:=$(shell /usr/bin/env which bash)
-# Enable stop on error, no undefined variables
+# Enable stop on error
 # the c flag is to run the script inline
-.SHELLFLAGS := -eu -c
+# https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
+# POSIX conforming mode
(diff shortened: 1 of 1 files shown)


tmontaigu (dismissed):

Makefile
```diff
@@ -1,7 +1,7 @@
 SHELL:=$(shell /usr/bin/env which bash)
-# Enable stop on error, no undefined variables
-# the c flag is to run the script inline
-.SHELLFLAGS := -eu -c
+# https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html
```
soonum, line 4: Why removing completely the old comment ? I think we just have to remove the "no undefined variables" part and append the new comment to show that 's POSIX compatible.

IceTDrinker: sure

soonum (approved):

PR #3638, opened 2026-06-17
[feat] API for selective decompression of `CompressedXofKeySet`

dvdplm:
Currently users of `CompressedXofKeySet` calling `decompress()` must pay the cost of decompressing all nine key parts even though just a subset of keys are actually used.

This is a code sample from a test on the KMS `main` branch:

```rust
            let (pk, server_key) = compressed_keyset.decompress().unwrap().into_raw_parts();
            let (_, _, _, _, _, _, _, oprf_key, _) = server_key.clone().into_raw_parts();
```

After this PR which introduces a `decompress_parts()` API, the above code would look like this:

```rust
            let (pk, oprf_key): (
                tfhe::CompactPublicKey,
                Option<tfhe::integer::oprf::OprfServerKey>,
            ) = compressed_keyset.decompress_parts();
```

The space/time savings in the KMS are fairly substantive, here shown as measured with the `hotpath` profiler using default (production) parameters on a subset of tests that are representative of user/public decryption:

| flow | metric | original | current | delta |
|---|---|---:|---:|---:|
| user decrypt NoiseFlood | runtime | `12.07s` | `6.88s` | `-5.19s` / `-43%` |
| user decrypt NoiseFlood | Hotpath alloc | `13.9 GB` | `8.9 GB` | `-5.0 GB` / `-36%` |
| user decrypt NoiseFlood | sampled RSS | `15.4 GB` | `12.1 GB` | `-3.3 GB` / `-21%` |
| public decrypt NoiseFlood | runtime | `12.13s` | `9.84s` | `-2.29s` / `-19%` |
| public decrypt NoiseFlood | Hotpath alloc | `13.9 GB` | `10.8 GB` | `-3.1 GB` / `-22%` |
| public decrypt NoiseFlood | sampled RSS | `15.4 GB` | `13.5 GB` | `-1.9 GB` / `-12%` |

The PR contains a benchmark that is mostly there to prove that the overhead of selective decompression is low. One benchmark measures a full `decompress()` call and another shows that decompressing each component one by one has less than 10% overhead. For use cases that only select a subset of keys the overhead is definitely worth it.

As a hint of possible follow-up work, the benchmarks also include a couple of `rayon` variants, showing how this PR provides the machinery necessary for decompressing keys in parallel.

Results on my laptop:

| variant | time | × full |
|---|---:|---:|
| `full` (`decompress()`) | 1.97 s | 1.00× |
| `all_parts_one_by_one` | 2.02 s | 1.03× |
| `all_parts_parallel` | 1.10 s | 0.56× |
| `kms_subset` | 1.41 s | 0.72× |
| `kms_subset_parallel` | 0.94 s | 0.48× |

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638)

### DIFF
--- a/tfhe-benchmark/Cargo.toml
+++ b/tfhe-benchmark/Cargo.toml
@@ -95,6 +95,12 @@ harness = false
 required-features = ["integer", "internal-keycache"]
 
 [[bench]]
+name = "hlapi-xof-keyset-decompress"
+path = "benches/high_level_api/xof_keyset_decompress.rs"
+harness = false
+required-features = ["integer", "internal-keycache"]
+
+[[bench]]
 name = "hlapi-bitonic-shuffle"
 path = "benches/high_level_api/bitonic_shuffle.rs"
 harness = false
--- a/tfhe-benchmark/benches/high_level_api/xof_keyset_decompress.rs
+++ b/tfhe-benchmark/benches/high_level_api/xof_keyset_decompress.rs
@@ -1,0 +1,137 @@
+//! Sanity-check microbench for selective decompression of a `CompressedXofKeySet`.
+//!
+//! `full` decompresses the whole key set; `all_parts_one_by_one` fetches every part through the
+//! selective API; `kms_subset` fetches only what `NoiseFloodSmall` needs. The `_parallel` variants
+//! fetch concurrently via rayon — the independent per-part walks need no coordination.
+
+use benchmark::params_aliases::{
+    BENCH_COMP_NOISE_SQUASHING_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+    BENCH_COMP_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+    BENCH_NOISE_SQUASHING_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+    BENCH_PARAM_KEYSWITCH_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+    BENCH_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+    BENCH_PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+};
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::hint::black_box;
+use tfhe::core_crypto::prelude::NormalizedHammingWeightBound;
+use tfhe::integer::ciphertext::NoiseSquashingCompressionKey;
+use tfhe::integer::compression_keys::{CompressionKey, DecompressionKey};
+use tfhe::integer::key_switching_key::KeySwitchingKeyMaterial;
+use tfhe::integer::noise_squashing::NoiseSquashingKey;
+use tfhe::integer::oprf::OprfServerKey;
+use tfhe::integer::ServerKey;
+use tfhe::xof_key_set::CompressedXofKeySet;
+use tfhe::{CompactPublicKey, ConfigBuilder, ReRandomizationKey, Tag};
+
+fn xof_keyset_decompress(c: &mut Criterion) {
+    let config =
+        ConfigBuilder::with_custom_parameters(BENCH_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128)
+            .use_dedicated_compact_public_key_parameters((
+                BENCH_PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+                BENCH_PARAM_KEYSWITCH_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+            ))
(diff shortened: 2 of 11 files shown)


tmontaigu (changes requested):
> @tmontaigu reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on dvdplm, IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs` line 173 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638#-Oviw6Gx7zDYwdN9UX1n:-Oviw6Gx7zDYwdN9UX1o:b5c536f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/512143230462044135572b9b2cc9304a7bd04a8e/tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs#L173)):*
> ```XML
> 
>     /// Advance the generator past `fork_config` worth of mask bytes, producing nothing.
>     pub(crate) fn skip(&mut self, fork_config: MaskRandomGeneratorForkConfig) {
> ```

API wise, I think is better for this method to take an `EncryptionMaskByteCount` it is more suited to what the function does, and does not leak the fact that forking is using to do the skipping because that's an implementation detail

dvdplm:
*[`tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs` line 173 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638#-Oviw6Gx7zDYwdN9UX1n:-OvpEVNA9hoYaaIUpCKm:b-hzzmzg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/512143230462044135572b9b2cc9304a7bd04a8e/tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs#L173)):*

API wise, I think is better for this method to take an `EncryptionMaskByteCount` it is more suited to what the function does, and does not leak the fact that forking is using to do the skipping because that's an implementation detail
</blockquote></details>

After 7260707ecb32b50b1041a824b793fad9526bf801 `skip` now takes an `EncryptionMaskByteCount`.

tmontaigu (changes requested):
> @tmontaigu reviewed 1 file and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638)* status: 1 of 11 files reviewed, 3 unresolved discussions (waiting on dvdplm, IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs` line 89 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638#-Ow2DXfL-2saK0X8EvPK:-Ow2DXfL-2saK0X8EvPL:b-dycpnj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7260707ecb32b50b1041a824b793fad9526bf801/tfhe/src/core_crypto/commons/generators/encryption/mask_random_generator.rs#L89)):*
> ```XML
>                 .checked_mul(self.mask_byte_count_per_child.0)
>                 .expect("mask byte count overflow"),
>         )
> ```

minor  style suggestion: 



_Suggestion:_
```XML
         self.children_count
                .checked_mul(self.mask_byte_count_per_child.0)
                .map(EncryptionMaskByteCount)
                .expect("mask byte count overflow")
```
___
*[`tfhe/src/high_level_api/xof_key_set/parts.rs` line 67 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638#-Ow2DDik-bFxaUHbTKqX:-Ow2DDik-bFxaUHbTKqY:b780260) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7260707ecb32b50b1041a824b793fad9526bf801/tfhe/src/high_level_api/xof_key_set/parts.rs#L67)):*
> ```XML
>     /// skipping every present component before it.
>     fn generator_at(&self, target: Slot) -> MaskRandomGenerator<DefaultRandomGenerator> {
>         use Slot::*;
> ```

Clippy complains about the glob import, wa can add `#[allow(clippy::enum_glob_use)]` locally
___
*[`tfhe/src/high_level_api/xof_key_set/parts.rs` line 247 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3638#-Ow2K8_ND-nYyAL9Egl_:-Ow2K8_ND-nYyAL9Egla:b-shh7hc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/7260707ecb32b50b1041a824b793fad9526bf801/tfhe/src/high_level_api/xof_key_set/parts.rs#L247)):*
> ```XML
>         let mut gen = keyset.generator_at(Slot::ReRand);
> 
>         // Mirrors the inline expansion in
> ```

Instead of mirroring by copy-pasting we should share the code, so probalby add a decompress_with_pre_seed_generator on theCompressedRandomizationKey

tmontaigu (approved):

Issue #3629, opened 2026-06-14
c_api: Add shortint_ciphertext_clone to shortint module


rdlopes:
**What is the problem you want to solve and can not with the current version?**

I'm always frustrated when I need to clone or duplicate a `ShortintCiphertext` in foreign language bindings (such as `tfhe-java`). 
Currently, the high-level integer API exposes native cloning functions (e.g. `fhe_uint8_clone`), but the `shortint` module C FFI does not expose any cloning or copy function for its ciphertexts. 

To work around this, we are forced to simulate cloning by serializing the ciphertext into a byte buffer and immediately deserializing it back to allocate a new native object:
```java
// Java workaround for cloning a ShortintCiphertext
@Override
public ShortintCiphertext clone() {
  return deserialize(serialize());
}
```
This round-trip serialization is computationally heavy, involves unnecessary memory allocations, and complicates performance profiling.

**Describe the solution you'd like**

Expose a native cloning function in `tfhe/src/c_api/shortint/ciphertext.rs` matching the pattern used in the high-level integer API:
```rust
#[no_mangle]
pub unsafe extern "C" fn shortint_ciphertext_clone(
    sself: *const ShortintCiphertext,
    result: *mut *mut ShortintCiphertext,
) -> c_int {
    catch_panic(|| {
        check_ptr_is_non_null_and_aligned(result).unwrap();
        let cloned = get_ref_checked(sself).unwrap().0.clone();
        *result = Box::into_raw(Box::new(ShortintCiphertext(cloned)));
    })
}
```

**Describe alternatives you've considered**

An alternative is to expose an in-place copy function that copies the internal state from a source ciphertext into an already allocated destination ciphertext, thereby avoiding dynamic memory allocation within Rust:
```rust
#[no_mangle]
pub unsafe extern "C" fn shortint_ciphertext_copy(
    src: *const ShortintCiphertext,
    dest: *mut ShortintCiphertext,
) -> c_int {
    catch_panic(|| {
        let src_val = get_ref_checked(src).unwrap();
        let dest_val = get_mut_checked(dest).unwrap();
        dest_val.0 = src_val.0.clone();
    })
}
```

**Additional context**

This issue was identified during the FFM Java binding generation work in `tfhe-java`, where ciphertext duplication is required in generic expression builders.

tmontaigu:
Yes this makes sense, we will add it

tmontaigu:
Closing as it's merged in main

PR #3642, opened 2026-06-18
feat(capi): add missing clone/clone_from

tmontaigu:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3642)

### DIFF
--- a/tfhe/src/c_api/shortint/ciphertext.rs
+++ b/tfhe/src/c_api/shortint/ciphertext.rs
@@ -7,6 +7,9 @@ use std::os::raw::c_int;
 pub struct ShortintCiphertext(pub(in crate::c_api) shortint::Ciphertext);
 pub struct ShortintCompressedCiphertext(pub(in crate::c_api) shortint::CompressedCiphertext);
 
+impl_wrapper_type!(ShortintCiphertext, shortint::Ciphertext);
+impl_wrapper_type!(ShortintCompressedCiphertext, shortint::CompressedCiphertext);
+
 #[no_mangle]
 pub unsafe extern "C" fn shortint_ciphertext_set_degree(
     ciphertext: *mut ShortintCiphertext,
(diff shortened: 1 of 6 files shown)


tmontaigu:
closes https://github.com/zama-ai/tfhe-rs/issues/3629

tfhe/src/c_api/shortint/ciphertext.rs
```diff
@@ -37,6 +37,64 @@ pub unsafe extern "C" fn shortint_ciphertext_get_degree(
...
+        *result = Box::into_raw(boxed);
+    })
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn shortint_compressed_ciphertext_clone_from(
+    dest: *mut ShortintCompressedCiphertext,
+    src: *const ShortintCompressedCiphertext,
+) -> c_int {
+    catch_panic(|| {
```
mayeul-zama: I think we should have a generic function 

```
unsafe fn generic_c_api_clone_from<T>(
    dest: *mut T,
    src: *const T,
) -> c_int {
    catch_panic(|| {
        let dest = get_mut_checked(dest).unwrap();
        let src = get_ref_checked(src).unwrap();

        dest.0.clone_from(&src.0);
    })
}
```

And simply call it in all specific functions

```
#[no_mangle]
pub unsafe extern "C" fn shortint_compressed_ciphertext_clone_from(
    dest: *mut ShortintCompressedCiphertext,
    src: *const ShortintCompressedCiphertext,
) -> c_int {
    unsafe {shortint_compressed_ciphertext_clone_from(dest, src)}
}
```

Also applies to the clone function, and maybe others

mayeul-zama (commented):
I'm wondering if tests could be useful to add

tfhe/src/c_api/shortint/ciphertext.rs
tmontaigu: done

tfhe/src/c_api/utils.rs
```diff
@@ -13,6 +13,76 @@ where
...
+/// `sself` must point to a valid `T`, and `result` must point to a writable `*mut T` slot.
+pub(in crate::c_api) unsafe fn generic_c_api_clone<T>(sself: *const T, result: *mut *mut T) -> c_int
+where
+    T: WrapperType,
+    T::Wrapped: Clone,
+{
+    catch_panic(|| {
+        check_ptr_is_non_null_and_aligned(result).unwrap();
+
+        *result = std::ptr::null_mut();
```
mayeul-zama, line 60: Could add comment 
```
// First fill the result with a null ptr so that if we fail and the return code is
// checked, then any access to the result pointer will segfault (mimics malloc on failure)
```
as in other parts of the c api

mayeul-zama (approved):
Thanks!

PR #3643, opened 2026-06-18
fix(gpu): make cuda_bind.rs stream/event bindings environment-indepent

pdroalves:
## Problem

`tfhe-cuda-common/build.rs` blocklisted the `CUstream_st` / `CUevent_st` structs but kept the `cudaStream_t` / `cudaEvent_t` typedefs that reference them. On machines where the CUDA headers resolve those structs (e.g. an H100 with `/usr/local/cuda/include` populated), bindgen regenerated `cuda_bind.rs` with `pub type cudaStream_t = *mut CUstream_st;` — a reference to a blocklisted (undefined) type — failing to compile (`E0412: cannot find type CUstream_st`) and breaking every build of `tfhe-cuda-common` and everything downstream (`tfhe-cuda-backend`, `tfhe` with `gpu`).

This blocklists the typedefs as well and provides them via `raw_line` as `*mut ffi::c_void`, so the handles resolve to a stable opaque form regardless of whether the CUDA headers are on the clang include path. The generated file is now byte-identical on every machine, and the `*mut c_void` form (the universal handle cast by `CudaStreams.ptr` and `zk-cuda-backend` at their FFI boundaries) is preserved.

---

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-common/build.rs
+++ b/backends/tfhe-cuda-common/build.rs
@@ -114,9 +114,13 @@ fn generate_cuda_bind_bindings(manifest_dir: &str, include_dir: &PathBuf) {
     if headers_modified > bindings_modified {
         let bindings = bindgen::Builder::default()
             .header(header_path.to_str().unwrap())
-            .allowlist_function("cuda_.*")
-            .blocklist_type("CUstream_st")
-            .blocklist_type("CUevent_st")
+            .allowlist_function(concat!(
+                "cuda_(create_stream_ffi|destroy_stream|synchronize_stream|",
+                "is_available|malloc|malloc_async|check_valid_malloc|",
+                "device_total_memory|memcpy_async_to_gpu|memcpy_async_gpu_to_gpu|",
+                "memcpy_gpu_to_gpu|memcpy_async_to_cpu|memset_async|",
+                "get_number_of_gpus|get_number_of_sms|synchronize_device|drop)",
+            ))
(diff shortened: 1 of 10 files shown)


guillermo-oyarzun (dismissed):
thanks for solving that annoying issue

andrei-stoian-zama (changes requested):
I think we should take a different approach:

- remove functions that are not called from Rust from extern C
- use the pattern we have in the tfhe backend: export functions with c_void / void* that need structs at the FFI. Use a wrapper which casts the void* to the right struct in a C++ backend function that is not exported. 
- avoid exceptions in `buid.rs` which is less maintanable

pdroalves:
@andrei-stoian-zama done. 

Removed from the binding:

```:
- cuda_get_device
- cuda_set_device
- cuda_create_event
- cuda_event_record
- cuda_stream_wait_event
- cuda_event_destroy
- cuda_malloc_with_size_tracking_async
- cuda_memcpy_with_size_tracking_async_to_gpu
- cuda_memcpy_with_size_tracking_async_gpu_to_gpu
- cuda_memset_with_size_tracking_async
- cuda_drop_with_size_tracking_async
- cuda_drop_async
````

none of these had rust callers.

The remaining functions had their `cudaStream_t` replaced by `void*` and a wrapper for C++ callers was added.

andrei-stoian-zama (approved):
Thanks for making the changes so fast !

PR #3658, opened 2026-06-26
feat(gpu): throughput oriented pbs

guillermo-oyarzun:

### Check-list:


backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
```diff
@@ -61,10 +65,15 @@ void cuda_convert_lwe_programmable_bootstrap_key_specialized_2_2_64_async(
                (size_t)(glwe_dim + 1), (size_t)level_count);
   // We will force the conversion that's why we set the boolean to true.
   constexpr bool use_specialized_fft_2_2 = true;
+  auto max_shared_memory = cuda_get_max_shared_memory(gpu_index);
+  bool use_throughput_oriented =
```
pdroalves, line 65: The formula to compute `use_throughput_oriented` repeats many times. You should encapsulate it.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
```diff
@@ -347,6 +356,134 @@ void cuda_fourier_polynomial_mul_async(void *stream_v, uint32_t gpu_index,
   cuda_drop_async(buffer, stream, gpu_index);
 }
 
+// Test-only entry point: negacyclic polynomial multiplication driven by the
+// throughput-oriented FFT16x4x16 cores (the FFT used by the specialized
+// 2_2_params PBS). Hardcoded to polynomial_size == 2048 and requires sm_90
```
pdroalves, line 360: What do you mean by "requires sm_90"? If it doesn't compile on other archs we need a preprocessor conditional clause here. Otherwise, we need a runtime conditional within the function to confirm the architecture.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
```diff
@@ -347,6 +356,134 @@ void cuda_fourier_polynomial_mul_async(void *stream_v, uint32_t gpu_index,
...
+    void const *_input2, void *_output, uint32_t polynomial_size,
+    uint32_t total_polynomials) {
+
+  if (polynomial_size != 2048)
+    PANIC("cuda_fourier_polynomial_mul_fft16x4x16_async only supports "
+          "polynomial_size == 2048");
+
+  auto stream = static_cast<cudaStream_t>(stream_v);
+  cuda_set_device(gpu_index);
+  auto input1 = (const double2 *)_input1;
```
pdroalves: We should be using `static_cast` everywhere to cast.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
```diff
@@ -347,6 +356,134 @@ void cuda_fourier_polynomial_mul_async(void *stream_v, uint32_t gpu_index,
...
+  check_cuda_error(cudaFuncSetCacheConfig(
+      batch_polynomial_mul_fft16x4x16<params>, cudaFuncCachePreferShared));
+
+  batch_polynomial_mul_fft16x4x16<params>
+      <<<gridSize, blockSize, shared_memory_size, stream>>>(input1, input2,
+                                                            output);
+  check_cuda_error(cudaGetLastError());
+}
+
+// Test-only entry point: forward-only negacyclic FFT using the classic
```
pdroalves, line 399: In which tests these methods are used?

pdroalves: I just noticed you added new tests to the C++ test suite. If these tests are only used there, perhaps it would be better to move them to the setup files there to not pollute this file.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -228,19 +226,41 @@ void cuda_convert_lwe_programmable_bootstrap_key(cudaStream_t stream,
...
       if (use_specialized_fft_2_2) {
-        blockSize = polynomial_size / choose_opt(polynomial_size);
-        check_cuda_error(
-            cudaFuncSetAttribute(batch_NSMFFT_classical_specialized<
-                                     FFTDegree<Degree<2048>, ForwardFFT>>,
-                                 cudaFuncAttributeMaxDynamicSharedMemorySize,
-                                 2 * shared_memory_size));
-        check_cuda_error(
-            cudaFuncSetCacheConfig(batch_NSMFFT_classical_specialized<
+        if (use_throughput_oriented) {
```
pdroalves, line 229: I feel weird these nested conditionals. Since `use_throughput_oriented` always depends on `use_specialized_fft_2_2`, maybe it would be better to rewrite this as

```c++
if(use_throughput_oriented) {
   ...
} else if (use_specialized_fft_2_2) {
   ...
}
   ...
}

File: backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh [hunk 6]
```diff
@@ -304,6 +304,45 @@ __device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_no_tbc(
...
+          uint32_t level_count>
+__device__ __forceinline__ void
+mul_ggsw_glwe_in_fourier_domain_2_2_params_fused_no_tbc(
+    double2 *fft, double2 *fft_regs, double2 *buffer_regs, double2 *other_fft,
+    const double2 *__restrict__ bootstrapping_key, int iteration) {
+  // Continues multiplying fft by every polynomial in that particular bsk level
+  // Each y-block accumulates in a different polynomial at each iteration
+  // We accumulate in registers to free shared memory
+  // In 2_2 params we only have one level
+  constexpr uint32_t level_id = 0;
```
pdroalves: Perhaps rename this variable to capital letters to highlight it's a constant?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -304,6 +304,45 @@ __device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_no_tbc(
...
+  auto self_bsk_slice =
+      get_ith_mask_kth_block_2_2_params<double2, polynomial_size,
+                                        glwe_dimension, level_count, level_id>(
+          bootstrapping_key, iteration, threadIdx.y);
+  auto self_bsk_poly = self_bsk_slice + threadIdx.y * polynomial_size / 2;
+
+  constexpr uint32_t glwe_id = 1;
```
pdroalves: Perhaps rename this variable to capital letters to highlight it's a constant?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -304,6 +304,45 @@ __device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_no_tbc(
...
 
+// We need a different version for the default specialized 2_2 params because
+// in that case we don't need to use dsm and then no cluster sync
+// We want to have the sync outside the function, so the product can leverage
+// the complex ops to overlap the bsk loads. This is an empirical observation
+// after trying both options.
+template <class params, uint32_t polynomial_size, uint32_t glwe_dimension,
+          uint32_t level_count>
+__device__ __forceinline__ void
+mul_ggsw_glwe_in_fourier_domain_2_2_params_fused_no_tbc(
```
pdroalves, line 317: This method expects a 2-dimension kernel. You must add a comment explaining that and what should be found in each dimension.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -27,13 +28,108 @@
...
+//   2. cross-y communication buffer for the GGSW pointwise multiply
+//      (other_fft_ptr in the GGSW helper);
+//   3. final uint64 accumulator on the last loop iteration, read by
+//      sample_extract_mask / sample_extract_body.
+// All three uses are sequenced by __syncthreads / mbarrier syncs in the
+// kernel; they never overlap in time.
+// ─────────────────────────────────────────────────────────────────────────────
+static constexpr int
+    SPECIALIZED_2_2_PARAMS_MIXPRECISION_SMEM_ACC_OFFSET_DOUBLES =
+        FFT16x4x16_DUAL_TWIST_OFFSET + 2 * 513;
```
pdroalves: Define other `constexpr` to avoid the magic numbers `2` and `513` perhaps?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap.cuh
```diff
@@ -304,6 +304,45 @@ __device__ void mul_ggsw_glwe_in_fourier_domain_2_2_params_classical_no_tbc(
...
+    double2 *fft, double2 *fft_regs, double2 *buffer_regs, double2 *other_fft,
```
pdroalves, line 318: You are not using `fft` here. You can remove it.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+
+  // Because our thread block is 2D, we need to use linear_tid to make a
+  // cooperative calculation of modswitch and the precalc coefs. Keep in mind
+  // that both glwe use the same rotation so modswitch is reusable. The same
+  // block also initialises precalc_coefs: 3 concatenated bands of 32 int8_t
+  // with values [-1, +1, -1] (one entry per 64-position chunk). Used by the
+  // compact precalc rotation helper to read the negacyclic sign each PBS
+  // iteration.
+  {
+    const int linear_tid = threadIdx.x + threadIdx.y * 64;
```
pdroalves: Why 64? two warps?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -27,13 +28,108 @@
 #include "polynomial/polynomial_math.cuh"
 #include "types/complex/operations.cuh"
 
+#ifdef __CUDACC_DEBUG__
+#define SPECIALIZED_2_2_PARAMS_THROUGHPUT_LAUNCH_BOUNDS
+#else
+#define SPECIALIZED_2_2_PARAMS_THROUGHPUT_LAUNCH_BOUNDS                        \
+  __launch_bounds__(128, 2)
```
pdroalves, line 39: Can you add a comment explaining 128 and 2? I already don't remember where they come from.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_tbc_classic.cuh
```diff
@@ -568,6 +569,68 @@ __host__ void host_programmable_bootstrap_tbc_with_mode(
...
+              max_shared_memory)) {
+        using mp_params = AccumulatorDegree<2048>;
+        int mp_thds = polynomial_size / mp_params::opt; // 64
+        dim3 mp_grid(input_lwe_ciphertext_count, 1, level_count);
+        dim3 mp_block(mp_thds, glwe_dimension + 1, 1); // (64, 2, 1)
+        uint64_t mp_smem =
+            get_buffer_size_full_sm_programmable_bootstrap_specialized_2_2_params_throughput<
+                Torus>(lwe_dimension);
+
+#define LAUNCH_TBC_SPECIALIZED_2_2_MIXPRECISION(BL)                            \
```
pdroalves: This is byte-identical to `LAUNCH_SPECIALIZED_2_2_MIXPRECISION`. We could generalize both on a single macro.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+  // block also initialises precalc_coefs: 3 concatenated bands of 32 int8_t
+  // with values [-1, +1, -1] (one entry per 64-position chunk). Used by the
+  // compact precalc rotation helper to read the negacyclic sign each PBS
+  // iteration.
+  {
+    const int linear_tid = threadIdx.x + threadIdx.y * 64;
+    for (int idx = linear_tid; idx < static_cast<int>(lwe_dimension);
+         idx += 128) {
+      Torus a_hat_full = 0;
+      modulus_switch(block_lwe_array_in[idx], a_hat_full, log_modulus);
```
pdroalves, line 501: Do you think we would gain anything by converting `modulus_switch` to a macro so we inline it here? We have it defined with `__forceinline__` but I'm not sure the effect is the same.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+  const double2 tw_cache_r3 = tw_shared[3 * 64 + threadIdx.x];
+  const double2 tw_cache_r11 = tw_shared[11 * 64 + threadIdx.x];
+  const double2 tw_cache_r1 = tw_shared[1 * 64 + threadIdx.x];
+
+  // ── Main PBS loop ──────────────────────────────────────────────────────────
+  for (int i = 0; i < lwe_dimension; i++) {
+    // Syncs only the warps participating on each fft
+    // Effectively means that only syncs threads with the same threadIdx.y
+    // that only takes values 0 or 1. So it syncs groups of 64 threads.
+    sync_coupled_warps();
```
pdroalves, line 528: I cannot say enough how much I loved this thing. You found a way to have a cluster within a single block.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+
+    FFT16x4x16_fwd_optimized_for_pbs(fft_out_regs, tw_shared, smem_twist,
+                                     smem_xpose, smem_xpose_pong, compact_w1,
+                                     compact_w2, compact_w3, tw_cache_r7,
+                                     tw_cache_r3, tw_cache_r11, tw_cache_r1);
+
+    // Copy fft from regs to shared memory to be used in the external product
+#pragma unroll
+    for (int j = 0; j < params::opt / 2; j++)
+      smem_comm[threadIdx.x + j * 64] = fft_out_regs[j];
```
pdroalves, line 568: You don't synchronize before this. I don't recall how many banks we have currently in the shared memory, but the last time I checked it was 16, and 64 is a multiple of it. Couldn't this cause a bank conflict if all threads arrive to this part together?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+
+    if (is_last_iter) {
+      // We store the output accumulation into shared memory, we are reusing
+      // regions so we must sync to make sure no thread is reading from it
+      // anymore.
+      __syncthreads();
+      constexpr int upshift = sizeof(Torus) * 8 - 32;
+#pragma unroll
+      for (int j = 0; j < params::opt / 2; j++) {
+        Torus torus_real = 0, torus_imag = 0;
```
pdroalves, line 603: Why do you initialize them? They are initialized by `typecast_double_round_to_torus`.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -223,6 +319,375 @@ device_programmable_bootstrap_specialized_2_2_params(
...
+            reg_acc_running[k];
+      }
+    }
+
+    // At the end of the iteration we need to swap buffers to use a ping pong
+    // strategy that save us extra syncs.
+    // Swap fwd <-> inv buffer roles for the next iteration so its forward FFT
+    // writes the set THIS y-group's inverse FFT just wrote (within-y WAW),
+    // which keeps the top-of-loop barrier a 64-thread named sync.
+    double *tmp_xpose = smem_xpose;
```
pdroalves: I think you can move all this block up to the end of the `if (is_last_iter) ` conditional.

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cu
```diff
@@ -17,7 +17,7 @@ const int THRESHOLD_MULTI_GPU_WITH_MULTI_BIT_PARAMS = 12;
 const int THRESHOLD_MULTI_GPU_WITH_CLASSICAL_PARAMS_U128 = 12;
 
 // Returns the threshold for multi-GPU with classical params.
-// Computed once based on GPU 0's compute capability and SM count.
+// Computed once based on GPU 0's SM count.
 // We are assuming that 2_2 params are going to be used.
 int get_threshold_multi_gpu_classical() {
```
pdroalves, line 22: I don't know if we want to change this method like that. What happens when we are not running the throughput-oriented version?

pdroalves (commented):
I checked only the files under `pbs/` as you requested. It's a long and dense PR so I want to do another careful read of the code but I'm sending the first batch of comments. Nothing serious.

I still see a lot of magic numbers hardcoded. I would prefer to avoid that at most so code becomes more readable.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
guillermo-oyarzun: no, i meant it is tuned for H100 and we wont get nice performance on other architectures, it compiles in other archs though so we don't need the clause, we just need to gate it out.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
guillermo-oyarzun: yup you are right, in fact it will never enter in the specialized_2_2 anymore, i changed it, but i want to keep the specialiazed implementation cause i have some plans for it in the near future ;)

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: well we can do it only for 2_2 params cause otherwise it is not as efficient as the tbc one, it is something :)

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: I don't need to synchronize before there because the threads that arrive will write on a "free" shared memory thanks to the ping pong buffers. I will add a comments that explains it better

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
guillermo-oyarzun: right i will change them

backends/tfhe-cuda-backend/cuda/src/utils/helper_multi_gpu.cu
guillermo-oyarzun: this one only affects classical, before we had 2 different thresholds, one for tbc and one for non-tbc, now we will only use the non-tbc one, since this replaces tbc.

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
guillermo-oyarzun: is already encapsulated in  specialized_2_2_use_throughput_oriented function

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
guillermo-oyarzun: I followed our pattern, the previous function is also used on cuda backend thats why i placed it here, the next one is used in both rust and cuda backend

guillermo-oyarzun: Indeed we probably could refactor them in a separated file, maybe in other PR we look for all those with that pattern?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: sure i added it, this is used to force the compiler to keep at least 2 blocks of 128 threads per SM. We need 128 threads cause each fft uses 64 threads and we want to keep both glwes on the same sm, so in total 128 threads

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: right, this 64 actually represents the blockDim.x,  i forgot to change the hardcoded formula, is just a linearlization to have a global thread id , since our block is 2D of size (64,2,1), and i want to have the 128 threads collaborating.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: not sure if i get it, with forceinline the modulus_switch code is embedded here at compilation time, don't see much benefit from the macro

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: it is just to be extra safe that all bits are zero before doing the casting, it is the same way we do it in add_to_torus

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: yup, you are right we don't need them for the last iteration

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_tbc_classic.cuh
guillermo-oyarzun: right, i fixed it, now we have a single one.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: sure, added!

guillermo-oyarzun (commented):
thanks! i replied all your comments

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
pdroalves: Maybe reframe the comment then and replace "requires sm_90" by "works better on sm_90".

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cu
pdroalves: Yeah, agreed. We could do this later. There are many things we could do to improve organization in the backend.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
```diff
@@ -27,13 +28,114 @@
...
+// Residency budget for the mixprecision specialized kernel: 114 KiB/block keeps
+// 2 blocks/SM on H100 (228 KiB carveout). Promoted from the old launch guard so
+// the applicability predicate and the launch path share one value.
+static constexpr uint64_t
+    SPECIALIZED_2_2_PARAMS_MIXPRECISION_SMEM_BUDGET_BYTES = 114ull * 1024ull;
+
+// True if the FFT16x4x16 mixprecision specialized kernel is applicable for the
+// given parameters.
+template <typename Torus>
+__host__ bool specialized_2_2_use_throughput_oriented(
```
bbarbakadze, line 127: Don't we need to check  compute capability here?

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16_twiddles.cuh
```diff
@@ -0,0 +1,1602 @@
...
+extern __device__ __constant__ double2 tw_1024[15][64];
+#endif
+
+// ============================================================================
+//  Backup: recompute twiddle table at runtime (overwrites hardcoded values)
+// ============================================================================
+#include <cmath>
+#include <cstdio>
+
+static inline void init_twiddle_tables() {
```
bbarbakadze, line 1036: I guess this implementations is fine in terms of precision, but later we can have a look to a different method of generating twiddles, which might be more precise.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
```diff
@@ -0,0 +1,834 @@
...
+// Inverse twiddles are conjugates of the forward ones (exp(+2*pi*i))
+static inline __device__ void mul_itwiddles_16(double2 *__restrict__ a, int tid,
+                                               const double2 *__restrict__ tw) {
+#pragma unroll
+  for (int j = 1; j < 16; j++) {
+    int r =
+        (((j & 1) << 3) | ((j & 2) << 1) | ((j & 4) >> 1) | ((j & 8) >> 3)) - 1;
+    a[j] = a[j] * conjugate(tw[r * 64 + tid]);
+  }
+}
```
bbarbakadze: I guess this function is not used anymore; maybe you used it during implementation process.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
```diff
@@ -0,0 +1,834 @@
...
+  a[13] = a[13] * w1;
+  a[2] = a[2] * w2;
+  a[6] = a[6] * w2;
+  a[10] = a[10] * w2;
+  a[14] = a[14] * w2;
+  a[3] = a[3] * w3;
+  a[7] = a[7] * w3;
+  a[11] = a[11] * w3;
+  a[15] = a[15] * w3;
+}
```
bbarbakadze: I could not find any use of this function in codebase.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
```diff
@@ -0,0 +1,834 @@
...
+  int tid = threadIdx.x;
+  FFT16(a);
+  mul_twiddles_16(a, tid, tw);
+  permute_16x64_mbarrier(a, smem_xpose, barrier);
+  FFT4x4(a);
+  mul_twiddles_4x4(a, compact_twiddles);
+  fft16x4x16_mbarrier_sync(barrier);
+  permute_4x4_mbarrier(a, smem_xpose, barrier);
+  FFT16(a);
+}
```
bbarbakadze, line 681: it is still used in BSK transform function but do we need to keep this and optimized one together?

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
```diff
@@ -0,0 +1,834 @@
...
+    int ri = (i & 12) | ((i & 1) << 1) | ((i & 2) >> 1);
+    smem_c[sb + i * 69] = a[ri];
+  }
+  sync_coupled_warps();
+#pragma unroll
+  for (int i = 0; i < 16; i++) {
+    a[i] = smem_c[lb + i];
+  }
+}
+
```
bbarbakadze: Can we add a bit more explanation about those hardcoded indexes.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
```diff
@@ -0,0 +1,834 @@
...
+  FFT2(a0, a2);
+  FFT2(a1, a3);
+  FFT2(a0, a1);
+  FFT2_mul_exp_1_4(a2, a3);
+}
+
+static inline __device__ void FFT4(double2 *a) { FFT4(a[0], a[1], a[2], a[3]); }
+
+// 16-point DFT on 16 registers (column pass for FFT16x4x16's outer stage)
+static inline __device__ void FFT16(double2 *a) {
```
bbarbakadze, line 154: maybe add split radix diagram somewhere, fft4 parts are easy to understand, 

it is hard to understand why specific twiddle is needed during intermediate step.

bbarbakadze (commented):
Amazing work. I could not find anything critical.  Just adding little bit more comments in low level parts would be good, also changed ordering polynomial in bsk and fft took me a while to confirm it was the same. maybe also comment this part.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
guillermo-oyarzun: Sure I will add it

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
guillermo-oyarzun: right, i changed it by the cached version then forgot to remove it.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
guillermo-oyarzun: indeed, it was again the first version that was later optimized, i removed it

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
guillermo-oyarzun: sure, i add it.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
guillermo-oyarzun: we kind of need this one, cause the other is tuned for the pbs flow, with the ping pong buffers and considering we allocate a lot more memory. If you are ok , i can keep this one, and evaluate the change in another branch.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16_twiddles.cuh
guillermo-oyarzun: sure! i was using the cuda intrinsics, but we can update those with your script on another branch

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic.cuh
guillermo-oyarzun: not really, since we are not using tbc anymore, checking that we have enough shared memory should be enough, since this is very memory hungry smaller gpus are filtered.

backends/tfhe-cuda-backend/cuda/src/fft/fft16x4x16.cuh
bbarbakadze: Yes sure, lets keep it.

bbarbakadze (dismissed):

bbarbakadze (approved):

PR #3672, opened 2026-06-30
feat(c_api): improve payload of error message

### LINKED ISSUE #3632: [C-API] Enhance panic/error handling: support String payloads

When consuming the `tfhe-rs` C-API through higher-level wrappers (such as Java FFM, Python, Go, etc.), debugging native library panics is difficult because the actual error message is often lost and replaced with a generic `"panic occurred"`.

In `tfhe/src/c_api/error.rs`, the `panic_payload_to_error` function currently downcasts the caught panic payload to a static string slice (`&str`):
```rust
payload
    .downcast_ref::<&str>()
    .map_or_else(|| b"panic occurred".to_vec(), |s| s.as_bytes().to_vec())
```

However, in modern Rust (especially Rust 2021 edition), many panic payloads are of type `String` instead of `&str` (e.g., panics originating from `unwrap()`, format strings, etc.). When a `String` payload is received, the current downcast to `&str` fails, and the C-API defaults to returning the generic `"panic occurred"` message, discarding the actual panic explanation.

We would like `panic_payload_to_error` to handle both `&str` and `String` payloads by checking for both types during the downcast:

```rust
fn panic_payload_to_error(payload: &Box<dyn Any + Send>) -> LastError {
    // Add a catch panic as technically the to_vec could fail
    let result = std::panic::catch_unwind(AssertUnwindSafe(|| {
        let msg = payload
            .downcast_ref::<&str>()
            .copied()
            .or_else(|| payload.downcast_ref::<String>().map(|s| s.as_str()))
            .unwrap_or("panic occurred");
        msg.as_bytes().to_vec()
    }));
    // ...
}
```

* **Benefits**:
  * **High Expressiveness**: Higher-level wrappers will receive the actual descriptive Rust panic error message (e.g., `called Result::unwrap() on an Err value: ...`).
  * **Zero Overhead**: This is a simple 4-line adjustment inside an existing FFI function that only runs when a panic has already occurred.
  * **Completely Safe**: It requires no panic hooks, no thread-local variables, and has zero impact on pure Rust consumers of `tfhe-rs`.

* **Setting a custom panic hook**: We considered defining a custom panic hook to capture the panic message, but that introduces unnecessary complexity (registering hooks, thread-local storage, etc.). Modifying the existing downcast logic inside `panic_payload_to_error` is much cleaner and fully sufficient.

**Additional context**

None.

tmontaigu:
Fixes https://github.com/zama-ai/tfhe-rs/issues/3632

### DIFF
--- a/tfhe/c_api_tests/test_high_level_error.c
+++ b/tfhe/c_api_tests/test_high_level_error.c
@@ -4,6 +4,9 @@
 #include <assert.h>
 #include <stdio.h>
 #include <string.h>
+
+const char *expected_panic_message = "The server key was not properly initialized.\nDid you forget to call `set_server_key` in the current thread ?";
+
 int main(void) {
   tfhe_error_disable_automatic_prints();
 
@@ -47,6 +50,7 @@ int main(void) {
   last_error = tfhe_error_get_last();
   assert(last_error != NULL);
   printf("Error message Received from tfhe-rs: '%s'\n", last_error);
+  assert(strcmp(last_error, expected_panic_message) == 0);
 
   // Destroy the ciphertexts
   fhe_uint128_destroy(lhs);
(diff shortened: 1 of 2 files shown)


IceTDrinker (approved):
thanks !

PR #3680, opened 2026-07-01
fix(zk): check c1 len in verify

nsarlin-zama:
add an additional length check in zk verify. c1 is the mask vector and d is lwe_dimension


IceTDrinker (approved):
Thanks !

pdroalves (approved):

PR #3682, opened 2026-07-01
chore: remove legacy compactPublicKey bench in wasm for Small/PBS_KS params

IceTDrinker:
- fixing the parsing would be more expensive than removing this bench which is now essentially useless

fixes the crash found here: https://github.com/zama-ai/tfhe-rs/actions/runs/28518365069/job/84536206498


SouchonTheo (dismissed):

IceTDrinker:
made a mistake while removing the bench

SouchonTheo (approved):

PR #3685, opened 2026-07-02
HPU backend fixes 

Bapt-Roux:
Summary

- Fix mmap ptr lifetime — replaced explicit munmap call with a Drop trait implementation, ensuring correct memory cleanup via Rust's drop glue
- Fix HpuDevice init — updated missed function calls left behind from a prior HPU backend refactor
- Fix DMA queue write guard — now reads and validates the max queue value before attempting to write it, preventing out-of-bounds writes
- Fix DMA queue PATH — added missing PATH entry required to update DMA queues

Chores

- Updated psi64.hpu with latest bitstream and firmware from main
- Extended the HPU bench GitHub workflow to run on a broader set of HPU targets

### DIFF
--- a/.github/workflows/benchmark_documentation.yml
+++ b/.github/workflows/benchmark_documentation.yml
@@ -138,8 +138,8 @@ jobs:
       op_flavor: default
       bench_type: both
       precisions_set: documentation
-      v80_pcie_dev: 24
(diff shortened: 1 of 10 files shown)


pgardratzama (dismissed):
bench from github workflow are running and manual tests look good

pgardratzama (approved):

PR #3702, opened 2026-07-07
feat(gpu): add re-randomization support to GPU OPRF bitonic shuffle

pdroalves:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1473

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cu
+++ b/backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cu
@@ -62,17 +62,24 @@ uint64_t scratch_cuda_integer_oprf_bitonic_shuffle_64_async(
     CudaLweKeyswitchKeyParamsFFI ksk_params, uint32_t key_num_radix_blocks,
     uint32_t data_num_radix_blocks, uint32_t num_values,
     uint32_t message_modulus, uint32_t carry_modulus, bool allocate_gpu_memory,
-    PBS_MS_REDUCTION_T noise_reduction_type) {
+    PBS_MS_REDUCTION_T noise_reduction_type, bool apply_rerand,
+    CudaLweKeyswitchKeyParamsFFI rerand_ksk_params, RERAND_MODE rerand_mode) {
 
   PUSH_RANGE("scratch oprf bitonic shuffle")
   int_radix_params params(bsk_params, ksk_params, message_modulus,
                           carry_modulus, noise_reduction_type);
 
+  int_radix_params rerand_params(
+      PBS_TYPE::CLASSICAL, 0, 0, rerand_ksk_params.input_lwe_dimension,
+      rerand_ksk_params.output_lwe_dimension, rerand_ksk_params.level_count,
+      rerand_ksk_params.base_log, 0, 0, 0, message_modulus, carry_modulus,
+      PBS_MS_REDUCTION_T::NO_REDUCTION);
+
   uint64_t ret = scratch_cuda_integer_oprf_bitonic_shuffle<uint64_t>(
       CudaStreams(streams),
       (int_oprf_bitonic_shuffle_buffer<uint64_t> **)mem_ptr,
       key_num_radix_blocks, data_num_radix_blocks, num_values, params,
-      allocate_gpu_memory);
+      apply_rerand, rerand_params, rerand_mode, allocate_gpu_memory);
   POP_RANGE()
   return ret;
 }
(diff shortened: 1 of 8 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cu
```diff
@@ -93,14 +100,57 @@ uint64_t scratch_cuda_integer_oprf_bitonic_shuffle_64_async(
...
+    host_oprf_bitonic_shuffle<uint64_t, params>(
+        cuda_streams, values, num_values,
+        static_cast<const uint64_t *>(seeded_lwe_input),
+        static_cast<const uint64_t *>(
+            lwe_flattened_encryptions_of_zero_compact_array_in),
+        reinterpret_cast<uint64_t *const *>(rerand_ksks), buf, oprf_bsks, bsks,
+        reinterpret_cast<uint64_t *const *>(ksks));
+  };
+
+  if (buf->applies_rerand()) {
```
andrei-stoian-zama: is this necessary here ? I remember this was added for re-rand. but looking at the re-rand code, it would seem the template param could be created using the switch just before the call to

```
  host_lwe_expand<Torus, params>(streams.stream(0), streams.gpu_index(0),
                                 expanded_zero_lwes, d_expand_jobs, num_lwes);
```
in  `host_rerand_inplace`

which would avoid duplication of this switch here and in rerand.cu ?

I'm not 100% sure, could you look into it ?

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
```diff
@@ -787,40 +788,59 @@ template <typename Torus>
...
-    int_radix_params params, bool allocate_gpu_memory) {
+    int_radix_params params, bool apply_rerand, int_radix_params rerand_params,
+    RERAND_MODE rerand_mode, bool allocate_gpu_memory) {
 
   uint64_t size_tracker = 0;
-  *mem_ptr = new int_oprf_bitonic_shuffle_buffer<Torus>(
-      streams, params, key_num_blocks, data_num_blocks, num_values,
-      allocate_gpu_memory, size_tracker);
+  if (apply_rerand) {
+    *mem_ptr = new int_oprf_bitonic_shuffle_buffer<Torus>(
```
andrei-stoian-zama, line 796: maybe assert that rerand_mode is sane here ?

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
```diff
@@ -787,40 +788,59 @@ template <typename Torus>
...
+        streams, params, rerand_params, key_num_blocks, data_num_blocks,
+        num_values, rerand_mode, allocate_gpu_memory, size_tracker);
+  } else {
+    *mem_ptr = new int_oprf_bitonic_shuffle_buffer<Torus>(
```
andrei-stoian-zama, line 800: what value should rerand_mode have ? maybe assert it ?

andrei-stoian-zama (changes requested):
look good overall! just a few remarks

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cu
pdroalves: That's a much better approach. We usually run the switch in the `.cu` file but in this case that's only increasing the compiling time.

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
pdroalves: `rerand_mode` has `RERAND_MODE` type. I think an assert would be redundant, don't you?

pdroalves:
@andrei-stoian-zama Done. Thanks!

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
andrei-stoian-zama: you're right, it's an enum

backends/tfhe-cuda-backend/cuda/src/integer/shuffle.cuh
```diff
@@ -804,23 +812,74 @@ __host__ uint64_t scratch_cuda_integer_oprf_bitonic_shuffle(
...
   host_integer_grouped_oprf<Torus>(
       streams, mem_ptr->keys_storage, seeded_lwe_input,
       num_values * key_num_blocks, mem_ptr->oprf_memory, oprf_bsks);
 
+  if (mem_ptr->applies_rerand()) {
+    auto *rerand_mem = mem_ptr->rerand_memory;
+    auto *keys_ptr = static_cast<Torus *>(mem_ptr->keys_storage->ptr);
+
+    switch (rerand_mem->params.big_lwe_dimension) {
+    case 256:
```
andrei-stoian-zama, line 841: see the other PR about this switch/case please

andrei-stoian-zama (approved):
This can be merged and in the other PR rebase and see if you can refactor the swithc/case in the shuffle rerand call as well.

PR #3703, opened 2026-07-08
fix(shortint): make sure non native power of 2 moduli LUTs work ok

### LINKED ISSUE #3559: Inconsistent encoding modulus between fill_accumulator_with_encoding and fill_many_lut_accumulator
Hi Zama team,

I noticed what looks like an inconsistency in the shortint accumulator construction code, and I would like to understand whether this is intentional.

In the current tfhe-rs codebase, fill_accumulator_with_encoding forces the output encoding modulus to native:

```
let output_encoding = ShortintEncoding {
    ciphertext_modulus: CiphertextModulus::new_native(),
    message_modulus: output_message_modulus,
    carry_modulus: output_carry_modulus,
    padding_bit: PaddingBit::Yes,
};
```
This is around line 101 in:
`tfhe/src/shortint/engine/mod.rs`

However, fill_many_lut_accumulator uses the accumulator ciphertext modulus instead:

```
let encoding = ShortintEncoding {
    ciphertext_modulus: accumulator.ciphertext_modulus(),
    message_modulus,
    carry_modulus,
    padding_bit: PaddingBit::Yes,
};
```
This is around line 183 in the same file:
`tfhe/src/shortint/engine/mod.rs`

So single-LUT accumulator generation appears to always encode LUT outputs in the native torus, while many-LUT accumulator generation encodes them using the accumulator modulus.

My questions are:

1. Is this difference intentional?
2. If yes, what is the reason why single-LUT and many-LUT accumulators use different encoding moduli?
3. For non-native ciphertext moduli, such as 2^63, should many-LUT accumulators also use CiphertextModulus::new_native() for consistency with the single-LUT path?
4. Are there correctness assumptions in PBS/blind rotation/sample extraction that require LUT accumulator plaintexts to be encoded in native torus even when the ciphertext modulus is non-native?

Context: while experimenting with a KS32 PBS path using ciphertext modulus 2^63 and an NTT-based blind rotation backend, single-LUT PBS behaved correctly, but many-LUT PBS only worked after changing fill_many_lut_accumulator to use CiphertextModulus::new_native() as well. I am not sure whether this indicates a real inconsistency, or whether the NTT path has an additional requirement that should be handled separately.

Thanks!

IceTDrinker:
- add non reg test for both single and many LUT

fixes https://github.com/zama-ai/tfhe-rs/issues/3559

### DIFF
--- a/tfhe/src/shortint/engine/mod.rs
+++ b/tfhe/src/shortint/engine/mod.rs
@@ -93,10 +93,12 @@ where
 {
     assert_eq!(accumulator.polynomial_size(), polynomial_size);
     assert_eq!(accumulator.glwe_size(), glwe_size);
+    // For now shortint does not really support non power of 2 moduli, so assert here to make sure
+    // the rest of the function makes sense
+    assert!(accumulator.ciphertext_modulus().is_power_of_two());
 
-    // NB: Following path will not go `power_of_two_scaling_to_native_torus`
-    // Thus keep value MSB aligned without considering real delta
-    // i.e force modulus to be native
+    // Power of 2 moduli are represented on the native torus to re-use the existing core code
+    // WARNING: do not forget to update fill_many_lut_accumulator in case of updates
     let output_encoding = ShortintEncoding {
         ciphertext_modulus: CiphertextModulus::new_native(),
         message_modulus: output_message_modulus,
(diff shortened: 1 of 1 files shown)


tfhe/src/shortint/engine/mod.rs
```diff
@@ -428,4 +439,42 @@ mod test {
...
+        assert!(!non_native_params.ciphertext_modulus.is_native_modulus());
+        let (cks, sks) = gen_keys(non_native_params);
+
+        for clear in 0..non_native_params.message_modulus.0 {
+            let enc = cks.encrypt(clear);
+
+            // Single lut case
+            {
+                let id_lut = sks.generate_lookup_table(|x| x);
+                let res = sks.apply_lookup_table(&enc, &id_lut);
```
mayeul-zama, line 460: It seems to me that `res` has `ciphertext_modulus = 2^63` but its lowest bits are not guaranteed to be 0

IceTDrinker: it's guaranteed by the core code tfhe/src/core_crypto/fft_impl/fft64/crypto/bootstrap.rs:367

IceTDrinker: I can add a check on the raw LWE if you want, but with f64 FFTs we would mostly see 53 bits of data in the high bits anyways

mayeul-zama: Indeed, my bad
Looks good to me

mayeul-zama (approved):
Thanks!

tfhe/src/shortint/engine/mod.rs
IceTDrinker: No worries, it's a legit question ! :)

PR #3743, opened 2026-07-22
chore(gpu): more tests on h100 on push

andrei-stoian-zama:
Run more tests more often on h100


enzodimaria (approved):
Everything looks good to me.

PR #3747, opened 2026-07-23
chore: update comparison table

IceTDrinker:


IceTDrinker:
@pgardratzama you may be missing Min/Max values in your tables, kept the previous one for HPU

@andrei-stoian-zama for now this is based on the updated multi bit tables it seems, is it the correct thing to showcase here or does it need another update ?

IceTDrinker:
will port to main once this is agreed upon/merged here

pgardratzama (approved):
Thanks a lot

PR #3748, opened 2026-07-23
chore(gpu): adding oprf+rerand to the long run tests

enzodimaria:
Adding oprf+rerand to the long run tests

### DIFF
--- a/tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
+++ b/tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
@@ -1,11 +1,16 @@
 use crate::core_crypto::commons::generators::DeterministicSeeder;
 use crate::core_crypto::gpu::{get_number_of_gpus, CudaStreams};
+use crate::integer::ciphertext::PrfReRandomizationContext;
 use crate::integer::gpu::ciphertext::boolean_value::CudaBooleanBlock;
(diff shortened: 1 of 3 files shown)


tfhe/src/integer/gpu/server_key/radix/tests_long_run/mod.rs
```diff
@@ -1800,3 +1805,278 @@ where
...
+        self.setup_keys(cks, sks, seeder);
+    }
+
+    fn execute(&mut self, input: (Seed, u64, u64, u64)) -> RadixCiphertext {
+        let context = self.context();
+        let oprf_key = context.oprf_key.as_ref().expect("OPRF key not set");
+        let rerand_key = self.rerand_key();
+        let rerand_context = PrfReRandomizationContext::default();
+
+        let excluded_upper_bound =
```
guillermo-oyarzun, line 1975: I am not that familiar with oprf+rerand, why in here we care about an upper bound and in the others no?

enzodimaria: It's for non power of 2 inputs
If you want a random between 0 and 2^b, the bound is already given by 'b', but if you want a random between 0 and N, N isn't a power of 2, so you need an explicit upper bound constant

guillermo-oyarzun (approved):

PR #3756, opened 2026-07-24
fix(gpu): optimize CI to parallelize on 2x gpu

andrei-stoian-zama:
1. Groups pairs of h100 tests: 3 new workflows to run in parallel on gpus 0 and 1 of a 2x gpu machine
- signed and unsigned integer 
  - classical
  - multi-bit
- hlapi and core

2. Moves classical tests to "PR on push" and multi-bit tests to approved
3.  One new h100 workflow: `gpu_h100_long_run_tests` -> on PR push and on CRON
4. Disables non 2x-gpu workflows: gpu_signed_integer_h100_tests and gpu_unsigned_integer_h100_tests -> replaced with the 2x gpu variant (multi-bit)

Bottom line: no cost savings but more tests

### DIFF
--- a/.github/workflows/gpu_core_h100_tests.yml
+++ b/.github/workflows/gpu_core_h100_tests.yml
@@ -20,11 +20,7 @@ env:
   EXTERNAL_CONTRIBUTION_RUNNER: "gpu_ubuntu-22.04"
 
 on:
-  # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
-  pull_request:
-  schedule:
-    - cron: "0 1 * * MON-FRI"
 
 permissions:
   contents: read
--- a/.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
+++ b/.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
@@ -1,5 +1,5 @@
-# Compile and test tfhe-cuda-backend on an H100 VM on hyperstack
-name: gpu_hlapi_h100_tests
+# Run signed and unsigned integer GPU classic tests in parallel on a 2xH100 machine
+name: gpu_integer_classic_tests_h100_2gpu_par
 
 env:
   CARGO_TERM_COLOR: always
@@ -15,12 +15,13 @@ env:
   IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
   PULL_REQUEST_MD_LINK: ""
   CHECKOUT_TOKEN: ${{ secrets.REPO_CHECKOUT_TOKEN || secrets.GITHUB_TOKEN }}
-  # Secrets will be available only to zama-ai organization members
   SECRETS_AVAILABLE: ${{ secrets.JOB_SECRET != '' }}
   EXTERNAL_CONTRIBUTION_RUNNER: "gpu_ubuntu-22.04"
 
 on:
   workflow_dispatch:
+  pull_request:
+    types: [ labeled ]
 
 permissions:
   contents: read
@@ -29,7 +30,7 @@ permissions:
 
 jobs:
   should-run:
-    name: gpu_hlapi_h100_tests/should-run
+    name: gpu_integer_classic_tests_h100_2gpu_par/should-run
     runs-on: ubuntu-latest
     permissions:
       pull-requests: read  # Needed to check for file change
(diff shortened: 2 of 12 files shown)


.github/workflows/gpu_core_h100_tests.yml
```diff
@@ -20,11 +20,7 @@ env:
   EXTERNAL_CONTRIBUTION_RUNNER: "gpu_ubuntu-22.04"
 
 on:
-  # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
-  pull_request:
-  schedule:
-    - cron: "0 1 * * MON-FRI"
```
enzodimaria, line 27: These tests aren't scheduled anymore then?

.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
```diff
@@ -0,0 +1,212 @@
...
+              - '.github/actions/gpu_setup/action.yml'
+              - ci/slab.toml
+            core_crypto:
+              - tfhe/src/core_crypto/gpu/**
+
+  setup-instance:
+    name: gpu_integer_classic_tests_h100_2gpu_par/setup-instance
+    needs: should-run
+    if: github.event_name != 'pull_request' ||
+      (github.event.action != 'labeled' && needs.should-run.outputs.core_crypto_changed == 'true') ||
```
enzodimaria: This condition seems unreachable.
Isn't `github.event.action != 'labeled'` always false because of line 24?

.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
```diff
@@ -0,0 +1,212 @@
...
+          CUDA_VISIBLE_DEVICES=1 BIG_TESTS_INSTANCE=TRUE make test_unsigned_integer_gpu_ci \
+            > /tmp/gpu1.log 2>&1 &
+          PID1=$!
+          wait $PID0; RC0=$?
+          wait $PID1; RC1=$?
+          echo "=== GPU 0: signed integer ==="
+          cat /tmp/gpu0.log
+          echo "=== GPU 1: unsigned integer ==="
+          cat /tmp/gpu1.log
+          [ $RC0 -eq 0 ] && [ $RC1 -eq 0 ]
```
enzodimaria: Are we sure that when a test is failing, we do print the details of the error?

.github/workflows/gpu_signed_integer_classic_tests.yml
```diff
@@ -23,7 +23,6 @@ on:
   # Allows you to run this workflow manually from the Actions tab as an alternative.
   workflow_dispatch:
   pull_request:
-    types: [ labeled ]
```
enzodimaria, line 26: Since this workflow is concatened with the unsigned one into `gpu_integer_classic_tests_h100_2gpu_par`, shouldn't we exclude it from the pull requests?

enzodimaria (changes requested):

enzodimaria:
Thanks for the PR @andrei-stoian-zama :100: 
I just have a few questions.

I put the 'approved' label to see how the new workflows work.

.github/workflows/gpu_core_h100_tests.yml
andrei-stoian-zama: yes, they're disabled. same tests will run in parallel in the 2par workflows

.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
andrei-stoian-zama: good point! I have not tested it no!

.github/workflows/gpu_signed_integer_classic_tests.yml
andrei-stoian-zama: I want both classical & MB to run unconditionally (see the PR message above)

enzodimaria: Indeed! So that was intented, perfect :+1:

.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
enzodimaria: It was the case: **fixed**

.github/workflows/gpu_integer_classic_tests_h100_2gpu_par.yml
enzodimaria: There was no printing infos: **fixed**

enzodimaria (approved):

PR #3764, opened 2026-07-27
chore: use better GPU params for 1_1 classic tuniform 2^-64 core bench

IceTDrinker:
- also fix bench param naming which contain trailing commas


guillermo-oyarzun (dismissed):

IceTDrinker:
updated to also manage the multi bit case @guillermo-oyarzun

guillermo-oyarzun (approved):

PR #3766, opened 2026-07-27
chore: update GPU bench data in docs

IceTDrinker:
target 1.7.x


tfhe/docs/.gitbook/assets/gpu-pbs-benchmark-tuniform-2m64.svg
IceTDrinker, line 1: looks like the same thing as for classical, 2^-64 params are actually using 2^-128 for GPU since they are specifically optimized

guillermo-oyarzun (approved):

PR #3767, opened 2026-07-28
chore(ci): Update test configuration for Hpu

Bapt-Roux:
Remove the clunky hlapi based regression and use native HPU tests instead

Aims of this is to check that nothing breaks the HPU interface; thus, all test are based on fast simulation model.

### DIFF
--- a/.github/workflows/hpu_tests.yml
+++ b/.github/workflows/hpu_tests.yml
@@ -1,5 +1,5 @@
-# Test HPU backend HLAPI layer
-name: hpu_hlapi_tests
+# Test HPU backend layer
+name: hpu_tests
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/hpu_tests.yml
```diff
@@ -68,7 +68,8 @@ jobs:
           toolchain: stable
           override: true
 
-      - name: Test HLAPI HPU
+      - name: Test HPU
         run: |
-          make HPU_CONFIG=sim test_high_level_api_hpu
+          make HPU_CONFIG=sim test_integer_hpu_sim_ci_fast
```
nsarlin-zama, line 73: it looks like this one is also running in `cargo_build_common.yml`, maybe we can remove it from cargo_build and keep it only here (hpu_tests) ?

Bapt-Roux: You are right; this was already raised some time ago by David.
I've removed it from the cargo_build_common workflow.

nsarlin-zama (dismissed):
Thanks!

Now we need some admin to update bpr, cc @IceTDrinker

IceTDrinker:
> Thanks!
> 
> Now we need some admin to update bpr, cc @IceTDrinker

Good to go !

nsarlin-zama (approved):

PR #3772, opened 2026-07-28
Ns/feat/otp transciphering

nsarlin-zama:

### PR content/description
Adds the OTP/OneTimePad transciphering method.

Here the idea is to directly generate, using the OPRF, enough keystream to be able to directly xor the full user input. The `next_keystream_bits` operation is a no-op. This is faster than the other transciphering options, since there is nothing to run in fhe except a xor, but one requirements is that the server/protocol needs to generate and decrypt enough bits to cover the full user input.

In this pr, I also updated the Transcipherer trait to allow to return an `KeystreamExhausted` error in `next_keystream_bits`. 
That way, the OTP cipher can cleanly error if the pad is smaller than the user input. I think it can be useful for other ciphers too, for example to indicate a maximum number of bits that we can generate for a given kreyvium key/iv.

This is breaking, but on a lower api layer, and on a trait that is not yet used downstream. I think it's worth the breakage, because it avoids a potential panic on a user controlled input.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772)

### DIFF
--- a/tfhe-benchmark/benches/transciphering/aes_transciphering.rs
+++ b/tfhe-benchmark/benches/transciphering/aes_transciphering.rs
@@ -96,7 +96,7 @@ pub fn cpu_aes_transciphering(c: &mut Criterion) {
             b.iter(|| {
                 let fhe_key = AesFheRoundKeys::new(&sks, &enc_key);
                 let mut stream = AesFheState::new(fhe_key, iv);
-                black_box(stream.next_keystream_bits(&sks, BLOCK_BITS));
+                black_box(stream.next_keystream_bits(&sks, BLOCK_BITS).unwrap());
             })
         },
     );
@@ -121,7 +121,7 @@ pub fn cpu_aes_transciphering(c: &mut Criterion) {
         |b| {
             b.iter(|| {
                 stream.seek(&sks, 0);
-                black_box(stream.next_keystream_bits(&sks, BLOCK_BITS));
+                black_box(stream.next_keystream_bits(&sks, BLOCK_BITS).unwrap());
             })
         },
     );
@@ -143,7 +143,7 @@ pub fn cpu_aes_transciphering(c: &mut Criterion) {
         |b| {
             b.iter(|| {
                 stream.seek(&sks, 0);
-                black_box(stream.next_keystream_bits(&sks, total_bits));
+                black_box(stream.next_keystream_bits(&sks, total_bits).unwrap());
             })
         },
     );
@@ -154,7 +154,7 @@ pub fn cpu_aes_transciphering(c: &mut Criterion) {
     // `seek(&sks, 0)` reset below.
     let total_bytes = BLOCK_BYTES * N_BLOCKS;
     let message = vec![0u8; total_bytes];
-    let sym_cipher = AesPlainState::new(key, iv).encrypt(&message);
+    let sym_cipher = AesPlainState::new(key, iv).encrypt(&message).unwrap();
 
     let benchmark_spec = BenchmarkSpec::<str>::new_transciphering(
         TranscipheringBench::Aes(AesFlavor::Transcipher16Blocks),
--- a/tfhe-benchmark/benches/transciphering/kreyvium_transciphering.rs
+++ b/tfhe-benchmark/benches/transciphering/kreyvium_transciphering.rs
@@ -104,7 +104,7 @@ pub fn cpu_kreyvium_transciphering(c: &mut Criterion) {
             b.iter_batched(
                 || warm_state.clone(),
                 |mut stream| {
-                    black_box(stream.next_keystream_bits(&sks, KEYSTREAM_BITS));
+                    black_box(stream.next_keystream_bits(&sks, KEYSTREAM_BITS).unwrap());
                 },
                 BatchSize::SmallInput,
             )
@@ -118,7 +118,7 @@ pub fn cpu_kreyvium_transciphering(c: &mut Criterion) {
             KreyviumPlainKey::from(key_bytes),
             KreyviumIV::from(iv_bytes),
         );
-        plain_stream.encrypt(&message)
+        plain_stream.encrypt(&message).unwrap()
     };
 
     let benchmark_spec = BenchmarkSpec::<str>::new_transciphering(
--- a/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_kreyvium.rs
+++ b/tfhe/src/integer/server_key/radix_parallel/tests_unsigned/test_kreyvium.rs
@@ -120,7 +120,7 @@ where
         let key_bool: [bool; 128] = std::array::from_fn(|i| key_bits[i] == 1);
         let iv_bool: [bool; 128] = std::array::from_fn(|i| iv_bits[i] == 1);
         let mut ref_kreyvium = KreyviumPlainState::new(key_bool, iv_bool);
-        let ref_bytes = ref_kreyvium.next_keystream_bits(num_steps);
+        let ref_bytes = ref_kreyvium.next_keystream_bits(num_steps).unwrap();
         let cpu_output = ref_bytes
             .iter()
             .flat_map(|&b| (0..8).map(move |i| (b >> i) & 1))
@@ -170,7 +170,7 @@ where
     let key_bool: [bool; 128] = std::array::from_fn(|i| key_bits[i] == 1);
     let iv_bool: [bool; 128] = std::array::from_fn(|i| iv_bits[i] == 1);
     let mut ref_kreyvium = KreyviumPlainState::new(key_bool, iv_bool);
-    let ref_bytes = ref_kreyvium.next_keystream_bits(total_steps);
+    let ref_bytes = ref_kreyvium.next_keystream_bits(total_steps).unwrap();
     let cpu_output = ref_bytes
         .iter()
         .flat_map(|&b| (0..8).map(move |i| (b >> i) & 1))
--- a/tfhe/src/shortint/client_key/mod.rs
+++ b/tfhe/src/shortint/client_key/mod.rs
@@ -179,6 +179,29 @@ impl<AP: EncryptionAtomicPattern> GenericClientKey<AP> {
         ShortintEngine::with_thread_local_mut(|engine| engine.encrypt(self, message))
     }
 
+    /// Encrypt a boolean message using the client key.
+    ///
+    /// The input message is converted to 1u64 if true, 0u64 otherwise.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use tfhe::shortint::parameters::PARAM_MESSAGE_2_CARRY_2_KS_PBS;
+    /// use tfhe::shortint::ClientKey;
+    ///
+    /// let cks = ClientKey::new(PARAM_MESSAGE_2_CARRY_2_KS_PBS);
+    ///
+    /// for msg in [false, true] {
+    ///     let ct = cks.encrypt_bool(msg);
+    ///     assert_eq!(ct.degree.get(), 1);
+    ///     let dec = cks.decrypt(&ct);
+    ///     assert_eq!(u64::from(msg), dec);
+    /// }
+    /// ```
+    pub fn encrypt_bool(&self, message: bool) -> Ciphertext {
+        ShortintEngine::with_thread_local_mut(|engine| engine.encrypt_bool(self, message))
+    }
+
     /// Encrypt a integer message using the client key returning a compressed ciphertext.
     ///
     /// The input message is reduced to the encrypted message space modulus
--- a/tfhe/src/shortint/engine/client_side.rs
+++ b/tfhe/src/shortint/engine/client_side.rs
@@ -80,6 +80,17 @@ impl ShortintEngine {
         )
     }
 
+    pub fn encrypt_bool<AP: EncryptionAtomicPattern>(
+        &mut self,
+        client_key: &GenericClientKey<AP>,
+        message: bool,
+    ) -> Ciphertext {
+        let mut ct = self.encrypt(client_key, message.into());
+        ct.degree = Degree::new(1);
+
+        ct
+    }
+
     pub fn encrypt_compressed<AP: EncryptionAtomicPattern>(
         &mut self,
         client_key: &GenericClientKey<AP>,
--- a/tfhe/src/transciphering/ciphers/aes/fhe.rs
+++ b/tfhe/src/transciphering/ciphers/aes/fhe.rs
@@ -1,6 +1,6 @@
 use crate::shortint::{Ciphertext, ServerKey};
 use crate::transciphering::ciphers::aes::AesIv;
-use crate::transciphering::{FheKeyStream, StreamCipherKind, Transcipherer};
+use crate::transciphering::{FheKeyStream, InsufficientKeystream, StreamCipherKind, Transcipherer};
 use rayon::prelude::*;
 
 use super::encrypt::encrypt_block;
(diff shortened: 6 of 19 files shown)


nsarlin-zama:
> ⚠️ **Backward-compat snapshot: suspicious changes detected**
> 
> Suspicious changes were detected that may hide breaking changes. Please review the warnings below carefully.
> ⚠️ Warnings
> 
> 
> If you encounter any errors or have doubts, you can verify locally by running:
> 
> ```
> make backward_correctness BASE_REF=<base_branch_or_commit>
> ```
> 
> Where `BASE_REF` is the reference branch or commit to check against.

This is ok since this is just a variant added at the end of the enum

IceTDrinker (commented):
> @IceTDrinker reviewed 19 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772)* status: all files reviewed, 5 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/transciphering/mod.rs` line 157 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-Oyi_R0cBujiKOJrImkw:-Oyi_R0cBujiKOJrImkx:bpvm9p8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/mod.rs#L157)):*
> ```XML
>     },
> 
>     KeystreamExhausted,
> ```

nit just wondering if it's really an exhausted keystream vs. keystream with insufficient remaining capacity for the particular request but still with capacity for a smaller request ?

e.g. the test for the error has 16 bits available IIRC and the request is 17, while the 16 are not consumed, the stream says "exhausted" but repeating the request for 16 bits would yield the 16 bits that are still available ?

https://translate.google.com/?sl=en&tl=fr&text=exhausted%20resource&op=translate
___
*[`tfhe/src/transciphering/ciphers/aes/fhe.rs` line 64 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiEVPSC4BAFCQexHXI:-OyiEVPSC4BAFCQexHXJ:b-ilgn6b) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/aes/fhe.rs#L64)):*
> ```XML
>         let blocks: Vec<[Ciphertext; 128]> = (0..n_blocks as u64)
>             .into_par_iter()
>             .map(|i| self.keystream_block(sks, (start_block + i) as u128))
> ```

I think there may be a slight bug we missed, if start_block and i are on u64 the sum could overflow, while the counter would be valid if the addition was made with two u128s ?

do we have a test at the limits for this one ?

If so could we have a follow up issue ?

looks like the plain code uses u128
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/mod.rs` line 6 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiNJig96oiCQBLmAKO:-OyiNJig96oiCQBLmAKP:bf63lhb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/mod.rs#L6)):*
> ```XML
> //! so the only FHE operation is the XOR with the input data.
> //! The pad is finite and single-use, and must be at least as long as the data. As a
> //! result, this cipher is worthwhile when decrypting a pad as large as the data is
> ```

last sentence is confusing to me
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/test.rs` line 224 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiaVEL2o4zMHw1XhQ6:-OyiaVEL2o4zMHw1XhQ7:bnz79y7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/test.rs#L224)):*
> ```XML
> /// FHE keystream == plain keystream for every (start offset, length) pair.
> ///
> /// `bit_count` is deliberately not a multiple of 8 so that
> ```

docstring is wrong :upside_down_face: 

lower:

    let max_bit_count = byte_count * 8;

I think I changed that test to exhaustive later on
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/test.rs` line 250 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyibCav3QQ73-Fs4xWT:-OyibCav3QQ73-Fs4xWU:bamkkjx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/test.rs#L250)):*
> ```XML
>         for start in 0..max_bit_count {
>             for output_bit_count in 0..=(max_bit_count - start) {
>                 plain_otp.seek(start as u64);
> ```

thinking about seek, since things changed, either let's make it a test util private to the lib or just get rid of it altogether ?

nsarlin-zama (commented):
> @nsarlin-zama made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772)* status: 9 of 19 files reviewed, 5 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/transciphering/mod.rs` line 157 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-Oyi_R0cBujiKOJrImkw:-Oym21q_4OtFKCLr1JwV:b-6qjvs) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/mod.rs#L157)):*

nit just wondering if it's really an exhausted keystream vs. keystream with insufficient remaining capacity for the particular request but still with capacity for a smaller request ?

e.g. the test for the error has 16 bits available IIRC and the request is 17, while the 16 are not consumed, the stream says "exhausted" but repeating the request for 16 bits would yield the 16 bits that are still available ?

https://translate.google.com/?sl=en&tl=fr&text=exhausted%20resource&op=translate
</blockquote></details>

Yes you are right, if you have an idea for a better name I'm taking it. Maybe InsufficientKeystream? Ideally it should cover other cases than just the otp.

For example, with kreyvium, at some point we were talking about setting a maximum number of bits that can be drawn securely, so we could raise this error too if we go past that.
___
*[`tfhe/src/transciphering/ciphers/aes/fhe.rs` line 64 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiEVPSC4BAFCQexHXI:-OymErzSAZDipSTMli2J:b-7mi1u1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/aes/fhe.rs#L64)):*

I think there may be a slight bug we missed, if start_block and i are on u64 the sum could overflow, while the counter would be valid if the addition was made with two u128s ?

do we have a test at the limits for this one ?

If so could we have a follow up issue ?

looks like the plain code uses u128
</blockquote></details>

start_block is `counter / 128`, counter is u64 so this cannot overflow ?

Added a test at the limits
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/mod.rs` line 6 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiNJig96oiCQBLmAKO:-Oym5Gwu5438YdmzWWGV:b-fd5kni) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/mod.rs#L6)):*

last sentence is confusing to me
</blockquote></details>

There are 3 available ciphers (aes, kryvium, otp). What I'd like is that each cipher explains why/when it should be chosen over the other ones, since I think it's the main information that will matter to someone reading this doc.

Here what I mean is that otp is cheaper for the transciphering part than the other ones, and should be chosen when generating and decrypting enough bits to cover the input is doable.
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/test.rs` line 224 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiaVEL2o4zMHw1XhQ6:-Oymrtsy3oz7EhpM5wkD:b-qb1vtp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/test.rs#L224)):*

docstring is wrong :upside_down_face: 

lower:

    let max_bit_count = byte_count * 8;

I think I changed that test to exhaustive later on
</blockquote></details>

yes indeed
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/test.rs` line 250 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyibCav3QQ73-Fs4xWT:-Oymub0J1Lfrj2Pmtq3W:b-16hlup) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/test.rs#L250)):*

thinking about seek, since things changed, either let's make it a test util private to the lib or just get rid of it altogether ?
</blockquote></details>

It was required at some point by the spec, and rust crypto also provides it: https://docs.rs/cipher/latest/cipher/stream/trait.StreamCipherSeek.html#method.seek.

It can be useful in real life scenarii:
- if a client generates an input that is not used/thrown away, server needs to seek forward. In that case it's actually a security requirement, since the client cannot reuse the stream for another input.
- IIRC there was a need to seek backward, to re-generate the stream in case of blockchain reorg. I think this assumed that the keystream was not cached by the coproc and that it would only be used on the exact same input

all of that is not needed if we don't keep stream cipher states but it might change again, and the current trait is designed with this possibility in mind

IceTDrinker (commented):
> @IceTDrinker reviewed 10 files and all commit messages, made 5 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772)* status: all files reviewed, 2 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/transciphering/mod.rs` line 157 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-Oyi_R0cBujiKOJrImkw:-Oyn0HLg2CdZxgoKu2Mp:b73por8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/mod.rs#L157)):*

Yes you are right, if you have an idea for a better name I'm taking it. Maybe InsufficientKeystream? Ideally it should cover other cases than just the otp.

For example, with kreyvium, at some point we were talking about setting a maximum number of bits that can be drawn securely, so we could raise this error too if we go past that.
</blockquote></details>

InsufficientKeyStream seems ok to me, it would also trigger in the exhausted case, for now I don't think we want to make a difference between both ?
___
*[`tfhe/src/transciphering/ciphers/aes/fhe.rs` line 64 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiEVPSC4BAFCQexHXI:-Oyn0fJF9XulwFNjVaqi:ba7m1n3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/aes/fhe.rs#L64)):*

start_block is `counter / 128`, counter is u64 so this cannot overflow ?

Added a test at the limits
</blockquote></details>

indeed, my bad
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/mod.rs` line 6 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyiNJig96oiCQBLmAKO:-Oyn13864SsvUdk2ruTp:bo5tori) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/mod.rs#L6)):*

There are 3 available ciphers (aes, kryvium, otp). What I'd like is that each cipher explains why/when it should be chosen over the other ones, since I think it's the main information that will matter to someone reading this doc.

Here what I mean is that otp is cheaper for the transciphering part than the other ones, and should be chosen when generating and decrypting enough bits to cover the input is doable.
</blockquote></details>

finally got the as large as possible part, was my bad
___
*[`tfhe/src/transciphering/ciphers/one_time_pad/test.rs` line 250 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-OyibCav3QQ73-Fs4xWT:-Oyn1Ry16T2nLnWx2CI3:b32wtd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0e5ef441fa07a58faf07274745c1b9a50cb56526/tfhe/src/transciphering/ciphers/one_time_pad/test.rs#L250)):*

It was required at some point by the spec, and rust crypto also provides it: https://docs.rs/cipher/latest/cipher/stream/trait.StreamCipherSeek.html#method.seek.

It can be useful in real life scenarii:
- if a client generates an input that is not used/thrown away, server needs to seek forward. In that case it's actually a security requirement, since the client cannot reuse the stream for another input.
- IIRC there was a need to seek backward, to re-generate the stream in case of blockchain reorg. I think this assumed that the keystream was not cached by the coproc and that it would only be used on the exact same input

all of that is not needed if we don't keep stream cipher states but it might change again, and the current trait is designed with this possibility in mind
</blockquote></details>

ok
___
*[`tfhe/src/transciphering/ciphers/kreyvium/test.rs` line 341 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-Oyn4X5LB7uuJZ4Tv3FZ:-Oyn4X5LB7uuJZ4Tv3F_:b-jghaiz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/50348f955102f0c901951a8c999a91b4869d5eb2/tfhe/src/transciphering/ciphers/kreyvium/test.rs#L341)):*
> ```XML
>     ));
>     assert_eq!(stream.current_counter(), start);
> }
> ```

nit sorry it seems it's the same sort of test as for aes at the limits, any reason for not having the same approach here ?

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/transciphering/ciphers/kreyvium/test.rs` line 341 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-Oyn4X5LB7uuJZ4Tv3FZ:-Oyn8mlsCbgPx_FJZo4x:b36sfz3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/50348f955102f0c901951a8c999a91b4869d5eb2/tfhe/src/transciphering/ciphers/kreyvium/test.rs#L341)):*

nit sorry it seems it's the same sort of test as for aes at the limits, any reason for not having the same approach here ?
</blockquote></details>

none that I see, modified to make it more similar to aes.
Only diff is that seeking near u64::MAX is not free for kreyvium, so we "cheat" by creating a dummy state with the required counter, since we are only interested in counter manipulations.

IceTDrinker (approved):
Thanks !



> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3772#-:-OynUCz_49ZTjPCI3WmH:b-cf187s)* status: all files reviewed, 1 unresolved discussion (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, SouchonTheo, and tmontaigu).

PR #3773, opened 2026-07-29
fix(eeprom): supporting I2C failure

lsainati:
Adding support of hwmod "-1" so that we can continue using boards without sensors due to I2C failure + debug message


Bapt-Roux (dismissed):

pgardratzama (dismissed):
Good to go

Bapt-Roux (approved):

pgardratzama (approved):

PR #3778, opened 2026-07-31
fix(gpu): scalar comparison bug due to pointer aliasing on two streams

andrei-stoian-zama:
Fixes a synchronization bug in scalar div:

Two streams are launched, but they work on aliased buffers (overlap of some blocks in a radix ciphertext). The fix is to ensure is no overlapping (aliasing). The common blocks were aliased by mistake.

Also adds many sanity checks.

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/aes/aes.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/aes/aes.cuh
@@ -722,15 +722,13 @@ __host__ void vectorized_mix_columns(CudaStreams streams,
     CudaRadixCiphertextFFI *col_copy_buffer =
         mem->round_workspaces->mix_columns_col_copy_buffer;
     for (uint32_t i = 0; i < BITS_PER_COLUMN; ++i) {
-      CudaRadixCiphertextFFI dest_slice, src_slice;
+      CudaRadixCiphertextFFI dest_slice;
       as_radix_ciphertext_slice<Torus>(&dest_slice, col_copy_buffer,
                                        i * num_aes_inputs,
                                        (i + 1) * num_aes_inputs);
-      as_radix_ciphertext_slice<Torus>(
-          &src_slice, s_bits, (col * BITS_PER_COLUMN + i) * num_aes_inputs,
-          (col * BITS_PER_COLUMN + i + 1) * num_aes_inputs);
-      copy_radix_ciphertext_async<Torus>(
-          streams.stream(0), streams.gpu_index(0), &dest_slice, &src_slice);
+      copy_radix_ciphertext_async<Torus>(streams.stream(0),
+                                         streams.gpu_index(0), &dest_slice,
+                                         &s_bits[col * BITS_PER_COLUMN + i]);
     }
 
     CudaRadixCiphertextFFI b_orig[BYTES_PER_COLUMN][BITS_PER_BYTE];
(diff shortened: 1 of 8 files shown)


guillermo-oyarzun (dismissed):

backends/tfhe-cuda-backend/cuda/src/integer/comparison.cuh
```diff
@@ -447,8 +447,11 @@ compare_radix_blocks(CudaStreams streams, CudaRadixCiphertextFFI *lwe_array_out,
   // Add one
   // Here Lhs can have the following values: (-1) % (message modulus * carry
   // modulus), 0, 1 So the output values after the addition will be: 0, 1, 2
-  host_add_scalar_one_inplace<Torus>(streams, lwe_array_out, message_modulus,
-                                     carry_modulus);
+  CudaRadixCiphertextFFI lwe_array_out_view;
+  as_radix_ciphertext_slice<Torus>(&lwe_array_out_view, lwe_array_out, 0,
+                                   num_radix_blocks);
+  host_add_scalar_one_inplace<Torus>(streams, &lwe_array_out_view,
```
andrei-stoian-zama, line 453: we didn't have a bug in comparisons but we're adding this for additional safety

backends/tfhe-cuda-backend/cuda/src/integer/scalar_comparison.cuh
```diff
@@ -72,14 +72,14 @@ __host__ void scalar_compare_radix_blocks(
...
-  // FIXME: without this sync signed scalar eq tests fail, I don't understand
-  // the reason
-  cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
   // Add one
   // Here Lhs can have the following values: (-1) % (message modulus * carry
   // modulus), 0, 1 So the output values after the addition will be: 0, 1, 2
-  host_add_scalar_one_inplace<Torus>(streams, lwe_array_out, message_modulus,
-                                     carry_modulus);
+  CudaRadixCiphertextFFI lwe_array_out_view;
+  as_radix_ciphertext_slice<Torus>(&lwe_array_out_view, lwe_array_out, 0,
```
andrei-stoian-zama, line 79: `scalar_compare_radix_blocks` adds 1 to all blocks of `lwe_array_out`. `lwe_array_out` is a radix ciphertext of the operation input `num_radix_blocks` but only `num_lsb_radix_blocks` blocks should be processed by `scalar_compare_radix_blocks`. the `host_add_scalar_one_inplace` wasn't using the right number of blocks (taking it from the ciphertext instead of the using the subset that was indicated to `scalar_compare_radix_blocks`, its caller.

backends/tfhe-cuda-backend/cuda/src/integer/scalar_comparison.cuh
```diff
@@ -181,11 +182,11 @@ __host__ void integer_radix_unsigned_scalar_difference_check(
     as_radix_ciphertext_slice<Torus>(&rhs, lhs, num_radix_blocks / 2,
                                      lhs->num_radix_blocks);
 
-    pack_blocks<Torus>(lsb_streams.stream(0), streams.gpu_index(0), lhs,
+    pack_blocks<Torus>(lsb_streams.stream(0), lsb_streams.gpu_index(0), lhs,
```
andrei-stoian-zama, line 185: just a minor cosmetic correction, packing is not multi-gpu so gpu 0 is used anyway

backends/tfhe-cuda-backend/cuda/src/aes/aes.cuh
```diff
@@ -722,15 +722,13 @@ __host__ void vectorized_mix_columns(CudaStreams streams,
     CudaRadixCiphertextFFI *col_copy_buffer =
         mem->round_workspaces->mix_columns_col_copy_buffer;
     for (uint32_t i = 0; i < BITS_PER_COLUMN; ++i) {
-      CudaRadixCiphertextFFI dest_slice, src_slice;
+      CudaRadixCiphertextFFI dest_slice;
```
andrei-stoian-zama, line 725: fixed because the new asserts were raised here

guillermo-oyarzun (approved):

PR #3822, opened 2026-08-13
fix(gpu): fixed setup of throughput oriented pbs

guillermo-oyarzun:

### PR content/description
Fixes issue when running multiple streams with different params each (like in the tests). In that case they were trying to setup different shared memory sizes based on the lwe_dimension.  This fix setups the shared memory only once and always up to a max of lwe_dimension =1024, that leaves margin for increasing params if need in future and covers all current potential calls to the throughput oriented pbs. 

Actually it was a problem when combining decompression params with regular pbs params too.
### Check-list:


enzodimaria (approved):
Thank you Guillermo!

PR #3825, opened 2026-08-13
fix(gpu): fix f128 FFT buffer leak and sizing, remove unused as_integer

pdroalves:
> **Warning:** removing the unused `as_integer` f128 FFT entry point is a breaking change to the public `tfhe-cuda-backend` API. This requires bumping the crate version from `0.15.0` to `0.16.0` (a minor bump, since `tfhe-cuda-backend` is pre-1.0 and breaking changes bump the second component). `Cargo.lock` and the `tfhe-cuda-backend` dependency in `tfhe/Cargo.toml` are updated accordingly. `scripts/check_scratch_cleanup.py`'s expected binding counts are also updated to match the removed bindings.

This PR fixes memory management issues in the f128 FFT host functions and in the 128-bit BSK conversion, found while investigating a memory leak in `host_fourier_transform_forward_as_torus_f128`. Following the review discussion, the unused `as_integer` f128 FFT entry point is removed instead of being fixed.

### Removed the unused `as_integer` f128 FFT path
`cuda_fourier_transform_forward_as_integer_f128_async` had the same leak and the same undersized scratch allocation as the `as_torus` variants, but it has no caller and no test: it was only exposed to `core_crypto` through the `fourier_transform_forward_as_integer_f128_async` wrapper, which nothing calls. It is deleted rather than fixed, along with everything it was the only user of: the `extern "C"` entry point and its declaration in `fft128.h`, `host_fourier_transform_forward_as_integer_f128`, the `convert_u128_to_f128_as_integer` and `batch_convert_u128_to_f128_as_integer` kernels, the generated Rust binding, and the `core_crypto` wrapper.

### Memory leak in the f128 FFT host functions
`host_fourier_transform_forward_as_torus_f128` and `host_fourier_transform_backward_as_torus_f128` allocated a global scratch `buffer` for the FFT kernel but never freed it. The cleanup block only released `d_standard` and `d_re0/d_re1/d_im0/d_im1`.

### Scratch buffer allocated in elements instead of bytes
The same `buffer` was sized with `safe_mul(number_of_samples, N/2, 4)`, but `cuda_malloc_async` takes bytes, so the allocation was 8x too small (missing the `sizeof(double)` factor). Any call taking the no-shared-memory (NOSM) kernel path wrote out of bounds. The same defect existed in `convert_u128_to_f128_and_forward_fft_128` (`bootstrapping_key.cuh`), which is used by the 128-bit BSK conversion.

### Device buffers not sized for batched calls
The two `as_torus` FFT host functions launch `number_of_samples` kernel blocks, each indexing its own sample slice, but all device buffers and host/device copies were sized for a single sample. Calls with `number_of_samples > 1` wrote out of bounds. Buffers and copies are now sized for the whole batch. (The only callers, the GPU FFT tests, pass 1, so this was latent.)

### Smaller cleanups
- The global scratch is now allocated only on the NOSM path instead of issuing a 0-byte allocation on the shared-memory path; `cuda_convert_lwe_programmable_bootstrap_key` also initializes its scratch pointer to `nullptr` instead of a 0-byte allocation whose pointer was overwritten (and leaked) by the NOSM branches.
- `cuda_drop_with_size_tracking_async` is now an explicit no-op for null pointers, since `cudaFreeAsync` (unlike `cudaFree`) is not documented to accept them.
- Added missing `check_cuda_error(cudaGetLastError())` after kernel launches in the two `as_torus` FFT variants, and replaced C-style casts with `static_cast` in the touched code.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
+++ b/backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
@@ -487,27 +487,6 @@ negacyclic_backward_fft_f128_tbc(double *dt_re_hi, double *dt_re_lo,
 
 // params is expected to be full degree not half degree
 template <class params>
-__device__ void convert_u128_to_f128_as_integer(
-    double *out_re_hi, double *out_re_lo, double *out_im_hi, double *out_im_lo,
-    const __uint128_t *in_re, const __uint128_t *in_im) {
-
-  Index tid = threadIdx.x;
-  // #pragma unroll
-  for (Index i = 0; i < params::opt / 2; i++) {
-    auto out_re = u128_to_signed_to_f128(in_re[tid]);
-    auto out_im = u128_to_signed_to_f128(in_im[tid]);
-
-    out_re_hi[tid] = out_re.hi;
-    out_re_lo[tid] = out_re.lo;
-    out_im_hi[tid] = out_im.hi;
-    out_im_lo[tid] = out_im.lo;
-
-    tid += params::degree / params::opt;
-  }
-}
-
-// params is expected to be full degree not half degree
-template <class params>
 __device__ void convert_u128_to_f128_as_torus(
     double *out_re_hi, double *out_re_lo, double *out_im_hi, double *out_im_lo,
     const __uint128_t *in_re, const __uint128_t *in_im) {
@@ -570,22 +549,6 @@ convert_f128_to_u128_as_torus(__uint128_t *out_re, __uint128_t *out_im,
 
     tid += params::degree / params::opt;
   }
-}
-
-// params is expected to be full degree not half degree
-template <class params>
-__global__ void
-batch_convert_u128_to_f128_as_integer(double *out_re_hi, double *out_re_lo,
-                                      double *out_im_hi, double *out_im_lo,
-                                      const __uint128_t *in) {
-
-  convert_u128_to_f128_as_integer<params>(
-      &out_re_hi[blockIdx.x * params::degree / 2],
-      &out_re_lo[blockIdx.x * params::degree / 2],
-      &out_im_hi[blockIdx.x * params::degree / 2],
-      &out_im_lo[blockIdx.x * params::degree / 2],
-      &in[blockIdx.x * params::degree],
-      &in[blockIdx.x * params::degree + params::degree / 2]);
 }
 
 // params is expected to be full degree not half degree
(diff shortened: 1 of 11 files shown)


backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
guillermo-oyarzun: actually this function seems like dead code, probably that's why we didn't notice the problem, better to remove it in my opinion, unless @bbarbakadze left it there for some reason?

backends/tfhe-cuda-backend/cuda/src/pbs/bootstrapping_key.cuh
```diff
@@ -164,7 +164,9 @@ void cuda_convert_lwe_programmable_bootstrap_key(
 
   auto max_shared_memory = cuda_get_max_shared_memory(gpu_index);
 
-  double2 *buffer = (double2 *)cuda_malloc_async(0, stream, gpu_index);
+  // global scratch is only allocated on the NOSM paths below; cuda_drop_async
+  // accepts the nullptr left by the FULLSM paths
+  double2 *buffer = nullptr;
```
guillermo-oyarzun, line 169: i guess this buffer thing here was not much of a problem in practice cause we don't use the NOSM flavor, and in the worse case we were allocating a 0 size. Your version seems cleaner

guillermo-oyarzun (changes requested):
Good tricky catches, if we end up removing the host calls, then i would suggest to squash the commits.

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
bbarbakadze: We are not using it anywhere so I guess we can remove it.

pdroalves: I think it's used in the C++ tests. Let me double check.

pdroalves: So, that method is only used in `cuda_fourier_transform_forward_as_integer_f128_async`, which is a method we expose to `core_crypto` to run the FFT. We don't use it anywhere, but I don't know if we shouldn't just keep it in case we want someday.

andrei-stoian-zama: in general we should keep a function:
- if it's used (thus tested indirectly) or
- if it's tested directly
if it's not used and not tested we can remove it.

I understand this function is not in the first case, is this function in the second case ?

pdroalves: No, it's not used or tested.

I will refactor this PR to remove this instead of fixing the leak.

pdroalves:
@guillermo-oyarzun @andrei-stoian-zama I just refactored this PR. Now we remove the `as_integer` path that's not tested or used. I kept the other functions and fixed the leaks since they have related tests.

andrei-stoian-zama (dismissed):

guillermo-oyarzun (approved):

PR #3833, opened 2026-08-17
refactor(hlapi): default to AES-256 for XofSeed

tmontaigu:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833)


nsarlin-zama (commented):
I'm wondering if we shouldn't accept having api breaking changes here, and try to refacto a bit the Seed/XofSeed/XofSeedStart/SeedKind objects to make things a bit less risky to use?

> @nsarlin-zama reviewed 12 files and all commit messages, and made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-:-P-oAlTN2mY2NfsOBUdj:b51jh9e)* status: 12 of 15 files reviewed, 10 unresolved discussions (waiting on mayeul-zama and tmontaigu).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 96 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nsdphEwI6pMNvWxr8:-P-nsdpiA_y_pVgPnOzL:b2ilhqi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L96)):*
> ```XML
> impl From<XofSeed> for XofSeedStart {
>     fn from(seed: XofSeed) -> Self {
>         Self::FirstByte(seed)
> ```

should this be FirstByteAes256? Some default conversion use Aes128 and others use Aes256, I think this should be unified (if we want to keep the conversions at all, see below)
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nqe4x-9GGRk-zU_Cq:-P-nqe4x-9GGRk-zU_Cr:bq7xnhk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L78)):*
> ```XML
> #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Versionize)]
> #[versionize(XofSeedStartVersions)]
> pub enum XofSeedStart {
> ```

name feels a bit off now that it encodes where to start and which algo to chose. Do we want to keep this? Or maybe we could have one enum for "where to start" and one enum for "which aes flavor" (for example the AesVariant from csprng?) ?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 466 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o5PBB8vQBbvCwK7il:-P-o5PBB8vQBbvCwK7im:b-6i1200) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L466)):*
> ```XML
>             .set_data(compressed_server_key.tag.data());
>         Self {
>             seed: pub_seed.into(),
> ```

Here with the current From impl, aes128 will be used?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 184 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o4bR77p8uqYZpiy8G:-P-o4bR77p8uqYZpiy8H:b-toe72l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L184)):*
> ```XML
>         // Relies on the `From<XofSeed>` default, which is Aes-256. Note that the private seed is
>         // never serialized, so regenerating a key set from the same `private_seed_bytes` with a
>         // version that used a different Aes variant yields a different ClientKey.
> ```

maybe we could have a version of this function that takes an AesVariant to allow someone with an old seed to re-generate its key? And should the commit be tagged as breaking because of this?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 417 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nsj_31aV8dPkH5_nJ:-P-nsj_31aV8dPkH5_nK:bxcb6l7) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L417)):*
> ```XML
>             // Aes-256 at the first byte, otherwise the key set decompresses against a different
>             // stream than the one it was generated with.
>             seed: XofSeedStart::new_aes256(pub_seed),
> ```

Is there a way to make this a bit less brittle ? At least if the from impl above is converted to Aes256 there would be only one source of truth
___
*[`tfhe/src/shortint/public_key/compact.rs` line 498 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nvw462gbdDhIdKjoj:-P-nvw462gbdDhIdKjok:b-im0p40) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/shortint/public_key/compact.rs#L498)):*
> ```XML
>         )));
>     }
>     Ok(SeedKind::xof_aes128(XofSeed::new(
> ```

Don't we want aes256 here too? I don't know if it's intended to be used between different versions of tfhe-rs
___
*[`tfhe-csprng/src/seeders/mod.rs` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o92d4BXh55ph8zOam:-P-o92d4BXh55ph8zOan:b6on3fb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/mod.rs#L152)):*
> ```XML
> /// This defaults to **Aes-256**, which is the current default for XOF derived streams.
> ///
> /// Call sites that must keep deriving the historical Aes-128 stream (because a seed they were
> ```

I wonder if we should keep this From impl at all? What about rerand and oprf?

Implicit conversions feel a bit risky here, maybe we should drop them and force the caller to be explicit about the kind of algo to use. IE, remove all the `seed: impl Into<SeedKind>` parameters and replace them by `seed: SeedKind`.

This is breaking but the commit is already breaking. And for reproducibility you need seed + algo to use so it makes sense to have this as params.
___
*[`tfhe-csprng/Cargo.toml` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o1_HCBwa28H_4eRu8:-P-o1_HCBwa28H_4eRu9:b-32lj1p) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/Cargo.toml#L33)):*
> ```TOML
> # disjoint sets of changes.
> bincode = { workspace = true }
> ciborium = "0.2"
> ```

do you use ciborium?
___
*[`tfhe-csprng/src/seeders/backward_compatibility/mod.rs` line 60 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o-rqqBQrszv-jz1dF:-P-o-rqqBQrszv-jz1dG:b-ovrnbh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/backward_compatibility/mod.rs#L60)):*
> ```XML
>         /// Bincode encoding of `SeedKindVersions::V0(SeedKind::Xof(XofSeed::new_u128(1,
>         /// b"abcdefgh")))` as written by tfhe-csprng 0.9 and earlier.
>         const LEGACY_XOF_BINCODE: &[u8] = &[
> ```

This could reuse the backward data test instead? 
If you don't want to make it too complicated I guess you can generate it along tfhe hl data, that's what we did for zk already.

It would make it easier to assess correctness if we have the code to generate it.
___
*[`tfhe-csprng/src/seeders/backward_compatibility/mod.rs` line 34 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nyHD974ApalbSNiJr:-P-nyHD974ApalbSNiJs:b9lfyvk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/backward_compatibility/mod.rs#L34)):*
> ```XML
>         Ok(match self {
>             Self::Ctr(seed) => SeedKind::Ctr(seed),
>             // Everything serialized by tfhe-rs 1.6 and earlier was derived by `xof_init_128`
> ```

since we are in the csprng crate I would put the csprng version here

tmontaigu (commented):
> @tmontaigu made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833)* status: 12 of 15 files reviewed, 10 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and nsarlin-zama).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nqe4x-9GGRk-zU_Cq:-P-oV4zr7C6XOuDltU2R:b-7da9yd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L78)):*

name feels a bit off now that it encodes where to start and which algo to chose. Do we want to keep this? Or maybe we could have one enum for "where to start" and one enum for "which aes flavor" (for example the AesVariant from csprng?) ?
</blockquote></details>

I kept it this way to limit the amount of breaking but if we can allow ourselfves to go fully breaking for this part then we should design a different enum/struct
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 96 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nsdphEwI6pMNvWxr8:-P-oWtzGDc5PAi8vMU5i:b-gd8xx6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L96)):*

should this be FirstByteAes256? Some default conversion use Aes128 and others use Aes256, I think this should be unified (if we want to keep the conversions at all, see below)
</blockquote></details>

yes I think it should be Aes256
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 466 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o5PBB8vQBbvCwK7il:-P-oX-3aCfUgS98yXBSx:b193ccb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L466)):*

Here with the current From impl, aes128 will be used?
</blockquote></details>

yes because the Into is still wrongly as Aes128
___
*[`tfhe/src/shortint/public_key/compact.rs` line 498 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nvw462gbdDhIdKjoj:-P-oVXQjDSH4gCBUUWco:b-keiebg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/shortint/public_key/compact.rs#L498)):*

Don't we want aes256 here too? I don't know if it's intended to be used between different versions of tfhe-rs
</blockquote></details>

I don't know, keeping Aes128 here felt the safest thing to do, as its not part of the functions fhevm/zama protocol uses it seemed fine to keep aes128 (just like regular non zama protocol usage is still using aes128)
___
*[`tfhe-csprng/Cargo.toml` line 33 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o1_HCBwa28H_4eRu8:-P-oUJK15PA_lNZvUo6f:b-kwqmjb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/Cargo.toml#L33)):*

do you use ciborium?
</blockquote></details>

nope that's some left over
___
*[`tfhe-csprng/src/seeders/mod.rs` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o92d4BXh55ph8zOam:-P-oW2hdAenCsMlh3y0Y:b-vfuuk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/mod.rs#L152)):*

I wonder if we should keep this From impl at all? What about rerand and oprf?

Implicit conversions feel a bit risky here, maybe we should drop them and force the caller to be explicit about the kind of algo to use. IE, remove all the `seed: impl Into<SeedKind>` parameters and replace them by `seed: SeedKind`.

This is breaking but the commit is already breaking. And for reproducibility you need seed + algo to use so it makes sense to have this as params.
</blockquote></details>

I made this few months ago when fixing other csprng bugs and try to not break too much
but yes it might be better to just break API so that at least there's some reassurance on what csprng is used because it's more explicit for the users

The idea of the from was also to make it an easy default for users, but maybe we can find other ways for that

I will try to remove it and see how bad it is in tfhe-rs hoping it wont be too bad

IceTDrinker (commented):
Generally agree with the remarks, the previous From implementation was welcome to try and keep the behavior unchanged, here the double legacy/From interactions look more scary and dangerous than useful

> @IceTDrinker reviewed 15 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-:-P-sjfVK35ppyt3IsKiM:b-byl4wu)* status: all files reviewed, 13 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, and tmontaigu).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nqe4x-9GGRk-zU_Cq:-P-seaB44K6eiHO3NKEB:bf2tde) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L78)):*

I kept it this way to limit the amount of breaking but if we can allow ourselfves to go fully breaking for this part then we should design a different enum/struct
</blockquote></details>

what's the breakage risk on this enum ?

can be checked with KMS to avoid any problems on what changes/breaks
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 96 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nsdphEwI6pMNvWxr8:-P-sfFo5CWbuHdzlhk2x:bi52v9a) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L96)):*

yes I think it should be Aes256
</blockquote></details>

maybe time to break this interface to avoid any weird behaviors and remove the From to make it easier to reason about it, current docstring earlier in the file indicates the From has the legacy behavior
___
*[`tfhe-csprng/src/seeders/mod.rs` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o92d4BXh55ph8zOam:-P-sjL4bCGNKrvJPP0vz:bhihafn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/mod.rs#L152)):*

I made this few months ago when fixing other csprng bugs and try to not break too much
but yes it might be better to just break API so that at least there's some reassurance on what csprng is used because it's more explicit for the users

The idea of the from was also to make it an easy default for users, but maybe we can find other ways for that

I will try to remove it and see how bad it is in tfhe-rs hoping it wont be too bad
</blockquote></details>

I agree we should try to remove the legacy on top of legacy, our client is the KMS team on that we can coordinate to make sure we have something that works for them
___
*[`tfhe-csprng/src/seeders/backward_compatibility/mod.rs` line 34 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nyHD974ApalbSNiJr:-P-oR6hYBTp0cRkiIR8B:b8khkdv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/backward_compatibility/mod.rs#L34)):*

since we are in the csprng crate I would put the csprng version here
</blockquote></details>

also 1.7 is the previous release (instead of 1.6), next version is 1.8 (we have to bump main), I think it's ok to have the version correspondance here
___
*[`tfhe-csprng/src/seeders/backward_compatibility/mod.rs` line 60 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o-rqqBQrszv-jz1dF:-P-sg_OB0h71FrW221Vo:b-nqw78c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/backward_compatibility/mod.rs#L60)):*

This could reuse the backward data test instead? 
If you don't want to make it too complicated I guess you can generate it along tfhe hl data, that's what we did for zk already.

It would make it easier to assess correctness if we have the code to generate it.
</blockquote></details>

agreed on having it be part of the backward compat dat apotentially
___
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 240 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-sgzKOAaXpaFYn7Jab:-P-sgzKOAaXpaFYn7Jac:bfmbvot) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/generators/aes_ctr/mod.rs#L240)):*
> ```XML
> impl From<XofSeed> for AesCtrParams {
>     fn from(seed: XofSeed) -> Self {
>         Self::from(SeedKind::from(seed))
> ```

let's maybe make the 256 explicit here ?

still reading and not entirely sure on how things interact with the SeedKind update
___
*[`tfhe/src/shortint/public_key/compact.rs` line 489 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-sgOn-28SGkstTK-74:-P-sgOn-28SGkstTK-75:b-y9a6za) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/shortint/public_key/compact.rs#L489)):*
> ```XML
> }
> 
> fn noise_generator_from_seed(
> ```

what is this function used for ?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 70 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-sesOi9t0QeA7e7FyM:-P-sesOi9t0QeA7e7FyN:bt4g7ks) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L70)):*
> ```XML
> /// byte it starts at, and which Aes variant derives the stream.
> ///
> /// Which Aes variant was used has to be recorded: key sets serialized by tfhe-rs 1.6 and earlier
> ```

1.7 and earlier, next version is 1.8 we need to bump main

IceTDrinker (commented):
also we need to be careful, PRF, ReRand and PKE need to call the 256 variant now instead of the 128 bit

essentially all prevous XofSeed behavior need to match the tech spec and therefore use the 256 path, the 128 path is strictly for backward compat on the keys to be able to load previous ones

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-:-P-sjv0I3Dc2pYuE1kg8:baf42h6)* status: all files reviewed, 13 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, and tmontaigu).

IceTDrinker:
Something unrelated to consider:

the warning about the xof keyset upgrade code looks to be a false positive caused by comments changing (?!?)

same kind of remark for the XofSeedStart the addition of the docstring is causing a hash change on the variant

@SouchonTheo I don't recall how those fields are hashed, but it feels like the comment and doc comment should not be impacting the overall hash of these objects, especially since the late linter should likely have gotten rid of them ?

do we hash the original code span explainign this oddity ?

tmontaigu:
> ⚠️ **Backward-compat snapshot: suspicious changes detected**
> 
> Suspicious changes were detected that may hide breaking changes. Please review the warnings below carefully.
> 
> ⚠️ Warnings
> * `high_level_api::backward_compatibility::xof_key_set::CompressedXofKeySetVersions` variant 1 (`high_level_api::xof_key_set::CompressedXofKeySet`):
>   struct hash changed
>   old: 300cd7c5963e04acb4f56dc8d2e85acb3909a397998d69eaedc89351880ac7a2
>   new: 855c59dac774128fa41ed427592a9d5a80dc864f85237354c6023a68a66603f6
> * `high_level_api::backward_compatibility::xof_key_set::CompressedXofKeySetVersions` upgrade `high_level_api::backward_compatibility::xof_key_set::CompressedXofKeySetV0` → `high_level_api::xof_key_set::CompressedXofKeySet`:
>   body hash changed
>   old: a0a7f8dfa98a16a05dcc01d35511666bcacc61bdccf690569ebe75db631edaf2
>   new: 69be81233614253bed285bdd158d41a106c94bb11fcff2a0450ed9c4927a7e07
> * Enum `high_level_api::backward_compatibility::xof_key_set::XofSeedStartVersions` removed
> 
> ➕ Neutral
> * New enum `high_level_api::backward_compatibility::xof_key_set::XofDerivationModeVersions` (1 variants)
> * New enum `high_level_api::backward_compatibility::xof_key_set::XofGenerationParamsVersions` (2 variants)
> * New enum `seeders::backward_compatibility::AesVariantVersions` (1 variants)
> * `seeders::backward_compatibility::SeedKindVersions`: new variant 1 (`seeders::SeedKind`)
> * `seeders::backward_compatibility::SeedKindVersions`: new upgrade `seeders::backward_compatibility::SeedKindV0` → `seeders::SeedKind`
> 
> If you encounter any errors or have doubts, you can verify locally by running:
> 
> ```
> make backward_correctness BASE_REF=<base_branch_or_commit>
> ```
> 
> Where `BASE_REF` is the reference branch or commit to check against.

Hash changed because XofSeedStart name changed, but backward compat tests are still ok because we are still able to honor the upgrade path

tmontaigu (commented):
> @tmontaigu made 8 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833)* status: 7 of 27 files reviewed, 11 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, and SouchonTheo).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nqe4x-9GGRk-zU_Cq:-P07abjf5ZVRvlQ5ClBc:b-3dgxhv) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L78)):*

what's the breakage risk on this enum ?

can be checked with KMS to avoid any problems on what changes/breaks
</blockquote></details>

I decided to completly break to force explicit selection
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 184 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o4bR77p8uqYZpiy8G:-P07bCBQ37Zux6BrZSqY:b-bkesan) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L184)):*

maybe we could have a version of this function that takes an AesVariant to allow someone with an old seed to re-generate its key? And should the commit be tagged as breaking because of this?
</blockquote></details>

It now takes XofGenerationParams as input so now its possible to select the generation algorithm
___
*[`tfhe/src/shortint/public_key/compact.rs` line 489 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-sgOn-28SGkstTK-74:-P-yM7ZI9PMLc4whWvbl:b-4gu8oo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/shortint/public_key/compact.rs#L489)):*

what is this function used for ?
</blockquote></details>

just to share the seed convertion + check in the test and seed functions
___
*[`tfhe-csprng/src/generators/aes_ctr/mod.rs` line 240 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-sgzKOAaXpaFYn7Jab:-P-yM1P860xziA1whZqG:b-hfpv0l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/generators/aes_ctr/mod.rs#L240)):*

let's maybe make the 256 explicit here ?

still reading and not entirely sure on how things interact with the SeedKind update
</blockquote></details>

removed
___
*[`tfhe-csprng/src/seeders/mod.rs` line 152 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o92d4BXh55ph8zOam:-P-yEbCXBuGcZcebjyNQ:bvezdnr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/mod.rs#L152)):*

I agree we should try to remove the legacy on top of legacy, our client is the KMS team on that we can coordinate to make sure we have something that works for them
</blockquote></details>

removed this From
___
*[`tfhe-csprng/src/seeders/backward_compatibility/mod.rs` line 60 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-o-rqqBQrszv-jz1dF:-P07aUr436Or1-UpD7d1:b-canj2h) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe-csprng/src/seeders/backward_compatibility/mod.rs#L60)):*

agreed on having it be part of the backward compat dat apotentially
</blockquote></details>

Added backward 1.8, 
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 158 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P07ajfsAliGu4B8XlDP:-P07ajfsAliGu4B8XlDQ:b6vca0l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L158)):*
> ```XML
>     ///
>     /// * `config` must use a dedicated public key
>     pub fn generate(
> ```

So far this is the only function without explicit xof generation params as it selects our new default (Aes256) is that ok or do we want to allow selection by taking the XofDerivationMode as input ?
___
*[`tfhe-csprng/Cargo.toml` line 30 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P07bOpD7FVfdzxzbMzQ:-P07bOpD7FVfdzxzbMzR:b4px8b4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe-csprng/Cargo.toml#L30)):*
> ```TOML
> ctr = "0.10.1"
> # Used to pin the serialized shape of `SeedKind` against hardcoded pre-0.10 bytes.
> bincode = { workspace = true }
> ```

bincode also no longer needed I think

nsarlin-zama (commented):
> @nsarlin-zama reviewed 17 files and all commit messages, made 7 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833)* status: 23 of 27 files reviewed, 11 unresolved discussions (waiting on IceTDrinker, mayeul-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 158 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P07ajfsAliGu4B8XlDP:-P0bYhsy3riOLmEyop1F:be6ctf9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L158)):*

So far this is the only function without explicit xof generation params as it selects our new default (Aes256) is that ok or do we want to allow selection by taking the XofDerivationMode as input ?
</blockquote></details>

The issue is that someone blindly updating tfhe-rs will have an unexpected result without notice, if you add the XofDerivationMode as parameter it would at least give a compile error?

And if someday we want to change something else (use a new algo, change something like table index) the user will have the control on when to do the switch
___
*[`tfhe/src/shortint/public_key/compact.rs` line 498 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nvw462gbdDhIdKjoj:-P0bcp3g8j7hO6aARKKT:b-32o52e) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/shortint/public_key/compact.rs#L498)):*

I don't know, keeping Aes128 here felt the safest thing to do, as its not part of the functions fhevm/zama protocol uses it seemed fine to keep aes128 (just like regular non zama protocol usage is still using aes128)
</blockquote></details>

it's used by the ProvenCompactList it looks like?

If you keep aes128 maybe add a comment explaining why
___
*[`tfhe-csprng/Cargo.toml` line 30 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P07bOpD7FVfdzxzbMzQ:-P0bdW1RDCo3bt_olSEa:b-7ym2sl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe-csprng/Cargo.toml#L30)):*

bincode also no longer needed I think
</blockquote></details>

to remove?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 75 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P0bZZKA5EeUGFFbzFGR:-P0bZZKA5EeUGFFbzFGS:b-6xfxsx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L75)):*
> ```XML
>     Aes256,
>     /// Stream derived with Aes-128, generator starts at the first byte.
>     LegacyAes128,
> ```

maybe have a `#[deprecated]` on these variants?
___
*[`utils/tfhe-backward-compat-data/crates/generate_1_8/Cargo.toml` line 2 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P0bdzt_2OlMa0NsRqNZ:-P0bdzt_2OlMa0NsRqN_:b64cfej) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/utils/tfhe-backward-compat-data/crates/generate_1_8/Cargo.toml#L2)):*
> ```TOML
> [package]
> name = "generate_1_8"
> ```

1_9 if it's in the next release?
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 76 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P0bYDoS95ao2ygTlCJL:-P0bYDoS95ao2ygTlCJM:bdrdyyx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L76)):*
> ```XML
>     /// Stream derived with Aes-128, generator starts at the first byte.
>     LegacyAes128,
>     /// Stream derived with Aes-128, replicating the tfhe-rs=1.5.4 csprng off by one.
> ```

Only this version? Or should be <=1.5.4?
___
*[`utils/tfhe-backward-compat-data/crates/generate_1_8/src/lib.rs` line 18 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P0beNerBzFCC_Triajo:-P0beNerBzFCC_Triajp:bh9jrl9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/utils/tfhe-backward-compat-data/crates/generate_1_8/src/lib.rs#L18)):*
> ```XML
> 
> // 1.8 changed `CompressedXofKeySet` to use AES-256 based random generator
> const HL_COMPRESSED_XOF_KEY_SET_TEST: HlCompressedXofKeySetTest = HlCompressedXofKeySetTest {
> ```

Just an idea, we could store the seed in HlCompressedXofKeySetTest, and in the corresponding test in `tfhe-rs/tests/backward_compatibility/high_level_api.rs`, try to generate from the seed and check that the keyset match?

IceTDrinker:
*[`utils/tfhe-backward-compat-data/crates/generate_1_8/Cargo.toml` line 2 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P0bdzt_2OlMa0NsRqNZ:-P1pEi6J85Tkt0kjzgSm:b-zb19ft) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/utils/tfhe-backward-compat-data/crates/generate_1_8/Cargo.toml#L2)):*

1_9 if it's in the next release?
</blockquote></details>

yeah, I think we need to bump main first though, the PR is fairly old

IceTDrinker (changes requested):
I think the backward compat of the xof seed as written is incorrect for keys generated right before this PR, there is no description of the keyset with XofSeedStart left in the code base (the old one V0 did not have it, the current struct does not contain a versioned XofSeedStart equivalent) meaning the deserializer would not be able to load the type at all

I think the reason this is hard to reason about is that most backward compatibility stuff we have handled is about adding fields, not so much modifying fields with types which can describe older states

may need some care and perhaps add a xof keyset from 1.8 to make sure we don't break anything

> @IceTDrinker reviewed 7 files and all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-:-P1pJ4Dk0lpQ59W5iNF9:bf8mqpu)* status: 23 of 27 files reviewed, 12 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 78 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nqe4x-9GGRk-zU_Cq:-P1pGn_P46IaHV5SFm5l:b-xml7xo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/high_level_api/xof_key_set/mod.rs#L78)):*

I decided to completly break to force explicit selection
</blockquote></details>

Fine on that, I just have a hard time wrapping my head around whether this is legal in terms of backward compatibility for what we do
___
*[`tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs` line 30 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pI19Y6sNQY5y6Ezfc:-P1pI19Y6sNQY5y6Ezfd:bam83ug) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs#L30)):*
> ```XML
>             // Start on second byte to keep backward compatibility with csprng bug. Aes-128 by
>             // construction, which is the only derivation this format could produce.
>             XofGenerationParams::new(seed, XofDerivationMode::LegacyAes128SecondByte),
> ```

this change must be illegal ?

XofSeedStart has been renamed to XofGenerationParams (its V0 is the old XofSeedStart)

If I have a CompressedXofKeySet from "V1" (i.e. the version before this PR)

then I stored essentially a discriminant + a Seed (XofSeedStart::First(Seed) or XofSeedStart::Second(Seed))

XofGenerationParams contains two fields, a Seed and a DerivationMode

so the layout has nothing in common with the "V1" layout

meaning we cannot load "V1" KeySets with this code ?

so here I think we have a problem in the way the backward compat is handled
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 140 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pGKhX0EGnZbrPj0-K:-P1pGKhX0EGnZbrPj0-L:b-nyj6m8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L140)):*
> ```XML
> #[versionize(CompressedXofKeySetVersions)]
> pub struct CompressedXofKeySet {
>     seed: XofGenerationParams,
> ```

I believe this requires a new variant in the enum dispatch ?

IceTDrinker:
*[`tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs` line 30 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pI19Y6sNQY5y6Ezfc:-P1pM9li1rRI_QQYU4vm:bxek5t6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs#L30)):*

this change must be illegal ?

XofSeedStart has been renamed to XofGenerationParams (its V0 is the old XofSeedStart)

If I have a CompressedXofKeySet from "V1" (i.e. the version before this PR)

then I stored essentially a discriminant + a Seed (XofSeedStart::First(Seed) or XofSeedStart::Second(Seed))

XofGenerationParams contains two fields, a Seed and a DerivationMode

so the layout has nothing in common with the "V1" layout

meaning we cannot load "V1" KeySets with this code ?

so here I think we have a problem in the way the backward compat is handled
</blockquote></details>

ok no I must be mistaken, the XofSeedStart gets auto upgraded during the load to the XofGenerationParams, but this code needs an update to be able to load into the latest form of that type, my bad

IceTDrinker:
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 140 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pGKhX0EGnZbrPj0-K:-P1pMaOp5ReRYZQSXGT3:b-9khl4z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L140)):*

I believe this requires a new variant in the enum dispatch ?
</blockquote></details>

incorrect, the type gets auto upgraded during unversionize since the enum dispatch has been kept

IceTDrinker:
*[`tfhe/src/shortint/public_key/compact.rs` line 498 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P-nvw462gbdDhIdKjoj:-P1pOPu905xoRBW2WSIv:bvr89ic) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/93c92b02e3c3c7945748a47b1fe6a8d59eb636bb/tfhe/src/shortint/public_key/compact.rs#L498)):*

it's used by the ProvenCompactList it looks like?

If you keep aes128 maybe add a comment explaining why
</blockquote></details>

256 here as far as I know

all manual XofSeed/AES uses (which were essentially coming from the NIST submission) should be upgraded to 256

as far as I remember we have the rerand and PRF stuff currently using the AES 128

IceTDrinker (commented):
> @IceTDrinker reviewed 4 files, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833)* status: 24 of 27 files reviewed, 12 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs` line 61 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pN7km8V-pBqZ1i21k:-P1pN7knBRumjWT04ApL:bi4cs4m) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/backward_compatibility/xof_key_set.rs#L61)):*
> ```XML
>         Ok(match self {
>             Self::FirstByte(seed) => {
>                 XofGenerationParams::new(seed, XofDerivationMode::LegacyAes128)
> ```

LegacyAes128FirstByte ?

strictly not needed, but at least symetrical with the other one
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 197 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pO8G29jVjZVzdSE9B:-P1pO8G29jVjZVzdSE9C:b-90ujbu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L197)):*
> ```XML
>         }
>         let mut private_generator =
>             RandomGenerator::<DefaultRandomGenerator>::new(private_seed.clone());
> ```

looks a bit weird, gives the impression we clone the private seed
___
*[`tfhe/src/high_level_api/xof_key_set/mod.rs` line 202 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3833#-P1pOAuh6w6HHn6YXm01:-P1pOAuh6w6HHn6YXm02:b-x3tc99) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/34d01b1926beb6a5c179e5279ac29c6b2dde2e54/tfhe/src/high_level_api/xof_key_set/mod.rs#L202)):*
> ```XML
>         private_generator.fill_slice_with_random_uniform(&mut public_seed_bytes);
>         let public_seed =
>             private_seed.with_seed(XofSeed::new(public_seed_bytes, public_seed_separator));
> ```

and then the private seed serves for the public seed

can likely grab the mode from the private param stuff instead of cloning all of it earlier

PR #3855, opened 2026-08-21
feat(gpu): better aes and new carry propagation (Sklansky)

enzodimaria:
image

image

image

---

**Comparision with FPGA latency shew that GPU performance were not enough.**

5 ideas to improve latency and a bit throughput:

- **1st Idea:** Flush only where the noise budget forces it.
The budget is five noise levels. We are now flushing only when we have to. The maximum we were using was two.

image

image

- **4th Idea:** New carry propagation algorithm. Trade-off: less PBS, more copies.

image

image


backends/tfhe-cuda-backend/cuda/include/aes/aes_utilities.h
```diff
@@ -2,152 +2,268 @@
...
+  int_aes_linear_tables(CudaStreams streams, bool allocate_gpu_memory,
+                        uint32_t sbox_parallelism, uint64_t &size_tracker,
+                        aes_buffer_scope scope) {
+
+    this->sbox_reorder_len = AES_BITS_PER_BYTE * sbox_parallelism;
+    this->has_round_tables = (scope == aes_buffer_scope::FULL_ENCRYPTION);
+
+    if (has_round_tables) {
+      const uint32_t shift_rows_map[AES_STATE_BYTES] = {
+          0, 5, 10, 15, 4, 9, 14, 3, 8, 13, 2, 7, 12, 1, 6, 11};
```
guillermo-oyarzun, line 161: In general, i am missing more comments or a diagrams that explain the new aes algorithm, is kind of hard to follow. For instance this looks like a random list of elements to me.

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
```diff
@@ -1782,6 +1783,253 @@ void host_resolve_group_carries_sequentially(
...
+ * @param t index of the pair inside the level, targets only, so
+ * run = t / half and base = (2 * run + 1) * half
+ * @param half level width, half of the block a level combines
+ * @param num_positions number of scanned positions
+ * @param forward true keeps q-space where the scan is a forward prefix,
+ * false mirrors both ends
+ * @param p_target receives the position that absorbs
+ * @param p_source receives the position it reads
+ */
+__device__ __forceinline__ void
```
guillermo-oyarzun: can you add a test (it can be a backend test), that verifies that this new carry prop works as it should be on some corner cases

guillermo-oyarzun (changes requested):

backends/tfhe-cuda-backend/cuda/include/aes/aes_utilities.h
enzodimaria: I added some comments! including a big header in aes.cuh to give a general overview

backends/tfhe-cuda-backend/cuda/src/integer/integer.cuh
enzodimaria: Test added in the Rust frontend
backend tests for this functions requiered too much refactoring

IceTDrinker (commented):
not sure why I was tagged here, a bad rebase of a PR stack ?

enzodimaria:
No it's based on Prince on purpose @IceTDrinker 
Because I'm using the same optimized struct/function in Prince and in AES

By reusing functions of Prince, I could delete a few thousands of lines in this PR

IceTDrinker:
is this PR supposed to touch the prince code ?

enzodimaria:
No, but it's based on Prince code

IceTDrinker:
ok well I think the reason we were pinged is this branch gets out of sync with the prince branch regularly so we get the prince PR content in this PR as well

enzodimaria:
Oh, maybe!

backends/tfhe-cuda-backend/cuda/include/aes/aes_utilities.h
```diff
@@ -2,152 +2,290 @@
...
-    constexpr uint32_t MIX_COLUMNS_MUL_WORKSPACE_BYTES = BYTES_PER_COLUMN + 1;
-
-    this->mix_columns_col_copy_buffer = new CudaRadixCiphertextFFI;
-    create_zero_radix_ciphertext_async<Torus>(
-        streams.stream(0), streams.gpu_index(0),
-        this->mix_columns_col_copy_buffer, BITS_PER_COLUMN * num_aes_inputs,
-        params.big_lwe_dimension, size_tracker, allocate_gpu_memory);
-
-    this->mix_columns_mul_workspace_buffer = new CudaRadixCiphertextFFI;
+    this->mix_columns_workspace = new CudaRadixCiphertextFFI;
```
andrei-stoian-zama: let's avoid `CudaRadixCiphertextFFI*` andjust use regular `CudaRadixCiphertextFFI`, avoiding `new`

PR #3856, opened 2026-08-21
chore(hpu): network trace dbg selftests & helpers

lsainati:
Network trace helpers & selftests.
Theses board tests are basically the replacement of the old bash scripts

### DIFF
--- a/backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
+++ b/backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
@@ -43,14 +43,14 @@ offset= 0x0
   field.retry_max_read_request = {size_b=8, offset_b=8, default={Cst=0xA}, description="Maximum number of retries to do before stopping and triggering error"}
 
 [section.mhdma_system.register.fsm_value]
-  description="Packed read-only snapshot of all internal FSM states: [26:24] formatter, [20:18] slave notify-rx, [17:15] slave ciphertext-emission, [11:9] master read-request, [5:3] master notify"
+  description="Live FSM snapshot: [22:20] formatter, [17:16] slave ciphertext-emission, [13:12] slave notify-rx, [9:8] master burst, [5:4] master read-request, [1:0] master notify"
   owner="Kernel"
   read_access="Read"
   write_access="None"
   default={Cst=0x0}
 
 [section.mhdma_system.register.errors]
-  description="Sticky error flags (cleared by reset_registers): [13] master read-request cmd queue overflow, [12] master notify cmd queue overflow, [11] formatter master cmd discard, [10] formatter slave cmd discard, [9] formatter payload gap, [8] decoder rx FIFO overflow, [7] slave read-request cmd queue overflow, [6] AXI4 HBM read response error, [5] master max retry read-request hit, [4] master max retry notify hit, [3] ciphertext sequence number mismatch, [2:1] per-PC AXI4 HBM write error, [0] multi-HPU one-hot ID violation"
+  description="Sticky error flags, cleared by reading this register: [31] software-injected error (see error_mask.inject_err), [13] master read-request cmd queue overflow, [12] master notify cmd queue overflow, [11] formatter master cmd discard, [10] formatter slave cmd discard, [9] formatter payload gap, [8] decoder rx FIFO overflow, [7] slave read-request cmd queue overflow, [6] AXI4 HBM read response error, [5] master max retry read-request hit, [4] master max retry notify hit, [3] ciphertext sequence number mismatch, [2:1] per-PC AXI4 HBM write error, [0] multi-HPU one-hot ID violation"
   owner="Kernel"
   read_access="ReadNotify"
   write_access="None"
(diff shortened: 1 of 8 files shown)


backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
```diff
@@ -64,6 +64,30 @@ offset= 0x0
   default={Cst=0x0}
   duplicate=["_0", "_1", "_2", "_3", "_4", "_5", "_6", "_7"]
 
+[section.mhdma_system.register.trace_ctrl]
+  description="Packet-trace control and status: write frozen=1 to freeze capture now, or frozen=0 to flush and re-arm."
```
JJ-hw: Not sure what "freeze" mean.
You mean when the bit is set to 1; we stop the capture ?
wptr and wrapped are status bits once the capture is stopped.

This is detail, but I won't mix command (which "freeze" is) and the status fields in the same register.

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
```diff
@@ -64,6 +64,30 @@ offset= 0x0
...
+  owner="Kernel"
+  read_access="Read"
+  write_access="WriteNotify"
+  field.frozen  = {size_b=1, offset_b=0, default={Cst=0x0}, description="An error stopped capture; the buffer is frozen and ready to read"}
+  field.wptr    = {size_b=6, offset_b=1, default={Cst=0x0}, description="Write pointer captured at freeze (= oldest entry index; rotate the ring by this to order oldest->newest when wrapped=1)"}
+  field.wrapped = {size_b=1, offset_b=7, default={Cst=0x0}, description="Ring wrapped since last re-arm: 1 = all TRACE_DEPTH entries valid (rotate by wptr); 0 = only entries [0 .. wptr-1] valid (partial fill, no rotation). Re-arm does not clear the RAM, so slots beyond a partial fill hold stale prior-session data"}
+
+[section.mhdma_system.register.trace_data]
+  description="Packet-trace read window: each read returns the next 32-bit word of the frozen trace and advances the read pointer."
```
JJ-hw: From which read pointer ?
Especially when the system has wrapped ?

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
```diff
@@ -64,6 +64,30 @@ offset= 0x0
...
+  write_access="None"
+  default={Cst=0x0}
+
+[section.mhdma_system.register.error_mask]
+  description="Packet-trace trigger mask + software error inject."
+  owner="User"
+  read_access="Read"
+  write_access="Write"
+  field.mask       = {size_b=31, offset_b=0,  default={Cst=0x7FFFFFFF}, description="Per-error trace-trigger enable, aligned to mhdma_system.errors[11:0]; bits above the mhdma error width are ignored"}
+  field.inject_err = {size_b=1,  offset_b=31, default={Cst=0x0}, description="Write 0->1 to inject a synthetic error: triggers the trace and sets mhdma_system.errors[31]. Sticky, cleared by reading mhdma_system.errors"}
```
JJ-hw: Is the injected error a pulse, or a leveled signal that equals one as long as the register is 1 (i.e. user has not read it) ?

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
```diff
@@ -64,6 +64,30 @@ offset= 0x0
...
+  read_access="ReadNotify"
+  write_access="None"
+  default={Cst=0x0}
+
+[section.mhdma_system.register.error_mask]
+  description="Packet-trace trigger mask + software error inject."
+  owner="User"
+  read_access="Read"
+  write_access="Write"
+  field.mask       = {size_b=31, offset_b=0,  default={Cst=0x7FFFFFFF}, description="Per-error trace-trigger enable, aligned to mhdma_system.errors[11:0]; bits above the mhdma error width are ignored"}
```
JJ-hw, line 100: I would not mix the mask, which is functional, with the inject_err, inside the same register.

backends/tfhe-hpu-backend/src/utils/mhdma/regs.rs
```diff
@@ -0,0 +1,161 @@
...
+// The opcode in req_id selects the type; see mhdma_notify / issue_reads.
+// =================================================================================================
+
+fn build_req_id(opcode: u32, node: u8, mode: u32) -> u32 {
+    // assert, not debug_assert: release drops those, and node >= 16 bleeds into the opcode nibble.
+    assert!(
+        node < 16 && mode < 4,
+        "req_id field overflow: node={node} mode={mode}"
+    );
+    (opcode << 20) | ((node as u32) << 16) | (mode << 14)
```
JJ-hw: Could the offsets be given by the toml ?
It would be less error prone.

backends/tfhe-hpu-backend/src/utils/mhdma/trace.rs
```diff
@@ -0,0 +1,436 @@
...
+            dst_addr: bitfield(w0, 0, 16),
+            src_addr: bitfield(w0, 16, 16),
+            iop: bitfield(w1, 0, 8),
+            req: bitfield(w1, 8, 4),
+            mode: bitfield(w1, 12, 2),
+            flag: bitfield(w1, 14, 6),
+            hpu: bitfield(w1, 28, 4),
+            seq: bitfield(w2, 0, 8),
+            src_mac: bitfield(w2, 8, 24),
+            errbits: bitfield(w3, 0, PKT_TRACE_ERR_W),
```
JJ-hw, line 93: Are these values available as constant ?
It would be less error prone, if they exist.

JJ-hw (dismissed):
Seems OK to me.

I'm not the more fitted to review the test part (not sure what is going on here, and if everything is tested).

General remarks:
* concerning register definition, avoid mixing fields of different natures. (ex. cmd vs status)
* I am not a big fan of clear-on-read register. The behavior is counter-intuitive (a read should be harmless). A reading "out of place" could erase the error registers, which may not be expected.
* I don't know if they exist, but if the constants are defined for register field size and offset exist, it is preferable to use them instead of numbers to avoid future errors or update if the registers are modified.
* Same for fields in packet.

backends/tfhe-hpu-backend/src/utils/mhdma/trace.rs
lsainati: no this cannot be inferred as is: there is no toml that expresses theses values.
If we were to write one we won't be able to use it in the RTL as is and we would need to create a new package system that generates from toml a pkg.sv file. Doable but a choice of how much time we want to use on this

backends/tfhe-hpu-backend/src/utils/mhdma/regs.rs
lsainati: yes this is true and this can be easily inferred from toml 
done in next commit

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
lsainati: done in next commit

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
lsainati: it is a level, once you read mhdma_errors it is cleared

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
lsainati: Here we always read in RTL from 0 to the size of the trace. 
Wrapped only means that there was data earlier: the trace does not only contain 0 up to the error trigger.
We can read the trace how much we want but there are helpers here in the rust to read the full trace directly without the hassle of thinking about sizes
the storage read pointer will be readable in the next commit

backends/tfhe-hpu-backend/config_store/regmap/hpu_regif_core_mhdma_2in3.toml
lsainati: when 1 it means the capture has stopped, can be triggered by an error or by the user. it is a status & command register
wptr and wrapped are status registers that are relevant if and only if freeze is "1"
technically it is both and because the two others depends strongly on the first I thought it would be nicer to do only one read instead of two
freeze status & command will be splitted in the next commit

pgardratzama (dismissed):

pgardratzama (approved):

PR #3859, opened 2026-08-24
Ns/feat/transciphering hl

nsarlin-zama:

### PR content/description
Add hl transciphering api on top of the transciphering layer. This is supposed to be used like this:

```rust
use tfhe::prelude::*;
  use tfhe::safe_serialization::{safe_deserialize, safe_serialize};
  use tfhe::shortint::parameters::TranscipheringParameters;                         
  use tfhe::transciphering::OneTimePadPlainState;
  use tfhe::{
      generate_keys, set_server_key, ConfigBuilder, FheUint64, HlStreamCipher, HlTranscipherer,
      OneTimePadFheSecretMask, Seed, StreamCiphertext, TranscipherSession,
  };

  let (client_key, server_key) = generate_keys(
      ConfigBuilder::default().enable_transciphering(TranscipheringParameters::SameAsCompute),
  );
  set_server_key(server_key);

  // server: generate a one-time pad with the OPRF
  let fhe_pad = OneTimePadFheSecretMask::new_random(Seed(0), 64).unwrap();

  // In production, this decryption is done by the KMS and the pad is sent to the
  // client encrypted under its own public key.
  let plain_pad = fhe_pad.decrypt(&client_key);

  // client: mask an input with the pad
  let input = 0xDEAD_BEEF_CAFE_BABEu64;
  let mut cipher = OneTimePadPlainState::new(plain_pad);
  let encrypted_input = cipher.try_encrypt(input).unwrap();

  // server: transcipher into an ordinary FheUint64
  // There is no IV: the pad is the keystream, consumed from bit 0 onwards.
  let mut session = TranscipherSession::one_time_pad(fhe_pad).unwrap();
  let ct: FheUint64 = session.transcipher(&encrypted_input).unwrap();

  // Ready to be used like any ciphertext
  let doubled = &ct + &ct;

  let recovered: u64 = ct.decrypt(&client_key);
  assert_eq!(recovered, input);
  ```

Remaining work (following pr):
- backward compat data
- supporting more than 1bit of keystream per lwe ct
- documentation
- gpu implem, currently there is only a stub
- input lists: packing multiple inputs in the same byte array (matters to follow closely the rfc, for sub byte objects like booleans)

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)

### CODE REVIEW

**Review by tmontaigu — CHANGES_REQUESTED** [review]
> @tmontaigu reviewed 13 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: 13 of 27 files reviewed, 1 unresolved discussion (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 334 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P070AT5EiAWjdJYmlQQ:-P070AT5EiAWjdJYmlQR:b-pl6h8t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L334)):*
> ```XML
> {
>     let integer_sks = &cpu_key.key.key;
>     let blocks = Transcipherer::transcipher(session, &integer_sks.key, input.inner())
> ```

Should this not use the transciphering_key now ?

**Review by tmontaigu — CHANGES_REQUESTED** [review]
> @tmontaigu reviewed 14 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 4 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072u_h8OzCtc8JUQqc:-P072u_h8OzCtc8JUQqd:b-hu5saa) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L396)):*
> ```XML
> 
>         // The key crosses the wire: round-trip it through versioned serialization.
>         let serialized = bincode::serialize(&fhe_kreyv_key.versionize()).unwrap();
> ```

Should probably call safe_serialize + safe_deserialize directly ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 472 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072mE7BOdiy_tdIir1:-P072mE7BOdiy_tdIir2:b-hu5saa) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L472)):*
> ```XML
> 
>         // The pad crosses the wire: round-trip it through versioned serialization.
>         let serialized = bincode::serialize(&fhe_mask.versionize()).unwrap();
> ```

Should probably call safe_serialize + safe_deserialize directly ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 528 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072rltBakg6ZrFtUkZ:-P072rltBakg6ZrFtUk_:b-hu5saa) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L528)):*
> ```XML
>         assert_eq!(fhe_mask.tag().as_u64(), KEY_SET);
> 
>         let serialized = bincode::serialize(&fhe_mask.versionize()).unwrap();
> ```

Should probably call safe_serialize + safe_deserialize directly ?

**Review by nsarlin-zama — COMMENTED** [review]
> @nsarlin-zama made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: 22 of 28 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 334 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P070AT5EiAWjdJYmlQQ:-P0MDjy60DFT5K6hsG0l:b-vmsxiy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L334)):*

Should this not use the transciphering_key now ?
</blockquote></details>

At the moment, the transciphering_key is only capable of doing the pad/stream key generation, it is just a wrapper for the oprf. In the future, we might want to have optimized parameters for specific ciphers (eg some work by R&D and kreyvium, implemented as "kreyvium fast" on the gpu)
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 396 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072u_h8OzCtc8JUQqc:-P0N20kiEGmSI3dA5SRw:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L396)):*

Should probably call safe_serialize + safe_deserialize directly ?
</blockquote></details>

done
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 472 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072mE7BOdiy_tdIir1:-P0N21vo8-sZfXm6nPOy:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L472)):*

Should probably call safe_serialize + safe_deserialize directly ?
</blockquote></details>

done
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 528 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P072rltBakg6ZrFtUkZ:-P0N22ujDHCnJsHnVamc:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L528)):*

Should probably call safe_serialize + safe_deserialize directly ?
</blockquote></details>

done

**Review by tmontaigu — DISMISSED** [review]
> @tmontaigu reviewed 6 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on IceTDrinker, mayeul-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 334 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P070AT5EiAWjdJYmlQQ:-P0N9KQVACLe7DkUPZQ8:b32wtd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f6cf489c015aa4998f9ca8b51b42853d5fdb9bf2/tfhe/src/high_level_api/transciphering/mod.rs#L334)):*

At the moment, the transciphering_key is only capable of doing the pad/stream key generation, it is just a wrapper for the oprf. In the future, we might want to have optimized parameters for specific ciphers (eg some work by R&D and kreyvium, implemented as "kreyvium fast" on the gpu)
</blockquote></details>

ok

**Review by IceTDrinker — COMMENTED** [review]
Missing just the code plugging integer into HL for review, can finish that afterwards

> @IceTDrinker reviewed 24 files and all commit messages, and made 10 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-:-P0RHY2u62uwhnuTqj-q:bcnkgv4)* status: all files reviewed, 9 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 60 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0R8aEw96d5_TJmV2BD:-P0R8aEw96d5_TJmV2BE:bt0xmj2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/.github/workflows/aws_tfhe_wasm_tests.yml#L60)):*
> ```YAML
>                 - tfhe/src/shortint/**
>                 - tfhe/src/integer/**
>                 - tfhe/src/transciphering/**
> ```

probably also necessary for whatever launches HL tests ?
___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 1 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RADKL4FhswqiX4_QR:-P0RADKL4FhswqiX4_QS:bvfu3xh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L1)):*
> ```XML
> use serde::{Deserialize, Serialize};
> ```

data PR open soon ? we should avoid merging ar at the very least releasing without it
___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAOet2Inf4U3Eg4rs:-P0RAOet2Inf4U3Eg4rt:bgt1wrj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L22)):*
> ```XML
> #[cfg_attr(dylint_lib = "tfhe_lints", allow(serialize_without_versionize))]
> pub struct KreyviumFheKeyVersionOwned {
>     pub(crate) key: <ShortintKreyviumFheKey as VersionizeOwned>::VersionedOwned,
> ```

why this pattern ? it departs from what we do everywhere else right ?
___
*[`tfhe/src/high_level_api/transciphering/stream_ciphertext.rs` line 36 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFyQv6MpU22NRmoUn:-P0RFsHz9TLaUUPnECNY:bicgiba) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/transciphering/stream_ciphertext.rs#L36)):*
> ```XML
>     /// Shape of the value this ciphertext encrypts, which decides the type it
>     /// can be transciphered into.
>     pub fn kind(&self) -> IntegerStreamCiphertextKind {
> ```

what's the difference with cipher_kind ?
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 69 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFMt64F0jdH59ejoh:-P0RFMt64F0jdH59ejoi:b-bqn178) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L69)):*
> ```XML
>     }
> 
>     /// Trans-cipher `input_stream` against `8 * input_stream.len()` bits of
> ```

what if it gets exhausted in the middle ? (like OTP may not have 8*input.len() bits ?
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 104 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFXtC3TdJ_p2LO3-V:-P0RFXtC3TdJ_p2LO3-W:b1fypwh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L104)):*
> ```XML
> /// `input_stream.len()` clear bytes are consumed (low-bit first within each
> /// byte), so the keystream must contain exactly `8 * input_stream.len()` bits.
> pub fn apply_keystream(
> ```

does that mean the GPU scaffolding does nothing today ?
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAxlt22npAPLhtsE1:-P0RAxlt22npAPLhtsE2:bxuckxk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L22)):*
> ```XML
> /// The `u32` payload is the width in bits (1 for `Boolean`). Returns an error
> /// for [`FheTypes::AsciiString`], which has no numeric interpretation.
> fn classify(kind: FheTypes) -> Result<(IntegerStreamCiphertextKind, u32), JsError> {
> ```

no way to have this on the enum somehow ? this makes the impl separate from the time which seems very odd

also don't we have a centralized enum for this ?
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 141 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RDgs70030S_OJJAoB:-P0RDgs70030S_OJJAoC:b-96jxki) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L141)):*
> ```XML
>         }
>         IntegerStreamCiphertextKind::Unsigned => {
>             let width = n_bits_override.unwrap_or(natural_bits as usize);
> ```

no risk of encrypting too many bits if override is < natural_bits ?
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 317 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0REVVT8Y41oLCkqWNG:-P0REVVT8Y41oLCkqWNH:b-bd7ofi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L317)):*
> ```XML
> 
>     #[wasm_bindgen]
>     pub fn safe_deserialize(buffer: &[u8], serialized_size_limit: u64) -> Result<Self, JsError> {
> ```

no conformant deserialize available for end user ?

same remark for the others

**Review by IceTDrinker — COMMENTED** [review]
I disagree with the new tagging logic which does not use the ServerKey as "tagging authority"

several questions/comments

tests look robot generated and don't always make sense around serialization IMO

> @IceTDrinker reviewed 5 files and all commit messages, and made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-:-P0S482OB7NquVc0kBv0:bmc1pdk)* status: all files reviewed, 27 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/kreyvium.rs` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RwanJ-9_3dkiGj9UD:-P0RwanJ-9_3dkiGj9UE:b-rxf1oo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/kreyvium.rs#L97)):*
> ```XML
> 
>     fn versionize(&self) -> Self::Versioned<'_> {
>         KreyviumFheKeyVersionedOwned::V0(KreyviumFheKeyVersionOwned {
> ```

same remark on versioning here, should the pattern be use the from/into pattern to avoid dealing with the device specific stuff ?
___
*[`tfhe/src/high_level_api/transciphering/one_time_pad.rs` line 127 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S42JhCg1gE00efCo7:-P0S42JhCg1gE00efCo8:bpl29lj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/one_time_pad.rs#L127)):*
> ```XML
> 
>     fn versionize(&self) -> Self::Versioned<'_> {
>         OneTimePadFheSecretMaskVersionedOwned::V0(OneTimePadFheSecretMaskVersionOwned {
> ```

versioning question
___
*[`tfhe/src/high_level_api/transciphering/aes.rs` line 80 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0Ruxrl-tDZFwIcrXqW:-P0Ruxrl-tDZFwIcrXqX:bm1fffo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/aes.rs#L80)):*
> ```XML
> }
> 
> impl Serialize for AesFheKey {
> ```

should this use the From/Into pattern you used in another instance @nsarlin-zama ?
___
*[`tfhe/src/high_level_api/transciphering/aes.rs` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RvSWCBiU0MCnpM2dY:-P0RvSWCBiU0MCnpM2dZ:b1dm9zb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/aes.rs#L97)):*
> ```XML
> 
>     fn versionize(&self) -> Self::Versioned<'_> {
>         AesFheKeyVersionedOwned::V0(AesFheKeyVersionOwned {
> ```

I see we sometimes do that but it seems some code works around managing those versions manually which seems very error prone

same for all code similar in this file around versioning

code I was checking delegating tfhe/src/high_level_api/compact_list.rs:257 and following on main
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 56 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RxRosCgO768cxrrp4:-P0RxRosCgO768cxrrp5:b3kjeid) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L56)):*
> ```XML
>         C: StreamCipher + ?Sized,
>     {
>         assert!(
> ```

maybe error out instead of assert since the function returns a result ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 304 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1HV6412aee4Or4J3:-P0S1HV6412aee4Or4J4:b-3d8knh) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L304)):*
> ```XML
>         // A session built straight from a lower-layer session carries no tag of
>         // its own, so fall back to the server key's, like every other HL op.
>         let tag = (!tag.is_empty()).then_some(&*tag);
> ```

I don't get this ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 338 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1_Ph-GuwejNfmJqy:-P0S1_Ph-GuwejNfmJqz:bjpqxh4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L338)):*
> ```XML
>     let kind = input.kind().to_data_kind(blocks.len())?;
>     let mut out = T::from_expanded_blocks(blocks, kind)?;
>     out.tag_mut().set_data(tag.unwrap_or(&cpu_key.tag).data());
> ```

shouldn't this always be the server key tag ?

like should the session have a tag at all ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 385 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1pAeCyDrJGWDu1_8:-P0S1pAeCyDrJGWDu1_9:baq2piu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L385)):*
> ```XML
>         set_server_key(server_key);
> 
>         // Client: pick a symmetric key + IV and encrypt a u64 with plain Kreyvium.
> ```

symmetric key and iv ? what does that mean ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 398 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2vvD7mWz51aRunch:-P0S2vvD7mWz51aRunci:b5ei5z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L398)):*
> ```XML
>         let fhe_kreyv_key = KreyviumFheKey::encrypt(plain_key, &client_key);
> 
>         // The key crosses the wire: round-trip it through versioned serialization.
> ```

not sure I understand the logic of the round trip on the "server side" ? since it keeps the fhe key materialized ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 437 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2qVMAdDLzqXIy8mE:-P0S2qVMAdDLzqXIy8mF:b5ei5z) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L437)):*
> ```XML
>         .unwrap();
> 
>         // The pad crosses the wire: round-trip it through versioned serialization.
> ```

not sure I understand the logic of the round trip on the "server side" ? since it keeps the fhe key materialized ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 485 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2hXq9Q31rcq3tLuH:-P0S2hXq9Q31rcq3tLuI:b-hkho3a) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L485)):*
> ```XML
> 
>         // Server: the OPRF-generated pad inherits the server key's tag.
>         let fhe_mask = OneTimePadFheSecretMask::new_random(Seed(0), 64).unwrap();
> ```

add a danger comment on hard coded seeds
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 497 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S37HH6f1P5nho9v3D:-P0S37HH6f1P5nho9v3E:b-aeqbjp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L497)):*
> ```XML
>         // Retag the pad so it no longer matches the server key: the output must
>         // follow the pad, since that is what decides which client key decrypts.
>         fhe_mask.tag_mut().set_u64(OTHER_KEY_SET);
> ```

this feels wrong tag authority previously has always been the server key, which should maybe reject invalidly tagged pads 
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 488 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2obi7Q-HATLub-XR:-P0S2obi7Q-HATLub-XS:b97rkl6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L488)):*
> ```XML
>         assert_eq!(fhe_mask.tag().as_u64(), KEY_SET);
> 
>         // The tag survives both wire formats.
> ```

both wire formats ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 503 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3N31EBODlJtfzZo0:-P0S3N31EBODlJtfzZo1:byxw9nz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L503)):*
> ```XML
> 
>         let mut fhe_stream = TranscipherSession::one_time_pad(fhe_mask).unwrap();
>         assert_eq!(fhe_stream.tag().as_u64(), OTHER_KEY_SET);
> ```

I disagree with the logic with historic behavior of the lib
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 506 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3PW8234d4sXPWzsL:-P0S3PW8234d4sXPWzsM:b-2guby3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L506)):*
> ```XML
> 
>         let transciphered: FheUint64 = fhe_stream.transcipher(&sym_cipher).unwrap();
>         assert_eq!(transciphered.tag().as_u64(), OTHER_KEY_SET);
> ```

same here
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 529 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3TQfBx0JQzrH6OZo:-P0S3TQfBx0JQzrH6OZp:bvix895) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L529)):*
> ```XML
> 
>         // Server: generate a random FHE pad.
>         let fhe_mask = OneTimePadFheSecretMask::new_random(Seed(0), 64).unwrap();
> ```

danger warning for hardcoded Seed(0)
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 532 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3ZWl5fVtmcDgX9o5:-P0S3ZWl5fVtmcDgX9o6:bkg7dk9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L532)):*
> ```XML
> 
>         // Client: decrypt the pad, then encrypt with it.
>         let plain_mask: OneTimePadPlainSecretMask = fhe_mask.decrypt(&client_key);
> ```

no safe serialize on Server then safe deserialize on client here ?
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 540 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3gqL0bH7NK1yEGDV:-P0S3gqL0bH7NK1yEGDW:beydboo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L540)):*
> ```XML
> 
>         // Server: transcipher against the pad it generated.
>         let mut fhe_stream = TranscipherSession::one_time_pad(fhe_mask).unwrap();
> ```

could simulate stateless one time pad by regening pad with same seed

**Review by nsarlin-zama — COMMENTED** [review]
> @nsarlin-zama reviewed 1 file and made 16 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 27 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, and SouchonTheo).

___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 60 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0R8aEw96d5_TJmV2BD:-P0RlBMl7217bR82IToj:b-gyb1gc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/.github/workflows/aws_tfhe_wasm_tests.yml#L60)):*

probably also necessary for whatever launches HL tests ?
</blockquote></details>

should be ok, hl filters have deny list:

```yaml
            high_level_api:
              - tfhe/src/**
              - '!tfhe/src/c_api/**'
              - '!tfhe/src/boolean/**'
              - '!tfhe/src/c_api/**'
              - '!tfhe/src/js_on_wasm_api/**'
```
___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 1 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RADKL4FhswqiX4_QR:-P0RI6nJCBcxpHNumJEe:b-l2ec72) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L1)):*

data PR open soon ? we should avoid merging ar at the very least releasing without it
</blockquote></details>

yes: https://github.com/zama-ai/tfhe-rs/pull/3882
___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAOet2Inf4U3Eg4rs:-P0RmXc-CfZkVKU9oYpQ:b7tiof0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L22)):*

why this pattern ? it departs from what we do everywhere else right ?
</blockquote></details>

This is the same pattern we use for the Signed/UnsignedRadixCiphertext, and all the "device-polymorphic" types that need to be moved to CPU before serialization.

IIRC this cannot be avoided today with Versionize because of some ownership issues. It might be possible to fix this by reworking the trait a bit.
___
*[`tfhe/src/high_level_api/transciphering/aes.rs` line 80 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0Ruxrl-tDZFwIcrXqW:-P0S4mBB9Rnfoi1rrk_d:b-omfkag) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/aes.rs#L80)):*

should this use the From/Into pattern you used in another instance @nsarlin-zama ?
</blockquote></details>

no that's what we also use for the signed/unsigned radix in tfhe/src/high_level_api/integers/unsigned/inner.rs, and AFAIK we don't have a choice
___
*[`tfhe/src/high_level_api/transciphering/aes.rs` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RvSWCBiU0MCnpM2dY:-P0S6Tku60tFMi1aFE2J:bmk0oes) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/aes.rs#L97)):*

I see we sometimes do that but it seems some code works around managing those versions manually which seems very error prone

same for all code similar in this file around versioning

code I was checking delegating tfhe/src/high_level_api/compact_list.rs:257 and following on main
</blockquote></details>

In compact list we delegate the version enum to the inner type, here we can't because of the tag. So I don't think the trick could work here.
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 56 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RxRosCgO768cxrrp4:-P0S5lkR5H2pe-Ixd9Ag:b-qb1vtp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L56)):*

maybe error out instead of assert since the function returns a result ?
</blockquote></details>

yes indeed
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 304 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1HV6412aee4Or4J3:-P0S91v36MJSG2gxqxtZ:b-824bro) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L304)):*

I don't get this ?
</blockquote></details>

my initial understanding was that the tag is used to identify what "produced the ct". Since here the ct are "produced" by the transciphering sessions, I thought it's were the tag should belong to.

But in the end the server key should be able to use the produced ct, so it means that the associated client key should decrypt them, so indeed we don't need to propagate the tag in the session

I reverted this
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 338 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1_Ph-GuwejNfmJqy:-P0SBpEpAEaOdES9W2gP:b-3gl447) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L338)):*

shouldn't this always be the server key tag ?

like should the session have a tag at all ?
</blockquote></details>

yes you're right, reverting this
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 385 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S1pAeCyDrJGWDu1_8:-P0S5RPlAucbMrHO3Bs6:b-cxmuya) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L385)):*

symmetric key and iv ? what does that mean ?
</blockquote></details>

symmetric cipher key, not fhe key. IE kreyvium key
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 398 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2vvD7mWz51aRunch:-P0SBwaj17c5FdkTCJ7x:b-dxmeai) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L398)):*

not sure I understand the logic of the round trip on the "server side" ? since it keeps the fhe key materialized ?
</blockquote></details>

this is part of the "client -> server" step, simulate serialization and sending to server
___
*[`tfhe/src/high_level_api/transciphering/stream_ciphertext.rs` line 36 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFyQv6MpU22NRmoUn:-P0RnLST8b1IfSJ9Hodc:bqelxdi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/transciphering/stream_ciphertext.rs#L36)):*

what's the difference with cipher_kind ?
</blockquote></details>

The kind of a stream **cipher** is *kreyvium, aes, otp*. 

The kind of a stream **ciphertext** is *signed, unsigned, boolean*. This is similar to DataKind. But we cannot use DataKind here, since DataKind also encodes a number of block, and these StreamCiphertext are not FHE encrypted, so block count means nothing to them.
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 69 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFMt64F0jdH59ejoh:-P0Rp6qm8KauBsOk7xjQ:b-v0uwl) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L69)):*

what if it gets exhausted in the middle ? (like OTP may not have 8*input.len() bits ?
</blockquote></details>

it will return an `InsufficientKeystream` error
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 104 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFXtC3TdJ_p2LO3-V:-P0RpB1QA2Q3gDB65x2t:b2g355x) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L104)):*

does that mean the GPU scaffolding does nothing today ?
</blockquote></details>

yes, they need to implement the xor between keystream and user input
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAxlt22npAPLhtsE1:-P0SCXbn0v9cDU2X3BXY:bdxs880) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L22)):*

no way to have this on the enum somehow ? this makes the impl separate from the time which seems very odd

also don't we have a centralized enum for this ?
</blockquote></details>

Reworked to use an intermediate type for that. Added some tests to check validity
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 141 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RDgs70030S_OJJAoB:-P0S-VJu3Kbq9nleGY4p:bix23xm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L141)):*

no risk of encrypting too many bits if override is < natural_bits ?
</blockquote></details>

from what I see, the behavior is the same as with the lwe encryption. If n_bits_override < natural_bits
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 317 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0REVVT8Y41oLCkqWNG:-P0SXEeW3yttzUbe16cY:b-k0qmh3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L317)):*

no conformant deserialize available for end user ?

same remark for the others
</blockquote></details>

intended use for this type in js is more in the other side:
client generate and encrypt a KreyviumFheKey, sends it to the server (not what will be used in prod, but a more usual transciphering approach without oprf). So as such there is no real need for this type.

I added it for the OneTimePadPlainSecretMask though, since this type is received by the client. For Kreyvium and Aes, they are just raw bytes, there is no metadata to check (except length that is checked at deserialization time).

nsarlin-zama (commented):
> @nsarlin-zama made 7 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: 19 of 33 files reviewed, 25 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 437 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2qVMAdDLzqXIy8mE:-P0WFSsA2QhxOI2IcVCw:b-d4h7k3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L437)):*

not sure I understand the logic of the round trip on the "server side" ? since it keeps the fhe key materialized ?
</blockquote></details>

this is not part of the server side, but more of the client->server part. Simulate the client sending the pad to the server. This is the initial transciphering scenario (without oprf) where pad is generated client side and fhe encrypted
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 485 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2hXq9Q31rcq3tLuH:-P0W60PI865izEb4JyZV:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L485)):*

add a danger comment on hard coded seeds
</blockquote></details>

done
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 488 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2obi7Q-HATLub-XR:-P0W61K47UMNmndC38z7:bulv7oa) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L488)):*

both wire formats ?
</blockquote></details>

initially it was using cbor and bincode, but I removed it and did not update the comment
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 497 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S37HH6f1P5nho9v3D:-P0W6HFW-3su6F8wIV7I:b-vs5jzi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L497)):*

this feels wrong tag authority previously has always been the server key, which should maybe reject invalidly tagged pads
</blockquote></details>

changed the behavior
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 503 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3N31EBODlJtfzZo0:-P0W6LHO9VzFvfiHhC7z:b-hfpv0l) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L503)):*

I disagree with the logic with historic behavior of the lib
</blockquote></details>

removed
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 529 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3TQfBx0JQzrH6OZo:-P0W6Nk15rEDrfn5OfDZ:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L529)):*

danger warning for hardcoded Seed(0)
</blockquote></details>

done
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 532 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3ZWl5fVtmcDgX9o5:-P0W6vbz9GEsatsqXrF9:bwwo0cr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L532)):*

no safe serialize on Server then safe deserialize on client here ?
</blockquote></details>

this is already in the test above, but I can add it for completeness

IceTDrinker (commented):
> @IceTDrinker reviewed 14 files and all commit messages, made 9 comments, and resolved 17 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 11 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`.github/workflows/aws_tfhe_wasm_tests.yml` line 60 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0R8aEw96d5_TJmV2BD:-P0ScfK_CrzxRUEsF0s2:bb7lbs2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/.github/workflows/aws_tfhe_wasm_tests.yml#L60)):*

should be ok, hl filters have deny list:

```yaml
            high_level_api:
              - tfhe/src/**
              - '!tfhe/src/c_api/**'
              - '!tfhe/src/boolean/**'
              - '!tfhe/src/c_api/**'
              - '!tfhe/src/js_on_wasm_api/**'
```
</blockquote></details>

nice
___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAOet2Inf4U3Eg4rs:-P0SdZTg9SFyZQwyFj95:b3sl4nm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L22)):*

This is the same pattern we use for the Signed/UnsignedRadixCiphertext, and all the "device-polymorphic" types that need to be moved to CPU before serialization.

IIRC this cannot be avoided today with Versionize because of some ownership issues. It might be possible to fix this by reworking the trait a bit.
</blockquote></details>

ok it seems the difference here is that you did not add the "inner" wrapping that some types have in the HL which moved the concern here

Should these owned types be stored here ? it seems in the past they were stored near their counterparts, I have 0 opinion on whether one is better than the other, just seems inconsistent at the moment
___
*[`tfhe/src/high_level_api/transciphering/aes.rs` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RvSWCBiU0MCnpM2dY:-P0Sf4z1EMXyYPQ1KIXe:b-fy511f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/aes.rs#L97)):*

In compact list we delegate the version enum to the inner type, here we can't because of the tag. So I don't think the trick could work here.
</blockquote></details>

yeah compact list has the inner dirty stuff implemented which means the HL struct can nicely derive the stuff it needs
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 398 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S2vvD7mWz51aRunch:-P0SfrjRAQ9KTXBnGsO7:b-7o8ju8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L398)):*

this is part of the "client -> server" step, simulate serialization and sending to server
</blockquote></details>

ah right thanks :facepalm: 
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 69 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFMt64F0jdH59ejoh:-P0SgB_cDY4-fxw5CJvl:b-pgl680) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L69)):*

it will return an `InsufficientKeystream` error
</blockquote></details>

something to keep an eye out/really be careful could mean the GPU would fail to process an otp stream of 12 bits today ? so could some very weird crashes that may not be easily caught if the coproc testing scenarii do rountrip only the simple types without grouping them with odd sizes in an input bundle
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAxlt22npAPLhtsE1:-P0SiYGB4cHyvhXUCeNL:b7vw76x) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L22)):*

Reworked to use an intermediate type for that. Added some tests to check validity
</blockquote></details>

what I meant was having the method as an impl block potentially for the type itself, even better would be to bridge with some rust type that serves as soure of truth for signedness/bit size, but I guess this is the simplest we can do at the moment
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 540 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WGttgAw5hXopxr0co:-P0WGttgAw5hXopxr0cp:b-d1wjcm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d83a4126a3506a2112af27d5e4bca8ad9eff6226/tfhe/src/high_level_api/transciphering/mod.rs#L540)):*
> ```XML
>         safe_serialize(&fhe_mask, &mut serialized, 1 << 24).unwrap();
>         let fhe_mask: OneTimePadFheSecretMask =
>             safe_deserialize(serialized.as_slice(), 1 << 24).unwrap();
> ```

robot change ?
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/integers.rs` line 653 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0SgS0fAMINVODQ-7xt:-P0SgS0fAMINVODQ-7xu:bjkcycw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d83a4126a3506a2112af27d5e4bca8ad9eff6226/tfhe/src/js_on_wasm_api/js_high_level_api/client/integers.rs#L653)):*
> ```XML
>     /// on both carrying the same discriminants, which this pins.
>     #[test]
>     fn fhe_types_mirror_hlapi() {
> ```

do we have something running those tests somewhere ? since it's in the wasm part of the lib I don't recall if we have actual "rust tests" in the wasm code running
___
*[`tfhe/src/integer/transciphering.rs` line 69 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WHE3FC35teNCZwdO_:-P0WHE3FC35teNCZwdOa:b-x82u47) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0921b154c42d94cad5b30ab924f9ee6664b9cd8/tfhe/src/integer/transciphering.rs#L69)):*
> ```XML
> 
>     fn is_conformant(&self, params: &Self::ParameterSet) -> bool {
>         self.kind == params.kind
> ```

destructuring ?

nsarlin-zama (commented):
> @nsarlin-zama made 7 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 9 unresolved discussions (waiting on IceTDrinker, mayeul-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/backward_compatibility/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAOet2Inf4U3Eg4rs:-P0WU8o80KGI6BSNHlpI:b-52q514) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/high_level_api/backward_compatibility/transciphering.rs#L22)):*

ok it seems the difference here is that you did not add the "inner" wrapping that some types have in the HL which moved the concern here

Should these owned types be stored here ? it seems in the past they were stored near their counterparts, I have 0 opinion on whether one is better than the other, just seems inconsistent at the moment
</blockquote></details>

indeed, I moved it closer to the base type
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 540 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0S3gqL0bH7NK1yEGDV:-P0WIBim4jk_Vh5bJ_sT:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L540)):*

could simulate stateless one time pad by regening pad with same seed
</blockquote></details>

done
___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 540 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WGttgAw5hXopxr0co:-P0WI8blDjowgZ3Zp4kk:b-6o2aqy) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d83a4126a3506a2112af27d5e4bca8ad9eff6226/tfhe/src/high_level_api/transciphering/mod.rs#L540)):*

robot change ?
</blockquote></details>

hmm yes, fixed
___
*[`tfhe/src/integer/transciphering.rs` line 69 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WHE3FC35teNCZwdO_:-P0WIbLw6sqLIGGip8Rl:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0921b154c42d94cad5b30ab924f9ee6664b9cd8/tfhe/src/integer/transciphering.rs#L69)):*

destructuring ?
</blockquote></details>

done
___
*[`tfhe/src/integer/gpu/transciphering/mod.rs` line 69 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RFMt64F0jdH59ejoh:-P0WL0GD7vJ0OVPUCqhY:b-1e4j3t) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/integer/gpu/transciphering/mod.rs#L69)):*

something to keep an eye out/really be careful could mean the GPU would fail to process an otp stream of 12 bits today ? so could some very weird crashes that may not be easily caught if the coproc testing scenarii do rountrip only the simple types without grouping them with odd sizes in an input bundle
</blockquote></details>

ah yes you're right, the comment was outdated
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/integers.rs` line 653 at r5](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0SgS0fAMINVODQ-7xt:-P0WUI_34sib6vBXGGcx:b-o9huyd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/d83a4126a3506a2112af27d5e4bca8ad9eff6226/tfhe/src/js_on_wasm_api/js_high_level_api/client/integers.rs#L653)):*

do we have something running those tests somewhere ? since it's in the wasm part of the lib I don't recall if we have actual "rust tests" in the wasm code running
</blockquote></details>

we wern't, I added a Make target and a job for this
___
*[`tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs` line 22 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RAxlt22npAPLhtsE1:-P0WMpAY0NhCih4ABcdP:b-az7cg2) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f956d29818b2355f891e9019c922552cb657c6ec/tfhe/src/js_on_wasm_api/js_high_level_api/client/transciphering.rs#L22)):*

what I meant was having the method as an impl block potentially for the type itself, even better would be to bridge with some rust type that serves as soure of truth for signedness/bit size, but I guess this is the simplest we can do at the moment
</blockquote></details>

Did the thing with the dedicated rust type, that works on crate::FheTypes.

This function is a small private helper transciphering related, since it returns a tuple and includes a transciphering type (IntegerStreamCiphertextKind), so maybe it's better if its stays that way?

IceTDrinker (commented):
> @IceTDrinker reviewed 10 files and all commit messages, made 2 comments, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 1 unresolved discussion (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 56 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RxRosCgO768cxrrp4:-P0WyUEtAf0ymxf7FKdH:bk7b8hr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L56)):*

yes indeed
</blockquote></details>

here @nsarlin-zama not necessary to move forward
___
*[`tfhe/src/integer/transciphering.rs` line 69 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WHE3FC35teNCZwdO_:-P0WyEcf0ttG0vIjQg88:b-igyty0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0921b154c42d94cad5b30ab924f9ee6664b9cd8/tfhe/src/integer/transciphering.rs#L69)):*

done
</blockquote></details>

was thinking of self, sorry

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: all files reviewed, 1 unresolved discussion (waiting on IceTDrinker, mayeul-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/high_level_api/transciphering/mod.rs` line 56 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0RxRosCgO768cxrrp4:-P0X3HxC2MzXukEQSlif:b-tn8l6w) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e4925f4e52ddc5953d5eb5d4dc6f7a5b1b9eec3a/tfhe/src/high_level_api/transciphering/mod.rs#L56)):*

here @nsarlin-zama not necessary to move forward
</blockquote></details>

sorry forgot this!
Fixed
___
*[`tfhe/src/integer/transciphering.rs` line 69 at r6](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-P0WHE3FC35teNCZwdO_:-P0X4Imm-mEk-gWVepSm:bbqu4ee) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/f0921b154c42d94cad5b30ab924f9ee6664b9cd8/tfhe/src/integer/transciphering.rs#L69)):*

was thinking of self, sorry
</blockquote></details>

done too!

IceTDrinker (approved):
Looks good to me, big thank you !

> @IceTDrinker reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859#-:-P0X764v6UtFl2toxByf:b-zahkr9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mayeul-zama, soonum, and SouchonTheo).

tmontaigu (approved):
> @tmontaigu reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mayeul-zama, soonum, and SouchonTheo).

IceTDrinker (approved):
> @IceTDrinker reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mayeul-zama, soonum, and SouchonTheo).

PR #3867, opened 2026-08-27
chore: robot update to avoid missing LFS objects in some cases

IceTDrinker:
- if an LFS object was updated in several commits and the commits were pushed at once, the previous method would not see the intermediate object

the 1024 limit according to the robot matches an internal limit from the lfs system which ignores larger blobs

the awk magic string is essentially parsing an object header and the rest matches an lfs pointer

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3867)

### DIFF
--- a/scripts/lfs_sync.sh
+++ b/scripts/lfs_sync.sh
@@ -1,17 +1,18 @@
 #!/usr/bin/env bash
 
 # This script tries to:
-# List all LFS references in a source remote on the given target branch and all tags
-# List all LFS references in a destination remote on the given target branch and all tags
-# Verify which LFS references are missing in the destination remote
-# Fetch ONLY the missing LFS references from the destination remote
-# Push all the LFS objects to the destination repo
+# List all LFS objects reachable from the given branch and all tags on a source remote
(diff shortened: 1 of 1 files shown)


SouchonTheo (commented):
seems clean ! do we need also to repair stuff that have been broke or we can forget about it and continue with your patch ?

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3867#-:-P07bNgI0WJZzCcbkNDo:bm0l11o)* status: 0 of 1 files reviewed, all discussions resolved (waiting on soonum).

SouchonTheo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3867#-:-P07bgpzA5v6mu7lnFi5:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on soonum).

IceTDrinker:
what's broken ? normally this is strictly better and when the problem was encoutered it was fixed as far as I can tell

PR #3872, opened 2026-08-28
feat(gpu): improve pbs128 further

guillermo-oyarzun:

### PR content/description
This PR improves latency an throughput of the SNS in classical params when executing in H100
Super tuned for current params work based on Pedro's finding that 1Block solving 3 FFTs gives beter throughput.

- Uses TBC to perform 1 iteration without moving global memory. Clusters of size 3. Each block calculates 3FFT + cross products and 1 IFFT. 
- Only one iteration cause the full `lwe_dimension` loops has worse throughput ( It seems like for throughput using clusters is better to have smalls kernels that can overlap with the comp and decomp of the protocol.
- 3FFTs are solved sequentially to avoid register to explote and polynomial copies are not unrolled for the same reason.
- Still uses Cooley Tukey FFT, warp syncs used to avoid shared in some stages.
- Torus accumulator can be narrowed cause after the first decomposition fits in 64-bit.
- Relaxed arithmetics used that provide ~103 bits of mantissa instead of 106.  In the emulated arithmetics the costly procedure of renormalization is delayed, so we perform fused operations instead of apply it each time. This saves a lot of calculations. According to R&D we could reduce at most to 102 bits without needing new params. 
- only 2 cluster syncs per iteration (minimal possible due to the reuse of shared memory).
- keeps 2 blocks residency per SM.
- Packing keyswitch kernel launcher tuned for the new params v1.7
- Some `syncthreads()` removed when using registers.
- Validated with noise test

The new pbs128 with relaxed arithmetics is not selected by default because we want to have more ci testing hours to fully validate it.  Its testing is triggered in `test_high_level_api_noise_squash_relaxed_pbs128_gpu` by using the `TFHE_RS_GPU_PBS128_RELAXED=1`  env.

Performance: 
  ```
┌──────────────────────────┬────────────┬───────────────┬───────────────┬────────┐
  │        benchmark         │   metric   │ old           │     final     │   Δ    │
  ├──────────────────────────┼────────────┼───────────────┼───────────────┼────────┤
  │ noise_squash             │ latency    │ 40.70 ms      │ 32.918 ms     │ ~1.2X  │
  ├──────────────────────────┼────────────┼───────────────┼───────────────┼────────┤
  │                          │ throughput │ 65 elem/s     │ 112.92 elem/s │ 1.7X   │
  ├──────────────────────────┼────────────┼───────────────┼───────────────┼────────┤
  │ decomp_noise_squash_comp │ latency    │ 50.1 ms       │ 41.355 ms     │  ~1.2X │
  ├──────────────────────────┼────────────┼───────────────┼───────────────┼────────┤
  │                          │ throughput │ 49 elem/s     │ 82.971 elem/s │ 1.67X  │
  └──────────────────────────┴────────────┴───────────────┴───────────────┴────────┘
```
As refence Q2 results for SNS were latency: 41ms throughput: 32 SNS/s
Noise test results

```
test integer::gpu::server_key::radix::tests_noise_distribution::dp_ks_pbs_128_packingks::test_gpu_noise_check_encrypt_dp_ks_standard_pbs128_packing_ks_noise_gpu_test_meta_param_cpu_2_2_ks_pbs_pke_to_small_zkv2_tuniform_2m128 ... measured_variance_after_packing=Variance(5.46156158303075e-39)
expected_variance_after_packing=Variance(7.001522210713421e-39)
variance_lower_bound=Variance(5.309243161732477e-39)
variance_upper_bound=Variance(5.620224452281175e-39)
measured_mean_after_packing=-4.585770038201851e-22
expected_mean_after_packing=0.0
mean_after_packing_lower_bound=-1.945763822206876e-21
mean_after_packing_upper_bound=1.0286098145665058e-21
PASS: measured_mean_after_packing confidence interval contains the expected mean
PASS: measured_variance_after_packing is smaller than expected variance.

==========
Warning: noise formula might be over estimating the noise.
==========

variance after noise squashing and packing = Variance(5.46156158303075e-39)
std_dev  after noise squashing and packing = StandardDev(7.390237873729607e-20)
modulus after noise squashing and packing = 340282366920938500000000000000000000000
std_dev_under_modulus = 2.5147676357814743e19
b_switch_squash=330691944105263900000
b_switch_squash_log2=68.16404979861058
bound_log2=70
PASS: noise after noise squashing and packing respects bound.
ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 7567 filtered out; finished in 1405.18s
```

Running it with 10x more runs still passes, i got 

```

running 1 test
test integer::gpu::server_key::radix::tests_noise_distribution::dp_ks_pbs_128_packingks::test_gpu_noise_check_encrypt_dp_ks_standard_pbs128_packing_ks_noise_gpu_test_meta_param_cpu_2_2_ks_pbs_pke_to_small_zkv2_tuniform_2m128 ... measured_variance_after_packing=Variance(5.546691666177022e-39)
expected_variance_after_packing=Variance(7.001522210713421e-39)
variance_lower_bound=Variance(5.496783057671275e-39)
variance_upper_bound=Variance(5.597252575054268e-39)
measured_mean_after_packing=2.0836336057201393e-22
expected_mean_after_packing=0.0
mean_after_packing_lower_bound=-2.6856765928663726e-22
mean_after_packing_upper_bound=6.852943804306652e-22
PASS: measured_mean_after_packing confidence interval contains the expected mean
PASS: measured_variance_after_packing is smaller than expected variance.

==========
Warning: noise formula might be over estimating the noise.
==========

variance after noise squashing and packing = Variance(5.546691666177022e-39)
std_dev  after noise squashing and packing = StandardDev(7.447611473604825e-20)
modulus after noise squashing and packing = 340282366920938500000000000000000000000
std_dev_under_modulus = 2.534290860145788e19
b_switch_squash=333259248109171100000
b_switch_squash_log2=68.17520680911885
bound_log2=70
PASS: noise after noise squashing and packing respects bound.
ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 7581 filtered out; finished in 12797.76s
```

### Check-list:

### CODE REVIEW

File: backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh [hunk 1]
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+  *buffer = new pbs_buffer_128<InputTorus, PBS_TYPE::CLASSICAL>(
+      stream, gpu_index, lwe_dimension, glwe_dimension, polynomial_size,
+      level_count, input_lwe_ciphertext_count,
+      PBS_VARIANT::TBC_SINGLE_ITERATION, allocate_gpu_memory,
+      noise_reduction_type, size_tracker);
+  return size_tracker;
+}
+
+template <typename InputTorus, class params>
+__host__ void host_programmable_bootstrap_single_iteration_tbc_128(
```
pdroalves, line 1179: I really don't like `single_iteration`. We are still running all the iterations, but from the host. I suggest renaming to `host_driven` or maybe `step_launch`.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
 }
 
+#if CUDA_ARCH >= 900
+// Single-iteration TBC flavor. Differs from
+// device_programmable_bootstrap_tbc_128
+// in three ways:
+//   * the blind-rotation loop runs on the host, one launch per iteration, so
+//     concurrent squashes interleave instead of each pinning a slice of the GPU
+//     for a whole rotation;
+//   * a cluster is (glwe_dimension + 1) blocks with no level dimension: every
```
pdroalves, line 943: Another thing that came to my mind is that by serializing the level decomposition we are reducing the cluster size. This probably delivers a cheaper clustrer synchronization and maybe other things.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+//     block decomposes all level_count levels and publishes them, so the GGSW
+//     terms come from the peers' shared memory and two cluster barriers
+//     suffice;
+//   * the accumulator stays in registers, which is what lets the three
+//     published buffers fit in 96 KB at two blocks per SM.
+
+template <typename InputTorus, class params, bool first_iter, bool last_iter,
```
pdroalves, line 950: `first_iter` and `last_iter` are checks to be used in conditionals, right? Let`s add the preffix `is_` so it's easier to reconize what they are? I don't think we do it usually but It's an idea that just came to my mind.

pdroalves: If you agree, in times of AI, we can easily modify the other places we use this approach.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+                                                   (uint32_t)a_hat);
+
+  // The relaxed fft version doesn't sync at enter nor exit to allow some
+  // overlapping between the levels.
+  __syncthreads();
+#pragma unroll 1
+  for (uint32_t l = 0; l < level_count; l++) {
+    double *level = get_current_fft_level<params>(accumulator_fft, l);
+    negacyclic_forward_fft_f128_relaxed<HalfDegree<params>>(
+        level + 0 * half_degree, level + 1 * half_degree,
```
pdroalves, line 1023: These are pointers to an specific level. I find it harder to read like this. Can you move these pointers to new variables as `level_0`, `level_1`, ... ? I think the compiler is smart enough not to increase register use.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+      cluster_group, params, level_count, glwe_dimension>(
+      acc_re_hi, acc_re_lo, acc_im_hi, acc_im_lo, accumulator_fft,
+      bootstrapping_key, lwe_iteration, cluster, this_block_rank);
+
+  // Consumes the accumulator straight from registers and takes the barrier that
+  // frees the published levels partway through, once its register-only levels
+  // are done.
+  // The relaxed ifft does a cluster sync before using the shared mem.
+  negacyclic_backward_fft_f128_relaxed<cluster_group, HalfDegree<params>>(
+      accumulator_fft + 0 * half_degree, accumulator_fft + 1 * half_degree,
```
pdroalves, line 1041: Same here regarding the new variables.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+
+  // Level 1 is only free after that barrier, so the torus accumulator is loaded
+  // here rather than alongside the transform.
+  __uint128_t *torus =
+      (__uint128_t *)get_current_fft_level<params>(accumulator_fft, 1);
+  // No barrier needed: the copy and add_to_torus_128 touch the same four
+  // positions in the same thread.
+  copy_polynomial_rolled<__uint128_t, params::opt, stride>(global_slice, torus);
+  add_to_torus_128<__uint128_t, params>(
+      accumulator_fft + 0 * half_degree, accumulator_fft + 1 * half_degree,
```
pdroalves, line 1053: Same here regarding the new variables.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
pdroalves: Or do you think that would bloat the code?

pdroalves: Or perhaps we could introduce a macro to encapsulate this. I see other repetitions in this function.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+template <class params, typename InputTorus>
+constexpr uint64_t get_buffer_size_single_iteration_tbc_128() {
+  // Three Fourier polynomials (the three published decomposition levels) plus
+  // two words per warp of scratch for the block-cooperative centered
+  // correction.
+  // 96 KB at the noise-squashing shape, which holds two blocks per SM AND stays
+  // under the ~99.3 KB shared-carveout cliff: the carveout is quantized, and
+  // crossing it drops L1 below the twiddle table's footprint, measured as a 20%
+  // throughput loss with occupancy unchanged.
+  return 3ul * (uint64_t)(params::degree / 2 * 4) * sizeof(double) +
```
pdroalves, line 1082: Here you hardcode "3", but in the kernel you use the content of `level_count`. This is not a good idea.

Perhaps we could have these values as constants in this file, since these methods as anyway specialized.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+template <typename InputTorus, typename params>
+__host__ uint64_t scratch_programmable_bootstrap_single_iteration_tbc_128(
+    cudaStream_t stream, uint32_t gpu_index,
+    pbs_buffer_128<InputTorus, PBS_TYPE::CLASSICAL> **buffer,
+    uint32_t lwe_dimension, uint32_t glwe_dimension, uint32_t polynomial_size,
+    uint32_t level_count, uint32_t input_lwe_ciphertext_count,
+    bool allocate_gpu_memory, PBS_MS_REDUCTION_T noise_reduction_type) {
+
+  cuda_set_device(gpu_index);
+  configure_single_iteration_tbc_128<InputTorus, params, 24, 3, 2>();
```
pdroalves, line 1167: Here we have the same problem as the `level_count`. We cannot leave these values hardocded like this.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+    pbs_buffer_128<InputTorus, PBS_TYPE::CLASSICAL> *buffer,
+    uint32_t glwe_dimension, uint32_t lwe_dimension, uint32_t polynomial_size,
+    uint32_t base_log_rt, uint32_t level_count_rt,
+    uint32_t input_lwe_ciphertext_count) {
+
+  // The flavor is compiled for exactly the noise-squashing shape: the GGSW
+  // product, the decomposition and the cluster geometry are all compile-time
+  // constants, and the three published Fourier buffers only fit in 96 KB at
+  // N = 2048.
+  constexpr uint32_t base_log = 24;
```
pdroalves, line 1192: Here we have the same problem as the `level_count`. We cannot leave these values hardcoded like this.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+                 "Cuda error: the single-iteration TBC implementation of the "
+                 "128-bit "
+                 "programmable bootstrap only supports the noise-squashing "
+                 "parameters (N=2048, k=2, l=3, Bg=2^24) on a GPU with "
+                 "distributed shared memory.");
+
+  cuda_set_device(gpu_index);
+  auto noise_reduction_type = buffer->noise_reduction_type;
+  auto *global_accumulator = buffer->global_accumulator;
+  configure_single_iteration_tbc_128<InputTorus, params, base_log, level_count,
```
pdroalves, line 1208: We already call this in the scratch function, do we need it here too? Otherwise we are adding many unneeded `cudaFuncSetAttribute` calls.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+  __uint128_t *global_slice =
+      global_accumulator +
+      (blockIdx.y + blockIdx.x * (glwe_dimension + 1)) * params::degree;
+
+  // Put "a" in [0, 2N[
+  InputTorus a_hat = 0;
+  modulus_switch<InputTorus>(block_lwe_array_in[lwe_iteration], a_hat,
+                             log_modulus);
+
+  if (first_iter) {
```
pdroalves, line 988: We use `if constexpr (...)` in other parts of this file.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
```diff
@@ -931,6 +933,315 @@ supports_thread_block_clusters_on_classic_programmable_bootstrap_128(
...
+      (__uint128_t *)get_current_fft_level<params>(accumulator_fft, 1);
+  // No barrier needed: the copy and add_to_torus_128 touch the same four
+  // positions in the same thread.
+  copy_polynomial_rolled<__uint128_t, params::opt, stride>(global_slice, torus);
+  add_to_torus_128<__uint128_t, params>(
+      accumulator_fft + 0 * half_degree, accumulator_fft + 1 * half_degree,
+      accumulator_fft + 2 * half_degree, accumulator_fft + 3 * half_degree,
+      torus);
+
+  if (last_iter) {
```
pdroalves, line 1057: We use `if constexpr (...)` in other parts of this file.

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -223,6 +223,192 @@ polynomial_accumulate_in_fourier_domain_128(double *result, double *x,
...
+template <bool negate>
+__device__ __forceinline__ void
+relaxed_fp128_accumulate_term(double &dst_hi, double &dst_lo, double xh,
+                              double xl, double yh, double yl) {
+  const double signed_xh = negate ? -xh : xh;
+  const double signed_xl = negate ? -xl : xl;
+
+  const double p = signed_xh * yh;
+  const double e = __fma_rn(signed_xh, yh, -p);
+  const double s = dst_hi + p;
```
pdroalves, line 277: The computation of `s` and `bb` here is problematic. We are using `+` / `-` instead of ` __dadd_rn` / `__dsub_rn`. This allows the compiler to contract this in a single `fma()`. If that happens, the precision changes and the result might differ.

We need to replace this by the intrinsics as done elsewhere.

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -223,6 +223,192 @@ polynomial_accumulate_in_fourier_domain_128(double *result, double *x,
...
+// four terms that OPEN a lane, which is why the caller peels one product and
+// not one level.
+template <bool negate>
+__device__ __forceinline__ void
+relaxed_fp128_open_term(double &dst_hi, double &dst_lo, double xh, double xl,
+                        double yh, double yl) {
+  const double signed_xh = negate ? -xh : xh;
+  const double signed_xl = negate ? -xl : xl;
+
+  const double p = signed_xh * yh;
```
pdroalves, line 300: Here we should be using `_dmul_rn` to match the convention. Don't think we would have any problem with precision in this one, though.

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -223,6 +223,192 @@ polynomial_accumulate_in_fourier_domain_128(double *result, double *x,
...
+// exactly ONE pair (the one at 2*threadIdx.x, relative to the pointers given).
+// opt only decides how many times the caller invokes it: opt=4 is one pair per
+// thread, opt=8 is two, at window offsets g * 2 * (degree/opt).
+template <class params, bool opening>
+__device__ void
+polynomial_product_accumulate_in_fourier_domain_128_pairs_relaxed_regs(
+    double2 &acc_re_hi, double2 &acc_re_lo, double2 &acc_im_hi,
+    double2 &acc_im_lo, double2 a_re_hi, double2 a_re_lo, double2 a_im_hi,
+    double2 a_im_lo, const double *__restrict__ second) {
+  static_assert(params::opt == 4 || params::opt == 8,
```
pdroalves, line 366: This assert allows `params::opt == 8` but I don't think this function implements the second half of operations for this case, does it?

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
```diff
@@ -223,6 +223,192 @@ polynomial_accumulate_in_fourier_domain_128(double *result, double *x,
...
+  relaxed_fp128_accumulate_complex_slot<opening>(
+      acc_re_hi.x, acc_re_lo.x, acc_im_hi.x, acc_im_lo.x, a_re_hi.x, a_re_lo.x,
+      a_im_hi.x, a_im_lo.x, b_re_hi.x, b_re_lo.x, b_im_hi.x, b_im_lo.x);
+  relaxed_fp128_accumulate_complex_slot<opening>(
+      acc_re_hi.y, acc_re_lo.y, acc_im_hi.y, acc_im_lo.y, a_re_hi.y, a_re_lo.y,
+      a_im_hi.y, a_im_lo.y, b_re_hi.y, b_re_lo.y, b_im_hi.y, b_im_lo.y);
+}
+
+// Same product with the left operand read from memory.
+template <class params, bool opening, bool first_is_shared = false>
```
pdroalves, line 390: I couldn't find a call to this method with `first_is_shared == false`. All of them set it to `true`. Do we really need this case covered?

pdroalves (changes requested):

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -892,6 +920,323 @@ __host__ void host_fourier_transform_backward_as_torus_f128(
...
+        v[i] = w;
+      } else {
+        u[i] = w;
+      }
+      w = NEG_TWID(tid / lane_mask + twiddle_shift);
+      f128::cplx_f128_relaxed_butterfly_assign(u[i].re, u[i].im, v[i].re,
+                                               v[i].im, w.re, w.im);
+      tid = tid + STRIDE;
+    }
+  }
```
pdroalves, line 1021: I'm not familiar enough with the FFT implementation to properly review this, so I will just leave my agent review suggestion for this line. It's saying we need a `_syncthreads()` here.

pdroalves: *HIGH — Missing `__syncthreads()` between Part 1 and the final store in `negacyclic_forward_fft_f128_relaxed`* 
- *Location:* Part 1 loop is lines 985-1021 (last iteration `l=6`, `lane_mask=32`); Part 2 loop is lines 1024-1061 (`__syncwarp()` only); final store loop is lines 1063-1069. No `__syncthreads()` between line 1021 (Part 1 ends) and line 1063 (final store begins).
- Mechanism: at `l=6` (Part 1's last iteration), each thread reads `dt_*[tid ^ 32]` — since a warp is 32 threads, this always crosses a warp boundary. That read is entered-guarded by a `__syncthreads()` but has no exit guard. Part 2 uses only `__syncwarp()` (intra-warp only). A fast warp can race through Part 2 and hit the final store (line 1063-1069, writing the entire `[0, DEGREE)` range) before a slower warp finishes its Part-1 l=6 read — a write-after-read hazard across warps with nothing to prevent it.
- Compare: `negacyclic_backward_fft_f128_relaxed` (starts line 1084) does have `__syncthreads()` before its own final store (line ~1227, right before "store registers in SM" at 1228) — the two new functions are inconsistent, suggesting this is an oversight.
- Fix: add `__syncthreads()`; either right after line 1021 or right before line 1063.

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -29,6 +29,34 @@ using Index = unsigned;
...
+// puts thread t at byte 16t, which is a 4-way shared-memory bank conflict; the
+// two coefficients are adjacent, so each plane becomes one 16-byte access
+// instead. Same bytes, same values, half the instructions, conflict free.
+#define F128x2_PAIR_TO_F64x4(lo_reg, hi_reg, pair_ind)                         \
+  *(double2 *)&dt_re_hi[pair_ind] = make_double2(lo_reg.re.hi, hi_reg.re.hi);  \
+  *(double2 *)&dt_re_lo[pair_ind] = make_double2(lo_reg.re.lo, hi_reg.re.lo);  \
+  *(double2 *)&dt_im_hi[pair_ind] = make_double2(lo_reg.im.hi, hi_reg.im.hi);  \
+  *(double2 *)&dt_im_lo[pair_ind] = make_double2(lo_reg.im.lo, hi_reg.im.lo)
+
+#define F64x4_TO_F128x2_PAIR(lo_reg, hi_reg, pair_ind)                         \
```
pdroalves, line 44: This is never used.

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
```diff
@@ -892,6 +920,323 @@ __host__ void host_fourier_transform_backward_as_torus_f128(
...
+    F128x2_TO_F64x4(v[i], tid + HALF_DEGREE);
+
+    tid = tid + STRIDE;
+  }
+  __syncthreads();
+}
+
 #undef NEG_TWID
 #undef F64x4_TO_F128x2
 #undef F128x2_TO_F64x4
```
pdroalves, line 1242: We need to undefine `F128x2_PAIR_TO_F64x4` and `F64x4_TO_F128x2_PAIR` (if we keep this last one).

pdroalves (changes requested):

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -187,11 +187,193 @@ __device__ Torus decompose_one(Torus &state, Torus mask_mod_b, int base_log) {
...
+//
+// Iterates pairs because coefficients i and i + opt/2 share one packed output
+// slot; they sit exactly degree/2 apart.
+template <typename T, class params, uint32_t base_log, uint32_t level_count>
+__device__ void
+rotate_and_decompose_first_level_128(double *result, uint64_t *state,
+                                     const T *accumulator, uint32_t j) {
+  constexpr int degree = params::degree;
+  constexpr int half_degree = degree / 2;
+  static_assert(base_log <= 52, "a decomposition digit must fit in 53 bits for "
```
pdroalves, line 247: Not very serious, but we have no bound for `level_count`. 

Perhaps change this assert to `base_log * (level_count - 1)` to make sure that it fits in 64 bits?

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -187,11 +187,193 @@ __device__ Torus decompose_one(Torus &state, Torus mask_mod_b, int base_log) {
...
+    out_re_hi[tid] = out_re.hi;
+    out_re_lo[tid] = out_re.lo;
+    out_im_hi[tid] = out_im.hi;
+    out_im_lo[tid] = out_im.lo;
+
+    tid += params::degree / params::opt;
+  }
+}
+
 // Follows the same logic than 2_2 params decomposition but for 128-bit
```
pdroalves, line 285: I think this comment got out of place.

pdroalves (changes requested):

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -427,6 +427,25 @@ struct pbs_buffer_128<InputTorus, PBS_TYPE::CLASSICAL>
...
     } break;
+    case PBS_VARIANT::TBC_SINGLE_ITERATION: {
+      // The step1 and step2 are fusioned using tbc but only performs a single
+      // iteration. This gives better throughput cause short kernels are better
+      // to spread on the SMs.
+      //
+      // d_mem and global_join_buffer are still allocated at size zero:
+      // release() frees them unconditionally, so leaving them unassigned makes
+      // the drop path free an uninitialized pointer.
+      d_mem = (int8_t *)cuda_malloc_with_size_tracking_async(
```
pdroalves, line 438: Shouldn't we use `d_mem = nullptr` in this case? I think we discussed this at some point but I don't remember the conclusion.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
```diff
@@ -427,6 +427,25 @@ struct pbs_buffer_128<InputTorus, PBS_TYPE::CLASSICAL>
...
+          0, stream, gpu_index, size_tracker, allocate_gpu_memory);
+      global_join_buffer = (double *)cuda_malloc_with_size_tracking_async(
```
pdroalves, line 440: Here too.

backends/tfhe-cuda-backend/cuda/src/crypto/packing_keyswitch.cuh
```diff
@@ -162,10 +162,9 @@ __host__ void host_packing_keyswitch_lwe_list_to_glwe(
   // pre-zeroing and the decomposed masks never round-trip through global
   // memory. Only level counts 1..9 are dispatched; the library's maximum is 4.
   //
-  // 32x32 tile with BK=4, TM=8 (128 threads/block): keeps occupancy high and
-  // avoids wasting output rows at the modest batch sizes (num_lwes) typical
-  // of the packing keyswitch.
-  launch_packing_gemm_1d<Torus, 32, 32, 4, 8>(
+  // This tiling attains better performance for the packing ks params with
```
pdroalves, line 165: Why did you remove the tile description?

pdroalves (changes requested):

pdroalves:
Another thing that just came to my mind: not sure if `TFHE_RS_GPU_PBS128_RELAXED` is affecting the relaxing-arithmetic used in the DEFAULT PBS FFT. I think we are just gating the TBC PBS with it, but if we merge this PR we would have the default PBS running the new arithmetic. Is that what we want?

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: yup i struggled to find a good naming, i am changing it to host driven.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: yup, cluster syncs of less blocks are cheaper indeed, and we can reduce the number of syncs too!

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: Done in the new kernel. Left the existing kernels alone to keep this diff reviewable

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: Those were the four planes of one Fourier polynomial (re_hi, re_lo, im_hi, im_lo), not levels. The relaxed transforms now take the polynomial base pointer and split internally.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: changed it too!

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: This one i kept it, cause it is shared with other flavors, and didn't want to modify those on this PR

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: Done

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: Factored into constants at the top of the new host_driven section

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: i reuse the new constants here now.

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: Indeed, removed it since the scratch already does it

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/pbs/programmable_bootstrap_classic_128.cuh
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
guillermo-oyarzun: Good catch! you are totally right, better use the intrinsics and force the compiler's hand

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
guillermo-oyarzun: done

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
guillermo-oyarzun: Removed, the left operand is always in shared memory

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
guillermo-oyarzun: ahh right, i forgot to port it to this one, it was the same error than on main.

backends/tfhe-cuda-backend/cuda/src/polynomial/polynomial_math.cuh
guillermo-oyarzun: right, removed it

backends/tfhe-cuda-backend/cuda/src/fft128/fft128.cuh
guillermo-oyarzun: indeed, it was one of the failed trials. removed!

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
guillermo-oyarzun: added the assert

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
guillermo-oyarzun: right it belong to the function below, moved it.

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
guillermo-oyarzun: right, changed it

backends/tfhe-cuda-backend/cuda/include/pbs/pbs_utilities.h
guillermo-oyarzun: changed it

backends/tfhe-cuda-backend/cuda/src/crypto/packing_keyswitch.cuh
guillermo-oyarzun: No reason, ai decided to remove it and didn't notice it (probably cause i ask it shorter comments), added it back.

guillermo-oyarzun:
> Another thing that just came to my mind: not sure if `TFHE_RS_GPU_PBS128_RELAXED` is affecting the relaxing-arithmetic used in the DEFAULT PBS FFT. I think we are just gating the TBC PBS with it, but if we merge this PR we would have the default PBS running the new arithmetic. Is that what we want?

Actually the default is not affected cause we are using a different fft/ifft in this flavor (i didn't want to touch the general ffts for safety)

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
```diff
@@ -187,6 +187,190 @@ __device__ Torus decompose_one(Torus &state, Torus mask_mod_b, int base_log) {
...
+  constexpr uint64_t mask_mod_b = (1ull << base_log) - 1ull;
+  static_assert(base_log <= 52, "a decomposition digit must fit in 53 bits for "
+                                "small_signed_to_f128 to be exact");
+
+  uint32_t tid = threadIdx.x;
+#pragma unroll
+  for (int i = 0; i < params::opt / 2; i++) {
+    const int j = i + params::opt / 2;
+    uint64_t in_re = state[i], in_im = state[j];
+    uint64_t res_re = in_re & mask_mod_b;
```
pdroalves, line 314: We have `decompose_one` to encapsulate this logic. Perhaps we could use it instead of duplicating it here?

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
```diff
@@ -18,6 +18,19 @@ __device__ void copy_polynomial(const T *__restrict__ source, T *dst) {
     tid = tid + block_size;
   }
 }
+// copy_polynomial with the unroll capped. A full unroll inflates the live
+// range,
+// which costs in kernels that are already spilling; measured 1.4% latency and
+// 2.9% throughput on the host-driven TBC 128-bit PBS.
+template <typename T, int elems_per_thread, int block_size>
+__device__ void copy_polynomial_rolled(const T *__restrict__ source, T *dst) {
```
pdroalves, line 26: Why do we need this new copy? It seems almost identical to `copy_polynomial`.

pdroalves (changes requested):
Just some minor observations.

backends/tfhe-cuda-backend/cuda/src/crypto/gadget.cuh
guillermo-oyarzun: here the base_log is a template param so resolved at compilation time.

backends/tfhe-cuda-backend/cuda/src/polynomial/functions.cuh
guillermo-oyarzun: The problem is the unroll, here we need to avoid the full unrolling, if not the registers explode. It was the simpler way i found, if you find a better way to enable or disable it, i am all ears

pdroalves (approved):

IceTDrinker:
just got a request for review but this was merged before I got a chance to look ? 😅 @pdroalves @guillermo-oyarzun did you need a review or not ?

guillermo-oyarzun:
> just got a request for review but this was merged before I got a chance to look ? 😅 @pdroalves @guillermo-oyarzun did you need a review or not ?

i was too excited and merged it right after Pedro's approval. From the rust lib i did minor changes related with adding some extra tests. Feel free to review those, i think there are on the safe side.

PR #3886, opened 2026-09-01
chore(ci): add version bump check for CUDA backend crates on PRs

pdroalves:
Bumping CUDA backend crate versions currently relies on manual awareness, which is error-prone. This PR adds a CI workflow that checks each PR for changes to the three CUDA backend crates (`tfhe-cuda-common`, `tfhe-cuda-backend`, `zk-cuda-backend`) and fails if the crate version was not bumped.

The check compares the PR's changes against the base branch (not the last release tag), so it only flags version bumps the PR itself needs to provide. On failure, it prints the full diff of detected changes to help the author identify what triggered the check.

### DIFF
--- a/scripts/check_cuda_version_bump.sh
+++ b/scripts/check_cuda_version_bump.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Each entry: tag_prefix:crate_dir:cargo_toml
+CRATES=(
+    "tfhe-cuda-common:backends/tfhe-cuda-common:backends/tfhe-cuda-common/Cargo.toml"
+    "tfhe-cuda-backend:backends/tfhe-cuda-backend:backends/tfhe-cuda-backend/Cargo.toml"
+    "zk-cuda-backend:backends/zk-cuda-backend:backends/zk-cuda-backend/Cargo.toml"
+)
+
+# In pull_request CI, actions/checkout produces a merge commit whose first
+# parent is the base branch. Compare against that so we only flag changes
+# introduced by the PR, not pre-existing unreleased changes on main.
+if [[ -n "${GITHUB_BASE_REF:-}" ]]; then
+    base_ref="HEAD^1"
+    echo "PR context: comparing against base branch (${GITHUB_BASE_REF})"
+else
+    base_ref=""
+    echo "Non-PR context: comparing against latest release tags"
+fi
+
+failed=0
+
+for crate_info in "${CRATES[@]}"; do
+    IFS=':' read -r tag_prefix crate_dir cargo_toml <<< "$crate_info"
+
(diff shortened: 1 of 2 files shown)


scripts/check_cuda_version_bump.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "Non-PR context: comparing against latest release tags"
+fi
+
+failed=0
+
+for crate_info in "${CRATES[@]}"; do
+    IFS=':' read -r tag_prefix crate_dir cargo_toml <<< "$crate_info"
+
+    current_version=$(grep '^version' "$cargo_toml" | head -1 | sed 's/.*"\(.*\)".*/\1/')
+
```
IceTDrinker, line 30: prefer cargo metadat with the Cargo.toml path and then jq to get the version

e.g.

for tfhe e.g.:

cargo metadata --format-version 1 --no-deps -m tfhe/Cargo.toml | jq -r 'first(.packages[] | select(.name == "tfhe")) | .version'

1.7.0

IceTDrinker: it seems since it's a workspace passing Cargo.toml works for all workspace deps

scripts/check_cuda_version_bump.sh
```diff
@@ -0,0 +1,92 @@
...
+        if git diff --quiet "$base_ref" HEAD -- "$crate_dir"; then
+            echo "[OK] ${tag_prefix}: no changes in this PR"
+            continue
+        fi
+
+        base_version=$(git show "${base_ref}:${cargo_toml}" 2>/dev/null \
+            | grep '^version' | head -1 | sed 's/.*"\(.*\)".*/\1/' || echo "")
+
+        if [[ -z "$base_version" ]]; then
+            echo "[OK] ${tag_prefix}: new crate added in this PR (version ${current_version})"
```
IceTDrinker: new crate added ?

andrei-stoian-zama:
Can we have a version on the release workflow ? it should work on main but also on custom release branches - do you think this could work @IceTDrinker ? e.g. releasing on branch `release/1.7.x` should check the latest tag for 1.7 -> so we bump only the minor. but releasing on main should check the whole semver.

IceTDrinker:
possible to parse the target branch and infer potential version, this way the workflow/script does not change

scripts/check_cuda_version_bump.sh
pdroalves: Thanks! `cargo metadata` is way more robust for this.

scripts/check_cuda_version_bump.sh
pdroalves: Yeah, we don't need that. I'm removing it.

pdroalves:
@andrei-stoian-zama Not sure I'm following this idea. I'm not very familiar with the release workflow.

andrei-stoian-zama (approved):

PR #3895, opened 2026-09-02
chore: fix CI node cache not being used right away

IceTDrinker:


IceTDrinker:
it seems a workflow with workflow call may not be able to access cache entries from the "base branch"... which may explain the weird behavior we encounter

.gitignore
```diff
@@ -50,6 +50,9 @@ ci/lattice_estimator.sage.py
 
 # Node
 safe_chain_install.sh
+# when setup on CI
+.nvm
+.npm
```
tmontaigu, line 55: not sure this is actually useful as CI does not do commits, but its not harmful either

tmontaigu (approved):

.gitignore
IceTDrinker: Just trying to think ahead if we ever need some automated commit to some branches by a bot, it should not try to grab those dirs

PR #3901, opened 2026-09-03
chore(gpu): make scratch_cuda_* functions truly asynchronous and enforce it in check_scratch_cleanup.py

pdroalves:
Part of https://github.com/zama-ai/tfhe-rs-internal/issues/1482 (split 2 of 3, replaces #3896).

### What

69 of the 80 `scratch_cuda_*_async` FFI functions reached `cuda_synchronize_stream` through their buffer constructors, at 17 synchronization sites. This PR removes 12 of them. The 5 remaining sites are inside the `generate_device_accumulator*` primitives and are kept on purpose (see below). No FFI signature changes.

A new check in `scripts/check_scratch_cleanup.py` (Check 7, implemented in `scripts/check_scratch_sync.py`) walks the C++ call graph from every `scratch_cuda_*` symbol in `bindings.rs` and fails if any of them can reach a host synchronization outside an explicit whitelist.

### Changes

- **`int_radix_lut` multi-GPU index broadcast** (68 scratch functions): host sync replaced by `multi_gpu_broadcast_barrier.local_streams_wait_for_stream_0`, guarded on `allocate_gpu_memory && count() > 1`. Single GPU needs no ordering since the copy runs on stream 0. The barriers are now created before the buffers that use them.
- **Constructor temporaries that were sync-then-freed** (`int_seq_group_prop_memory`, `int_prop_simu_group_carries_memory`, `int_sum_ciphertexts_vec_memory`, `div_rem.h`, `oprf.h`, `rerand_utilities.h`, `kv_store_utilities.h`, `zk_utilities.h`): each host staging buffer becomes a member freed in `release()` after the sync that already exists there. These buffers are written once, in the constructor, and never rewritten. Two syncs that protected nothing are deleted (`int_bit_extract_luts_buffer`, `zk_expand_mem`, `RERAND_NO_KS` branch).
- **`int_sum_ciphertexts_vec_memory` dry-run LUT**: the size probe (built with `allocate_gpu_memory == false`) is kept as a member and released in `release()`. The syncs in `int_radix_lut::release()` are guarded on `gpu_memory_allocated`, since nothing was ever enqueued for a probe.

### What is not changed: `generate_device_accumulator*`

The 6 `generate_device_accumulator*` primitives still allocate a host LUT, upload it, synchronize and free it. A first version of this PR replaced that with a per-LUT host staging slot and an asynchronous upload. That was reverted after review: some `host_*` functions regenerate a LUT into the same `int_radix_lut` on every call (`are_all_comparisons_block_true`, `tree_sign_reduction`, `integer_radix_unsigned_scalar_difference_check`, `integer_radix_signed_scalar_difference_check`, `reduce_signs`, and `setup_lookup_tables` from `host_integer_partial_sum_ciphertexts_vec`). An asynchronous upload from a reused host buffer would race with the next regeneration, and relying on `cudaMemcpyAsync` staging pageable memory internally is not acceptable.

As a result, 65 of the 80 scratch functions still synchronize once per LUT they build. The follow-up (https://github.com/zama-ai/tfhe-rs-internal/issues/1557) is to generate every LUT variant in scratch and select among them in the host function, after which the primitives can become asynchronous and the entry points leave the whitelist below.

### Bugs fixed along the way

- `zk_utilities.h` and general-path `create_indexes_for_overflow_sub` freed upload sources with no sync (now members).
- `int_prop_simu_group_carries_memory::release()` deleted `h_scalar_array_cum_sum` before its sync (moved after).
- `setup_lookup_tables` leaked a `luts_message_carry` per partial sum (now reused when large enough, rebuilt otherwise).

### Static analysis (Check 7)

- Implemented with tree-sitter-cpp in `scripts/check_scratch_sync.py`, separate from the other checks so it can be disabled by removing its call from `check_scratch_cleanup.py`. CUDA attributes (`__host__`, `__global__`, ...) are blanked before parsing.
- Follows plain calls, member calls (receiver type resolved through local declarations, struct fields and `using` aliases), member-initializer lists, calls inside macro arguments such as `DISPATCH_POLY_SIZE`, and `new T<...>(...)` constructor calls, which is how every scratch function builds its buffer.
- Scans `tfhe-cuda-common` in addition to the backend, because the memcpy wrappers there call `cuda_set_device`.
- `SYNC_SITE_NAMES` (in the sync module) says what a host synchronization looks like; meeting one is a violation. `SYNC_WALK_STOP_CALLS` (same module) holds the runtime helpers the walker does not enter: `cuda_setup_mempool` (one-time pool warm-up reached by every `cuda_set_device`) and the two allocation wrappers whose `cudaMalloc`/`cudaFree` fallback depends on device support.
- `CHECK7_SYNC_WHITELIST` in `check_scratch_cleanup.py` is passed to the check and names the top-level scratch entry points known to synchronize. It holds entry points, never the subfunction that synchronizes: skipping a subfunction would hide the synchronization and let every caller above it pass as asynchronous. Whitelisted chains are still walked, and a whitelisted name that does not reach a synchronization is reported as stale.
- It currently lists the 65 entry points whose buffers build a LUT through the `generate_device_accumulator*` primitives. With the whitelist emptied, Check 7 reports those 65 of 80, all through the five sync sites in those primitives.
- The expected number of scratch entry points (80) is pinned.

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
@@ -621,8 +621,14 @@ struct int_radix_lut_custom_input_output {
     multi_gpu_alloc_array_async(active_streams, lwe_trivial_indexes_vec,
                                 num_radix_blocks, size_tracker,
                                 allocate_gpu_memory);
-    cuda_synchronize_stream(active_streams.stream(0),
-                            active_streams.gpu_index(0));
+
+    // lwe_trivial_indexes was filled from the host on stream 0 by
+    // setup_lwe_trivial_indices, while the copies below run on the worker
+    // streams, so those must wait for stream 0. On a single GPU the only copy
+    // runs on stream 0 itself and needs no ordering.
+    if (allocate_gpu_memory && active_streams.count() > 1)
+      multi_gpu_broadcast_barrier.local_streams_wait_for_stream_0(
+          active_streams);
 
     // This call will not copy if allocate_gpu_memory is false
     // thus it's safe to call it on a null source pointer
@@ -674,14 +680,17 @@ struct int_radix_lut_custom_input_output {
 
   void setup_multi_gpu(int_radix_params params, uint32_t num_radix_blocks,
                        bool allocate_gpu_memory, uint64_t &size_tracker) {
-    alloc_and_init_multi_gpu_buffers(params, num_radix_blocks,
-                                     allocate_gpu_memory, size_tracker);
-
+    // The barriers are created first because
+    // alloc_and_init_multi_gpu_buffers orders its peer copies with
+    // multi_gpu_broadcast_barrier.
     if (active_streams.count() > 1) {
       multi_gpu_gather_barrier.create_on(active_streams);
       multi_gpu_broadcast_barrier.create_on(active_streams);
       multi_gpu_scatter_barrier.create_on(active_streams);
     }
+
+    alloc_and_init_multi_gpu_buffers(params, num_radix_blocks,
+                                     allocate_gpu_memory, size_tracker);
   }
 
   int_radix_lut_custom_input_output(CudaStreams streams,
@@ -955,10 +964,10 @@ public:
     multi_gpu_broadcast_barrier.local_streams_wait_for_stream_0(
         new_active_streams);
 
-    // The LUT and its indexes reside on GPU 0
-    // these were filled by calls to generate_device_accumulator
-    // due to the previous synchronization, we're sure these buffers have
-    // finished copying to GPU 0 from CPU
+    // The LUT and its indexes reside on GPU 0, filled by calls to
+    // generate_device_accumulator on stream 0. The event barrier above was
+    // recorded on that same stream, so the worker streams below only start
+    // reading once those host to device copies have completed.
     auto src_lut = lut_vec[0];
     auto src_lut_indexes = lut_indexes_vec[0];
 
@@ -1148,6 +1157,10 @@ public:
       std::vector<OutputTorus *> cpu_prealloc_buffers = {},
       std::optional<int> factor = std::nullopt) {
     // streams should be a subset of active_streams
+    GPU_ASSERT(
+        cpu_prealloc_buffers.empty() || gpu_memory_allocated,
+        "LUT Generation with pre-allocated CPU buffer expects "
+        "gpu_memory_allocated==True ");
 
     if constexpr (!std::is_same_v<IndexGenerator, std::nullptr_t>) {
       set_lut_indexes(streams, index_generator, num_blocks);
@@ -1211,8 +1224,12 @@ public:
         active_streams.gpu_index(0), gpu_memory_allocated);
     lwe_trivial_indexes = nullptr;
 
-    cuda_synchronize_stream(active_streams.stream(0),
-                            active_streams.gpu_index(0));
+    // Nothing was ever enqueued when gpu_memory_allocated is false: this
+    // object is then only used to measure the memory it would take, so there
+    // is no in-flight copy the host frees below could race with.
+    if (gpu_memory_allocated)
+      cuda_synchronize_stream(active_streams.stream(0),
+                              active_streams.gpu_index(0));
     lut_vec.clear();
     lut_indexes_vec.clear();
     free(h_lwe_indexes_in);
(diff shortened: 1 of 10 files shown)


andrei-stoian-zama (changes requested):
some more comments:

- don't combine the generate_ with the generate_.._prealloc it makes things less clear (and they're used in few places so we're not really gaining on verbosity)
- we need to fix comparisons first to avoid relying on non-pinned memory for correctness I don't yet understand the usage in sum ciphertexts, we'll need a review by @bbarbakadze 
- for the script, parse C++ code with an AST parser instead of text parsing

pdroalves:
@andrei-stoian-zama done.

pdroalves:
@andrei-stoian-zama This PR adds a new check in `check_scratch_cleanup.py` to make sure we don't have scratch functions with the `async` suffix but with hidden host synchronizations inside of it. It simply parses the C++ code using brace matching and looks for `cuda_synchronize_stream`. It won't follow macros, though, since them are solved after compilation and this script validates only source code.

backends/tfhe-cuda-backend/cuda/include/integer/integer_utilities.h
```diff
@@ -480,6 +515,29 @@ struct int_radix_lut_custom_input_output {
   std::vector<OutputTorus *> lut_vec;
   std::vector<InputTorus *> lut_indexes_vec;
   InputTorus *h_lut_indexes = nullptr;
+  /// Host staging area for the LUT values, one slot per LUT, mirroring the
+  /// layout of lut_vec[0]. The host to device upload is asynchronous, so the
+  /// source must stay alive until the copy completed: it is freed in release()
+  /// after the stream synchronization.
+  ///
+  /// During scratch each slot is written exactly once, so scratch-time safety
```
andrei-stoian-zama: This is the main problem with this approach: without sync the CPU LUT values may be re-written while an upload is in progress. The solution is to generate all possible luts in scratch and use the right one in the host. 

we can not rely on 

```
 /// int_sum_ciphertexts_vec_memory::setup_lookup_tables regenerates its slots
  /// on every operation. Those rewrites are safe only because
  /// cudaMemcpyAsync from pageable host memory returns after the driver has
  /// copied the source into its own staging buffer. This buffer must therefore
  /// stay pageable malloc memory; making it pinned would turn every one of
  /// those host-side rewrites into a race.
```

andrei-stoian-zama (changes requested):
There are some issues with this:

1. the check_scratch_cleanup.py script is unmaintainable by a human. It's not vital, it's just a helper script. One solution is to move all the parsing stuff to a separate file, so we can easily disable this new check if we find ourselves unable to maintain this
2. the LUT changes with a CPU side staging buffer rely on a quirk to work properly -> memcpyasync is actually using its own staging buffer when not using pinned memory. it's dangerous. see the comment in the review

To accept this PR I would need:
- address point 1
- for point 2 -> I would prefer you don't change the LUT this way, and add an ignore list of host functions that re-generate luts in host_ which will be addressed later

pdroalves:
@andrei-stoian-zama done.

I added the ignore list, but I really don't like to have async functions that aren't async. I opened an issue to proper fix this in the future. https://github.com/zama-ai/tfhe-rs-internal/issues/1557

andrei-stoian-zama:
> @andrei-stoian-zama done.
> 
> I added the ignore list, but I really don't like to have async functions that aren't async. I opened an issue to proper fix this in the future. [zama-ai/tfhe-rs-internal#1557](https://github.com/zama-ai/tfhe-rs-internal/issues/1557)

This only partially addresses the second point of my previous comment and the first point is not addressed at all

pdroalves:
@andrei-stoian-zama the ignore list was moved to `check_scratch_cleanup.py` as `CHECK7_SYNC_WHITELIST`. Moreover, it now lists the scratch entry points that are known to synchronize (those whose buffers build a LUT) instead of the `generate_device_accumulator*` functions. Whitelisted entry points are still walked, and an entry that no longer reaches a sync is reported so the list shrinks as they get fixed. The `generate_*` functions themselves are unchanged and remain synchronous. 

Please see if it matches your expectations.

backends/tfhe-cuda-backend/cuda/include/integer/rerand_utilities.h
```diff
@@ -65,10 +70,6 @@ template <typename Torus> struct int_rerand_mem {
           lwe_trivial_indexes, h_lwe_trivial_indexes,
           safe_mul_sizeof<Torus>(num_lwes), streams.stream(0),
           streams.gpu_index(0), allocate_gpu_memory);
-      cuda_synchronize_stream(streams.stream(0), streams.gpu_index(0));
-      free(h_lwe_trivial_indexes);
-    } else {
```
andrei-stoian-zama, line 70: why did we synchronize here before ?

PR #3906, opened 2026-09-04
chore(gpu): simplify and remove unused fft functions

guillermo-oyarzun:

### PR content/description
Mbarriers used on the bsk key conversion are not really needed since we can use the final version based on named barriers. 
Mbarriers can trigger fake race conditions cause the sanitizer doesn't know how to handle them.
using the same name barriers used in the pbs execution simplifies the bsk conversion. 
no change in data just a different way to perform the syncs
in that way we avoid triggering fake errors from the racecheck and we can test it more often
### Check-list:


andrei-stoian-zama:
Moved to #3909

PR #3909, opened 2026-09-07
fix(gpu): add racecheck on selected kernels, optimize CI

andrei-stoian-zama:
- fix racecheck false alarm in `batch_FFT16x4x16_classical_specialized` (@guillermo-oyarzun)
- move valgrind to use GPU release kernels and CPU debug
- combine memcheck & racecheck workflows on 2xH100 in parallel
- full stress test in gpu full tests manually run
- add stress test to disabled 1xh100 core tests
- add racecheck in addition to memcheck on tests that run selected kernels:

```
batch_NSMFFT (FULLSM)                                                                                                                                
  batch_NSMFFT (NOSM)                                                                                                                                  
  batch_NSMFFT_classical_specialized                                                                                                                   
  batch_FFT16x4x16_classical_specialized
  
  device_programmable_bootstrap_step_one                                                                                                               
  device_programmable_bootstrap_step_two
  device_programmable_bootstrap_cg
  device_programmable_bootstrap_tbc
  device_programmable_bootstrap_tbc_2_2_params
  device_programmable_bootstrap_specialized_2_2_params
  device_programmable_bootstrap_step_one_128
  device_programmable_bootstrap_step_one_128_regs
  device_programmable_bootstrap_step_two_128
  device_programmable_bootstrap_cg_128
  device_programmable_bootstrap_tbc_128
  
  device_multi_bit_programmable_bootstrap_keybundle
  device_multi_bit_programmable_bootstrap_keybundle_2_2_params
  device_multi_bit_programmable_bootstrap_accumulate_step_one
  device_multi_bit_programmable_bootstrap_accumulate_step_two
  device_multi_bit_programmable_bootstrap_cg_accumulate
  device_multi_bit_programmable_bootstrap_tbc_accumulate
  device_multi_bit_programmable_bootstrap_tbc_accumulate_2_2_params
  device_multi_bit_programmable_bootstrap_keybundle_128
  device_multi_bit_programmable_bootstrap_accumulate_step_one_128
  device_multi_bit_programmable_bootstrap_accumulate_step_two_128
  device_multi_bit_programmable_bootstrap_cg_accumulate_128

  keyswitch_zero_output_with_output_indices
  keyswitch_gemm_copy_negated_message_with_indices
  tgemm_all_levels_split_k
  keyswitch_negate_with_output_indices
  cleartext_multiplication
  tgemm_all_levels
  polynomial_accumulate_monic_monomial_mul_many_neg_and_add_C
  accumulate_glwes
[...]

PR #3915, opened 2026-09-07
fix(gpu): add timestamps to compute sanitizer

andrei-stoian-zama:
Add timestamps to compute sanitizer workflow


andrei-stoian-zama:
Cherry picked in #3909

PR #3917, opened 2026-09-08
feat: support dedicated transciphering oprf params

nsarlin-zama:

### PR content/description
Add support for dedicated params for the transciphering oprf. By lowering oprf lwe dimension, we can improve oprf performance.

Some design choices:
- I added a new variant of TranscipheringParameters. This new variant is for the case where we have a dedicated oprf but that still outputs under the general purpose compute key. In the future, I want to add another variant for the case where we also have dedicated params for the transciphering evaluation. This will be needed to support some kreyvium improvements.
- This new variant holds a new type: `OprfParameters`. This will allow to also change the params of the general purpose oprf later. Everything is ready to support this, this is just not "plugged in".
- Backward data for the new variant is added through a client key. This will only test that the parameters load, not that they are correct. In the future, I think we should store the meta params in the HlClientKeyTest, not just the pbs params. But I considered this out of scope for this pr.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)

### DIFF
--- a/tfhe/src/high_level_api/keys/inner.rs
+++ b/tfhe/src/high_level_api/keys/inner.rs
@@ -26,6 +26,7 @@ use crate::integer::public_key::CompactPublicKey;
 use crate::integer::CompressedCompactPublicKey;
 use crate::shortint::atomic_pattern::AtomicPatternParameters;
 use crate::shortint::key_switching_key::KeySwitchingKeyConformanceParams;
+use crate::shortint::oprf::OprfKeyConformanceParams;
 use crate::shortint::parameters::list_compression::CompressionParameters;
 use crate::shortint::parameters::{
     CompactPublicKeyEncryptionParameters, NoiseSquashingCompressionParameters,
@@ -1253,7 +1254,9 @@ impl ParameterSetConformant for IntegerServerKey {
         let oprf_is_ok = match (parameter_set.dedicated_oprf_key, oprf_key.as_ref()) {
             // We have to have a dedicated oprf key
             // Make sure it's there and that it's conformant
-            (true, Some(key)) => key.is_conformant(&parameter_set.sk_param),
+            (true, Some(key)) => key.is_conformant(&OprfKeyConformanceParams::same_as_compute(
+                parameter_set.sk_param,
+            )),
             (true, None) => false,
             // The config says to not use a dedicated oprf key but we have one
             // while it works, it is not strictly conformant
@@ -1266,9 +1269,12 @@ impl ParameterSetConformant for IntegerServerKey {
             transciphering_key.as_ref(),
         ) {
             // The key bootstraps into the compute key, so it is checked against the compute
-            // parameters.
-            (Some(TranscipheringParameters::SameAsCompute), Some(key)) => {
-                key.is_conformant(&parameter_set.sk_param)
+            // parameters, plus the oprf parameters it was generated with.
+            (Some(transciphering_params), Some(key)) => {
+                key.is_conformant(&OprfKeyConformanceParams::new(
+                    parameter_set.sk_param,
+                    transciphering_params.oprf_parameters(parameter_set.sk_param),
+                ))
             }
             (Some(_), None) => false,
             // The config says to not use transciphering but we have a key
@@ -1427,7 +1433,9 @@ impl ParameterSetConformant for IntegerCompressedServerKey {
         let oprf_is_ok = match (parameter_set.dedicated_oprf_key, oprf_key.as_ref()) {
             // We have to have a dedicated oprf key
             // Make sure it's there and that it's conformant
-            (true, Some(key)) => key.is_conformant(&parameter_set.sk_param),
+            (true, Some(key)) => key.is_conformant(&OprfKeyConformanceParams::same_as_compute(
+                parameter_set.sk_param,
+            )),
             (true, None) => false,
             // The config says to not use a dedicated oprf key but we have one
             // while it works, it is not strictly conformant
@@ -1440,9 +1448,12 @@ impl ParameterSetConformant for IntegerCompressedServerKey {
             transciphering_key.as_ref(),
         ) {
             // The key bootstraps into the compute key, so it is checked against the compute
-            // parameters.
-            (Some(TranscipheringParameters::SameAsCompute), Some(key)) => {
-                key.is_conformant(&parameter_set.sk_param)
+            // parameters, plus the oprf parameters it was generated with.
+            (Some(transciphering_params), Some(key)) => {
+                key.is_conformant(&OprfKeyConformanceParams::new(
+                    parameter_set.sk_param,
+                    transciphering_params.oprf_parameters(parameter_set.sk_param),
+                ))
             }
             (Some(_), None) => false,
             // The config says to not use transciphering but we have a key
--- a/tfhe/src/high_level_api/keys/server.rs
+++ b/tfhe/src/high_level_api/keys/server.rs
@@ -856,12 +856,14 @@ mod test {
     use crate::prelude::ParameterSetConformant;
     use crate::shortint::parameters::test_params::TEST_META_PARAM_CPU_2_2_KS_PBS_PKE_TO_SMALL_ZKV2_TUNIFORM_2M128;
     use crate::shortint::parameters::{
-        TranscipheringParameters, COMP_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
+        OprfParameters, TranscipheringParameters,
+        COMP_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
         NOISE_SQUASHING_PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
         PARAM_KEYSWITCH_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
         PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
         PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128,
     };
+    use crate::shortint::prelude::LweDimension;
     use crate::shortint::ClassicPBSParameters;
     use crate::{ClientKey, CompressedServerKey, ConfigBuilder, ServerKey};
 
@@ -965,18 +967,8 @@ mod test {
 
                 modifier(&mut sk_param);
 
-                let sk_param = sk_param.into();
-
-                let conformance_params = IntegerServerKeyConformanceParams {
-                    sk_param,
-                    cpk_param: None,
-                    compression_param: None,
-                    noise_squashing_param: None,
-                    noise_squashing_compression_param: None,
-                    cpk_re_randomization_params: None,
-                    dedicated_oprf_key: true,
-                    transciphering_parameters: None,
-                };
+                let mut conformance_params = IntegerServerKeyConformanceParams::from(config);
+                conformance_params.sk_param = sk_param.into();
 
                 assert!(!sk.is_conformant(&conformance_params));
             }
@@ -995,54 +987,60 @@ mod test {
             let ck = ClientKey::generate(config);
             let sk = ServerKey::new(&ck);
 
-            let sk_param = PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128.into();
-
             cpk_params.encryption_lwe_dimension.0 += 1;
 
-            let conformance_params = IntegerServerKeyConformanceParams {
-                sk_param,
-                cpk_param: Some((cpk_params, casting_params)),
-                compression_param: None,
-                noise_squashing_param: None,
-                noise_squashing_compression_param: None,
-                cpk_re_randomization_params: None,
-                dedicated_oprf_key: true,
-                transciphering_parameters: None,
-            };
+            let mut conformance_params = IntegerServerKeyConformanceParams::from(config);
+            conformance_params.cpk_param = Some((cpk_params, casting_params));
 
             assert!(!sk.is_conformant(&conformance_params));
         }
         {
             let params = PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+            let with_transciphering = ConfigBuilder::with_custom_parameters(params)
+                .enable_transciphering(TranscipheringParameters::SameAsCompute)
+                .build();
+            let without_transciphering = ConfigBuilder::with_custom_parameters(params).build();
+
+            let with_params = IntegerServerKeyConformanceParams::from(with_transciphering);
+            let without_params = IntegerServerKeyConformanceParams::from(without_transciphering);
+
+            let ck = ClientKey::generate(with_transciphering);
+            let sk = ServerKey::new(&ck);
+
+            // Key has transciphering, parameters do not.
+            assert!(!sk.is_conformant(&without_params));
+            assert!(sk.is_conformant(&with_params));
+
+            // Parameters ask for transciphering, key does not have it.
+            let sk_without = ServerKey::new(&ClientKey::generate(without_transciphering));
+            assert!(!sk_without.is_conformant(&with_params));
+        }
+        {
+            let params = PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
+            let dedicated = OprfParameters {
+                lwe_dimension: LweDimension(600),
+            };
+
             let config = ConfigBuilder::with_custom_parameters(params)
+                .enable_transciphering(TranscipheringParameters::DedicatedOprf(dedicated))
+                .build();
+            let same_as_compute_config = ConfigBuilder::with_custom_parameters(params)
                 .enable_transciphering(TranscipheringParameters::SameAsCompute)
                 .build();
 
             let ck = ClientKey::generate(config);
             let sk = ServerKey::new(&ck);
+            let compressed_sk = CompressedServerKey::new(&ck);
 
-            let sk_param = params.into();
-            let mut conformance_params = IntegerServerKeyConformanceParams {
-                sk_param,
-                cpk_param: None,
-                compression_param: None,
-                noise_squashing_param: None,
-                noise_squashing_compression_param: None,
-                cpk_re_randomization_params: None,
-                dedicated_oprf_key: true,
-                transciphering_parameters: None,
-            };
+            let dedicated_oprf_params = IntegerServerKeyConformanceParams::from(config);
+            let same_as_compute_params =
+                IntegerServerKeyConformanceParams::from(same_as_compute_config);
 
-            // Key has transciphering, parameters do not.
-            assert!(!sk.is_conformant(&conformance_params));
-            conformance_params.transciphering_parameters =
-                Some(TranscipheringParameters::SameAsCompute);
-            assert!(sk.is_conformant(&conformance_params));
+            assert!(sk.is_conformant(&dedicated_oprf_params));
+            assert!(compressed_sk.is_conformant(&dedicated_oprf_params));
 
-            // Parameters ask for transciphering, key does not have it.
-            let without = ConfigBuilder::with_custom_parameters(params).build();
-            let sk_without = ServerKey::new(&ClientKey::generate(without));
-            assert!(!sk_without.is_conformant(&conformance_params));
+            assert!(!sk.is_conformant(&same_as_compute_params));
+            assert!(!compressed_sk.is_conformant(&same_as_compute_params));
         }
     }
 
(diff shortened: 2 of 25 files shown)


tmontaigu (commented):
> @tmontaigu reviewed 21 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 1 unresolved discussion (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, and SouchonTheo).

___
*[`tfhe/src/high_level_api/keys/server.rs` line 1064 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P1-uEj38jBig5g7a4LX:-P1-uEj38jBig5g7a4LY:b-aahm7c) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/high_level_api/keys/server.rs#L1064)):*
> ```XML
> 
>             let sk_param = params.into();
>             let mut conformance_params = IntegerServerKeyConformanceParams {
> ```

Should use `IntegerServerKeyConformanceParams::from(config)` ?

IceTDrinker (changes requested):
First batch of comments since I hit something that prevents from merging this IMO

> @IceTDrinker reviewed 6 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-:-P103x7uDJuiToXDDM-d:b7zaohc)* status: all files reviewed, 3 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/parameters/transciphering.rs` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103Y5B2m3h4Fudg6SJ:-P103Y5B2m3h4Fudg6SK:b-gi3xp3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/transciphering.rs#L23)):*
> ```XML
>     /// The parameters of the OPRF key these describe, resolved against the compute parameters
>     /// they are used alongside.
>     pub const fn oprf_parameters(self, compute_params: AtomicPatternParameters) -> OprfParameters {
> ```

just wondering if there OprfParameters used for key gen should contain everything, while the ones for the metaparams only the config/missing info, it's the approach I took for the rerand IIRC, not saying this approach is not good, just wondering
___
*[`tfhe/src/shortint/parameters/v1_8/meta/cpu.rs` line 604 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103qa113SfHGC7AULQ:-P103qa113SfHGC7AULR:b-afp5wk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/v1_8/meta/cpu.rs#L604)):*
> ```XML
>             ReRandomizationConfiguration::DerivedCompactPublicKeyWithoutKeySwitch,
>         ),
>         transciphering_parameters: Some(V1_8_TRANSCIPHERING_PARAM_DEDICATED_OPRF),
> ```

I don't think that covers all params that need the update, notably prod params, since this parameter set is ZKV1 and to big, ie.e. really not what we use in prod

IceTDrinker:
*[`tfhe/src/shortint/oprf.rs` line 198 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P105MvnClfgAAooQT41:-P105MvnClfgAAooQT42:b-2mg6s0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/oprf.rs#L198)):*
> ```XML
>         target_bsk: &ShortintBootstrappingKey<T>,
>     ) {
>         assert!(target_bsk.input_lwe_dimension() >= self.input_lwe_dimension());
> ```

nit since we'll have an update to the pr, indicate this is because this inequality guarantees prf output noise compatibility with the compute AP

IceTDrinker (commented):
> @IceTDrinker reviewed 15 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 6 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/parameters/v1_8/transciphering.rs` line 6 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P104d8t5nWP63FA-0d0:-P104d8t5nWP63FA-0d1:bbn8wd6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/v1_8/transciphering.rs#L6)):*
> ```XML
> pub const V1_8_TRANSCIPHERING_PARAM_DEDICATED_OPRF: TranscipheringParameters =
>     TranscipheringParameters::DedicatedOprf(OprfParameters {
>         lwe_dimension: LweDimension(600),
> ```

I think we may want to add a check in the validate function of the metaparameters that this lwe_dim is <= to the compute param dim potentially ? since currently it makes sense
___
*[`tfhe/src/transciphering/keys.rs` line 242 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P10AG5O0Wzn1MrKAKZb:-P10AG5O0Wzn1MrKAKZc:b-31008x) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/transciphering/keys.rs#L242)):*
> ```XML
>                 // A dedicated key must not pass for a key mirroring the compute parameters.
>                 (OprfKeyConformanceParams::same_as_compute(compute), false),
>             ],
> ```

could have the positive case as well here ?

nsarlin-zama (commented):
> @nsarlin-zama made 5 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker, mayeul-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/high_level_api/keys/server.rs` line 1064 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P1-uEj38jBig5g7a4LX:-P10CCTG2FTPfDgmLDKw:b1ok5hp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/high_level_api/keys/server.rs#L1064)):*

Should use `IntegerServerKeyConformanceParams::from(config)` ?
</blockquote></details>

yes indeed, and applied to other places at the same time
___
*[`tfhe/src/shortint/oprf.rs` line 198 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P105MvnClfgAAooQT41:-P10E0k-BPmdZ-GiApYg:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/oprf.rs#L198)):*

nit since we'll have an update to the pr, indicate this is because this inequality guarantees prf output noise compatibility with the compute AP
</blockquote></details>

done
___
*[`tfhe/src/shortint/parameters/v1_8/transciphering.rs` line 6 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P104d8t5nWP63FA-0d0:-P10EfJ86MJ-NIyzf_IN:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/v1_8/transciphering.rs#L6)):*

I think we may want to add a check in the validate function of the metaparameters that this lwe_dim is <= to the compute param dim potentially ? since currently it makes sense
</blockquote></details>

done
___
*[`tfhe/src/shortint/parameters/v1_8/meta/cpu.rs` line 604 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103qa113SfHGC7AULQ:-P10KNqb2qe3mrkMftDA:b-sp3q5q) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/v1_8/meta/cpu.rs#L604)):*

I don't think that covers all params that need the update, notably prod params, since this parameter set is ZKV1 and to big, ie.e. really not what we use in prod
</blockquote></details>

yes indeed! Was a bit to fast when I did this
___
*[`tfhe/src/transciphering/keys.rs` line 242 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P10AG5O0Wzn1MrKAKZb:-P10YsRq94eMyN92v3nc:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/transciphering/keys.rs#L242)):*

could have the positive case as well here ?
</blockquote></details>

done

IceTDrinker (commented):
Have you considered the question about the different pattern being used here for OprfParameters which are not full parameters for an oprf key ?

additional nit question on naming convention for the transciphering param 

> @IceTDrinker reviewed 7 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-:-P10hxRJAlyv980-whDh:b-jdr4pq)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/parameters/v1_8/transciphering.rs` line 4 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P10gm8S1wvu9gFHffvG:-P10gm8S1wvu9gFHffvH:b-grwytb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6ff9d2fe3447e393dc81f44ada7c2655520e48d2/tfhe/src/shortint/parameters/v1_8/transciphering.rs#L4)):*
> ```XML
> use crate::shortint::prelude::LweDimension;
> 
> pub const V1_8_TRANSCIPHERING_PARAM_DEDICATED_OPRF: TranscipheringParameters =
> ```

sorry kind of missed this, does it maybe need the extra MESSAGE_CARRY, etc metadata in the name here ?

the way the param is structured is a bit odd so maybe not, but it is a departure from previous patterns, so yeah

metaparams use the validate pattern though, so it might be ok

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker, mayeul-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/shortint/parameters/transciphering.rs` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103Y5B2m3h4Fudg6SJ:-P10nCgi-JgJBIVl0nN8:b1jpo8n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/transciphering.rs#L23)):*

just wondering if there OprfParameters used for key gen should contain everything, while the ones for the metaparams only the config/missing info, it's the approach I took for the rerand IIRC, not saying this approach is not good, just wondering
</blockquote></details>

I'm not sure I understand correctly, but it would duplicate the struct and make it possible to create incompatible params? 

If we store additional data and we have to check that they are correctly in sync it would be error prone?
___
*[`tfhe/src/shortint/parameters/v1_8/transciphering.rs` line 4 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P10gm8S1wvu9gFHffvG:-P10nlC88MbvvpgYTh_9:b-suqgwn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6ff9d2fe3447e393dc81f44ada7c2655520e48d2/tfhe/src/shortint/parameters/v1_8/transciphering.rs#L4)):*

sorry kind of missed this, does it maybe need the extra MESSAGE_CARRY, etc metadata in the name here ?

the way the param is structured is a bit odd so maybe not, but it is a departure from previous patterns, so yeah

metaparams use the validate pattern though, so it might be ok
</blockquote></details>

Here I have the feeling that these "minimalistic" params could be reused with params of different precisions, that they are not tied to 2_2 ?

Question is, would it make sense to have different "OprfParameters" for every compute params? If the values would always (or almost always) be the same I would keep it like this.

IceTDrinker:
*[`tfhe/src/shortint/parameters/transciphering.rs` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103Y5B2m3h4Fudg6SJ:-P143Avt07updpxejftI:bvmcfcn) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/transciphering.rs#L23)):*

I'm not sure I understand correctly, but it would duplicate the struct and make it possible to create incompatible params? 

If we store additional data and we have to check that they are correctly in sync it would be error prone?
</blockquote></details>

not really what I had in mind is that there is an OprfParameters that contains everything needed to generate an OprfServerKey

however the metaparam does not contain the OprfParameters struct, only the "config" (that's the name I had for rerand) withe modified LweDimension, and there is a function on the MetaParameters struct like fn oprf_parameters(&self) -> OprfParameters that manages to return the correctly fully populated parameters

so the function is matching on the config and then returning the OprfParameters with everything in it, it avoids the mix of atomic pattern params/additional enum

not saying it's best but at least it means managing the OprfParameters happens in a single function, while currently the logic is spread in all the function taking the oprf lwe dim + pbs params

IceTDrinker:
*[`tfhe/src/shortint/parameters/v1_8/transciphering.rs` line 4 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P10gm8S1wvu9gFHffvG:-P143u6iBvNHYTclx_Iw:b5rcq4d) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/6ff9d2fe3447e393dc81f44ada7c2655520e48d2/tfhe/src/shortint/parameters/v1_8/transciphering.rs#L4)):*

Here I have the feeling that these "minimalistic" params could be reused with params of different precisions, that they are not tied to 2_2 ?

Question is, would it make sense to have different "OprfParameters" for every compute params? If the values would always (or almost always) be the same I would keep it like this.
</blockquote></details>

no that's a fair point, 600 seems to be like the minimal value and since params don't seem to be allowed to be smaller than 600 this is likely ok

tmontaigu (commented):
> @tmontaigu reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 2 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, and SouchonTheo).

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/parameters/transciphering.rs` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103Y5B2m3h4Fudg6SJ:-P14BL-U3BoiaLGtW-oz:blc2433) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/transciphering.rs#L23)):*

not really what I had in mind is that there is an OprfParameters that contains everything needed to generate an OprfServerKey

however the metaparam does not contain the OprfParameters struct, only the "config" (that's the name I had for rerand) withe modified LweDimension, and there is a function on the MetaParameters struct like fn oprf_parameters(&self) -> OprfParameters that manages to return the correctly fully populated parameters

so the function is matching on the config and then returning the OprfParameters with everything in it, it avoids the mix of atomic pattern params/additional enum

not saying it's best but at least it means managing the OprfParameters happens in a single function, while currently the logic is spread in all the function taking the oprf lwe dim + pbs params
</blockquote></details>

Here we have the private key generation that only need the lwe dim (so only OprfParameters, no pbs params) and the server/pbs key generation that needs the other params (levels, base log,...), so it does not use OprfParameters.

We never have to build full "hybrid" params with the lwe dim from oprf and the levels/base log/etc. from the compute params

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/parameters/transciphering.rs` line 23 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P103Y5B2m3h4Fudg6SJ:-P14XKh5DoZNOvJbxZtn:b9cfpo6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/063f12edb9fb1fc78f2cd99d3367a1bf5fcf2c25/tfhe/src/shortint/parameters/transciphering.rs#L23)):*

Here we have the private key generation that only need the lwe dim (so only OprfParameters, no pbs params) and the server/pbs key generation that needs the other params (levels, base log,...), so it does not use OprfParameters.

We never have to build full "hybrid" params with the lwe dim from oprf and the levels/base log/etc. from the compute params
</blockquote></details>

Following our discussion, I added some checks for the lwe dimension

IceTDrinker (commented):
last question

> @IceTDrinker reviewed 8 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-:-P14zSK73L-DoDzEZ0yZ:b-18lghh)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/oprf.rs` line 102 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P14yzR431GRNN8vFKyn:-P14yzR431GRNN8vFKyo:b7t056k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/12c19662ae9623f125d5bb36bb0703033fb81453/tfhe/src/shortint/oprf.rs#L102)):*
> ```XML
> 
>     pub fn new_with_params(target_ck: &ClientKey, params: OprfParameters) -> Self {
>         assert!(
> ```

nit should this be in new_with_lwe_dimension maybe ? since it seems to be the lowest function

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and SouchonTheo).

___
*[`tfhe/src/shortint/oprf.rs` line 102 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-P14yzR431GRNN8vFKyn:-P157wmv501fDdQwJyhI:b-286rr8) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/12c19662ae9623f125d5bb36bb0703033fb81453/tfhe/src/shortint/oprf.rs#L102)):*

nit should this be in new_with_lwe_dimension maybe ? since it seems to be the lowest function
</blockquote></details>

you are right. I removed new_with_lwe_dimension so that new_with_params is the new lowest function

IceTDrinker (commented):
looks good to me, @tmontaigu if you can have a final look since it touches some of the xof internal stuff

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-:-P15F20KC798hGN94Hl-:b-afndat)* status: all files reviewed, 1 unresolved discussion (waiting on mayeul-zama, nsarlin-zama, and SouchonTheo).

IceTDrinker (approved):
Approving to unblock, still taker of a second pair of eyes



> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3917#-:-P15GQTu9gZnoBXxx-Hn:b-j5vv2h)* status: all files reviewed, 1 unresolved discussion (waiting on mayeul-zama, nsarlin-zama, and SouchonTheo).

PR #3925, opened 2026-09-09
feat(gpu): add custom size lhs rhs mul and mul_add

bbarbakadze:
Saved to /tmp/claude-1000/.../scratchpad/pr-mul-add.md. For PR #3925:

# feat(gpu): fixed-point fused multiply-add with asymmetric operands

## What

Adds one new GPU operation in two shapes: a fused multiply-add where the two operands
have very different widths, the low columns of the product are never computed, and the
result is as wide as the left operand.

- `mul_add_fixed_point_with_rescaling` — `result[L] = trunc_beta^s(lhs[L] * rhs[R] + added[L] * beta^s)`
- `mul_low_partial_sum` — low half of `lhs[n] * rhs[n]` plus caller-supplied addends

Both are exposed on `CudaServerKey`. Nothing existing changes behaviour; `mul` and
`div_rem` are untouched.

## Why

This is the multiplication a Goldschmidt divider needs, and the existing square, exact
multiplication is the wrong shape for it three times over:

- The operands are **asymmetric** — a 34-block running value against a 5, 9 or 16-block
  iteration factor, not 34 x 34.
- Most of the product is **discarded**. The result is a fixed point of the same width as
  the left operand, so everything below one output ulp is computed and then thrown away.
- The **addition is separate**, costing another pass over the accumulator.

Doing it as `mul` + shift + `add` means paying for bits that never reach the result.

### Latency

H100 SXM5, 64-bit operands, milliseconds. `mul`  gpu backend multiplication with 64 bit operands

Classical:

| GPUs | mul | mul_add 34x5 | mul_add 34x9 | mul_add 34x16 | mul_low 32x32 |
| ---- | --- | ------------ | ------------ | ------------- | ------------- |
| 1    | 76  | 49           | 58           | 62            | 72            |
| 2    | 59  | 46           | 50           | 54            | 57            |
| 4    | 51  | 44           | 48           | 53            | 54            |
| 8    | 50  | 44           | 49           | 52            | 55            |

Multi-bit:

| GPUs | mul | mul_add 34x5 | mul_add 34x9 | mul_add 34x16 | mul_low 32x32 |
| ---- | --- | ------------ | ------------ | ------------- | ------------- |
| 1    | 98  | 43           | 56           | 56            | 92            |
| 2    | 59  | 32           | 36           | 37            | 52            |
| 4    | 40  | 24           | 28           | 29            | 33            |
| 8    | 31  | 23           | 25           | 26            | 31            |

### Check-list:

### DIFF
--- a/backends/tfhe-cuda-backend/cuda/include/integer/shift_and_rotate.h
+++ b/backends/tfhe-cuda-backend/cuda/include/integer/shift_and_rotate.h
@@ -7,6 +7,13 @@ template <typename Torus> struct int_shift_and_rotate_buffer {
   SHIFT_OR_ROTATE_TYPE shift_type;
   bool is_signed;
 
+  /// @brief Selects the algorithm. When true the block-level barrel shifter
+  /// (host_block_shift_and_rotate_inplace) runs, otherwise the bit-level one
+  /// (host_shift_and_rotate_inplace). Only the buffers of the selected path
+  /// are allocated; the other path's pointers stay null.
+  bool use_block_path;
+
+  // ---- bit-level path ----
   CudaRadixCiphertextFFI *tmp_bits;
   CudaRadixCiphertextFFI *tmp_shift_bits;
   CudaRadixCiphertextFFI *tmp_rotated;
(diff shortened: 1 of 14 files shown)


backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -637,4 +637,222 @@ __host__ uint64_t scratch_cuda_integer_mult_radix_ciphertext(
...
+    PANIC_IF_FALSE(lhs->degrees[a] < message_modulus,
+                   "Cuda error (mul_add_fixed_point): left operand carries a "
+                   "block with a non-empty carry");
+  for (uint32_t r = 0; r < R; r++)
+    PANIC_IF_FALSE(rhs->degrees[r] < message_modulus,
+                   "Cuda error (mul_add_fixed_point): right operand carries a "
+                   "block with a non-empty carry");
+
+  // ---- block products: gather, one bootstrap per surviving pair, scatter ----
+  uint32_t block_size = mem->params.big_lwe_dimension + 1;
```
andrei-stoian-zama, line 740: extract the kernel + degrees handling as a separate function. ideally we can enforce all functions calling kernels must modify degrees.

backends/tfhe-cuda-backend/cuda/src/integer/multiplication.cuh
```diff
@@ -637,4 +637,222 @@ __host__ uint64_t scratch_cuda_integer_mult_radix_ciphertext(
...
+    mem->pair_products->noise_levels[j] =
+        rhs->noise_levels[mem->h_pair_rhs_idx[j]];
+  }
+
+  integer_radix_apply_bivariate_lookup_table<Torus>(
+      streams, mem->pair_products, mem->pair_products, mem->pair_lhs, bsks,
+      ksks, mem->luts_array, mem->num_pairs, message_modulus);
+
+  cuda_set_device(gpu_index);
+  mul_add_scatter_products<Torus><<<mem->num_pairs, 256, 0, stream>>>(
```
andrei-stoian-zama, line 761: same here, extract kernel call + degrees handling as separate function

andrei-stoian-zama (changes requested):
Some Preliminary findinds

andrei-stoian-zama (changes requested):
Some more review; three aesthetic suggestions :

```
  2. shift_and_rotate.cuh grew 75% (+233 lines)

  Two independent algorithms — bit-level and block-level barrel shift — now live in one file. Extract host_block_shift_and_rotate_inplace and
  its buffer setup to block_shift_and_rotate.cuh. This is mandatory at 75% growth.

  3. multiplication.h is now 546 lines holding two independent algorithms

  int_mul_memory and int_mul_add_fixed_point_memory have completely different lifecycles (one-shot vs. long-lived), yet share a header. Split
  into multiplication.h and mul_add_fixed_point.h.

7. Pair enumeration inline in constructor

  The trapezoid-shape pair enumeration logic (23 lines, lines 379–417 in multiplication.h) is buried in the constructor. Extract to
  enumerate_lsb_pairs() — this makes the skip/accumulator logic testable and clarifies intent.
```

And one more simplificaton for which I'll let you decide (and check as I'm not 100% sure):

```
  The mul_add_scatter_products kernel already knows which output slot each pair lands in (d_pair_dst_idx[i]). It also knows whether the pair is
  LSB or MSB — that boundary is num_lsb_pairs, which could be passed as a kernel param. So the kernel can write the degree and noise directly:

  __global__ void mul_add_scatter_products(..., uint32_t num_lsb_pairs,
                                            uint32_t lsb_degree, uint32_t msb_degree,
                                            uint64_t *degrees, uint64_t *noise_levels) {
      uint32_t pair_idx = blockIdx.x;
      uint32_t dst = dst_idx[pair_idx];
      // ... scatter data ...
      degrees[dst]      = (pair_idx < num_lsb_pairs) ? lsb_degree : msb_degree;
      noise_levels[dst] = 1;
  }

  This eliminates:
  - h_product_degrees (host allocation, initialization, per-call memcpy)
  - h_product_noise_levels (same)
```

backends/tfhe-cuda-backend/cuda/include/integer/shift_and_rotate.h
```diff
@@ -2,45 +2,180 @@
...
-  /// @brief The cmux itself: returns the shifted bit when the control bit is
-  /// set, the unshifted one otherwise.
   int_radix_lut<Torus> *mux_lut;
+  int_radix_lut<Torus> *cleaning_lut;
+
+  // ---- block-level path ----
+  // Working array holding the result being built, plus the two "donor" arrays
+  // of the fused first round (what each block hands to the block one and two
+  // positions away).
+  CudaRadixCiphertextFFI *blk_messages;
```
andrei-stoian-zama: Please transform these to normal members, not pointers

backends/tfhe-cuda-backend/cuda/include/integer/shift_and_rotate.h
```diff
@@ -112,155 +247,95 @@ template <typename Torus> struct int_shift_and_rotate_by_bits_buffer {
...
+    GPU_ASSERT(blk_num_amount_blocks <= num_radix_blocks - 1,
                "Cuda error: not enough shift amount blocks for the block "
                "barrel shifter");
 
     uint32_t block_modulus = message_modulus * params.carry_modulus;
     uint32_t box_size = params.polynomial_size / block_modulus;
-    lut_stride = (block_modulus / 2) * box_size;
+    blk_lut_stride = (block_modulus / 2) * box_size;
 
     auto alloc = [&](CudaRadixCiphertextFFI **ct, uint32_t n) {
```
andrei-stoian-zama: avoiding pointers here, we can just call create_zero directly

andrei-stoian-zama (changes requested):

PR #3935, opened 2026-09-11
fix(strings)!: support 1_1 and properly reject 3_3

### LINKED ISSUE #3846: [Bug report] `strings::ClientKey` accepts a key generated with built-in `MESSAGE_3/CARRY_3` parameters but `encrypt_ascii()` later panics
I would like to report a tfhe-rs strings API compatibility bug.

At the tested commit below, the public lower-level `tfhe::strings::ClientKey::new(...)` accepts an integer client key generated with the publicly exported built-in `PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128` parameter, but `encrypt_ascii("A", None)` subsequently panics.

The direct strings encryption path derives the number of radix blocks per ASCII character by requiring `log2(message_modulus)` to divide `8` exactly:

```rust
assert_eq!(8 % message_modulus.ilog2(), 0);
8 / message_modulus.ilog2() as usize
```

For `message_modulus == 8`, the bit width is `3`, so the assertion evaluates to `assert_eq!(2, 0)` and fails. The adjacent built-in `MESSAGE_2/CARRY_2` control succeeds and decrypts back to `"A"`.

This strings-specific compatibility condition is neither validated by `tfhe::strings::ClientKey::new(...)` nor documented by `encrypt_ascii(...)`. Separately, the same strings module already uses a ceiling-based `7u32.div_ceil(message_modulus.0.ilog2())` rule for compact-list packing and expansion, so the module currently applies divergent ASCII block-width derivations.

## Summary

I reproduced the issue at tfhe-rs commit:

```text
4927335578a3ea73cb8c130f4f39f0ba0fd2d04c
```

Observed behavior:

| Profile | Built-in parameter set | Result |
| ------- | ---------------------- | ------ |
| dev/debug | `PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128` | `encrypt_ascii("A", None)` succeeds and decrypts to `"A"` |
| release | `PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128` | same successful result |
| dev/debug | `PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128` | panics at `assert_eq!(8 % message_modulus.ilog2(), 0)` |
| release | `PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128` | same panic |

I also checked the high-level route. A high-level client key configured with the same built-in `MESSAGE_3/CARRY_3` parameter is rejected earlier by the high-level API's separate `MessageModulus == 2 || 4` policy, so the confirmed failing route here is the public lower-level `tfhe::strings::ClientKey` API.

## Environment

- tfhe-rs tested commit: `4927335578a3ea73cb8c130f4f39f0ba0fd2d04c`
- Crate built locally as `tfhe v1.7.0`
- OS: Linux x86_64
- Rust toolchain: `rustc 1.91.1`
- Cargo: `1.91.1`
- Probe features: `default-features = false`, `features = ["integer", "strings", "software-prng"]`
- Standalone probe manifest parsed successfully with `cargo metadata --no-deps`; the probe type-checked successfully with `cargo check --offline`

## Minimal reproduction

Save this as `src/main.rs`:

```rust
use tfhe::integer::ClientKey as IntegerClientKey;
use tfhe::shortint::parameters::{
    ClassicPBSParameters, PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128,
    PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128,
};
use tfhe::strings::ClientKey as StringClientKey;

fn run(label: &str, params: ClassicPBSParameters) {
    let integer_ck = IntegerClientKey::new(params);
    let string_ck = StringClientKey::new(&integer_ck);
    let encrypted = string_ck.encrypt_ascii("A", None);
    let decrypted = string_ck.decrypt_ascii(&encrypted);

    println!("label={label}");
    println!(
        "message_modulus={}",
        integer_ck.parameters().message_modulus().0
    );
    println!("carry_modulus={}", integer_ck.parameters().carry_modulus().0);
    println!("char_count={}", encrypted.chars().len());
    println!("decrypted={decrypted:?}");
}

fn main() {
    match std::env::args().nth(1).as_deref() {
        Some("message2") => run(
            "PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128",
            PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128,
        ),
        Some("message3") => run(
            "PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128",
            PARAM_MESSAGE_3_CARRY_3_KS_PBS_GAUSSIAN_2M128,
        ),
        other => panic!("unknown mode: {other:?}"),
    }
}
```

Use this `Cargo.toml` for the local probe crate:

```toml
[package]
name = "tfhe-rs-strings-message3-probe"
version = "0.1.0"
edition = "2021"

[dependencies.tfhe]
path = "/path/to/tfhe-rs/tfhe"
default-features = false
features = ["integer", "strings", "software-prng"]
```

Commands from that local probe crate:

```bash
cargo run -- message2
cargo run -- message3

cargo run --release -- message2
cargo run --release -- message3
```

Representative outputs:

```text
label=PARAM_MESSAGE_2_CARRY_2_KS_PBS_GAUSSIAN_2M128
message_modulus=4
carry_modulus=4
char_count=1
decrypted="A"
```

```text
thread 'main' panicked at /tmp/tfhe-rs-main-492733/tfhe/src/strings/ciphertext.rs:458:5:
assertion `left == right` failed
  left: 2
 right: 0
```

Representative abbreviated backtrace from a run with `RUST_BACKTRACE=1`:

```text
tfhe::strings::ciphertext::num_ascii_blocks
  at tfhe/src/strings/ciphertext.rs:458:5
tfhe::strings::client_key::ClientKey<T>::num_ascii_blocks
  at tfhe/src/strings/client_key.rs:104:9
tfhe::strings::client_key::ClientKey<T>::encrypt_ascii
  at tfhe/src/strings/client_key.rs:75:31
```

## Expected behavior

If the strings API only supports message-bit widths that divide the per-character representation exactly, that compatibility rule should be validated and documented at a clear strings API/key boundary before `encrypt_ascii()` reaches an internal assertion.

If non-dividing widths are intended to be supported, tfhe-rs should define one canonical blocks-per-ASCII-character rule and use it consistently across direct encryption, trivial encryption, compact-list packing and expansion, integer conversion, and server-key helpers.

## Root cause and related representation inconsistency

The public `strings::ClientKey::new(...)` constructor is just:

```rust
pub fn new(inner: T) -> Self {
    Self { inner }
}
```

so it does not perform any strings-specific compatibility validation.

`encrypt_ascii(...)` only documents panics for non-ASCII input or null characters, then computes its per-character radix width through:

```rust
fn num_ascii_blocks(&self) -> usize {
    let ck = self.inner.borrow();

    assert_eq!(
        ck.parameters().message_modulus().0,
        ck.parameters().carry_modulus().0
    );

    num_ascii_blocks(ck.parameters().message_modulus())
}
```

The direct helper is:

```rust
pub(super) fn num_ascii_blocks(message_modulus: MessageModulus) -> usize {
    let message_modulus = message_modulus.0;

    assert!(message_modulus.is_power_of_two());
    assert_eq!(8 % message_modulus.ilog2(), 0);
    8 / message_modulus.ilog2() as usize
}
```

For the built-in `MESSAGE_3/CARRY_3` parameter, `message_modulus == 8`, so `message_modulus.ilog2() == 3` and the exact-divisibility assertion fails.

Source inspection also shows that the same strings module already uses a different block-width rule for compact strings:

```rust
let blocks_per_char = 7u32.div_ceil(message_modulus.0.ilog2());
```

That ceiling-based `7`-bit rule is used during compact packing and expansion, while the direct client-key path requires exact divisibility of `8`. So the current problem is not only a missing precondition check; it is also an internal inconsistency in how ASCII block widths are derived.

I am not claiming that the compact-list path alone proves that every built-in `MESSAGE_3/CARRY_3` strings operation is intended to be supported. The point is narrower: the strings module currently contains two non-equivalent width derivations, and the direct `encrypt_ascii(...)` route fails through an internal assertion instead of a documented compatibility check.

## Impact

The confirmed impact is a public API panic for a valid ASCII input when the strings client key is built from a publicly exported built-in shortint/integer parameter set.

This is primarily a strings API compatibility and robustness issue. I have not established memory corruption, information disclosure, code execution, ciphertext corruption, or cryptographic key-recovery consequences.

## Relevant source locations

- Built-in parameter aliases:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/shortint/parameters/aliases.rs#L45-L50>
- Public lower-level strings constructor and encryption path:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/client_key.rs#L17-L19>
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/client_key.rs#L61-L104>
- Direct strings ASCII block-width helper:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/ciphertext.rs#L453-L460>
- Compact strings block-width calculation and expansion:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/ciphertext.rs#L61-L68>
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/ciphertext.rs#L163-L180>
- Server-key bridge to the shared direct helper:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/strings/server_key/mod.rs#L51-L58>
- High-level earlier `MessageModulus == 2 || 4` rejection:
  <https://github.com/zama-ai/tfhe-rs/blob/4927335578a3ea73cb8c130f4f39f0ba0fd2d04c/tfhe/src/high_level_api/keys/inner.rs#L371-L378>

## Suggested direction

Define one canonical blocks-per-ASCII-character rule and apply it consistently across direct encryption, trivial encryption, compact-list packing and expansion, integer conversion, and server-key helpers.

If non-dividing widths are intentionally unsupported, reject them through a documented strings-specific compatibility check at key construction or at the start of the relevant strings APIs.

If they are intended to be supported, the implementation should use a ceiling-based representation and define the unused high bits consistently, instead of mixing an exact `8 % bits == 0` rule with the current compact-path `7u32.div_ceil(bits)` rule.

Reported by Jiang Chao, Beijing University of Posts and Telecommunications

nsarlin-zama:
BREAKING_CHANGE: strings stored in compact lists with 1_1 params before this commit will now be rejected, as they yielded incorrect results.

closes: https://github.com/zama-ai/tfhe-rs/issues/3846

Since the string module was initially designed with 8bit/char, I decided to generalize this, and change the compact list code that used 7bits

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)


tmontaigu (commented):
> @tmontaigu reviewed 10 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 1 unresolved discussion (waiting on IceTDrinker, mayeul-zama, and nsarlin-zama).

___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 369 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_kEtVExdPGq6tliHY:-P1_kEtVExdPGq6tliHZ:b-iwgy84) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/test_functions/test_compact.rs#L369)):*
> ```XML
>     // Sanity check of the test itself: the alteration must be visible on 2 bits blocks
>     if bits_per_block == 2 {
>         assert_eq!(expected, "pqrs");
> ```

Just to be sure I get it, if bits_per_block was 4 we would have the same alteration ?

mayeul-zama (commented):
> @mayeul-zama reviewed 3 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 6 unresolved discussions (waiting on IceTDrinker and nsarlin-zama).

___
*[`tfhe/src/shortint/atomic_pattern/mod.rs` line 618 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_xCWWAfQj3m2DXDro:-P1_xCWWAfQj3m2DXDrp:bp9xk8f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/shortint/atomic_pattern/mod.rs#L618)):*
> ```XML
> 
>     pub fn is_compatible_with_strings(&self) -> bool {
>         self.message_modulus().0 == self.carry_modulus().0
> ```

Could call `MessageModulus::is_compatible_with_strings`
___
*[`tfhe/src/strings/ciphertext.rs` line 68 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_tImO3jcFabiI96nL:-P1_tImO3jcFabiI96nM:b-8im3u6) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L68)):*
> ```XML
>         num_blocks: Option<usize>,
>     ) -> Option<DataKind> {
>         let blocks_per_char = message_modulus.num_blocks_per_ascii_char().ok()?;
> ```

Maybe we should return a Result<Option<DataKind >> to manage errors.
None is ignored and not treated as an error in `push_and_check_kind_coherence`

Or even a `Result<DataKind>` (not clear if Option was really useful here) 
___
*[`tfhe/src/strings/ciphertext.rs` line 112 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_vdc3EW3nlKZIkllm:-P1_vdc3EW3nlKZIklln:buhc9qk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L112)):*
> ```XML
>         let blocks_per_char = message_modulus
>             .num_blocks_per_ascii_char()
>             .expect("The parameters of the CompactPublicKey are not compatible with strings");
> ```

Could add check on carry modulus
___
*[`tfhe/src/strings/ciphertext.rs` line 142 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_vjSkBknDKTqjw0qc:-P1_vjSkBknDKTqjw0qd:buhc9qk) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L142)):*
> ```XML
>         let blocks_per_char = message_modulus
>             .num_blocks_per_ascii_char()
>             .expect("The parameters of the CompactPublicKey are not compatible with strings");
> ```

Could add check on carry modulus
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 27 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_xTaU3CW25au2HuUh:-P1_xTaV4sEgx5z5s09V:biu867m) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/test_functions/test_compact.rs#L27)):*
> ```XML
>     let pke_params = PARAM_PKE_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
>     let ksk_params = PARAM_KEYSWITCH_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
>     let fhe_params = PARAM_MESSAGE_2_CARRY_2_KS_PBS_TUNIFORM_2M128;
> ```

Any reason not to test with 1_1 params?
Maybe the 3 kinds of params don't exist

nsarlin-zama (commented):
> @nsarlin-zama made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 6 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and tmontaigu).

___
*[`tfhe/src/shortint/atomic_pattern/mod.rs` line 618 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_xCWWAfQj3m2DXDro:-P1dc9uZAxPjMa_txeZ0:b-qb1vtp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/shortint/atomic_pattern/mod.rs#L618)):*

Could call `MessageModulus::is_compatible_with_strings`
</blockquote></details>

yes indeed
___
*[`tfhe/src/strings/ciphertext.rs` line 68 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_tImO3jcFabiI96nL:-P1ddjRW180_PRG-KTjL:b1qgg5n) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L68)):*

Maybe we should return a Result<Option<DataKind >> to manage errors.
None is ignored and not treated as an error in `push_and_check_kind_coherence`

Or even a `Result<DataKind>` (not clear if Option was really useful here)
</blockquote></details>

that's a trait that is used by every data types that we can put in a compact list (bool, integer,...) so changing the signature is out of scope for this fix IMO. But on the long run I agree with you.

I converted it into a panic so it would at least not silently fail. This is what is done elsewhere, and since this is client side I guess it's a good temporary solution...

I made an issue to update this trait
___
*[`tfhe/src/strings/ciphertext.rs` line 112 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_vdc3EW3nlKZIkllm:-P1e87tvAmsm3e5oSfI7:b-6ynh33) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L112)):*

Could add check on carry modulus
</blockquote></details>

the check is done on the string client and server key creation. 

The carry modulus check is only needed for compute. Here we only have access to the public key, which could in theory have different moduli (and a keyswitch), so we wouldn't be checking the correct thing I think?
___
*[`tfhe/src/strings/ciphertext.rs` line 142 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_vjSkBknDKTqjw0qc:-P1e8XpwDrntLI13c46a:b-urb88d) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/ciphertext.rs#L142)):*

Could add check on carry modulus
</blockquote></details>

same answer
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 27 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_xTaU3CW25au2HuUh:-P1e9u6-0OgDcQkXI8nN:b-dnxdxu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/test_functions/test_compact.rs#L27)):*

Any reason not to test with 1_1 params?
Maybe the 3 kinds of params don't exist
</blockquote></details>

no, we don't have the casting params for 1_1
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 369 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1_kEtVExdPGq6tliHY:-P1eEfb1CttHzH5NcIo3:basjxks) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/9fd0cbb64e0c75e5c3301870238eee0eb6666b9e/tfhe/src/strings/test_functions/test_compact.rs#L369)):*

Just to be sure I get it, if bits_per_block was 4 we would have the same alteration ?
</blockquote></details>

yes exactly, but that's a bit of a chance. The alteration sets the last block to its max value (full of 1s) (line 354).
The goal is to check that the last bit of a char is correctly cleared by sanitization.

with msg mod 2, the alteration will only set bit 6 and 7.
with msg mod 4, the alteration will set bits 4,5,6 and 7. But, on the input digits, bit 4 and 5 are already set, so only 6 and 7 will be flipped.
 7 is cleared by sanitization. So only 6 remains. But with different inputs, msg mod 2 and 4 could produce different altered strings.

I made the sanity check a bit more robust now.

mayeul-zama (dismissed):
Thanks!

> @mayeul-zama reviewed 8 files and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1iTWem6qs64b0YgXL5:b-ucfknh)* status: all files reviewed, 1 unresolved discussion (waiting on IceTDrinker and tmontaigu).

IceTDrinker (commented):
some questions for now, haven't read everything

> @IceTDrinker reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1iVFSTBvxLoACHzOVG:biwi7hz)* status: all files reviewed, 2 unresolved discussions (waiting on nsarlin-zama and tmontaigu).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 657 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iV17f8qFxjbODAYUX:-P1iV17f8qFxjbODAYUY:bc6kdxi) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L657)):*
> ```XML
>                         push_functions(blocks_per_char - 1, &self.msg_extract, &self.carry_extract);
>                         push_functions(
>                             1,
> ```

wrong if blocks_per_char == 1 ?

should blocks_per_char be some new type that handles all the logic so we can avoid the random "- 1" in some locations ?

IceTDrinker:
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 657 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iV17f8qFxjbODAYUX:-P1iWBaE-irOpnDT5Dhr:bs9covr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L657)):*

wrong if blocks_per_char == 1 ?

should blocks_per_char be some new type that handles all the logic so we can avoid the random "- 1" in some locations ?
</blockquote></details>

I am the one not able to read or do basic maths, nothing to see here

IceTDrinker (commented):
> @IceTDrinker reviewed 9 files and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 7 unresolved discussions (waiting on nsarlin-zama and tmontaigu).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 556 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iXEcnEY2n3rGecuPf:-P1iXEcnEY2n3rGecuPg:bdimfdu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L556)):*
> ```XML
>     /// Number of blocks used to store one ASCII char, `None` if the parameters do not support
>     /// strings
>     blocks_per_char: Option<usize>,
> ```

Option<NonZeroUsize> maybe ?
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 566 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iWxYiAxGBa9IrRbWg:-P1iWxYiAxGBa9IrRbWh:bkdiamr) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L566)):*
> ```XML
>         // sanitized to 0.
>         let bits_of_last_char_block = blocks_per_char.map_or(0, |blocks_per_char| {
>             7u32.saturating_sub((blocks_per_char as u32 - 1) * message_modulus.0.ilog2())
> ```

should the 7u32 be a cosnt the same way you have the bits per ASCII char ?

also naming is important to understand the difference of "useful bits" vs. "stored bits"
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 279 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZ5281Dfga9_s29CU:-P1iZ5281Dfga9_s29CV:b-7eb2rm) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L279)):*
> ```XML
>     let direct = str_cks.encrypt_ascii(clear.str(), None);
>     assert_eq!(
>         direct.chars()[0].ciphertext().blocks().len(),
> ```

could be direct.chars().all(...)
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 298 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZG1A0g0fagYZIh6j:-P1iZG1A0g0fagYZIh6k:b-ty9yc5) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L298)):*
> ```XML
> 
>             assert_eq!(
>                 from_list.chars()[0].ciphertext().blocks().len(),
> ```

could be chars().all(...)
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 347 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZjQW5rxhCXStqTAM:-P1iZjQW5rxhCXStqTAN:b-jc1hqt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L347)):*
> ```XML
>     // clears the whole block instead of just bit 7 would be detected. With 1 bit blocks the last
>     // block only holds bit 7 and the string must come back unchanged.
>     let string = "0123";
> ```

is this is a safe choice for all param sets ? it mentions bit 6 and 7 which sound 2_2 centric ?

should this be raw u8s (with only 7 bits set) converted to chars ?
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 356 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZsJb2ldG-muzOdCi:-P1iZsJb2ldG-muzOdCj:b276g7m) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L356)):*
> ```XML
>     // Set all the bits of the last block of each char
>     for char_idx in 0..string.len() {
>         builder.messages[char_idx * blocks_per_char + blocks_per_char - 1] = message_modulus - 1;
> ```

are we interested in checking the case where the carry bits are filled too ? we assume padding bit is 0 as checked by the ZK

tmontaigu (commented):
> @tmontaigu reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 6 unresolved discussions (waiting on nsarlin-zama).

IceTDrinker (approved):
Integer test crash https://github.com/zama-ai/tfhe-rs/actions/runs/35199770299

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1jD7U6DgOzuAPtMF1G:bzdhfxn)* status: all files reviewed, 6 unresolved discussions (waiting on nsarlin-zama).

IceTDrinker (changes requested):
integer test crash https://github.com/zama-ai/tfhe-rs/actions/runs/35199770299

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1jDCEKC_412nZw8knu:b-qqcbp1)* status: all files reviewed, 6 unresolved discussions (waiting on nsarlin-zama).

nsarlin-zama (commented):
> @nsarlin-zama made 6 comments and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 556 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iXEcnEY2n3rGecuPf:-P1i_id8By6GDudgX3IG:b5qw4lj) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L556)):*

Option<NonZeroUsize> maybe ?
</blockquote></details>

That's already an option? What do you mean?
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 566 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iWxYiAxGBa9IrRbWg:-P1jUJUJ1krVAYOq8xCq:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L566)):*

should the 7u32 be a cosnt the same way you have the bits per ASCII char ?

also naming is important to understand the difference of "useful bits" vs. "stored bits"
</blockquote></details>

done
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 279 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZ5281Dfga9_s29CU:-P1jUmQ72KhwnY4u10yG:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L279)):*

could be direct.chars().all(...)
</blockquote></details>

done
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 298 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZG1A0g0fagYZIh6j:-P1jUn9a9kJ1sNEIMeUt:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L298)):*

could be chars().all(...)
</blockquote></details>

done
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 347 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZjQW5rxhCXStqTAM:-P1jVAJ63yMYBfE5ueak:bbodsqo) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L347)):*

is this is a safe choice for all param sets ? it mentions bit 6 and 7 which sound 2_2 centric ?

should this be raw u8s (with only 7 bits set) converted to chars ?
</blockquote></details>

yes bit 7 is the bit cleared by sanitization and bit 6 is a "canary" that detects if we sanitize too much.
What we need for this test is to have the last block full of 1s, so the actual char depend on the block size. This is done by the tempering below. Digits, with bit 6 at 0, are used because they allow to check that the tempering has been correctly done.

I reworded the comment.
___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 356 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZsJb2ldG-muzOdCi:-P1jbAUm5QxwP4AX_9_H:b-3f5oic) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L356)):*

are we interested in checking the case where the carry bits are filled too ? we assume padding bit is 0 as checked by the ZK
</blockquote></details>

Ideally yes, but I don't know how to do that easily since build use "encrypt_slice" which use the message modulus, so even if I set bits in the carry space they will be removed at encryption time.

IceTDrinker:
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 556 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iXEcnEY2n3rGecuPf:-P1jcZLgE10aOUN5Sn3R:bqkmeg0) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L556)):*

That's already an option? What do you mean?
</blockquote></details>

`Option<NonZeroUsize>`

sorry got got by markdown display

nsarlin-zama (commented):
> @nsarlin-zama made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935)* status: 3 of 10 files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and tmontaigu).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 556 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iXEcnEY2n3rGecuPf:-P1jqa7R-fccxIFfL0UK:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/integer/ciphertext/compact_list.rs#L556)):*

`Option<NonZeroUsize>`

sorry got got by markdown display
</blockquote></details>

done

nsarlin-zama (commented):
also fixed a bug where carry_modulus was used for carry_extract instead of message_modulus

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1jqeKp2RsWSc8ReFe1:b-2fi187)* status: 3 of 10 files reviewed, 2 unresolved discussions (waiting on IceTDrinker, mayeul-zama, and tmontaigu).

IceTDrinker (commented):
Some small things that can be done in follow-up, otherwise I think I'm good, I let the others re-check

> @IceTDrinker reviewed 7 files and all commit messages, made 5 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1judgy3MWqRhUohxbC:bhi4bym)* status: all files reviewed, 3 unresolved discussions (waiting on nsarlin-zama).

___
*[`tfhe/src/strings/test_functions/test_compact.rs` line 356 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1iZsJb2ldG-muzOdCi:-P1ju1EVD_a9_PF3gkQ1:b32wtd) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/99df1ef5a459f486caf6519818a7ae036cfb097c/tfhe/src/strings/test_functions/test_compact.rs#L356)):*

Ideally yes, but I don't know how to do that easily since build use "encrypt_slice" which use the message modulus, so even if I set bits in the carry space they will be removed at encryption time.
</blockquote></details>

ok
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 534 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1js2EXDwZdu1JLEfaK:-P1js2EXDwZdu1JLEfaL:b-jioien) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L534)):*
> ```XML
> 
> /// Number of bits needed to store one ASCII char, according to the spec
> const ASCII_USEFUL_BITS: u32 = 7;
> ```

nit: should this be in strings/const.rs e.g. ?

same as the shortint param 9 bits thing ?

can be done separately to move ahead with this PR, but a smalle follow up issue to not forget could be good
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 548 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1jsVSi8Zj7ccTEz7jy:-P1jsVSi8Zj7ccTEz7jz:bcipm3f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L548)):*
> ```XML
>     msg_extract_bool: Box<dyn Fn(u64) -> u64 + Sync>,
>     carry_extract_bool: Box<dyn Fn(u64) -> u64 + Sync>,
>     msg_extract_last_char_block: Box<dyn Fn(u64) -> u64 + Sync>,
> ```

nit for follow up should these be Options in case blocks_per_char is None ?
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 562 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1js_XZ6dh_wdTVpkW4:-P1js_XZ6dh_wdTVpkW5:b-d3zmj5) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L562)):*
> ```XML
>         // value stored in ASCII_CHAR_BITS bits (currently 8), so the upper bits of the last block
>         // must be sanitized to 0.
>         let bits_of_last_char_block = blocks_per_char.map_or(0, |blocks_per_char| {
> ```

nit for follow up: this should probably be an option ?

nsarlin-zama (commented):
I also fixed a small bug with carry_extract using carry_modulus, it only matters when carry_mod != message_mod

> @nsarlin-zama made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1nfdsC-ErSBFWkZaOh:b-jfa7lm)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on IceTDrinker).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 534 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1js2EXDwZdu1JLEfaK:-P1jv4rA5Ow2h17Mre1K:b87hsfc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L534)):*

nit: should this be in strings/const.rs e.g. ?

same as the shortint param 9 bits thing ?

can be done separately to move ahead with this PR, but a smalle follow up issue to not forget could be good
</blockquote></details>

The shortint one is needed as a method for the message_modulus so we would introduce a circular dep if we moved it to strings I guess.

This one is only ever used here so I introduced it for readability, but not sure it's worth being moved to a dedicated file.
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 548 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1jsVSi8Zj7ccTEz7jy:-P1nfOf-8eVYfrB1v3hJ:b1ln2tb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L548)):*

nit for follow up should these be Options in case blocks_per_char is None ?
</blockquote></details>

done, I created a single option for all the string only fields so there are no invalid states
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 562 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1js_XZ6dh_wdTVpkW4:-P1nfVkS6iIRclNZm0K8:bwg0iug) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L562)):*

nit for follow up: this should probably be an option ?
</blockquote></details>

reworked

IceTDrinker (commented):
I like the new shape of this, good on my end, approving to unblock but if the other reviewers can have a last look would appreciate :+1: 

> @IceTDrinker reviewed 1 file and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1ni8Tx2-UVi1H08CI2:bvall10)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nsarlin-zama).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 534 at r3](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-P1js2EXDwZdu1JLEfaK:-P1ng6iy7cBslrkV2wn8:bbjlyax) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/e5245dfa87e664a35484117b3cec7e3a11965f4b/tfhe/src/integer/ciphertext/compact_list.rs#L534)):*

The shortint one is needed as a method for the message_modulus so we would introduce a circular dep if we moved it to strings I guess.

This one is only ever used here so I introduced it for readability, but not sure it's worth being moved to a dedicated file.
</blockquote></details>

was just a thought, as for the circular dep it could be re-exported in the const files if we ever go down that route or feature gated

IceTDrinker (approved):
and messed up again, approving to unblock

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1niNZZBbcws3TMP7Z5:b-f0r8tt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nsarlin-zama).

nsarlin-zama:
It was already approved by Mayeul

IceTDrinker:
> It was already approved by Mayeul

Still it was reworked somewhat, warrants a re-approve

mayeul-zama (approved):
Thanks!

> @mayeul-zama reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3935#-:-P1nkk6J7FNWY2B-XNfs:b-ucfknh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on IceTDrinker, mayeul-zama, and tmontaigu).

PR #3940, opened 2026-09-11
chore: add a script to detect crates that need to be bumped

nsarlin-zama:

### PR content/description
Add a script to detect which version to bump before a release. It answers the following questions:
- which crates changed since their last release and need a version bump?
- which crates are bumped but not yet tagged or published on crates.io?

It parses the make_release_XXX workflows to get the list of releasable crates, so there is no hardcoded list.

As an additional feature, it can also generate the mermaid diagram used in the release doc notion page

example run:
```
❯ make release_status
python3 ./scripts/release_status.py
crate                    manifest  crates.io  baseline tag                   changes  status
tfhe-cuda-common         0.1.1     0.1.1      tfhe-cuda-common-0.1.1         0        OK
tfhe-fft                 0.10.1    0.10.1     tfhe-fft-0.10.1                0        OK
tfhe-hpu-backend         0.7.0     0.7.0      tfhe-hpu-backend-0.7.0         0        OK
tfhe-ntt                 0.7.1     0.7.1      tfhe-ntt-0.7.1                 0        OK
tfhe-versionable-derive  0.8.0     0.8.0      tfhe-versionable-derive-0.8.0  0        OK
wasm-par-mq              0.1.1     0.1.1      wasm-par-mq-0.1.1              0        OK
tfhe-cuda-backend        0.16.0    0.16.0     tfhe-cuda-backend-0.16.0       9        BUMP
tfhe-versionable         0.8.0     0.8.0      tfhe-versionable-0.8.0         0        OK
zk-cuda-backend          0.2.1     0.2.1      zk-cuda-backend-0.2.1          0        OK
tfhe-csprng              0.10.0    0.10.0     tfhe-csprng-0.10.0             0        OK
tfhe-safe-serialize      0.1.1     0.1.1      tfhe-safe-serialize-0.1.1      0        OK
tfhe-zk-pok              0.9.0     0.9.0      tfhe-zk-pok-0.9.0              0        OK
tfhe                     1.9.0     1.8.0      tfhe-rs-1.8.0                  4        TAG

Changed since their last release, version bump needed:
  tfhe-cuda-backend: 9 changed file(s) since tfhe-cuda-backend-0.16.0, bump backends/tfhe-cuda-backend/Cargo.toml

Bumped but not tagged, create the tags in this order:
  git tag -s -a -m "tfhe 1.9.0 release" tfhe-rs-1.9.0 && git push origin tag tfhe-rs-1.9.0
      then run make_release_tfhe.yml on tag tfhe-rs-1.9.0
```

Script was generated using claude

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940)


IceTDrinker:
Think I mentioned scripts/check_cuda_version_bump.sh ?

does not work for us ?

IceTDrinker:
not saying it's the right solution, but is there a world where we can have the same tool for both our needs ? (one can dream) 🥲

nsarlin-zama:
Shell script is made for ci (it errors if you run it locally), and only answers question 1.
This new script is more made for human releasers, telling what needs to be done.

The script from this pr is a superset of the cuda one. If we want to mutualize I think we could use only this one, making it errors if there is a crate in the "bump" state. (but it means we would need to change a bit our process)

But as a first step I didn't want to interact with gpu processes

IceTDrinker:
> Shell script is made for ci (it errors if you run it locally), and only answers question 1. This new script is more made for human releasers, telling what needs to be done.
> 
> The script from this pr is a superset of the cuda one. If we want to mutualize I think we could use only this one, making it errors if there is a crate in the "bump" state. (but it means we would need to change a bit our process)
> 
> But as a first step I didn't want to interact with gpu processes

fair, do talk with Pedro though, to see if we can make something that works well for them as a second step

nsarlin-zama:
cc @pdroalves if we can have your opinion on this

pdroalves:
I really like this output. Perhaps it's only useful for who is in charge of releasing a new version, but I like it. 

I think we should merge this and the CUDA script, though. I think the core is the same and it would be bad to have two different implementations of the same thing. We could generalize the shell script to handle more crates and perhaps return the result to a Python parser.

pdroalves:
The expected behavior of a Github Action that targets a PR and this make target is probably different, but we still can have the same engine behind both.

nsarlin-zama:
New version:
- added a `--pr-gate` flag that returns an error status if the pr changes a crate without a bump, and removed the bash script. Cuda workflow now calls this. In the future, we might want to add this for the other crates too.
- added a `--release-gate` flag, that runs in the releases workflows. This flag checks that:
  - the release is done on the correct tag (ie: correct crate, correct version)
  - the crate is ready for release

I think this hardens the release process. But to avoid any bad surprise, I would suggest to backport this on the 1.8 branch to test it a bit on smaller releases before 1.9

I also added, in another commit, a toggle to allow publishing tfhe-versionable without the derive crate. That way, if somehow derive succeeds but tfhe-versionable fails, we are able to retry later.

IceTDrinker (commented):
python file is very dense, some remarks

> @IceTDrinker reviewed 7 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-:-P1_WFjkD_m8Mm4gReLJ:bcov201)* status: all files reviewed, 6 unresolved discussions (waiting on nsarlin-zama, pdroalves, soonum, and SouchonTheo).

___
*[`.github/workflows/make_release_tfhe_versionable.yml` line 43 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_MZtp5kbpECL0YLpT:-P1_MZtq78r8YXkwXJWz:b-ar74t4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/.github/workflows/make_release_tfhe_versionable.yml#L43)):*
> ```YAML
>     needs: make-release-derive
>     # Skipping the derive release must not skip this one, but a failed derive must.
>     if: ${{ always() && needs.make-release-derive.result != 'failure' }}
> ```

nit https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context

we likely only want success and skipped, if the previous job is cancelled this one is not guaranteed to fail if it's semver compatible, though we generally update those versions in lockstep, so may not be an issue, versionable is only releasable if derive is released

which makes me think we may want a check box to choose to not release versionable alone (if e.g. derive failed to publish 0.10.1 (0.10.0 is live) but versionable still is able to publish as 0.10.1 we would not be able to launch only the derive
___
*[`scripts/release_status.py` line 9 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_OHv3AiYnSIiWdvTL:-P1_OHv3AiYnSIiWdvTM:b-fvro5g) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L9)):*
> ```Python
> 
> Everything is derived, nothing is hand-maintained: the crate list comes from the
> `make_release_*.yml` workflows, versions and the dependency graph from `cargo
> ```

perhaps exploit cargo metadata and mark all unpblished crates with publish = false so that the signal is not inferred from reelase machinery ?
___
*[`scripts/release_status.py` line 65 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_PBxb1FrT-Anja_n0:-P1_PBxb1FrT-Anja_n1:b7xe8dx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L65)):*
> ```Python
> 
> 
> def version_key(version):
> ```

I think you can pip freeze > requirements.txt and allow a venv setup for this in CI

same for git above potentially though a subprocess in that case is perfectly fine

https://pypi.org/project/semver/
___
*[`scripts/release_status.py` line 95 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_QaqJ0zqkKwlWvJlm:-P1_QaqJ0zqkKwlWvJln:bpslp9d) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L95)):*
> ```Python
>     crates = {}
>     for workflow in sorted(WORKFLOW_DIR.glob("make_release_*.yml")):
>         for name in re.findall(
> ```

We can do better as shared elsewhere with cargo metadata IMO
___
*[`scripts/release_status.py` line 123 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_ULhGBTwvGs-cJ_KO:-P1_ULhH69hde_3dkiDo:bslc3gp) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L123)):*
> ```Python
>     """Versions of `name` on crates.io, oldest first."""
>     lowered = name.lower()
>     if len(lowered) <= 2:
> ```

what's this ? the previous curl version did not seem to have this weird thing ?
___
*[`scripts/release_status.py` line 146 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_UigDBhaCP6GcMY9z:-P1_UigDBhaCP6GcMYA-:b-clmhnu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L146)):*
> ```Python
> 
> 
> def consumer_facing(manifest_text):
> ```

strip_dev_deps_from_maniftest ? dunno

IceTDrinker:
also why python if most of the primitives might benefit from rust tooling like cargo toml for manifest, semver etc. ?

nsarlin-zama (commented):
The idea is to be able run this easily and quickly locally, so python was chosen because there is no compilation. It's just a small quality of life script for releasers, so I expect to to run as smoothly as possible: no install, compilation, setup, etc. 
Just `make release_status`

> @nsarlin-zama made 7 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-:-P1dIRQo6Bcq-YNMn1IE:brjyb8h)* status: 6 of 9 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, pdroalves, soonum, and SouchonTheo).

___
*[`.github/workflows/make_release_tfhe_versionable.yml` line 43 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_MZtp5kbpECL0YLpT:-P1dYSSh6MtAKSejN3Mw:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/.github/workflows/make_release_tfhe_versionable.yml#L43)):*

nit https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context

we likely only want success and skipped, if the previous job is cancelled this one is not guaranteed to fail if it's semver compatible, though we generally update those versions in lockstep, so may not be an issue, versionable is only releasable if derive is released

which makes me think we may want a check box to choose to not release versionable alone (if e.g. derive failed to publish 0.10.1 (0.10.0 is live) but versionable still is able to publish as 0.10.1 we would not be able to launch only the derive
</blockquote></details>

done
___
*[`scripts/release_status.py` line 9 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_OHv3AiYnSIiWdvTL:-P1dDkJ5AyRfMNEnL69i:bkcc0td) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L9)):*

perhaps exploit cargo metadata and mark all unpblished crates with publish = false so that the signal is not inferred from reelase machinery ?
</blockquote></details>

I used the workflow as a ground truth for what needs to be released since we sometimes forget the "publish = false", whereas no workflow => not releasable will always be true.

I updated it to use cargo metadata instead, we just need to commit to be stricter about `publish = false`.
___
*[`scripts/release_status.py` line 65 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_PBxb1FrT-Anja_n0:-P1dEOyy8zAFIpglOi0F:byj6kqb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L65)):*

I think you can pip freeze > requirements.txt and allow a venv setup for this in CI

same for git above potentially though a subprocess in that case is perfectly fine

https://pypi.org/project/semver/
</blockquote></details>

I explicitly did not want to add any dependency to keep this as simple to use as possible. It's only 8 lines of code. Main use case is to run this locally.
___
*[`scripts/release_status.py` line 95 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_QaqJ0zqkKwlWvJlm:-P1dYaK13OChh6ITUgtG:bb7f2mf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L95)):*

We can do better as shared elsewhere with cargo metadata IMO
</blockquote></details>

done
___
*[`scripts/release_status.py` line 123 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_ULhGBTwvGs-cJ_KO:-P1dF5Kv6MmD_p4x5l-4:buwys0k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L123)):*

what's this ? the previous curl version did not seem to have this weird thing ?
</blockquote></details>

It's the way crates.io index is organized: https://github.com/rust-lang/crates.io-index.

Small crates name are stored in directories based on the name length. So for example the "naive" version wouldn't work for `zhc`.

I added a comment about that.
___
*[`scripts/release_status.py` line 146 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_UigDBhaCP6GcMY9z:-P1dGhyu51JtS1_C4Mdb:b1k9amf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L146)):*

strip_dev_deps_from_maniftest ? dunno
</blockquote></details>

The goal of this is to avoid "false positives", reports that a crate should be bumped but the modifications actually don't require a new version.
I did only the `dev-dependencies` because that's the case I actually hit, but this function might be extended to strip other keys (ex: `example` ?). 
To avoid bloating it now with things that will never be used, I think we can add new filters when we actually hit these false positives.

I renamed it to `strip_not_consumer_facing_keys`, is it better?

IceTDrinker (commented):
if it ever becomes more complex, having a rust with cargo run in debug will avoid some potential headaches and more appropriate tools for manifest manipulations

> @IceTDrinker reviewed 4 files and all commit messages, made 5 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-:-P1j7QMmAhLewXC2QYg0:b-1vnowm)* status: all files reviewed, 2 unresolved discussions (waiting on nsarlin-zama, pdroalves, soonum, and SouchonTheo).

___
*[`scripts/release_status.py` line 9 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_OHv3AiYnSIiWdvTL:-P1j7v5r2WQZdO7HpuxQ:bu69vpx) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L9)):*

I used the workflow as a ground truth for what needs to be released since we sometimes forget the "publish = false", whereas no workflow => not releasable will always be true.

I updated it to use cargo metadata instead, we just need to commit to be stricter about `publish = false`.
</blockquote></details>

I'm perfectly fine with stricter publish enforcement
___
*[`scripts/release_status.py` line 65 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_PBxb1FrT-Anja_n0:-P1j8-F-0o1HA5z9FDYR:bcakywb) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L65)):*

I explicitly did not want to add any dependency to keep this as simple to use as possible. It's only 8 lines of code. Main use case is to run this locally.
</blockquote></details>

to be revisited if the script becomes more complex, having recoded utils by LLMs will not outlast well maintained and long standing packages IMO
___
*[`scripts/release_status.py` line 123 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_ULhGBTwvGs-cJ_KO:-P1j9TYcEAXqMCBaKoKx:b-r2prg3) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L123)):*

It's the way crates.io index is organized: https://github.com/rust-lang/crates.io-index.

Small crates name are stored in directories based on the name length. So for example the "naive" version wouldn't work for `zhc`.

I added a comment about that.
</blockquote></details>

I know why I was confused the previous script used https://crates.io/api/v1/crates
___
*[`scripts/release_status.py` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1j8A3FDPzJiJHeuQyy:-P1j8A3FDPzJiJHeuQyz:biknj2f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/035c1673db47fbb553b519b2d95276d359e770e8/scripts/release_status.py#L97)):*
> ```Python
>     that restriction is released and belongs in this report.
>     """
>     return sorted(name for name, p in packages.items() if p["publish"] is None)
> ```

publish is None or publish == True ?

nsarlin-zama (commented):
I largely prefer rust to python, but I also believe that no tool cover 100% of use cases, and I believe this is one of the cases where python shines more.
To be fair since we have tomllib, manifest parsing is not really an issue here. Maintainability could be, but with the current script size it's not worth losing the "reactivity" an interpreted language offers (no installation or compilation). 

I don't expect this script to grow a lot.

> @nsarlin-zama made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-:-P1jCHNz9Dj-5y2klkLc:blmyt5f)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker, pdroalves, soonum, and SouchonTheo).

___
*[`scripts/release_status.py` line 65 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_PBxb1FrT-Anja_n0:-P1jKXpd--BcxgSfNx3c:b-o6rnna) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L65)):*

to be revisited if the script becomes more complex, having recoded utils by LLMs will not outlast well maintained and long standing packages IMO
</blockquote></details>

agreed in general, but it's a matter of compromise. As said earlier, the no setup/install is something important for this script imo, and losing this property for 8 lines is not worth it.
___
*[`scripts/release_status.py` line 123 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1_ULhGBTwvGs-cJ_KO:-P1jJns3B-GzFLgz-bYc:bir4dqw) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1937226d5dc0f9af10d4ad05586d2a53a4665fad/scripts/release_status.py#L123)):*

I know why I was confused the previous script used https://crates.io/api/v1/crates
</blockquote></details>

bot used index for a bad reasons, I converted it to the api, which simplified this function. It just needs a user agent since api is rate limited, though. But for our use it should be ok.
___
*[`scripts/release_status.py` line 97 at r4](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-P1j8A3FDPzJiJHeuQyy:-P1jF61y0opS1oVyA1_1:b-pb9bue) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/035c1673db47fbb553b519b2d95276d359e770e8/scripts/release_status.py#L97)):*

publish is None or publish == True ?
</blockquote></details>

indeed

IceTDrinker (approved):
> @IceTDrinker reviewed 1 file and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3940#-:-P1jPuoL-KskLFirvyD4:b-15hw7n)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pdroalves, soonum, and SouchonTheo).

nsarlin-zama:
@pdroalves, is it ok for you? For gpu ci it should work the same way as before

pdroalves:
Perhaps I'm missing documentation, but I'm not having a great time reading the output of this script.

I just pulled the most recent version of the branch and ran:

```
➜  bump_detector_script git:(ns/chore/bump_detector_script) ✗ make release_status                          
python3 ./scripts/release_status.py
crate                    manifest  crates.io  baseline                       changes  status
tfhe-cuda-common         0.1.1     0.1.1      tfhe-cuda-common-0.1.1         0        OK
tfhe-fft                 0.10.1    0.10.1     tfhe-fft-0.10.1                0        OK
tfhe-hpu-backend         0.7.0     0.7.0      tfhe-hpu-backend-0.7.0         0        OK
tfhe-ntt                 0.7.1     0.7.1      tfhe-ntt-0.7.1                 0        OK
tfhe-versionable-derive  0.8.1     0.8.0      tfhe-versionable-derive-0.8.0  7        TAG
wasm-par-mq              0.1.1     0.1.1      wasm-par-mq-0.1.1              0        OK
tfhe-cuda-backend        0.16.0    0.16.0     tfhe-cuda-backend-0.16.0       9        BUMP
tfhe-versionable         0.8.1     0.8.0      tfhe-versionable-0.8.0         5        TAG
zk-cuda-backend          0.2.1     0.2.1      zk-cuda-backend-0.2.1          0        OK
tfhe-csprng              0.10.0    0.10.0     tfhe-csprng-0.10.0             1        BUMP
tfhe-safe-serialize      0.1.1     0.1.1      tfhe-safe-serialize-0.1.1      1        BUMP
tfhe-zk-pok              0.9.0     0.9.0      tfhe-zk-pok-0.9.0              1        BUMP
tfhe                     1.8.0     1.8.1      tfhe-rs-1.8.0                  8        BUMP

Changed without a version bump:
  tfhe-cuda-backend: 9 changed file(s) since tfhe-cuda-backend-0.16.0, bump backends/tfhe-cuda-backend/Cargo.toml
  tfhe-csprng: 1 changed file(s) since tfhe-csprng-0.10.0, bump tfhe-csprng/Cargo.toml
  tfhe-safe-serialize: 1 changed file(s) since tfhe-safe-serialize-0.1.1, bump utils/tfhe-safe-serialize/Cargo.toml
  tfhe-zk-pok: 1 changed file(s) since tfhe-zk-pok-0.9.0, bump tfhe-zk-pok/Cargo.toml
  tfhe: 8 changed file(s) since tfhe-rs-1.8.0, bump tfhe/Cargo.toml

Bumped but not tagged, create the tags in this order:
  git tag -s -a -m "tfhe-versionable-derive 0.8.1 release" tfhe-versionable-derive-0.8.1 && git push origin tag tfhe-versionable-derive-0.8.1
  git tag -s -a -m "tfhe-versionable 0.8.1 release" tfhe-versionable-0.8.1 && git push origin tag tfhe-versionable-0.8.1

Note: publish tfhe-versionable-derive before tfhe-versionable; cargo publish fails on a dependency that is not on crates.io yet.
```

I like how in the end of it I see a descriptive explanation of what needs to be done, but I cannot understand the status column in the crate list. In this first version I see crates with `OK`, which makes me think there is nothing to do with them (i.e. they were not updated or the version was already bumped and it's ok for a new release); and there is `tfhe-versionable` which has `TAG` in the status list. I don't know what that means and I don't see it explained when I try to run with `-h`.

Anyway, `tfhe-cuda-backend` status is `BUMP`. It says there are 9 changes, which I can see where thanks to `--files`. Local version matches crates.io, so intuitively this status means that I need to bump the crate version. So I go to `backends/tfhe-cuda-backend/Cargo.toml` and change it to `0.17.0`. Run the script again but this time I see:

```
➜  bump_detector_script git:(ns/chore/bump_detector_script) ✗ make release_status
error: failed to select a version for the requirement `tfhe-cuda-backend = "^0.16.0"`
candidate versions found which didn't match: 0.17.0
location searched: /home/zama/Code/worktree/bump_detector_script/backends/tfhe-cuda-backend
required by package `tfhe v1.8.0 (/home/zama/Code/worktree/bump_detector_script/tfhe)`
python3 ./scripts/release_status.py
crate                    manifest  crates.io  baseline                       changes  status
tfhe-cuda-common         0.1.1     0.1.1      tfhe-cuda-common-0.1.1         0        OK
tfhe-fft                 0.10.1    0.10.1     tfhe-fft-0.10.1                0        OK
tfhe-hpu-backend         0.7.0     0.7.0      tfhe-hpu-backend-0.7.0         0        OK
tfhe-ntt                 0.7.1     0.7.1      tfhe-ntt-0.7.1                 0        OK
tfhe-versionable-derive  0.8.1     0.8.0      tfhe-versionable-derive-0.8.0  7        TAG
wasm-par-mq              0.1.1     0.1.1      wasm-par-mq-0.1.1              0        OK
tfhe-cuda-backend        0.17.0    0.16.0     tfhe-cuda-backend-0.16.0       9        MISMATCH
tfhe-versionable         0.8.1     0.8.0      tfhe-versionable-0.8.0         5        TAG
zk-cuda-backend          0.2.1     0.2.1      zk-cuda-backend-0.2.1          0        OK
tfhe-csprng              0.10.0    0.10.0     tfhe-csprng-0.10.0             1        BUMP
tfhe-safe-serialize      0.1.1     0.1.1      tfhe-safe-serialize-0.1.1      1        BUMP
tfhe-zk-pok              0.9.0     0.9.0      tfhe-zk-pok-0.9.0              1        BUMP
tfhe                     1.8.0     1.8.1      tfhe-rs-1.8.0                  8        BUMP

!! tfhe-cuda-backend: HEAD declares version 0.16.0, commit the bump before tagging

Changed without a version bump:
  tfhe-csprng: 1 changed file(s) since tfhe-csprng-0.10.0, bump tfhe-csprng/Cargo.toml
  tfhe-safe-serialize: 1 changed file(s) since tfhe-safe-serialize-0.1.1, bump utils/tfhe-safe-serialize/Cargo.toml
  tfhe-zk-pok: 1 changed file(s) since tfhe-zk-pok-0.9.0, bump tfhe-zk-pok/Cargo.toml
  tfhe: 8 changed file(s) since tfhe-rs-1.8.0, bump tfhe/Cargo.toml

Bumped but not tagged, create the tags in this order:
  git tag -s -a -m "tfhe-versionable-derive 0.8.1 release" tfhe-versionable-derive-0.8.1 && git push origin tag tfhe-versionable-derive-0.8.1
  git tag -s -a -m "tfhe-versionable 0.8.1 release" tfhe-versionable-0.8.1 && git push origin tag tfhe-versionable-0.8.1

Note: publish tfhe-versionable-derive before tfhe-versionable; cargo publish fails on a dependency that is not on crates.io yet.
```

`BUMP` was replaced by `MISMATCH`, and there is an error at the beginning. Not the behavior I would expect for this script.

nsarlin-zama:
Thanks a lot for trying it!

I think there are 2 issues here:

1. the error
```
error: failed to select a version for the requirement `tfhe-cuda-backend = "^0.16.0"`
candidate versions found which didn't match: 0.17.0
location searched: /home/zama/Code/worktree/bump_detector_script/backends/tfhe-cuda-backend
required by package `tfhe v1.8.0 (/home/zama/Code/worktree/bump_detector_script/tfhe)`
```

This one is certainly that you bumped `tfhe-cuda-backend` but did not update it in `tfhe/Cargo.toml`. The error is coming from the Makefile, even before the script was started. I think any make command would return the same error in that state. So I cannot catch this from the script, but I can certainly improve the intructions about "bumping".

2. the MISMATCH

This one is explained in the output, but I agree that it can get lost in the noise. It's because the bump was not committed:
```
!! tfhe-cuda-backend: HEAD declares version 0.16.0, commit the bump before tagging
```

It's an error because in the release process you need to `git tag`, and if you forgot to commit you might tag the previous commit without the bump.

I agree that the documentation is not complete, I will try to add more info about that! And maybe try to make the important bits stand out more from the rest of the output.

PR #3943, opened 2026-09-14
fix(ntt): fix invalid logic in the NTT plan creation

IceTDrinker:
reduction only performs a single reduction in mul_accumulate
- primes needing two reduction can return invalid results
- previous criterion was derived on an empty/0 accumulator
- add a regression test with a worst case accumulator surfacing the issue
- remove check for internal fast primes: they are not used with the
mul_accumulate code path and so retain performance on the
mul_assign_normalize code associated with their use (e.g. primes32::P0) in
the various polymul plans (native32/64/128, native_binary32/64/128)
- TODO: consider selecting better primes for the barrett path in the 2^50
range
- TODO: consider having a double reduction code path gated by a const
generics so the code is easier to maintain and still allow much better
performance than a generic reduction code path


IceTDrinker:
problem is plan 52 relies on Plan64 which does not guarantee all underlying plans are propely initialized on 52 bits with the flawed logic of this patch (if prime is not fast reducible it fallsback to 64 bits making underlying plans a mix of incompatible 52/64 bits), there are questions around how this should be handled to have "nice primes" which have easy to write division via some magic constants + shifts

IceTDrinker:
The logic regarding the q/3 threshold was incorrect for the `mul_accumulate` path (the product code performs a single reduction only; the q/3 check stemmed from my misreading of the code).

The logic for choosing between 52-bit and 64-bit modes conflicted with a library invariant I had overlooked.
- Currently, certain CRT plans (such as `Plan52` in `native_binary_64`) require all underlying "prime" plans to be initialized identically.
- Consequently, we cannot select a fast plan for some 52-bit modules while using a different one for others; therefore, if 52-bit processing is possible (whether fast or not), we choose to instantiate the plan as-is.

It turns out that `mul_accumulate` isn't used by the CRT plan, so ultimately it has no impact; however, we could likely improve our choice of primes, so an update might be in order.

IceTDrinker:
there still is a perf regression currently, putting back into draft

PR #3947, opened 2026-09-15
Half rotate and half product

mayeul-zama:
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947)


IceTDrinker (commented):
stupid question to start

> @IceTDrinker made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-:-P1_uiwTCHXrI7xVlcn2:bu8rgxf)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs` line 262 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_ubDP3zkKWb4DSpw2:-P1_ubDP3zkKWb4DSpw3:bp6h446) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs#L262)):*
> ```XML
> /// bootstrap.
> #[inline]
> pub(crate) fn cmux_step<OutputScalar, ContGgsw>(
> ```

no impact on perf ?

should it be inline(always) ?
___
*[`tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs` line 61 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_uh6J8y6PfO80BjvX:-P1_uh6J8y6PfO80BjvY:bp6h446) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs#L61)):*
> ```XML
> /// [`Fourier128LweBootstrapKey::blind_rotate_assign_split`] and the half-rotate bootstrap.
> #[inline]
> pub(crate) fn cmux_step_split<ContGgsw>(
> ```

no impact on perf ?

should it be inline(always) ?

IceTDrinker (changes requested):
Surface remarks for now, I think some typing is misleading/wrong in the fourier half product part of things

if it's annoying to create too many types for this part perhaps something like 

`HalfProductGgswCiphertext<Content: HalfProductGgswMarker>`

with marker types MaskAndBody, Mask, Body

you can have an impl block for Content: HalfProductGgswMarker for the generic stuff like the shared metadata

and dedicated impl blocks (so that only the MaskAndBody variant can split into a Mask and Body HalfProd, but not the others) can help manage the code with essentially the same underlying struct

you would have 

```rust
impl<C: Container<Element = f64>> Fourier128HalfProductGgswCiphertext<MaskAndBody, C> {
...
    fn split_blocks(&self) -> (Fourier128HalfProductGgswCiphertext<Mask, C>, Fourier128HalfProductGgswCiphertext<Body, C>)
}
```

and only a single struct that would look like

```rust
struct Fourier128HalfProductGgswCiphertext<Content: HalfProductGgswMarker, C> {
...
    content: PhantomData<Content>,
}
```

I believe the PhantomData is required otherwise the generics Content is unused and I remember cases where it could cause issues with the compiler complaining about it. I could be wrong though and the PhantomData may not be required

> @IceTDrinker reviewed 1 file and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-:-P1pWFlr3FCwqGzzMMOH:bwxmklf)* status: 1 of 35 files reviewed, 5 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 18 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pR_uN4xctTzEx1zTa:-P1pR_uN4xctTzEx1zTb:b700t12) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L18)):*
> ```XML
> /// * the `k` mask GLev ciphertexts use `(base_log_mask, level_count_mask)`,
> /// * the body GLev ciphertext uses `(base_log_body, level_count_body)`.
> ///
> ```

maybe indicate we expect level_count_body < level_count_mask otherwise it makes little sense ?
___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 29 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pWBojDuEhaxq95waZ:-P1pWBojDuEhaxq95wa_:bdtmun) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L29)):*
> ```XML
> /// ```text
> /// [ level l_msk: rows 0..k-1 ] .. [ level 1: rows 0..k-1 ] [ level l_body ] .. [ level 1 ]
> /// |<---------------- k * l_msk GLWE ciphertexts ---------->|<---- l_body GLWE ciphertexts -->|
> ```

did you check what layout the GPU elected to use ?
___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw_half_product.rs` line 155 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pVObs7fqxjAyvStij:-P1pVObs7fqxjAyvStik:bntkx7q) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw_half_product.rs#L155)):*
> ```XML
> 
>     /// Split the containers into the mask block and the body block.
>     fn split_blocks(self) -> (Self, Self)
> ```

we agree that the resulting objects are not in fact HalfProductGgswCiphertext(s) ? since one contains all masks and the other all body ?

pdroalves (commented):
> @pdroalves made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947)* status: 1 of 35 files reviewed, 5 unresolved discussions (waiting on IceTDrinker, mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 29 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pWBojDuEhaxq95waZ:-P23T9RaC1p2VxrSna50:btgk94) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L29)):*

did you check what layout the GPU elected to use ?
</blockquote></details>

All GPU PBSs are level-major. That's not a problem.

mayeul-zama (commented):
> @mayeul-zama reviewed all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947)* status: 1 of 35 files reviewed, 4 unresolved discussions (waiting on IceTDrinker, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 18 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pR_uN4xctTzEx1zTa:-P278mwFEQ_yNvYZoRj3:b-8b11j9) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L18)):*

maybe indicate we expect level_count_body < level_count_mask otherwise it makes little sense ?
</blockquote></details>

Added
___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs` line 262 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_ubDP3zkKWb4DSpw2:-P278pcXEvMyWHfFWNei:b-eqqhin) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs#L262)):*

no impact on perf ?

should it be inline(always) ?
</blockquote></details>

I just tried to bench but the diff (if there is one) is too small compared to the variability of the bench

The cost of this function is not very small so I guess the cost function calling (which inline can remove) is negligible compared to it

I'd remove the `#[inline]`, what do you think?
___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw_half_product.rs` line 155 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pVObs7fqxjAyvStij:-P27DQkfDe2iTm50WHeq:b-t9a3w1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128/crypto/ggsw_half_product.rs#L155)):*

we agree that the resulting objects are not in fact HalfProductGgswCiphertext(s) ? since one contains all masks and the other all body ?
</blockquote></details>

Yes, that was not clean
Replaced by `mask_data` and `body_data` function to avoid adding new types for this private use
___
*[`tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs` line 61 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_uh6J8y6PfO80BjvX:-P27D-eL-e7g_ox9vX0D:bug1fom) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs#L61)):*

no impact on perf ?

should it be inline(always) ?
</blockquote></details>

Discussed (for another function, same scenario) in another comment

IceTDrinker (commented):
could we maybe update this PR to have the half rotate only + corresponding tests (currently missing) so that it's easier to review ?

then add the half product stuff and then the combination of both ?

> @IceTDrinker reviewed 15 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-:-P2HbwRCATKbBvPv3ymh:b-us39n6)* status: 16 of 35 files reviewed, 5 unresolved discussions (waiting on mayeul-zama, nsarlin-zama, soonum, and SouchonTheo).

___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 18 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pR_uN4xctTzEx1zTa:-P2HT_4J86Djgo-27rzu:b-qyrh8o) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L18)):*

Added
</blockquote></details>

does the code support level_count_body > level_count_mask ?
___
*[`tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs` line 29 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1pWBojDuEhaxq95waZ:-P2HTOHU6ljaH_9RQt8x:bz6v4w4) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/entities/half_product_ggsw_ciphertext.rs#L29)):*
<details><summary><i>Previously, pdroalves (Pedro Alves) wrote…</i></summary><blockquote>

All GPU PBSs are level-major. That's not a problem.
</blockquote></details>

the potential difference is the way mask vs bodies are stored, in classic GGSWs everything is like mask|body etc.

here the mask part as far as I understand is one chunk and the body body part is one chunk
___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs` line 262 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_ubDP3zkKWb4DSpw2:-P2HZKlwA6t5lXl1lhvV:bjhbi0b) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128/crypto/bootstrap.rs#L262)):*

I just tried to bench but the diff (if there is one) is too small compared to the variability of the bench

The cost of this function is not very small so I guess the cost function calling (which inline can remove) is negligible compared to it

I'd remove the `#[inline]`, what do you think?
</blockquote></details>

I think I would prefer keeping the existing PBS 128 code as close as possible to what it is, so in that case I'm thinking inline(always) for both relevant functions
___
*[`tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs` line 61 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P1_uh6J8y6PfO80BjvX:-P2HbUcR4k6P8O_uLR0H:b16fb80) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0b7de82cd7538e8c57b12e676016039b177f213f/tfhe/src/core_crypto/fft_impl/fft128_u128/crypto/bootstrap.rs#L61)):*

Discussed (for another function, same scenario) in another comment
</blockquote></details>

I'd prefer inline(always) to keep the original call site spirit as close as possible to what it was
___
*[`tfhe/src/core_crypto/fft_impl/fft128/crypto/tests.rs` line 82 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3947#-P2HYVK-0A9D9gBw0j_e:-P2HYVK-0A9D9gBw0j_f:b-c6kh32) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/0d78fb58a510134f9bc4125395c3a823ddae33f9/tfhe/src/core_crypto/fft_impl/fft128/crypto/tests.rs#L82)):*
> ```XML
> // tolerance is set well above the values observed in practice (around `2^40`) rather than at a
> // tight bound a rare draw could exceed. `2^56` out of `2^128` is still 67 bits below the `2^123`
> // spacing of the encoded messages, so a structural error cannot hide under it.
> ```

what's this about ?

2^40 difference sounds like something is going wrong in the code ? it means we essentially have 40 bits of noise on the lower limb ? this does not sound right

PR #3950, opened 2026-09-16
chore(hl): deprecate allow_unpacked compact list conformance params

nsarlin-zama:
closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1562

### Check-list:
Deprecate `allow_unpacked` in compact list conformance params. We currently cannot correctly guarantee that an unpacked compact list is conformant. For example, the degree could be above the message_modulus, or a boolean could have more than one bit set.

This is a deprecation to make it loud that this should be avoided, while still giving some time for migration. It also gives us a bit more time to decide what we do with this. If we agree that this is the correct solution, I will create an issue to completely remove the `allow_unpacked` method in the following release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3950)


IceTDrinker (commented):
as discussed maybe worth deprecating allow_unpacked on ProvenCompatCiphertextList, since we can't really trust CompactCiphertextList alone since we have no guarantee on the value of the padding bit

> @IceTDrinker reviewed 8 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3950#-:-P1jEUsAAEg3Bk2Al70x:b1zoiyo)* status: all files reviewed, 1 unresolved discussion (waiting on nsarlin-zama, SouchonTheo, and tmontaigu).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1313 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3950#-P1jDyRq30tm62mk4eJe:-P1jDyRq30tm62mk4eJf:b-1wb58w) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/82a6b6c55ea95451d191d5f045fbb6039c764d26/tfhe/src/integer/ciphertext/compact_list.rs#L1313)):*
> ```XML
>         note = "unpacked lists might not be sanitized during expansion, so they are unsafe to accept \
>                 from an untrusted source. For trusted inputs, skip the conformance check with \
>                 `DeserializationConfig::disable_conformance`. For untrusted inputs, require packed \
> ```

worth indicating that sanitizing is safe only if the list is Proven, otherwise we don't have a guarantee on the padding bit

IceTDrinker (commented):
generally for untrusted/unproven CompactList the only thing that can be checked is the parameters conformance, for the rest nothing meaningful can be done since you can not know if the noise is ok, the encoding of the plaintext is ok, etc.

> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3950#-:-P1jMK_PBMLe9uOCY8iP:bn52b0m)* status: all files reviewed, 1 unresolved discussion (waiting on nsarlin-zama, SouchonTheo, and tmontaigu).

nsarlin-zama:
replaced by https://github.com/zama-ai/tfhe-rs/pull/3957

PR #3954, opened 2026-09-17
fix(hlapi): fused_scalar_mul_scalar_div when scalar type is wider tha…

tmontaigu:
…n fhe

The macro that declared the wide fhe type to be used in the fused_scalar_mul_scalar_div was 'half' incorrect for non classical width types, width for which the clear type allowed has a bigger width than the FHE type. E.g for FheUint4 the wide fhe type was declared as FheUint8 and scalar type u8 (wide scalar type u16).
This is fine as long scalar value for the mul (encoded on a u8) was actually in range [0, 2^4[, but when the value was in the range [2^4, 2^8[, the internal computation the mul would not use a type wide enough, losing precision, leading to incorrect result

The immediate fix is to align the wide FheType to have the same width as the wide scalar type.

Another fix could have been not accepting values not in the same range as the fhe type. Or possibly create wrapper scalar types to have our own `u4`.

Fixes https://github.com/zama-ai/tfhe-rs-internal/issues/1564

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954)


IceTDrinker (commented):
Thanks a lot for the PR, looks good.

Some questions:
- only enc/scalar/scalar was affected ? (it feels like the enc/enc/scalar was probably already fine when looking at the width of the intermediate result
    - should we consider splitting the enc/enc/scalar from the other impl later on (in a follow up) to hope for better performance in the enc/enc case ?
- should we add a test for each width ?
- if so should it be something like max_enc * max_scalar / max_scalar -> max_enc for unsigned and the same with min_scalar for signed variants, should we have some cases where the mul scalar < div scalar and the other way around as well ?

> @IceTDrinker reviewed 8 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-:-P1oTEfsDc4uVgvlXcrw:bneb8xc)* status: all files reviewed, 2 unresolved discussions (waiting on nsarlin-zama and tmontaigu).

___
*[`tfhe/src/high_level_api/integers/signed/tests/mod.rs` line 598 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-P1oSnLc4rEAyMHJ2yzN:-P1oSnLc4rEAyMHJ2yzO:b-qen0kz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5cef2865cfbfced844f25fd7e1f752afad315e2f/tfhe/src/high_level_api/integers/signed/tests/mod.rs#L598)):*
> ```XML
>     // we use the closest larger clear type, here we make sure the results
>     // are as expected
>     check_fused_mul_div::<FheInt4, i8, i16>(cks, 3, 127, 127);
> ```

should we instantiate tests for all widths ? :face_with_peeking_eye: 
___
*[`tfhe/src/high_level_api/integers/signed/tests/mod.rs` line 672 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-P1oSOfx0Wg4sQ0jBd-y:-P1oSOfx0Wg4sQ0jBd-z:bjpnrqz) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5cef2865cfbfced844f25fd7e1f752afad315e2f/tfhe/src/high_level_api/integers/signed/tests/mod.rs#L672)):*
> ```XML
> 
>     let a = FheType::try_encrypt(clear_a, cks).unwrap();
>     // encrypted * scalar / scalar
> ```

only the scalar variant was affected ?

tmontaigu (commented):
1) Only scalar affected
   If we were to impl at the integer level, we would probably pick the correct size to use depending on the scalar to have the shortest  mul and divs possible
2) Somewhat your call, can be revisited while sorting out integer,graph api testing and 2_2, vs others params
3) I think scalar test would need two loops one drawing in same range as encrypted type, other drawing outside of the encrypted type range 

> @tmontaigu made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-:-P1ocylw-Qmh9dzYkNRH:b-y3md8n)* status: all files reviewed, 2 unresolved discussions (waiting on IceTDrinker and nsarlin-zama).

___
*[`tfhe/src/high_level_api/integers/signed/tests/mod.rs` line 598 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-P1oSnLc4rEAyMHJ2yzN:-P1ocLeM9LBrWfkv5vFc:b-vytm8f) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5cef2865cfbfced844f25fd7e1f752afad315e2f/tfhe/src/high_level_api/integers/signed/tests/mod.rs#L598)):*

should we instantiate tests for all widths ? :face_with_peeking_eye:
</blockquote></details>

Theorically we should, graph api will have a more exhaustive term
here we could use a macro to have test run on all sizes
But it will indeed impact tests timing
___
*[`tfhe/src/high_level_api/integers/signed/tests/mod.rs` line 672 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-P1oSOfx0Wg4sQ0jBd-y:-P1oaC0L23b9nzTltpQe:b-coy8k) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5cef2865cfbfced844f25fd7e1f752afad315e2f/tfhe/src/high_level_api/integers/signed/tests/mod.rs#L672)):*

only the scalar variant was affected ?
</blockquote></details>

Yes beause the whole problem was a mismatch in the wide size picked with regards to the narrow size

in enc/enc, the narrow size is the same for both enc, so picking the narrow size is just picking twice the size of narrow


in the scalar case the error came because in enc/scalar, for the type used for enc, there was no rust type matching the exact same with (e.g FheUint4 would need a u4 which rust does not provide so we use u8) and since we used the larger clear type possible users can use a value bigger than the enc type, and so we have to make sure to use correct wide type, so a size allowing to do a full multiplication of `u4 * u8` so the easiest is to pick as the wide width the witdh of the wide scalar type

IceTDrinker:
do we think the "worst cases" tests would significantly slow the HL tests ?

IceTDrinker (approved):
Thanks

> @IceTDrinker made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-:-P1ofFYY7zzIfLZGdPcw:b-zcjvte)* status: all files reviewed, 1 unresolved discussion (waiting on nsarlin-zama and tmontaigu).

___
*[`tfhe/src/high_level_api/integers/signed/tests/mod.rs` line 598 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-P1oSnLc4rEAyMHJ2yzN:-P1ofDEeD9O93IJ3x060:bxunx7j) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/5cef2865cfbfced844f25fd7e1f752afad315e2f/tfhe/src/high_level_api/integers/signed/tests/mod.rs#L598)):*

Theorically we should, graph api will have a more exhaustive term
here we could use a macro to have test run on all sizes
But it will indeed impact tests timing
</blockquote></details>

let's have it as part of the revamped tests

tmontaigu (commented):
Hard to tell, but since this ops chains the 2 most expensive scalar ops it will indeed have a cost

> @tmontaigu made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3954#-:-P1ohUGE6-sdgqP8ke_Y:bn0ddbg)* status: all files reviewed, 1 unresolved discussion (waiting on nsarlin-zama).

IceTDrinker:
if you can give it a shot with a robot to have an estimate of how much time it represents

tmontaigu:
I might just as well run it and measure

IceTDrinker:
up to you

IceTDrinker:
> I might just as well run it and measure

any report on how long the tests ran ?

PR #3955, opened 2026-09-17
fix(gpu): use msg mod for carry_extract in compact list expand

nsarlin-zama:

### PR content/description
compact list expansion used `x / carry_modulus` for the carry_extract, but it should be `x / message_modulus`. I found this because there is the same bug is present in the cpu code


nsarlin-zama:
~I bumped to 0.16.1 since this is just a minor fix, but maybe you would prefer a bump to 0.17.0~

bump to 0.17.0 already done on main, so I removed mine

enzodimaria (approved):

PR #3957, opened 2026-09-18
chore(hl)!: compact list expand now always sanitize

nsarlin-zama:
BREAKING_CHANGE: compact list expand now returns an error if called without a server key in scope.

closes: https://github.com/zama-ai/tfhe-rs-internal/issues/1562
supersedes: https://github.com/zama-ai/tfhe-rs/pull/3950

### PR content/description
At the hl, CompactList and ProvenCompactList expand methods now always go through the sanitization pbs. The escape hatch that allowed to skip sanitization was made for wasm tests only (as the comment stated). It has been also used by other tests. But it impacts every caller even outside of tests.
I think that, for prod case, we need to be confident that sanitization will run.

This pr removes this shortcut in every call to `expand`. To fix the tests, a new method `expand_without_key` has been added. This method can only be called from js or in `cfg(test)`.

This is technically breaking, since it is now not possible from the outside to expand a list without sanitization. But I think it's ok since it breaks an "unsafe" codepath.

As a side note, it also simplifies a bit the expand code. IntegerCompactCiphertextListExpansionMode has lost a variant, and the underlying function have one less case to handle.

This pr only touches the HL/Integer part. But I believe a similar logic could (should?) by applied at the shortint level. This would reduce the complexity of the expansion code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)


nsarlin-zama:
cc @zama-ai/gpu, it looks like this surfaced some gpu compact list expansions tests not running on gpu that were taking the no pbs shortcut. Can you have a look at these tests and check that the fix is correct?

IceTDrinker (commented):
> @IceTDrinker reviewed 15 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: all files reviewed, 3 unresolved discussions (waiting on nsarlin-zama and tmontaigu).

___
*[`tfhe/src/high_level_api/compact_list.rs` line 1164 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22LeZHAihSROR488jM:-P22LeZHAihSROR488jN:btck5kc) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2ac7562f3fc5bf48c22184f86849ef65eba5627c/tfhe/src/high_level_api/compact_list.rs#L1164)):*
> ```XML
>         // Since the server key is stored in a thread local, expand in a thread of our own to make
>         // sure that no other test has set one
>         std::thread::spawn(move || {
> ```

no need for a thread I think ? unless cargo re-uses them cross tests, otherwise I there is an unset_server_key function

https://docs.rs/tfhe/latest/tfhe/fn.unset_server_key.html
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1071 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22MIBv9ixn-gcdHUxh:-P22MIBv9ixn-gcdHUxi:bnduexu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2ac7562f3fc5bf48c22184f86849ef65eba5627c/tfhe/src/integer/ciphertext/compact_list.rs#L1071)):*
> ```XML
>     /// degree they advertise. The list needs to be unpacked.
>     /// Use [`expand`](Self::expand) if you need sanitization or unpacking.
>     pub fn expand_without_key(&self) -> crate::Result<CompactCiphertextListExpander> {
> ```

maybe gate by the same features as in HL ?

But I think I see the reinterpret_data needing this for the example :/ 
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1069 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22oGe0-0mJ9paE6f8X:-P22oGe0-0mJ9paE6f8Y:b-e2kpre) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1a7ed181666f12217ad19e338a09f2112e8efd82/tfhe/src/integer/ciphertext/compact_list.rs#L1069)):*
> ```XML
>     ///
>     /// No sanitizing lookup table is applied to its blocks, so they may hold values above the
>     /// degree they advertise. The list needs to be unpacked.
> ```

why does the list need to be unpacked, if it's unpacked to begin with ?

nsarlin-zama (commented):
> @nsarlin-zama made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: 13 of 15 files reviewed, 3 unresolved discussions (waiting on IceTDrinker and tmontaigu).

___
*[`tfhe/src/high_level_api/compact_list.rs` line 1164 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22LeZHAihSROR488jM:-P23RHysAnDiEE4WovgU:b3c7vtt) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2ac7562f3fc5bf48c22184f86849ef65eba5627c/tfhe/src/high_level_api/compact_list.rs#L1164)):*

no need for a thread I think ? unless cargo re-uses them cross tests, otherwise I there is an unset_server_key function

https://docs.rs/tfhe/latest/tfhe/fn.unset_server_key.html
</blockquote></details>

Yes thread reuse can happen, for example with `--test-threads=1`. But yes indeed, `unset_server_key` is better suited for that.
done
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1071 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22MIBv9ixn-gcdHUxh:-P23OpfGAu0EJOUs7zlk:btmg782) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/2ac7562f3fc5bf48c22184f86849ef65eba5627c/tfhe/src/integer/ciphertext/compact_list.rs#L1071)):*

maybe gate by the same features as in HL ?

But I think I see the reinterpret_data needing this for the example :/
</blockquote></details>

yes that's exactly why I did it like this. Since this is the integer level that is for more "advanced users" (ie: us in practice), I'm not against making it available.
___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1069 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22oGe0-0mJ9paE6f8X:-P23OxGU29wibFBPdWdO:boj982b) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1a7ed181666f12217ad19e338a09f2112e8efd82/tfhe/src/integer/ciphertext/compact_list.rs#L1069)):*

why does the list need to be unpacked, if it's unpacked to begin with ?
</blockquote></details>

I'm not sure I understand? Since packed/unpacked is not a type level info, this function could mistakenly be called with a packed list

tmontaigu (commented):
> @tmontaigu reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: all files reviewed, 3 unresolved discussions (waiting on IceTDrinker).

tfhe/src/high_level_api/integers/signed/tests/gpu.rs
```diff
@@ -129,34 +129,32 @@ fn test_min_max_gpu() {
...
+        crate::high_level_api::integers::unsigned::tests::gpu::setup_gpu_with_dedicated_cpk();
+    let public_key = CompactPublicKey::new(&client_key);
+    let compact_list = CompactCiphertextList::builder(&public_key)
+        .push(-123i8)
+        .build_packed();
+    let expanded = compact_list.expand().unwrap();
+    let a: FheInt8 = expanded.get(0).unwrap().unwrap();
+
+    let clear: i8 = a.decrypt(&client_key);
+    assert_eq!(clear, -123i8);
```
enzodimaria, line 157: Why not concatenate test_compact_public_key_small_gpu and test_compact_public_key_big_gpu?

enzodimaria: Same in gpu.rs from /unsigned/

nsarlin-zama: I didn't want to change more gpu code than what I needed for my pr, but I can do this now if you prefer !

enzodimaria: Well this is the only comment that I would do, everything else seems very consistant! So yeah, I think we should concatenate these functions, there are almost the same (with a small diff in /signed/)

nsarlin-zama: hmm actually, reading the name I think they should not be THAT similar, there is an issue here. Do you know if this is a simple copy paste error? Did we want at some point to actually test the gpu code that expands to something encrypted for the small key? Because despite their name they both expand to the big key. Is it even possible on gpu?

enzodimaria: Oh I see! @zama-ai/gpu should know better about this

IceTDrinker (commented):
Looks ok, but I've been reading so much code I would like another pair of eyes on this

@nsarlin-zama should we have a robot run on this ? we can ping Jordan e.g.

> @IceTDrinker reviewed 3 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-:-P28hyaL4-kLZ8TdgihO:b-c3l00j)* status: all files reviewed, 2 unresolved discussions (waiting on enzodimaria and nsarlin-zama).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1069 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22oGe0-0mJ9paE6f8X:-P28hGo_0e1asLf2vL8k:b-x9c8qg) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1a7ed181666f12217ad19e338a09f2112e8efd82/tfhe/src/integer/ciphertext/compact_list.rs#L1069)):*

I'm not sure I understand? Since packed/unpacked is not a type level info, this function could mistakenly be called with a packed list
</blockquote></details>

ok the "This list needs to be unpacked" seems to refer to the "unpacked list" cited at the start of the docstring

as for knowing if the list believes it is packed there is a function to verify it it's the is_packed that is called a bit below ?

nsarlin-zama (commented):
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: all files reviewed, 2 unresolved discussions (waiting on enzodimaria and IceTDrinker).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1069 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22oGe0-0mJ9paE6f8X:-P2CLPULCSlxzAUdwoGh:b-vanfbu) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1a7ed181666f12217ad19e338a09f2112e8efd82/tfhe/src/integer/ciphertext/compact_list.rs#L1069)):*

ok the "This list needs to be unpacked" seems to refer to the "unpacked list" cited at the start of the docstring

as for knowing if the list believes it is packed there is a function to verify it it's the is_packed that is called a bit below ?
</blockquote></details>

Yes it's the `is_packed` check. I reworded the comment, is it better?

IceTDrinker (commented):
@nsarlin-zama should we ping Jordan for an AI review ?

> @IceTDrinker reviewed 1 file and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-:-P2CaAPNCwiNUOSf3_zb:b-ukerfm)* status: all files reviewed, 1 unresolved discussion (waiting on enzodimaria and nsarlin-zama).

___
*[`tfhe/src/integer/ciphertext/compact_list.rs` line 1069 at r2](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957#-P22oGe0-0mJ9paE6f8X:-P2Ca5Tf5mjG_19a2C4x:bz7zmf) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1a7ed181666f12217ad19e338a09f2112e8efd82/tfhe/src/integer/ciphertext/compact_list.rs#L1069)):*

Yes it's the `is_packed` check. I reworded the comment, is it better?
</blockquote></details>

yep thanks!

nsarlin-zama:
@jfrery could you run an AI review on this pr?

jfrery (commented):
**Codex review: 1 finding** (1 minor; 1 security)

The unconditional sanitization path lets unchecked, attacker-controlled compact-list metadata trigger a process-level denial of service. Validate the metadata against the actual ciphertext block count before allocating sanitization LUTs.

Not on changed lines:
- **Minor (security):** Validate metadata before allocating sanitization LUTs (`tfhe/src/high_level_api/compact_list.rs:415`)
  When an unpacked, no-casting list comes from unchecked deserialization—including the WASM `safe_deserialize`, which disables conformance—removing the fast path makes this call enter `sanitize_blocks`. That path sizes its LUT vector from untrusted `info` before comparing it with the actual block count, so a tiny list containing one block but `DataKind::Unsigned(NonZeroUsize::MAX)` metadata triggers a capacity-overflow panic or allocation abort. Validate the metadata's checked block count against the ciphertext count before allocating or sanitizing.

<sub>Automated review by codex in an Umbra confidential VM, commit `3ae83b507e52`.</sub>

tmontaigu (approved):
> @tmontaigu reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: all files reviewed, 1 unresolved discussion (waiting on enzodimaria and nsarlin-zama).

nsarlin-zama:
> **Codex review: 1 finding** (1 minor; 1 security)
> 
> The unconditional sanitization path lets unchecked, attacker-controlled compact-list metadata trigger a process-level denial of service. Validate the metadata against the actual ciphertext block count before allocating sanitization LUTs.
> 
> Not on changed lines:
> 
>     * **Minor (security):** Validate metadata before allocating sanitization LUTs (`tfhe/src/high_level_api/compact_list.rs:415`)
>       When an unpacked, no-casting list comes from unchecked deserialization—including the WASM `safe_deserialize`, which disables conformance—removing the fast path makes this call enter `sanitize_blocks`. That path sizes its LUT vector from untrusted `info` before comparing it with the actual block count, so a tiny list containing one block but `DataKind::Unsigned(NonZeroUsize::MAX)` metadata triggers a capacity-overflow panic or allocation abort. Validate the metadata's checked block count against the ciphertext count before allocating or sanitizing.
> 
> 
> Automated review by codex in an Umbra confidential VM, commit `3ae83b507e52`.

Not considered a security issue, since it is already guarded by conformance. Also, the panic predates this pr.
However, a fix is introduced in the next pr in the stack, since it also touches this part of the code: https://github.com/zama-ai/tfhe-rs/pull/3964

tmontaigu (dismissed):
> @tmontaigu reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3957)* status: all files reviewed, 1 unresolved discussion (waiting on enzodimaria and nsarlin-zama).

tfhe/src/high_level_api/integers/signed/tests/gpu.rs
pdroalves: I already explained to @nsarlin-zama but will add here too just for the record.

These tests have bad naming, they are checking the edges of `i8` and `u8` through "big" and "small" inputs, not parameters. And looks like the unsigned version has a copy paste mistake not doing that. We can merge both as long as we keep these two cases.

pdroalves (commented):
As far I can tell, I don't think there is no need to change anything in the GPU path. We don't have that shortcut and sanitization is always applied.

I added a comment on the test question raised by @enzodimaria but I don't think there is anything else to be reviewed by us.

enzodimaria (approved):
Perfect then!

enzodimaria (approved):

PR #3961, opened 2026-09-21
fix(gpu): return errors instead of panicking on unsupported compact lists

bbarbakadze:
### PR content/description

Fixes #1514, and a few more things I found while looking into it.

  The reported bug: a proven compact list containing a string panics when moved to the GPU. Since a `ProvenCompactCiphertextList` comes straight from the user and that list is perfectly conformant, anyone could crash the coprocessor on demand.

  While tracing that path I found the string wasn't the only input that did this, so this PR covers the whole call graph rather than just the one line.

  ## The three types of inputs that crashed
  **A string anywhere in `info`.** Two separate sites: the `match` building the `is_boolean` array, and an `assert!` at the top of `expand()`. Both now go through one shared helper that returns an error. I kept the check in `expand()` as well as the constructor on purpose — `data_info` is `pub(crate)` and mutable, and one of the new tests swaps it *after* the list is already on the GPU.

**An empty list.**  `is_conformant` has an explicit early return saying an empty list is valid — and `is_packed()` returns `false` for it, so the guard `assert!(is_packed())` fired. The safety check was waving through exactly the input that crashed.

**The same empty list,  in lwer level**, at a `.first().unwrap()`. Needed fixing separately.

## Tests

Three new ones in `integer/gpu/zk/mod.rs`. Each builds a real proven list, tampers with the metadata, and **asserts the tampered list is still conformant** before asserting the GPU rejects it. That middle assertion is the important one — it's what proves these are actually reachable and not hypothetical.

`test_proven_list_without_ciphertexts_on_gpu` is the one that caught the CPU bug.

## Breaking change

One, and it's unavoidable:

```rust
- pub fn from_proven_compact_ciphertext_list(..) -> Self
+ pub fn from_proven_compact_ciphertext_list(..) -> crate::Result<Self>

The whole point is that it can now fail instead of aborting, so the return type has to change. It's in tfhe::integer::gpu::zk, which is publicly reachable — our own benchmark was a caller, hence the zk_pke.rs churn.

The high-level API doesn't move. Everything users actually call (verify_and_expand, expand, expand_with_key, …) already returned crate::Result, so no signature changed there. Behaviour does change — inputs that used to abort now come back as Err — but nothing stops compiling.

The other signature changes are all pub(crate) or private.

Worth a decision before merge

CPU and GPU now disagree on the empty list. CPU returns an empty expander; GPU returns an error. CudaFlattenedVecCompactCiphertextList has no LWE dimension to build an empty device buffer from, so matching the CPU means threading an Option through the type — more than I wanted to do in a security fix. An error beats the abort it replaces, but we should decide whether the coprocessor filters empty lists earlier instead.

One panic left in. ProvenCompactCiphertextList::message_modulus() still does proved_lists[0]. Not conformant-reachable, but it's public API that panics, and there's no sensible fallback modulus — the fix is -> Option<MessageModulus>, a breaking change that needs an API owner's call rather than me slipping it in here.

Config panics left alone deliberately. The atomic-pattern panic!, the BSK/KSK dimension asserts, and cpk_key_switching_key_material.as_ref().unwrap() all depend on how the operator built the server key, not on request data — they fail identically on every request, so they're startup misconfiguration, not a DoS vector. Cheap to align later (expand already returns Result), but I kept this diff scoped.

On fuzzing

@ picking up the earlier offer — the natural harness is verify_and_expand with a CUDA key set, seeded with valid lists whose info vector is mutated. That one mutation reproduces all three of the main findings. It needs a GPU runner, so probably nightly GPU CI rather than the existing CPU campaign.

Testing done

clippy with gpu,gpu-zk,zk-pok and -D warnings is clean. 24 tests pass — the 7 GPU zk tests plus the CPU integer::ciphertext suite, including the existing test_attack_proven_list_metadata, test_attack_list_info and the CPU↔GPU re-randomization test, which is what convinced me the is_packed() change didn't alter behaviour anywhere it was previously defined.

One caveat: that 24/24 run was on an identical tree rather than this branch (I verified the nine files byte-for-byte first). Clippy is confirmed green here. Worth a re-run in CI.


### Check-list:

### CODE REVIEW

File: tfhe/src/integer/gpu/ciphertext/compact_list.rs [hunk 1]
```diff
@@ -205,13 +205,37 @@ pub struct CudaFlattenedVecCompactCiphertextList {
     pub(crate) ciphertext_modulus: CiphertextModulus<u64>,
 }
 
+/// Checks that the metadata of a compact ciphertext list describes something the GPU backend is
+/// able to expand.
+fn check_data_info_is_supported_on_gpu(data_info: &[DataKind]) -> crate::Result<()> {
+    if data_info
+        .iter()
+        .any(|kind| matches!(kind, DataKind::String { .. }))
```
andrei-stoian-zama, line 213: please use a whitelist instead of a blacklist here. We support 

```
    /// The held value is a number of radix blocks.
    Unsigned(NonZeroUsize),
    /// The held value is a number of radix blocks.
    Signed(NonZeroUsize),
    Boolean,
```
right ? I'm not so sure about Boolean actually

andrei-stoian-zama: there's also a number of blocks for DataKind ? 

```

impl DataKind {
    pub fn num_blocks(self, message_modulus: MessageModulus) -> usize {
        match self {
            Self::Unsigned(n) | Self::Signed(n) => n.get(),
```

we should whitelist some number of blocks too

andrei-stoian-zama (changes requested):

andrei-stoian-zama (commented):
Does this have a test that reproduces the issue and checks the new error message is raised ?

tfhe/src/integer/gpu/ciphertext/compact_list.rs
```diff
@@ -205,13 +205,37 @@ pub struct CudaFlattenedVecCompactCiphertextList {
...
-        vec_compact_list: Vec<crate::shortint::ciphertext::CompactCiphertextList>,
+        vec_compact_list: &[crate::shortint::ciphertext::CompactCiphertextList],
         data_info: Vec<DataKind>,
         streams: &CudaStreams,
-    ) -> Self {
-        let first = vec_compact_list.first().unwrap();
+    ) -> crate::Result<Self> {
+        check_data_info_is_supported_on_gpu(&data_info)?;
+
+        // An empty list is conformant, so this has to be an error and not an unwrap
```
nsarlin-zama: I don't know how this would be handled at higher levels, but IIRC coproc had the case of empty lists, that's why we accept them on cpu.

I don't know what it would mean for gpu to support them, but that's the kind of edge cases where we need to align cpu/gpu behavior.

PR #3962, opened 2026-09-21
chore(bench): implement a visualization tooling for bench

SouchonTheo:
first version of an ui to display value from benchmark

we can use it as a gui on desktop
- you will be able to load/store value from db
- also load and store view

or you can also use it as a website
- you can only load data (no db access)
- you can load and store view

this was fully vibe coded but I took care to split the code in many file to make the code clear as possible


IceTDrinker (commented):
Do we have clippy/fmt running on the added code
Is it worth keeping the web thing if we can't deploy this one ? (since we can always launch a native gui)

Did not review much since the gui stuff was mostly working (we'll have some refinements to add, e.g. more than 8 items, maybe have a panel on the left with everything being displayed, on the right we have the top right quadrant with the raw timings (with a selector like last/all) and bottom right quadrant is a graph

> @IceTDrinker reviewed 8 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/zama-ai/tfhe-rs/3962#-:-P2D00Jw66BP9l2WVcrJ:bh0bywq)* status: 8 of 27 files reviewed, 1 unresolved discussion (waiting on nsarlin-zama and SouchonTheo).

___
*[`utils/tfhe-bench-viewer/Cargo.toml` line 22 at r1](https://reviewable.io/reviews/zama-ai/tfhe-rs/3962#-P2Cxgw889uWUuj5sy1f:-P2Cxgw889uWUuj5sy1g:b-pnuuc1) ([raw file](https://github.com/zama-ai/tfhe-rs/blob/1295110e9a2167bcf184c09fc27c505b86584acb/utils/tfhe-bench-viewer/Cargo.toml#L22)):*
> ```TOML
> # socket, which a page does not have, and tokio's multi-threaded runtime needs
> # threads, which wasm does not have either.
> [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
> ```

why do we have the web thing if it cannot work standalone ?

were there issues with the native build ?

